PR #10, opened 2023-10-21
ref!: enable static builds

coolaj86:
- switch from `adapter-node` to `adapter-static`
- rename `ENDPOINT` to `API_BASE_URL` to follow common convention and avoid confusing with webserver vs api server
- match on `/api` prefix to more easily allow control of Reverse Proxy routes from Web Server to API Server

# Preview Documentation

https://github.com/coolaj86/ollama-webui/tree/feat-static#how-to-build-for-static-deployment

## How to build

```sh
git clone git@github.com:ollama-webui/ollama-webui.git
pushd ./ollama-webui/
npm run build
```

## Example `Caddyfile`

- Reverse Proxy to `localhost:11434`
- Serve `./build/`

```perl
# replace localhost with example.com or whatever
localhost {
    handle /api/* {
        reverse_proxy localhost:11434
    }

    file_server {
        root ./build/
    }
}
```

```sh
caddy run --config ./Caddyfile
```

### with Working CORS Snippet

```perl
# CORS Preflight (OPTIONS) + Request (GET, POST, PATCH, PUT, DELETE)
(cors-api) {
	@match-cors-api-preflight method OPTIONS
	handle @match-cors-api-preflight {
		header {
			Access-Control-Allow-Origin "{http.request.header.origin}"
			Access-Control-Allow-Methods "GET, POST, PUT, PATCH, DELETE, OPTIONS"
			Access-Control-Allow-Headers "Origin, Accept, Authorization, Content-Type, X-Requested-With"
			Access-Control-Allow-Credentials "true"
			Access-Control-Max-Age "3600"
			defer
		}
		respond "" 204
	}

	@match-cors-api-request {
		not {
			header Origin "{http.request.scheme}://{http.request.host}"
		}
		header Origin "{http.request.header.origin}"
	}
	handle @match-cors-api-request {
		header {
			Access-Control-Allow-Origin "{http.request.header.origin}"
			Access-Control-Allow-Methods "GET, POST, PUT, PATCH, DELETE, OPTIONS"
			Access-Control-Allow-Headers "Origin, Accept, Authorization, Content-Type, X-Requested-With"
			Access-Control-Allow-Credentials "true"
			Access-Control-Max-Age "3600"
			defer
		}
	}
}

# replace localhost with example.com or whatever
localhost {
	## HTTP Basic Auth
	## (uncomment to enable)
	# basicauth {
	# 	# see .example.env for how to generate tokens
	# 	{env.OLLAMA_API_ID} {env.OLLAMA_API_TOKEN_DIGEST}
	# }

	handle /api/* {
		# Comment to disable CORS
		import cors-api

		reverse_proxy localhost:11434
	}

	# Same-Origin Static Web Server
	file_server {
		root ./build/
	}
}
```


tjbck:
Looking good! Great work on enabling the static build. Could you also please verify that it functions seamlessly with Docker? It would be beneficial to ensure compatibility across different environments. Thanks!

tjbck:
Additionally, I also suggest providing users with the option to bypass the reverse proxy, allowing direct usage of the ollama API_ENDPOINT (https://localhost:11434/api). This will ensure flexibility for users while maintaining a straightforward setup process. Thank you for your efforts on this!

coolaj86:
I don't use Docker, but I can verify that both `npm run dev` and `npm run build` are working, which means that the docker bits should still be working since it's just a node image, which I believe is just an ubuntu image.

## Help Needed

Here's where I need help:

I don't understand how the svelte environment system works and it seems to behave differently between `adaptor-node` and `adapter-static` as well as between `npm run dev` and `npm run build`.

This is what _should_ happen:

```sh
# this should build from `src/lib/constants.ts`
npm run build
```

```sh
# this should use the ENV
API_SERVER=https://localhost/api npm run build
```

```sh
# this should have access to `${location.protocol}` and `${location.host}`
npm run dev
```

I'm not sure how to get that behavior. That needs someone with Svelte expertise, which I absolutely am not.

I'm very familiar with Linux and Node and JavaScript, but this is doing proprietary things, not following any normal conventions. (I believe Svelte is is completely its own language that just happens to use JS syntax)

tjbck:
@coolaj86 

I've just pushed a few commits and attempted to align the behaviour as per your requirements. If there are any discrepancies or if I've misunderstood any part of your request, please don't hesitate to let me know.

tjbck:
Forgot to mention, to build with the environment variable, please run:

```bash
OLLAMA_API_ENDPOINT="http://[Your Ollama URL]/api" npm run build
```

This should run with the specified environment variable. Let me know if you have any questions.

.gitignore
```diff
@@ -3,7 +3,6 @@ node_modules
 /build
 /.svelte-kit
 /package
-.env
```
coolaj86, line 6: This file should be named example.env and referenced with a flag (typically `-r ./example.env), otherwise a person can't ignore their real .env from a copy of the repo.

README.md
```diff
@@ -7,15 +7,25 @@ ChatGPT-Style Web Interface for Ollama 🦙
...
-- 📤📥 **Import/Export Chat History**: Seamlessly move your chat data in and out of the platform.
+
+- 📤📥 **Import/Export Chat History**: Seamlessly move your chat data in and out of the platform.\
+
 - ⚙️ **Fine-Tuned Control with Advanced Parameters**: Gain a deeper level of control by adjusting parameters such as temperature and defining your system prompts to tailor the conversation to your specific preferences and needs.
+
 - 💻 **Code Syntax Highlighting**: Enjoy enhanced code readability with our syntax highlighting feature.
-- 🔗 **External Ollama Server Connection**: Link to the model when Ollama is hosted on a different server via the environment variable -e OLLAMA_ENDPOINT="http://[insert your Ollama address]".
+
+- 🔗 **External Ollama Server Connection**: You can seamlessly connect to an external Ollama server hosted on a different address by setting the environment variable during the Docker build process. Execute the following command to include the Ollama API endpoint in the Docker image: `docker build --build-arg OLLAMA_API_ENDPOINT="http://[Your Ollama URL]/api" -t ollama-webui .`.
```
coolaj86, line 27: My understanding is that there is no Ollama API, but rather Ollama uses the ChatGPT API, so calling it OLLAMA may be a bit of a misnomer.

That said, I may be wrong, and I could see a future in which OLLAMA begins to add its own features to create a separate API (maybe that's already the case with /api/tags? Or maybe that's how GPT distinguishes between 3, 3.5, 4, etc too?)

tjbck: I'm uncertain about your point regarding Ollama's use of the ChatGPT API. Could you please provide further clarification? Ollama functions locally and relies on API calls to access the downloaded models.

.gitignore
tjbck: Just renamed the current .env file to .env.local, had to include .env.local to prevent any conflicts with the npm run dev and npm run build commands. 

Let me know if it still causes an issue on your end. Additionally, if you have any suggestions on how we can handle this more effectively, I'd be happy to incorporate them.

tjbck: On second thought, It's probably safer to remove all .env files to prevent future conflicts. Renamed .env.local to .env.example.

tjbck:
@coolaj86 Is everything working as it should? everything lgtm! Will merge this branch once you confirm everything is working as intended for you.

coolaj86:
I'm just now taking a look at this again.

README.md
coolaj86: My point is that it's like S3. Whether you're using Amazon or Digital Ocean or Minio, the ENVs are always named according to S3 because S3 is a protocol more than it is a specific product.

```sh
S3_REGION=...
S3_BUCKET=...
S3_PREFIX=...
S3_KEY=...
```
It would be _more_ confusing to do something like

```sh
DIGITAL_OCEAN_BUCKET=...
```

Because everyone just knows it as "S3" regardless of who is actually providing the service.

Likewise, the GPT is generic. Lots and lots of tools are being built that either provide as a service or consume as a client the GPT API.

So naming it `OLLAMA_API_ENDPOINT` is actually confusing - it makes it sound like it's NOT compatible with GPT, but actually it is.

So if someone had another client product that works with GPT and they wanted to try it with ollama, it would be much more consistent for it to either not carry the protocol name at all `API_ENDPOINT`, or to carry the generic protocol name `GPT_API_ENDPOINT`.

coolaj86: I had seen is some of the ollama or mistral documentation that they use the GPT API, but in looking at https://platform.openai.com/docs/api-reference/fine-tuning, it doesn't appear to be the case from what I can see.

tjbck: I see what you mean but I think you might be misunderstanding what Ollama is offering. FYI, Ollama does **_not_** make any requests to proprietary OpenAI GPT-related APIs and everything is processed on a local machine, hence needing to download LLMs locally beforehand (it can also be operated offline!). Additionally, Ollama WebUI only supports Ollama APIs at the moment, so I believe it's appropriate to use the name "OLLAMA_API_ENDPOINT" for the env variable. If you have any other questions regarding Ollama to help you understand how it works, feel free to ask, I'll do my best to answer to the best of my ability! Thanks for the explanation though!

timbk: @tjbck your discord handle appears to be my GitHub account name. If you tag it with the @ I automatically get notifications for this whole discussion. Just wanted to make you aware and commenting here was the only option I saw.
Cool project btw and have a nice Sunday guys!

tjbck: Oops, so sorry 😅. Thanks for your understanding. Have a good day as well!

coolaj86:
Okay, I believe I've adequately tested and documented everything.

I also changed `*_API_ENDPOINT` to `*_API_BASE_URL` to match the common convention, which is also what's in the svelt documentation.

README.md
coolaj86: I know it doesn't make requests to OpenAI's GPT, but I was under the impression that it used the same API signatures.

After checking the documentation and API calls, it looks like that was not correct, or I misunderstood.

Agreed, `OLLAMA_*` is the best way to go.

coolaj86:
One last thing I didn't test or check:
- if `API_BASE_URL=''` (empty) it should default to `http://localhost:11434`
- if `API_BASE_URL='/api'` (path only) then it should be relative to the current scheme/host/port

Then the main release could be built for `''` and embedded releases could be built for `'/api'`

It would also be nice (in another iteration) to have a setting to change it at runtime.

src/lib/constants.ts
```diff
@@ -6,6 +6,8 @@ export const API_BASE_URL =
 		? browser
 			? `http://${location.hostname}:11434/api`
 			: `http://localhost:11434/api`
+		: PUBLIC_API_BASE_URL === '/api'
```
coolaj86, line 9: this is the default browser behavior and should not be set

whether the user sets `/api` or `/ollama-api` or whatever else, it already should and does work with the previous behavior

coolaj86: Also, `.origin` won't always exist - such as in private browsing mode, IIRC

tjbck: Agreed, I'll revert the commit and merge this branch to main! Thanks!

README.md
```diff
@@ -48,9 +48,10 @@ OLLAMA_HOST=0.0.0.0 OLLAMA_ORIGINS=* ollama serve
 
 ### Using Docker 🐳
 
-```bash
-docker build --build-arg OLLAMA_API_BASE_URL='http://localhost:11434/api' -t ollama-webui .
```
coolaj86: Better to be explicit rather than implicit, so that users have to reference less documentation.

It _looks_ "cool" to copy and paste fewer characters, but it's much more frustrating to have to reference documentation for things that aren't intuitive.

That said, now that we know that `''` will work correctly, showing `OLLAMA_API_BASE_URL=''` would both be fewer characters and still be mostly self-documenting.

src/lib/constants.ts
coolaj86: If you revert this change, I think it's all good to go. I did test it, apparently at the same time you were making these changes.

README.md
tjbck: Agreed!

coolaj86:
I just rebased this on the current `main` and squashed both of our most recent doc commits into one.

I think this is ready for your final review.

tjbck:
LGTM! Great work, Thanks a lot!

PR #13, opened 2023-10-22
chore: add fmt to github actions

coolaj86:
I noticed that the other PR had whitespace changes in it, which probably came from my editor being incorrect, or a mismatch of `prettier@3` on your local machine and `prettier@2` in the repo.

I didn't update this to `prettier@3` because it's pretty new (less than 3 months, I think) I don't know if v3 is compatible with all the svelte plugins.

Anyway, you can set it such that PRs need to pass lint and fmt before they can be rebased into main.

> [x] Require status checks to pass before merging

I was going to add the lint and type checking as well, but there many type and lint errors that would need to be fixed first.

https://github.com/ollama-webui/ollama-webui/branch_protection_rules/

Screenshot 2023-10-21 at 10 39 01 PM


tjbck:
LGTM! Thanks!

Issue #15, opened 2023-10-22
Differents images same container


loranger:
Hi,

I try to run ollama and (this awesome) ollama-webui on the same container using docker compose, but I fail.
The aim is to run ollama as an API using a dedicated hostname routed by the traefik reverse proxy and an ui on another dedicated hostname also routed by traefik.

Here is my env file. `*.docker` is now resolved locally on 127.0.0.1 since I would like to make it work before using a real public hostname and tld.

<details>
<summary>.env</summary>

```
APP_PROJECT=ollama
APP_DOMAIN=ollama.docker

OLLAMA_HOST=0.0.0.0
OLLAMA_ORIGINS=*
OLLAMA_ENDPOINT="http://api.${APP_DOMAIN}"
```

</details>

And here is my docker-compose file. The project folder container an ollama folder (where are stored ssh keys and models) and an ollama-webui which is a clone of this repository.

<details>
<summary>docker-compose.yml</summary>

```yaml
version: '3'
services:

  ollama:
    container_name: ${APP_PROJECT}-api
    hostname: ${APP_PROJECT}-api
    image: ollama/ollama
    env_file:
      - .env
    volumes:
      - ./ollama:/root/.ollama
    command: serve
    entrypoint: ['ollama']
    labels:
      - "traefik.http.routers.${APP_PROJECT}.rule=Host(`api.${APP_DOMAIN}`)"
      - "traefik.http.services.${APP_PROJECT}-service.loadbalancer.server.port=11434"

  ollama-webui:
    container_name: ${APP_PROJECT}-webui
    image: ${APP_PROJECT}-webui
    build:
      context: ./ollama-webui/
      dockerfile: Dockerfile
    env_file:
      - .env
    labels:
      - "traefik.http.routers.${APP_PROJECT}-webui.rule=Host(`${APP_DOMAIN}`)"
      - "traefik.http.services.${APP_PROJECT}-webui-service.loadbalancer.server.port=3000"

networks:
  default:
      name: traefik-network
      external: true

```

</details>

Everything seems to run as I get a response from ollama:

```shell
$ curl http://api.ollama.docker/
Ollama is running
$ curl http://api.ollama.docker/api/tags
[...]

Issue #2, opened 2023-10-17
403 OPTIONS  "/api/generate"


Chillance:
Hi!

Just tested this and noticed it does a  OPTIONS  "/api/generate" request. From what I can tell this doesn't exist in latest ollama code...

Thoughts?

tjbck:
Hi there,

Browsers often send an OPTIONS request to verify CORS, which might be the case here. Were there any other issues caused by the OPTIONS request?

Thanks.

Chillance:
Well, I get that asking something trying to generate something, but it seems to stop there so nothing else happens.

tjbck:
Could you please provide any specific error messages or logs that you encounter during the process? This information would greatly assist me in diagnosing the problem more accurately and providing you with the appropriate guidance.

Thanks.

Chillance:
How do I get better logs? I tried typing something, pressed enter, and then nothing.

tjbck:
Hi there,

To obtain more detailed logs, you can use the following command for Docker:

```bash
docker logs ollama-webui
```

Additionally, it would be helpful for diagnosing your issue if you could provide a screenshot of your console logs from your browser's developer tools. This will allow us to examine any client-side errors or issues that might not be visible in the server logs.

Please feel free to share the Docker logs and the browser console logs screenshot, and we'll do our best to assist you in resolving the problem.

Thanks.

coolaj86:
@Chillance Part of the reason that browsers invented CORS Preflight Requests (the OPTIONS issue) is to prevent people from stumbling into security issues.

Do you already have the access protected with an API token or HTTP Basic Auth?

Check out https://github.com/jmorganca/ollama/issues/849#issuecomment-1773697189 and the CORS section at https://webinstall.dev/caddy.

coolaj86:
I have an example of how I did it in #10

Chillance:
I don't do anything particular but just staring the ollama serve. And, on the same machine I run this in docker.

`docker run --network=host -p 3000:3000 --name ollama-webui --restart always ollama-webui`

I actually  got chatbot-ollama (other repo) working fine. But here I can see this in the console log:

> 
> e87e0c1f-4d67-4015-959a-0e2b59659483
> 2.fb1b6367.js:52 submitPrompt
[...]

PR #20, opened 2023-10-22
WIP: more docker-friendly proxying

coolaj86:
Just added an example that uses `http://*:8080` as the host that could be used to solve some of the issues of #2  and #15.

It would probably be used something like this:

```Dockerfile
COPY Caddyfile.docker Caddyfile
START caddy run --config ./Caddyfile
```

Then it would:
- serve the files with CORS from any internal / private domain
- hit the API container through its own proxy
- without the need for internal CORS between Docker containers

TODO:
- [ ] use `reverse_proxy {$OLLAMA_API_HOST}` in place of `reverse_proxy localhost:11434`


tjbck:
Seems like this would be ideal for https://github.com/ollama-webui/ollama-webui-lite now (stripped down version of ollama-webui without the backend part), I'll think of ways to migrate everything to the other repo. Thanks!

PR #21, opened 2023-10-22
feat: pull/delete model

tjbck:


tjbck:
image

Pull/Delete model features been added

PR #94, opened 2023-11-11
feat: ui improvement

tjbck:


brogan-wilding:
## Review Comment Responses

All four comments addressed in 655e50b:

**1. Dev Tika URL (development.tfvars:29)** — Fixed. Updated FQDN to `thankfulstone-90e0563b` (confirmed via `az containerapp show`), added port `:9998`, and corrected the comment from "internal" to "external" to match the actual ingress config.

**2. KV access policy bootstrap (main.tf:180)** — Added documentation block explaining the chicken-and-egg situation. Since we're importing existing container apps that already have KV access policies, this is a non-issue for our case. Added a note that net-new environments would need a two-phase apply or pre-provisioned user-assigned identity.

**3. ENABLE_PERSISTENT_CONFIG hardcoded (main.tf:46)** — Converted to a variable (`enable_persistent_config`) that defaults to `"true"` (current live behavior). This prevents an accidental behavior change during initial import/apply. Operators explicitly set it to `"false"` in tfvars only after the critical fixes (Tika URL, KV cross-refs, plaintext secrets) are resolved.

**4. OneDrive string vs bool (main.tf:159)** — Converted `enable_onedrive_integration` from `string` to `bool` type. The condition now uses native bool comparison instead of case-sensitive string matching. Updated the dev tfvars value from `"true"` to `true`.

brogan-wilding:
## Round 2 Review Responses

Addressed in 95ee476:

**1. `enable_websocket_support` as bool (variables.tf:250)** — Intentionally left as string. The live dev value is `"True"` (capital T). Converting to `bool` would produce `"true"` (lowercase), causing unnecessary drift in `terraform plan`. Open WebUI reads this as a raw string env var, so it's a passthrough — Terraform shouldn't opinionated about casing.

**2. Ingress validation message (variables.tf:96)** — Fixed. Error message now lists all five accepted values.

**3. `enable_persistent_config` as bool (variables.tf:108)** — Fixed. Converted to `bool` (defaults `true`), uses `tostring()` when building the env var.

**4. `kv_secrets` description (variables.tf:262)** — Fixed. Clarified the mapping direction: key = CA secret name, value = KV secret name.

**5. `enable_direct_connections` as bool (variables.tf:187)** — Fixed. Converted to `bool` (defaults `false`), uses `tostring()` when building the env var. Updated both tfvars to use native bool values.

PR #165, opened 2023-11-30
Adding num_ctx to the model settings page

AnthonyCucci:
I have added an option to change the default context length of 2048 up to 16000!

tested on my local and it was pulling proper context lengths.


src/lib/components/chat/SettingsModal.svelte
```diff
@@ -30,6 +30,7 @@
 	let repeat_penalty = 1.1;
 	let top_k = 40;
 	let top_p = 0.9;
+	let num_ctx = 2048;
```
tjbck, line 33: Thanks for the PR, is the value of num_ctx 2048 the default for most models?

src/lib/components/chat/SettingsModal.svelte
```diff
@@ -732,8 +734,24 @@
...
+								<label for="steps-range" class=" mb-2 text-sm font-medium flex justify-between">
+									<div>Context Length</div>
+									<div>
+										{num_ctx}
+									</div></label
+								>
+								<input
+									id="steps-range"
+									type="range"
+									min="0"
```
tjbck, line 747: Are there better values for min/max? Min of 0 doesn't make all that much sense to me.

AnthonyCucci: I guess it just depends on how much previous context someone wants during a chat

tjbck: The min value for OpenAI Playground value seems to be 1. Once you update the min value to 1, I'll try to run it on couple of my machines then merge to main, Thanks!

AnthonyCucci: Updated

tjbck:
LGTM, Thanks!

PR #167, opened 2023-11-30
feat: add PWA manifest.json

### LINKED ISSUE #150: feat: PWA support
Currently the app can't be installed as a PWA which could make it more feel like a native application.

Make it possible to install the app as an PWA.

**Additional context**
https://blog.logrocket.com/how-to-build-a-progressive-web-app-pwa-with-node-js/

ntimo:
This MR adds a super basic manifest.json for pwa support.

Closes: #150

### DIFF
--- a/static/manifest.json
+++ b/static/manifest.json
@@ -1,0 +1,16 @@
+{
+  "name": "Ollama Web UI",
+  "short_name": "Ollama",
+  "start_url": "/",
+  "display": "standalone",
+  "background_color": "#343541",
+  "theme_color": "#343541",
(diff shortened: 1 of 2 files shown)


static/manifest.json
tjbck, line 1: Thanks for the PR! Questions I have for this is, how does this handle updating and does it reflect right away when the webui gets updated?

hammeiam: afik this doesn't have a service worker, so no requests / resources are cached. Ergo this just makes the webpage look like an app visually.

PR #196, opened 2023-12-09
feat: horizontal scroll for code block

### LINKED ISSUE #194: Horizontal scrolling for code blocks
**Horizontal scrolling for code blocks**
The line wrapping in the code blocks makes it very difficult to read Python code in place.   I am not sure this is the only way to handle the issue. It is just what the GPT-4 web-ui does.    
Personally, it feels as if the margins on either side of the message area could be reduced to allow a bit more growth on larger screens. i.e. I have a 32’ screen, which I don’t get any benefit from. I know the pattern is normal for dynamic content centred on the page.  While the proportions seem ok for chat, in my opinion it does not work that well when you have code output.

tjbck:
Resolves  #194

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -37,8 +37,8 @@ math {
 }
 
 ::-webkit-scrollbar {
-	height: 1rem;
-	width: 0.5rem;
+	height: 0.35rem;
+	width: 0.35rem;
 }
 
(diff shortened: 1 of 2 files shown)


Phuket2:
Thank you

PR #205, opened 2023-12-11
feat: dropdown for deleting models

samuelskottenborg:
Added a dropdown menu that shows all the local models available for deletion along with their disk size.

Screenshot from 2023-12-11 17-23-13


tjbck:
LGTM, Thanks!

PR #215, opened 2023-12-14
WIP: chat enhancements

tjbck:
Resolves #188 #131 #141

- Adds general enhancements to chat as well as chat share feature

TODO

- [x] Clipboard image paste support
- [x]  Response Notification using browser Notification API

image

- [x] Notify user when the model has been downloaded via Notification API
 
image

- [x] Ollama version check
- [x] Reponse auto copy to clipboard
- [x] Share chat w/ OllamaHub

PR #216, opened 2023-12-14
feat: full backend support (including auth/rbac)

### LINKED ISSUE #68: feat: role-based access control w/ multi-user support
**Description:**
I'd like to request a user system, with simply antonymous as default, I like to have ui's like this web facing so i may use this on the go, or maybe even share it with friends, however no barrier for entry this allows anyone who discovers the url to spam requests, so I wish for a user system with a disable-able anonymous default, to not intrude on users who do not with to partake in this feature.

**Alternatives:**
A simple password would do fine for the initial problem, but would not be preferable.

### LINKED ISSUE #281: feat: copy code
This is a minor feature request but it is a feature the Refact.ai guys put in their Jetbrains client that I found extremely useful. 

All of the code segment boxes have a small copy to clipboard button in the top right had corner of the code window and it copies the contents to the clipboard. 

I've seen the feature on a lot of websites so i'm guessing the behaviour is well supported by web browsers or at least easy to implement. It makes using any generated code segements and getting them into an editor very straight forward, and those are the sort of segments you want to cut and paste typically. 

This seems like a straight forward feature, and it would greatly enhance the utility of the web-ui when asking it coding style questions.

tjbck:
Resolves #68, #281

Auth (backend) will be **REQUIRED** from now on, even for local deployment, just like JupyterLab.

For those who wish to use frontend only for the UI, please check out our stripped down version of ollama-webui forked repo: https://github.com/ollama-webui/ollama-webui-lite

TODO

- [x] migrate DB to sqlite
- [x] require Auth for security
- [x] full multi-user chat history support
- [x] overall code refac
- [x] local indexeddb -> backend db migration guide
- [x] save modelfiles to the backend
- [x] local modelfiles -> backend migration option
- [x] disable modelfiles edit for users
- [x] update backend error page
- [x] whitelist ollama version '0.0.0' (#263)
- [x] doc update
- [x] testing
- [x] **HELP NEEDED**: better doc?

Blockers

- #215


backend/apps/web/models/auths.py
```diff
@@ -86,17 +97,24 @@ def insert_new_auth(
...
-            if verify_password(password, auth["password"]):
-                user = self.db.users.find_one({"id": auth["id"]})
-                return UserModel(**user)
+        print("authenticate_user", email)
+        try:
+            auth = Auth.get(Auth.email == email, Auth.active == True)
+            print(auth.email)
+
+            if auth:
+                print(password, str(auth.password))
```
anuraagdjain, line 106: thoughts: Should we avoid logging or only log them in debug mode, at least for the passwords?

tjbck: Good catch, forgot about it 😅 It's been removed!

backend/apps/web/models/auths.py
```diff
@@ -12,15 +13,23 @@
...
-DB = config.DB
+from apps.web.internal.db import DB
 
 ####################
 # DB MODEL
 ####################
 
 
+class Auth(Model):
+    id = CharField(unique=True)
```
anuraagdjain, line 24: question: we're saving uuid in the `id` columns that's why its CharField?

backend/apps/web/routers/auths.py
```diff
@@ -104,8 +104,8 @@ async def signup(form_data: SignupForm):
                     "profile_image_url": user.profile_image_url,
                 }
             else:
-                raise HTTPException(500, detail=ERROR_MESSAGES.DEFAULT(err))
+                raise HTTPException(500, detail=ERROR_MESSAGES.DEFAULT())
```
anuraagdjain, line 107: suggestion: I'd log the `err` , might be useful for debugging.

backend/apps/web/models/auths.py
tjbck: Yep, basically uuid casted as a string!

backend/apps/web/routers/auths.py
tjbck: Sounds good!

backend/apps/web/routers/chats.py
```diff
@@ -0,0 +1,161 @@
...
+    user = Users.get_user_by_token(token)
+
+    if user:
+        chat = Chats.get_chat_by_id_and_user_id(id, user.id)
+
+        if chat:
+            return ChatResponse(**{**chat.model_dump(), "chat": json.loads(chat.chat)})
+        else:
+            raise HTTPException(
+                status_code=status.HTTP_401_UNAUTHORIZED,
```
anuraagdjain, line 105: nit (minor): the status code could be 500 and not 401, as `chat` would be None when fetching the db if user-A tried to access chat-3 which belonged to user-B.

tjbck: Hmm, I believe 401 would be more appropriate in this situation but thanks for your feedback regardless! If anything we could change it to 500 later down the line.

tjbck:
@anuraagdjain, any issues with the migration or setup?

0xb15d3:
Everything worked seamlessly for me, Good work!

tjbck:
Alright, Merging it to main! Thanks guys!

PR #246, opened 2023-12-19
Refactor docker-compose configuration for modularity

gnscc:
Split the original docker-compose.yml into three separate files for enhanced modularity and ease of use. Created docker-compose.api.yml for API exposure configuration and docker-compose.gpu.yml for GPU support. This change simplifies the management of different deployment environments and configurations, making it easier to enable or disable specific features such as GPU support and API access without modifying the main docker-compose file.

Additionally, this separation ensures that personal or environment-specific configurations do not affect the main docker-compose file, thus avoiding unnecessary changes in Git tracking. This means that switching between different configurations (e.g., with or without GPU support) won't lead to modifications in tracked files, keeping the Git history cleaner and more focused on meaningful changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,14 @@ If you don't have Ollama installed yet, you can use the provided Docker Compose 
 docker compose up -d --build
 ```
 
-This command will install both Ollama and Ollama Web UI on your system. Ensure to modify the `compose.yaml` file for GPU support and Exposing Ollama API outside the container stack if needed.
+This command will install both Ollama and Ollama Web UI on your system. 
+Enable GPU support or Exposing Ollama API outside the container stack with the following command:
+```bash
+docker compose -f docker-compose.yml \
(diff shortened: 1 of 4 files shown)


moDal7:
As a recent user, while trying to do this I also thought it could be better to split the docker .yaml file this way.

justinh-rahb:
mustgofaster
Screenshot 2023-12-23 at 12 50 08 AM

tjbck:
Hi @gnscc, Thanks for your PR! I think we should have a separate section for enabling gpu and another section for exposing api in the doc for better clarity, but otherwise LGTM!

gnscc:
Hey @tjbck, thank you for your feedback. Following your suggestion I have updated the README with a section for each option.

tjbck:
LGTM, Thanks a lot!

PR #253, opened 2023-12-20
Update TROUBLESHOOTING.md

YasushikoX:
Removed confusion with --network=host, as it did not specify that the port would be changed.


tjbck:
LGTM, Thanks!

PR #268, opened 2023-12-23
fix: decode the message to render them onto UI

anuraagdjain:
Fixes this bug https://github.com/ollama-webui/ollama-webui/issues/267. The received response contains raw characters which we have to decode to convert "\u003c" to "<". 

```json
{
    "model": "deepseek-coder:6.7b",
    "created_at": "2023-12-23T14:16:01.764806Z",
    "message": {
        "role": "assistant",
        "content": "\u003c"
    },
    "done": false
}
```

**Before**
image

**After**

image

### DIFF
--- a/src/routes/(app)/+page.svelte
+++ b/src/routes/(app)/+page.svelte
@@ -240,7 +240,7 @@
 								if (responseMessage.content == '' && data.message.content == '\n') {
 									continue;
 								} else {
-									responseMessage.content += data.message.content;
+									responseMessage.content += decodeURIComponent(data.message.content);
 									messages = messages;
 								}
 							} else {


tjbck:
Hi, Thanks for the PR! Could you also verify for us that the result of the second screen shot wasn't just because of a different response? Setting the seed value would be helpful in this situation, keep us updated!

anuraagdjain:
**Seed settings**
Screenshot 2023-12-23 at 8 45 09 PM


**First attempt**
image


**Second attempt, by clicking on regenrate**
image


**New prompt for twitter**
Screenshot 2023-12-23 at 8 47 04 PM


@tjbck - Can you confirm if this what you were looking for?

tjbck:
Hi! I meant if you could make a comparison with the current version and this branch, that would verify it's working as it should. Thanks!

tjbck:
Just manage to check it for myself, doesn't seem to resolve the issue :/  
But thanks a lot for your effort, it hasn't gone unnoticed!

image

@oliverbob @anuraagdjain, I believe the issue stems from the local models not properly generating formatted html tag blocks (not using three backticks), and there isn't much we can do from the webui part. We could try to detect incorrectly formatted code blocks from our end but I'm not sure if that's the right way and even if we do implement the fix, we would have an extremely convoluted solution, so I'll be closing this issue. However, if you manage find a simple fix, feel free to make a PR again! Thanks a lot!

anuraagdjain:
You are right, I am unable to replicate this anymore with mistral or deepseek-coder models. I'm wondering if we should add hacky solution to add backticks if missing. Since the user can choose any model, it might be better for them to specify it back to the model to regenerate in plain text.

tjbck:
I think the hacky method probably is the best solution we have at the moment, I'm just very cautious to implement it as there might be other edge-cases where it will fail to function properly. Let's continue our discussion here: https://github.com/ollama-webui/ollama-webui/pull/268

PR #308, opened 2023-12-28
fix(model-selector): prevent saving empty default model

anuraagdjain:
- throw a toast error when user is trying to click on `set as default` without actually choosing the model(s).
- feat: added hmr for local development. use `npm run dev` and you get page reload + backend functionality without needing to build every time for small changes. 

https://github.com/ollama-webui/ollama-webui/assets/4705103/e72ea83a-4854-4439-9bf3-c325c51ab820

### DIFF
--- a/src/lib/components/chat/ModelSelector.svelte
+++ b/src/lib/components/chat/ModelSelector.svelte
@@ -6,6 +6,11 @@
 	export let disabled = false;
 
 	const saveDefaultModel = () => {
+		const hasEmptyModel = selectedModels.filter(it => it === '');
+		if(hasEmptyModel.length){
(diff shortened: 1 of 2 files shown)


tjbck:
Thanks for your PR and your amazing work! everything LGTM, but I believe hmr should be turned on by default? was that not the case for you?

anuraagdjain:
> Thanks for your PR and your amazing work! everything LGTM, but I believe hmr should be turned on by default? was that not the case for you?

Did a clean install again and it worked without explicitly needing the `hmr` in vite.config, so reverted that change.

tjbck:
LGTM, Thanks!

PR #311, opened 2023-12-28
feat(auth): add auth middleware

anuraagdjain:
refac: use dependencies to verify token

- feat: added new util to get the current user when needed. Middleware was adding authentication logic to all the routes. let's revisit if we can move the non-auth endpoints to a separate route. 
- refac: update the routes to use new helpers for verification and retrieving user
- chore: added black for local formatting of py code


anuraagdjain:
@tjbck  - I appreciate your review on the PR. I want to try fastapi and hence I am taking minor improvement tasks which could help me.

backend/.gitignore
```diff
@@ -4,4 +4,5 @@ _old
 uploads
 .ipynb_checkpoints
 *.db
-_test
\ No newline at end of file
+_test
+Pipfile
```
anuraagdjain, line 8: I'm using `pipenv shell` and that created the `Pipfile`.

backend/apps/web/middlewares/auth.py
```diff
@@ -0,0 +1,27 @@
...
+from fastapi import Request, status
+from starlette.authentication import (
+    AuthCredentials, AuthenticationBackend, AuthenticationError, 
+)
+from starlette.requests import HTTPConnection
+from utils.utils import verify_token
+from starlette.responses import JSONResponse
+from constants import ERROR_MESSAGES
+
+class BearerTokenAuthBackend(AuthenticationBackend):
```
anuraagdjain, line 11: source https://www.starlette.io/authentication/#custom-authentication-error-responses

anuraagdjain:
@tjbck  - verified models, users, auth, chat and the non-authenticated endpoints. everything seems to be working fine for me. Would appreciate your review and merge it if no improvements.

tjbck:
Taking a look!

backend/utils/utils.py
```diff
@@ -53,16 +55,23 @@ def extract_token_from_auth_header(auth_header: str):
...
+    else:
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail=ERROR_MESSAGES.UNAUTHORIZED,
+        )
+
+
+def get_current_user(auth_token: HTTPAuthorizationCredentials = Depends(HTTPBearer())):
+    data = decode_token(auth_token.credentials)
+    return Users.get_user_by_email(data["email"])
```
tjbck, line 77: I believe it would be better to merge `verify_auth_token` and `get_current_user`, so that we don't need additional dependencies for `/chat`, `/users`, and `/modelfiles` api routes. Any reasoning behind the separation of the two? Seems like to me that there isn't any reason for the backend to make the same user query to db twice. Here's my suggestion:


```suggestion
def get_current_user(auth_token: HTTPAuthorizationCredentials = Depends(HTTPBearer())):
    data = decode_token(auth_token.credentials)

    if data != None and "email" in data:
        user = Users.get_user_by_email(data["email"])
        if user:
            return user
        else:
            raise HTTPException(
                status_code=status.HTTP_401_UNAUTHORIZED,
                detail=ERROR_MESSAGES.INVALID_TOKEN,
            )
    else:
        raise HTTPException(
            status_code=status.HTTP_401_UNAUTHORIZED,
            detail=ERROR_MESSAGES.UNAUTHORIZED,
        )
```

Thanks!

backend/apps/web/main.py
```diff
@@ -16,13 +17,26 @@
...
+    prefix="/chats",
+    tags=["chats"],
+    dependencies=[Depends(verify_auth_token)],
+)
+app.include_router(
+    modelfiles.router,
+    prefix="/modelfiles",
+    tags=["modelfiles"],
+    dependencies=[Depends(verify_auth_token)],
+)
```
tjbck, line 39: Refer to previous comment!

backend/utils/utils.py
anuraagdjain:
```python
app.include_router(
    users.router,
    prefix="/users",
    tags=["users"],
    dependencies=[Depends(get_current_user)],
)
```

If I do the above and remove the second function, then in my actual route I don't have access to the user. I would have to refetch it anyways, as `Depends` doesn't inject the value into `request` object. We could do `request.state.user = user # fetch user from db` in the util function, but that becomes implicit. I don't think a second call to DB to fetch the user by id is going to affect much when everything is local.

tjbck: I meant getting rid of `dependencies` like such:

```python
app.include_router(
    users.router,
    prefix="/users",
    tags=["users"]
)
```

and leaving the router functions as they are!

anuraagdjain: Hmm, I thought having token validation at the base route as above would be better and less copy-paste of the same dependency on users.router when adding new endpoints. I can make the change as suggested in the original comment.

tjbck: Yep, I believe that would be best, as we don't know what we'll be adding to those api routes in the future 😅 We can always come back to this and adopt your approach if necessary, Thanks!

anuraagdjain: Pushed the changes as discussed.

tjbck:
LGTM, Thanks a lot for all the hard work and Happy New Year!

anuraagdjain:
> LGTM, Thanks a lot for all the hard work and Happy New Year!

Happy new year 🥳 to you too

PR #275, opened 2023-12-24
Docker compose modularity improvements + kubernetes support

dnviti:
As per pull request: [Refactor docker-compose configuration for modularity](https://github.com/ollama-webui/ollama-webui/pull/246).

- Improved docker compose overrides specialization.
- Added an installation script with complete functionalities.
- Updated README.md accordingly
- Fixed .gitignore which was too generic.
- Added support for kubernetes (kustomize and helm)

Note: the script "run-compose.sh" contains a function (for linux) to automatically detect the GPU installed on the host system, i'm not be able to test it on AMD and Intel GPUs cause i do not have any. On Nvidia works flawlessly.

PR #317, opened 2023-12-29
Docker compose fix

Morethanevil:
Fixed docker-compose.yml, Built tag and image tag may not be enabled at the same time, I included the other two files too for easier deployment


dnviti:
Thank you for the pull request and for working at this amazing project!
I think your codebase is not updated with latest changes on repo.
Sorry can you please explain why the deletion of the gpu and api override?
We are working on removing the commented sections to better modulate compose files, please follow #275

Morethanevil:
I just saw the current version of the docker-compose.yml in main and it will not work this way, because image tag and build tag are used at the same time. So I fixed that and in this step I wanted to include gpu support and ollama api expose from the other both files. One docker-compose.yml which can be configured as needed. That was my intention :)

dnviti:
Main is not yet updated with latest commits, we are working on dev, you'll not see much differences though as the #275 is currently under review.
You are welcome to check #275 about latest docker and kubernetes updates and continue the review on that PR.
Many thanks for working on the project.

PR #357, opened 2024-01-03
feat: set default models globally

### LINKED ISSUE #266: How do you set/load the first default model as default?
When the UI loads, users expect to be able to chat directly (just like in Chat GPT), coz it is annoying to receive a "Model not selected" message on first impression chat experience.

What file will I need to modify to preselect orca-mini:3b for instance?

Can it be modefied in ModelSelector.svelte?

I tried <option class=" text-gray-700" value="orca-mini:3b" selected>Select a model</option>

But this only adds it to the list and still have to be selected.

Your help is appreciated.

tjbck:
Resolves #266

### DIFF
--- a/src/lib/apis/configs/index.ts
+++ b/src/lib/apis/configs/index.ts
@@ -4,7 +4,7 @@ export const setDefaultModels = async (token: string, models: string) => {
 	let error = null;
 
 	const res = await fetch(`${WEBUI_API_BASE_URL}/configs/default/models`, {
-		method: 'GET',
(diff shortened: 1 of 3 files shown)


avi7611:
Hey, can anyone tell me how can I actually set a global model? do I need to set an environment variable while running the container or what ?

HCharlie:
> Hey, can anyone tell me how can I actually set a global model? do I need to set an environment variable while running the container or what ?

open the ui and search for text `set as default`

PR #371, opened 2024-01-03
fix: :hammer: start script permissions & environment

ThatOneCalculator:
Fixes start.sh to have execution permission & to use script env


tjbck:
LGTM, Thanks!

PR #407, opened 2024-01-06
feat: parallel model downloads

### LINKED ISSUE #221: feat: enable concurrent model downloads
**Describe the bug**
Right now I can download multiple models at the same time and end up seeing sha hashes collisions in the UI.

**To Reproduce**
You can just type in 2 different models and import at the same time

**Expected behavior**
parallel model imports should show up on different lines along with a label of the model name + download progress bars for each

**Screenshots**
Screenshot 2023-12-14 at 5 39 01 PM

**Desktop (please complete the following information):**
 - OS: Mac M1
 - Browser Brave

**Additional context**
N/A

anuraagdjain:
- Allow downloading multiple models in parallel, addresses https://github.com/ollama-webui/ollama-webui/issues/221
- Currently the queue is limited to 3 downloads in parallel. 
- Refactored the pullModel return type.
- Fixed the progress bar color in Light mode, it was non-existent and made it harder to read the progress %. 

https://github.com/ollama-webui/ollama-webui/assets/4705103/3a913c25-848c-435c-89ba-e0426ce17c6e


src/lib/apis/ollama/index.ts
```diff
@@ -249,8 +249,7 @@ export const deleteModel = async (token: string, tagName: string) => {
 };
 
 export const pullModel = async (token: string, tagName: string) => {
```
anuraagdjain, line 251: _Note:_ Refactored this because `res` would never be `null` because we explicitly assign `err` to `error` and then throw it from line 269. With the refactor, the function's return type has changed from `Promise<Response | null>` to `Promise<Response>` making it easier to read / use and no need `if` to check if `res` is valid.

tjbck:
Wow, amazing work! I'll change the design a bit and we can merge, Thanks for the hard work!

src/lib/apis/ollama/index.ts
tjbck: Will be reverting to the old code, as it should also catch and raise error when the backend proxy can't access Ollama.

tjbck: image

Updated to display this error message

src/lib/components/chat/SettingsModal.svelte
```diff
@@ -330,11 +335,10 @@
 
 				if (!data.success) {
 					toast.error(data.error);
-					return;
+				} else {
+					toast.success(`Model ${modelName} was successfully downloaded`);
```
anuraagdjain, line 339: suggestion: the idea is not to report toast notification from the queue processor, instead return the final state and handle them from the job completion callback. So UI logic and business logic don't get mixed together

src/lib/components/chat/SettingsModal.svelte
```diff
@@ -258,92 +267,89 @@
...
 									} else {
-										pullProgress = 100;
+										downloadProgress = 100;
 									}
+									modelDownloadStatus[opts.modelName] = {
+										pullProgress: downloadProgress,
+										digest: data.digest
+									};
+								} else {
+									toast.success(data.status);
```
anuraagdjain, line 298: suggestion: the idea is not to report toast notification from the queue processor, instead return the final state and handle them from the job completion callback. So UI logic and business logic don't get mixed together

tjbck: Hmm, I can't really think of good ways to report Ollama responses directly to the user other than including in the business logic. If we decide to remove this toast, the user won't know what's going on (at least for a second or two) as it violates visibility of system status heuristics.

anuraagdjain: Hmm, I didn't see any delay in notifications after my download was successful with existing logic. Any specific steps to reproduce?

tjbck: I meant, when you first click on the download button, for a first few seconds, it seems like the button click hasn't been registered.

anuraagdjain: Ah, I think I could improve that tomorrow. Provide faster feedback so the user doesn't try to add the same model again to the queue. As that would give an error message 😅

tjbck: You already did an amazing work, so if anything we can merge this for now and get back to it later :)

ThatOneCalculator:
Nice! A couple of things:

- The progress bars seem to be the same length despite being at different points
- The Rose Pine themes need to be updated with the green and red buttons

image

tjbck:
@ThatOneCalculator, The minimum length for the progress bars are set to 15% by default for styling issues, you'll notice once it's over 15% the progress bar length will differ!

ThatOneCalculator:
Instead of using a div like this: https://github.com/ollama-webui/ollama-webui/blob/acd290431a7d53f4ada0aa9fd3a24242840eb820/src/lib/components/chat/SettingsModal.svelte#L1174C13-L1174C13

Maybe it'd be easier to use and style a [`<progress />`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress) element?

anuraagdjain:
> @ThatOneCalculator, The minimum length for the progress bars are set to 15% by default for styling issues, you'll notice once it's over 15% the progress bar length will differ!

Is it just the % value and its color getting lost / not visible with the background color?

src/lib/components/chat/SettingsModal.svelte
anuraagdjain: Pushed the change. It does exactly what you intended. If it's complicated, then I can revert it.

tjbck: @anuraagdjain Thanks for the hard work! It does seem a bit over-engineered imo, let's just revert to the original solution for now and come back to this later! Everything else LGMT!

anuraagdjain: reverted the change. Should be good to merge.

tjbck:
LGTM, Thanks!

PR #436, opened 2024-01-09
Improved Docker docs

justinh-rahb:
Improved Docker installation instructions for better readability and comprehension for inexperienced users.


justinh-rahb:
I added some instructions for installing Docker, and broke up the docker run commands to multi-line so that you can see them without scrolling.

tjbck:
Thanks, LGTM! Although, I believe we should add more thorough update guide in the '/docs' directory as well but we could create a separate PR for it!

justinh-rahb:
> Thanks, LGTM! Although, I believe we should add more thorough update guide in the '/docs' directory as well but we could create a separate PR for it!

Sounds like a plan, I'll work on an UPDATING.md next.

PR #454, opened 2024-01-11
feat: custom port for server

### LINKED ISSUE #453: Change the default 8080 port or use an environment variable
When running the webui directly on the host with `--network=host`, the port 8080 is troublesome because it's a very common port, for example phpmyadmin uses it.

It would be nice to change the default port to 11435 or being able to change it with an environment variable.

For now I can't run the webui since the default docker run command doesn't work (Server connection error) and making it run on the host conflicts with phpmyadmin.

anuraagdjain:
Allow using a different port for the backend.. Default to `8080`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,7 @@ RUN npm run build
 FROM python:3.11-slim-bookworm as base
 
 ENV ENV=prod
+ENV PORT ""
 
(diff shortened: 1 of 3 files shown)


tjbck:
LGTM, Thanks!

PR #466, opened 2024-01-13
feat: Add epub support

baumandm:
Enables RAG with [EPUB](https://python.langchain.com/docs/integrations/document_loaders/epub) documents.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,11 +28,6 @@ ENV WEBUI_JWT_SECRET_KEY "SECRET_KEY"
 
 WORKDIR /app
 
-# Install pandoc
-RUN apt-get update \
-    && apt-get install -y pandoc \
-    && rm -rf /var/lib/apt/lists/*
-
 # copy embedding weight from build
 RUN mkdir -p /root/.cache/chroma/onnx_models/all-MiniLM-L6-v2
 COPY --from=build /app/onnx.tar.gz /root/.cache/chroma/onnx_models/all-MiniLM-L6-v2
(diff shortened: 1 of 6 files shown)


tjbck:
image

Facing python dependency issue on macos :/ Taking a look!

tjbck:
image

Facing another issue, was it working from your end?

baumandm:
Yes, I was testing via the Docker container (on Linux), and `pypandoc_binary` was working successfully.  I originally tried the `pandoc`/`pypandoc` combo, but it wasn't able find pandoc so I switched to the binary package.

I tried a variety of different epub files and they all worked without issues once the dependency got sorted out.

tjbck:
image


`pypandoc_binary` does not seem to support macos (apple silicon), we're going to have to find other alternatives before we merge.

baumandm:
Alright, so pip's `pandoc` package is just a wrapper library, not Pandoc itself.  Pypandoc has an auto-download feature, but it didn't work for me. I think the best thing to do is to require Pandoc to be [installed separately](https://pandoc.org/installing.html) to enable epub support.

I updated the Dockerfile to install it during the build, this works.  I haven't tested outside of Docker, but the installation is going to depend on the OS.

If Pandoc is not installed, it throws a specific error message.  I also updated the frontend to display a toast and remove the file if any errors occur during the upload.  I saw there was a `doc.error` field but I didn't see it being used so I left that alone.

Employee9833:
Hi, I am also interested in it
From my experiments with llama index it uses: 
ebooklib
html2text
https://github.com/run-llama/llama-hub/tree/956aa44b6dfa3e085b9b9a80c3caec0144b1bbbf/llama_hub/file/epub
Is this a better option?

baumandm:
Rebased to fix merge conflicts with other changes.

tjbck:
Just tested everything, seems to be working fine! everything LGTM, Thanks!

kamjin1996:
First of all, thank you very much for your efforts and current results, I'm very excited to see the support for this feature, the first time I tried it with an epub file of 174M size, it is a document containing html content ([GodotEngine.epub](https://nightly.link/godotengine/godot-docs/workflows/build_offline_docs/master/godot-docs-epub-master.zip)), when I uploaded it the page prompted me:
Unknown File Type 'application/epub', but accepting and treating as plain text

I checked the code related to type judgment, and it supports "application/epub+zip", but not "application/epub". Since I'm not familiar with the implementation mechanism of these functions, I'd like to ask you guys if it's possible to support "application/epub"?

PR #495, opened 2024-01-17
feat: Add RAG support for XML and various programming languages

Marclass:
Allows XML, python, css, javascript, java, golang, sql, c/c++/c#, ruby, rust, php, perl, and more files to be uploaded and parsed for RAG.

### DIFF
--- a/backend/apps/rag/main.py
+++ b/backend/apps/rag/main.py
@@ -166,7 +166,7 @@ def store_doc(
         "css", "cpp", "hpp","h", "c", "cs", "sql", "log", "ini",
         "pl" "pm", "r", "dart", "dockerfile", "env", "php", "hs",
         "hsc", "lua", "nginxconf", "conf", "m", "mm", "plsql", "perl",
-        "rb", "rs", "db2", "scala", "bash", "swift", "vue"
+        "rb", "rs", "db2", "scala", "bash", "swift", "vue", "svelte"
         ]
     file_ext=file.filename.split(".")[-1].lower()
     if file.content_type == "application/octet-stream" and file_ext not in (octet_markdown + octet_plain):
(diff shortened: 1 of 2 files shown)


bhulston:
Let's go! this is super helpful for me

tjbck:
Amazing stuff! You'd also want to update the frontend logic to accept those file formats!

Marclass:
I updated the supported file types in the JS constants. Let me know if there are any other changes needed.

justinh-rahb:
Thank you for your contribution, @Marclass! It holds significant potential and will greatly benefit the project once completed. To maintain repository organization with more contributors getting onboard, would you mind marking it as a draft if you're still polishing it up? We're getting more and more PRs from more and more people now, which is fantastic 🚀👏🎉

Marclass:
@justinh-rahb I'm done with the current version unless you see any issues with it.

I'll likely refactor it at some point though in a later PR because the current file upload and parsing workflow relies on whitelisting files based off of their inferred file type. That type isn't guaranteed to be constant over time and requires manually declaring every possible source file extension ahead of time. 

Most source files are marked as "application/octet-stream" but there are some special cases like javascript (application/x-javascript) and python (text/x-python). At some point in the future other languages might get the same special treatment as python and that would require an update to the white list. In the mean time source files for that language which were fine before would be rejected.   

A workaround might be to allow any file to be uploaded and anything of an unknown format is assumed to be a text file for parsing.

justinh-rahb:
@Marclass your concerns about potential maintainability issues are valid, if this implementation is working for now, I agree let's get it merged and handle this concern in another pull request. Please keep up the great work! 💪

Marclass:
@justinh-rahb It works as is. I fed it in some of it's own code and it proposed one way to rewrite itself.

example

Marclass:
Arbitrary file support added with PR #524 which incorporates the changes in this PR, so closing this in favor of the other.

PR #496, opened 2024-01-17
Feat: Enable imports of ChatGPT chat history

bhulston:
For this issue: https://github.com/ollama-webui/ollama-webui/issues/337

Enable imports of ChatGPT histories (conversations.json), and the ability to choose a new model to use upon import.

1. Added several functions to `src/lib/utils/index.ts` that are used at import time before sending to the `createNewChat` api
        a) Identify if import is a chatGPT one or Ollama-webui one (as they are very different)
        b) Parses and iterates through the chatGPT json to match the ollama-webui chat structure
2. Adjusted the "disabled" tag for model selector element at `src/routes/(app)/c/[id]/+page.svelte`
        a) This makes sure that we can select a new model to chat with when a model is not currently selected (upon import)
        
        
** Note **: When we import from ChatGPT, we don't default to a model so 1. users aren't forced to use chatgpt and 2. because actually collecting which chatGPT model was used is not clear from the chatgpt json.
Does not import context (i.e. pdf RAG) info from GPT chats. Maybe would be nice in future but would take some investment.

### DIFF
--- a/src/lib/components/chat/SettingsModal.svelte
+++ b/src/lib/components/chat/SettingsModal.svelte
@@ -21,7 +21,7 @@
 	import { WEB_UI_VERSION, WEBUI_API_BASE_URL } from '$lib/constants';
 
 	import { config, models, settings, user, chats } from '$lib/stores';
-	import { splitStream, getGravatarURL } from '$lib/utils';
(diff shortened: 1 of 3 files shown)


bhulston:
For importing the conversations.json

madman2012:
Can't wait to use this. How do we allow users to get bing/copilot imports as well?

PR #504, opened 2024-01-17
This script can now be run from any directory

mslinn:
The current version of `backend/start.sh` requires that it be invoked from the `backend` directory.


tjbck:
Thanks for the PR! LGTM!

PR #535, opened 2024-01-19
Add validation for chatGPT imports, stopping any breaking issues with imports of incompatible chats

bhulston:
Tested on 1000 chats from 3 separate people...

chatGPT imports might use external tools like Dall-E or there also might be corrupted chats that were interrupted several times from users.

To deal with this, I basically added a new validation function to make sure the chat imports look good and a new try catch block for any bad imports. In the case that they do not, we just won't include them.

### DIFF
--- a/src/lib/utils/index.ts
+++ b/src/lib/utils/index.ts
@@ -206,25 +206,32 @@ const convertOpenAIMessages = (convo) => {
 	const mapping = convo['mapping'];
 	const messages = [];
 	let currentId = '';
+	let lastId = null;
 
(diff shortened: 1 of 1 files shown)


justinh-rahb:
> Sorry about the multiple commits. Hard to get all the ways chatgpt might structure their chat conversations with all their newer features:(

No worries, this is heroic work you're doing. I think a lot of people are going to really love this feature once it's merged.

tjbck:
LGTM, Thanks!

PR #536, opened 2024-01-19
Unhardcode some stuff, flexibilize dotenv requirement

lucasew:
- Why dotenv as optional dependency is useful?

Today Docker and Systemd already have ways of mimicking what dotenv does, but for anything that runs under it. It's a detail.

- Unhardcode some paths

I am trying to Nix it. This PR started as I being lazy to patch stuff up to work with my requisites lol.

Even though one can customize some extra values, like the data directory, the defaults are still the same.

I tested the backend using a virtualenv environment to see if it finds stuff and works. I need to test if the ollama integration properly works.

### DIFF
--- a/backend/apps/web/routers/utils.py
+++ b/backend/apps/web/routers/utils.py
@@ -11,7 +11,7 @@ import json
 
 from utils.misc import calculate_sha256
 
-from config import OLLAMA_API_BASE_URL
+from config import OLLAMA_API_BASE_URL, DATA_DIR, UPLOAD_DIR
 from constants import ERROR_MESSAGES
 
 
@@ -96,8 +96,7 @@ async def download(
     file_name = parse_huggingface_url(url)
 
     if file_name:
-        os.makedirs("./uploads", exist_ok=True)
-        file_path = os.path.join("./uploads", f"{file_name}")
+        file_path = str(UPLOAD_DIR / file_name)
 
         return StreamingResponse(
             download_file_stream(url, file_path, file_name),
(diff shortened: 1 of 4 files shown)


backend/apps/web/routers/utils.py
```diff
@@ -109,16 +108,15 @@ async def download(
 
 @router.post("/upload")
 def upload(file: UploadFile = File(...)):
-    os.makedirs("./data/uploads", exist_ok=True)
-    file_path = os.path.join("./data/uploads", file.filename)
+    file_path = f"{UPLOAD_DIR}/{file.filename}"
 
     # Save file in chunks
-    with open(file_path, "wb+") as f:
+    with file_path.open("wb+") as f:
```
tjbck, line 114: Any reason for this change?

backend/config.py
```diff
@@ -14,17 +13,24 @@
...
 ####################################
 # File Upload
 ####################################
 
+DATA_DIR = Path(os.getenv("DATA_DIR", "./data")).resolve()
 
-UPLOAD_DIR = "./data/uploads"
-Path(UPLOAD_DIR).mkdir(parents=True, exist_ok=True)
+UPLOAD_DIR = f"{DATA_DIR}/uploads"
+UPLOAD_DIR.mkdir(parents=True, exist_ok=True)
```
tjbck, line 31: Why is Path removed? Won't this break the code?

backend/apps/web/routers/utils.py
```diff
@@ -96,8 +96,7 @@ async def download(
     file_name = parse_huggingface_url(url)
 
     if file_name:
-        os.makedirs("./uploads", exist_ok=True)
-        file_path = os.path.join("./uploads", f"{file_name}")
+        file_path = f"{UPLOAD_DIR}/{file_name}"
```
tjbck, line 99: I refactored code myself but next time if you make a PR, please stick to the f-string convention used! Thanks!

tjbck (changes requested):

TanvirOnGH:
Oh Yes Nix finally!!

backend/apps/web/routers/utils.py
lucasew: fixed

backend/config.py
lucasew: fixed

backend/apps/web/routers/utils.py
lucasew: fixed

BTW pathlib's / operator normalizes the path joins, for example using \ instead of / on Windows.

backend/apps/web/routers/utils.py
tjbck: The change seem to have been introduced in [5b26d2a](https://github.com/ollama-webui/ollama-webui/pull/536/commits/5b26d2a686d47822d5959e90a1d0b64ebec88d67) :)

tjbck:
LGTM, Thanks!

PR #537, opened 2024-01-19
Add max_tokens workaround for gpt-4-vision-preview model

CreatorGhost:
This commit adds a workaround for the gpt-4-vision-preview model in the OpenAI API. 

Due to an issue with this model, it's necessary to set the 'max_tokens' parameter  in the request body. The changes in this commit check if the model is 'gpt-4-vision-preview' and, if so, set 'max_tokens' to 4000. 

This workaround is temporary and should be removed once the issue with the model is fixed by OpenAI.

A screenshot has been added as a reference to show that the gpt-4-vision-preview model is now working with this workaround. The screenshot can be found in the related pull request or issue.

image


tjbck:
LGTM, Thanks!

PR #538, opened 2024-01-20
feat: add guard clause to improve signup process

Shiyinq:
Refactor the signup functionality by implementing a guard clause.


tjbck:
Much needed refac! LGTM, Thanks!

PR #547, opened 2024-01-23
feat: Add reStructuredText specific parser for RAG

Marclass:
Add specific parser for common python documentation format.


tjbck:
LGTM, Thanks!

PR #549, opened 2024-01-23
Bugfix: Fix toast error popup when front end can't figure out file type.

Marclass:
Was getting error toasts with RAG because file['type'] on front end was empty string sometimes in doc upload. Back end parsed them as expected though. 

Copied list of file ext from back end to front end added in #547 and also added typescript extension to both lists.


tjbck:
LGTM, Thanks!

Issue #60, opened 2023-11-05
Add Support for Text File Upload


TanvirOnGH:
**Is your feature request related to a problem? Please describe.**
Currently, the project lacks the capability to handle text file uploads, such as markdown and PDF files, for the purpose of summarization and Q&A. This limitation restricts users from efficiently extracting information from documents.

**Describe the solution you'd like**
I propose the addition of a feature that allows users to upload text files in formats like markdown, PDF. ollama-webui should be able to process these files to provide summaries and answer user queries based on the content within the uploaded documents. This enhancement would significantly improve the versatility and usefulness of the ollama-webui.

**Describe alternatives you've considered**
An alternative to supporting file uploads could be manual copying and pasting of text content into the chat interface, but this approach is time-consuming and error-prone. Another alternative is integrating with external document management systems, but this might be less user-friendly and may introduce privacy concerns.

**Additional context**
In many use cases, users need to extract specific information or summaries from existing documents in their possession. Adding support for text file uploads, especially in common formats like markdown and PDF, would greatly enhance the capabilities and make it a valuable tool for content processing and information retrieval. This feature aligns with the growing demand for efficient document processing and information extraction tools in various domains.

TanvirOnGH:
Possibly support all the types of file formats same as gpt4all localdocs as mentioned [here](https://docs.gpt4all.io/gpt4all_chat.html#how-localdocs-works):
```txt
["txt", "doc", "docx", "pdf", "rtf", "odt", "html", "htm", "xls", "xlsx", "csv", "ods", "ppt", "pptx", 
"odp", "xml", "json", "log", "md", "org", "tex", "asc", "wks",
"wpd", "wps", "wri", "xhtml", "xht", "xslt", "yaml", "yml", "dtd", "sgml", "tsv", "strings", "resx",
"plist", "properties", "ini", "config", "bat", "sh", "ps1", "cmd", "awk", "sed", "vbs", "ics", "mht",
"mhtml", "epub", "djvu", "azw", "azw3", "mobi", "fb2", "prc", "lit", "lrf", "tcr", "pdb", "oxps",
"xps", "pages", "numbers", "key", "keynote", "abw", "zabw", "123", "wk1", "wk3", "wk4", "wk5", "wq1",
"wq2", "xlw", "xlr", "dif", "slk", "sylk", "wb1", "wb2", "wb3", "qpw", "wdb", "wks", "wku", "wr1",
[...]

PR #554, opened 2024-01-23
feat: Add excel parser for RAG

Marclass:
Adds xls and xlsx parsers for RAG, but they're pretty slow, so users would probably want to convert to csv first.


justinh-rahb:
I could have sworn support for Excel had already been added but I can't find a related issue or PR other than #60 which only added support for uploading the filetype, doesn't appear it does any other processing beyond that.

tjbck:
LGTM, Thanks!

PR #555, opened 2024-01-24
Changed the sizing of the users table

jannikstdl:
Sometimes, if your users have a too long mail / name the table cuts off at the right end. Changed the size.

Looked like this with 3xl:

image

Now its bigger and shows the full content (Exept someone has a way too big mail or name)

image


tjbck:
Hi, Thanks for the PR! I'll close this PR for now as you should be able to scroll horizontally to access the part that's been cut off, but I'll see what I can do to improve the user experience!

PR #558, opened 2024-01-24
fix: use bash when starting the Dockerfile's entrypoint script (as th…

peperunas:
…e script uses BASH_SOURCE that is bash only)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,4 +55,4 @@ COPY --from=build /app/build /app/build
 # copy backend files
 COPY ./backend .
 
-CMD [ "sh", "start.sh"]
+CMD [ "bash", "start.sh"]


peperunas:
`BASH_SOURCE` is a bash-only variable that is used in `start.sh`. `start.sh` is used as entrypoint in the docker image: if you check the logs, once the image is started there is an error reporting that on line 3 there is a `bad substitution` which refers to `BASH_SOURCE`.

Calling the script with `bash` fixes the issue.

justinh-rahb:
In my opinion, it is crucial to consistently follow one approach when deciding how to run a script. This is because there can be significant side effects and unintended consequences when running a script directly with bash, as opposed to allowing the system command interpreter to infer it. A well-known example of this is Valve's Linux Steam client, which accidentally performed an `rm -rf /` on some users' machines when they ran `bash steam.sh` instead of `./steam.sh`.

jnkstr:
> In my opinion, it is crucial to consistently follow one approach when deciding how to run scripts in a project. This is because there can be significant side effects and unintended consequences when running a script directly with bash, as opposed to allowing the system command interpreter to infer it. A well-known example of this is Valve's Linux Steam client, which accidentally performed an `rm -rf /` on some users' machines when they ran `bash steam.sh` instead of `./steam.sh`. Whatever is decided here, it should be applied repo-wide.


Haha funny, saw the video a couple days ago on YouTube😂

justinh-rahb:
> Haha saw the video a couple days ago on YouTube😂

Indeed, we don't want to pull a Valve here!

peperunas:
I agree @justinh-rahb, the interpreter is set to use `bash` but the code was run as `sh`. Hence why I forced `bash` but that should not be the case anyhow and let the shell use the interpreter defined in the shebang.

tjbck:
I suppose we could either remove the shebang in the sh file or go with your suggested approach. As others have mentioned, even though I'm pretty sure there won't be any side effects, I'm concerned it might cause some issues of some sort later down the line. If everyone concurs to merge this PR, I really don't have anything against it and will merge this PR promptly!

tjbck:
I'll just merge this, Thanks!

PR #559, opened 2024-01-24
fix: use exec the uvicorn process to handle Docker's SIGTERM correctly

peperunas:
Replace the shell with the uvicorn process, so it becomes PID 1 and receives the signals directly


tjbck:
LGTM, Thanks!

PR #569, opened 2024-01-25
Reduce container image size

yeahdongcn:
Move `untar` onnx to build phase and copy the extracted directory in runtime stage.

Container image size gets reduced.

```bash
<none>                                <none>        4dee6ba8ae1c   32 seconds ago      2.4GB
ghcr.io/ollama-webui/ollama-webui     main          610f0e54854e   About an hour ago   2.56GB
```


tjbck:
LGTM, Thanks!

PR #578, opened 2024-01-26
add missing OLLAMA_WEBUI_PORT override

gryphius:
`run-compose.sh` does not actually handle a non-default webui port. it just shows the changed port but does not set the `OLLAMA_WEBUI_PORT` environment variable to the custom value. This PR fixes this.


tjbck:
LGTM, Thanks!

PR #582, opened 2024-01-26
feat: Add netcat to Dockerfile for healthcheck

Trimutex:
Adds the netcat-openbsd package while building the image to allow creating a healthcheck.

I couldn't find anything related to Docker's healthcheck implemented. Ollama has commits to use curl and I'm not sure if there's any advantage to use it for healthcheck instead.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,10 +35,10 @@ COPY ./backend/requirements.txt ./requirements.txt
 RUN pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu --no-cache-dir
 RUN pip3 install -r requirements.txt --no-cache-dir
 
-# Install pandoc
+# Install pandoc and netcat
 # RUN python -c "import pypandoc; pypandoc.download_pandoc()"
 RUN apt-get update \
-    && apt-get install -y pandoc \
+    && apt-get install -y pandoc netcat-openbsd \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


tjbck:
Not sure why/how this is useful, could you elaborate a bit more on its use cases? Thanks!

Trimutex:
This would mainly be for monitoring purposes for the container. The healthcheck would be for if the container was up, but the application was not working or having issues and did not successfully open the port for connections. 

Perhaps adding the healthcheck as a part of the image by default would make more sense?

Also any monitoring services would also be able to match the criteria of utilizing the healthcheck on the container.

tjbck:
Ideally I'd want to keep the image size as small as possible, so I'm a bit hesitant to merge this. Could you also check the container size for us? Thanks!

Trimutex:
I definitely understand that! This is why I chose netcat-openbsd over netcat-traditional or curl as it seemed like the smallest addition.

As for the container size, I'm not exactly sure which part about the container that you want so I will post the image and size of an active container (non-virtual amount) below.

Without netcat-openbsd:
active container size | image size
-- | -- 
655kB | 2.09GB

With netcat-openbsd:
active container size | image size
-- | --
655kB | 2.09GB

Pulled from ghcrio (main):
active container size | image size
-- | --
650kB | 1.91GB

I'm not too clear as to why my locally created image differs so much from the main image though as I haven't changed anything else. The container without netcat that I built from my own image was confirmed to not have netcat internally.

tjbck:
Doesn't seem all that much an increase. LGTM, Thanks!

PR #594, opened 2024-01-28
Added username in chat instead of 'You'

jannikstdl:
The UserMessage Component now loads the username and renders it in the chat layout. 

Looks more professional / personal imo.

Previous:
image

Now:
image


justinh-rahb:
Personally I'm not against having an *option* to set the shown username to something other than 'You', but I disagree with **forcing** this change which diverges from the appearance of this project's well-known and used inspiration.

jnkstr:
Ok i will try to implement this as an option if i find time for that.

tjbck:
Yep, you might want to add an option to toggle in Settings > Interface.

jnkstr:
Ok added now the option in settings -> interface. Local test worked. I hate html and css now, spent more than half the time getting the Save button inside the Popup but outside the scroll area.. 🤪

jnkstr:
Default is off.

https://github.com/ollama-webui/ollama-webui/assets/69747628/1856b122-4558-4233-967c-5ecd0611108b

jnkstr:
Ok so this is not working fully yet. I need to make this option somehow globally for all users.

tjbck:
I'll take a look in a bit, in the mean time if you could resolve conflicts that would be great!

jnkstr:
Should this be a global or user option? @tjbck

tjbck:
I don't think it's necessary, if anything we can always add the option in the future.

PR #618, opened 2024-02-02
Disable Chroma telemetry

ghost:
Chroma makes outbound connections to `app.posthog.com` by default. This should prevent such connections. It's also possible to set an environment variable `ANONYMIZED_TELEMETRY=False` which works with the Docker image. See https://docs.trychroma.com/telemetry

Thanks for this software! It's one of the best I've encountered.


tjbck:
Nice catch! LGTM, Thanks!

PR #656, opened 2024-02-06
feat: openai tts support

tjbck:


oliverbob:
I presume that adding an OPENAI key at the settings will activate the tts. Correct? But I notice that it only give me new models to interact with. How do I activate openai tts?

oliverbob:
Found it, I have to click the default voice a second time. Thanks.

oliverbob:
image

For some reason the audio doesn't play.

tjbck:
@oliverbob Hmm, it might be a browser cache issue. Try using it with other browsers or in incognito mode and see if the issue persists.

tjbck:
@oliverbob https://stackoverflow.com/questions/75041580/openai-api-giving-error-429-too-many-requests

PR #663, opened 2024-02-06
Enable/disable signup in env 

kylemclaren:
This PR will get `ENABLE_SIGNUP` from env, and if it's not set will keep the default of `True`.

I noticed that I'm not able to persist the setting between container restarts.


tjbck:
LGTM, Thanks!

Issue #665, opened 2024-02-06
feat: openai api compatible model presets (profiles)


keriati:
**Is your feature request related to a problem? Please describe.**
As the modelfiles from the hub are pretty good, it would be nice if we could use also them with external models from openAI api compatible services.

**Describe the solution you'd like**
Option A:
When adding a modelfile, we could check that it should use a model from the external provider and select it.
Option B:
When selecting a modelfile, we should be able to exchange then the actual model to one from the external service.

**Describe alternatives you've considered**
Setting the default prompt according to the modelfile works for now, just the profile pick and name is wrong and the extra settings.

**Additional context**
My ollama instance has limited vram available. External services can provide bigger models. I think it would be good if the same characters would be usable when using an external service.

explorigin:
Ollama.ai is not directly associated with this project. Your feature request belongs with them.

keriati:
Thanks for taking a look. I already implemented this and it works for me. See the referenced PR.

explorigin:
OK first, thanks for adding features. It's awesome to have a community that's adding to a project.

Per the PR, I have some suggestions that I'll post over there.

explorigin:
Changed my mind, I'll post here:

Modelfiles are an Ollama.ai-specific feature to describe settings that belong to a specific model. OpenAI doesn't have such a thing because most of the provided functionality it is wrapped up in the product offering. Of course we live in a world where the OpenAI API is becoming the standard even for interacting with custom models and local LLM servers. That leaves a functionality hole that you're trying to address here.

However, I think the over-arching feature request (added metadata for non-local, non-ollama.ai models) needs to be thought of a bit more holistically. For example, in the UI, we set an Ollama server and optionally and OpenAI key and endpoint. the WebUI then gives you access to both.

I think the right way to solve this issue is to unify the idea of external LLM servers much like how LiteLLM handles it. (That and even Ollama.ai is working on adopting OpenAI's API.) Your PR provides metadata to all in an agnostic way but maintains the type-difference segregation.

The whole idea of multiple models/servers/configurations is tied up in other feature requests that should be considered since they all would (should?) share the same solution: https://github.com/ollama-webui/ollama-webui/issues/278 https://github.com/ollama-webui/ollama-webui/issues/619

explorigin:
[...]

PR #667, opened 2024-02-06
feat: Use modelfiles with external API instead of ollama

keriati:
With this change a modelfile from the ollama hub can be imported in a way that it uses an external model from openAI compatible API instead of ollama.

A pre requirement is to provide an API key and refresh the page so the models are loaded.

In my testing to achieve this we have to copy an existing model's name to two places:
- the Tag name
- the FROM part of the content

This works in my local instance.

I am rather not so sure about this changes as I never touched svelte before.

Probably a checkbox called: use external and a dropdown list to select from external models would be a more complete solution.

Should relate to #665

### DIFF
--- a/src/routes/(app)/+page.svelte
+++ b/src/routes/(app)/+page.svelte
@@ -512,6 +512,10 @@
 	const sendPromptOpenAI = async (model, userPrompt, parentId, _chatId) => {
 		let responseMessageId = uuidv4();
 
+		const myModel = $models.find((m) => m.name === model);
+		let external = myModel?.external === true;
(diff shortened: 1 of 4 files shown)


explorigin:
I haven't reviewed the frontend code enough to have an opinion on the correctness of this, but I posted some thoughts as to why we should probably do this whole thing differently in the related issue.

tjbck:
I agree with @explorigin, we might want to have our own "modelfile" format that's extensible and compatible with other LLM runners. I'll close this for now, let's continue our discussion here: https://github.com/ollama-webui/ollama-webui/issues/665

PR #682, opened 2024-02-09
Simplify endpoint role checking

explorigin:
Endpoint role-checking was redundantly applied but FastAPI provides a nice abstraction mechanic...so I applied it.

There should be no logical changes in this code. I did swap some of the HTTP 40X codes to be more appropriate to the situation.

This is mostly an incremental PR that moves things forward while I'm working on the larger authentication rewrite.

image

😁


justinh-rahb:
I've always said the best PRs are ones which remove more than they add without compromising function.. **\*chef's kiss\***

tjbck:
LGTM, Thanks for the hard work!

PR #721, opened 2024-02-13
Added support for setting the new KeepAlive property to be sent to Ollama

### LINKED ISSUE #596: feat: Support ollama's keep_alive request parameter
Ollama unloads models after 5 minutes by default. A new parameter, `keep_alive`, allows the user to set a custom value.
It would be nice to have an option in the UI where a value for this parameter can be set.

Here is the relevant PR: [#2146](https://github.com/ollama/ollama/pull/2146)

zabirauf:
The change supports the new KeepAlive property from Ollama which keeps the model loaded for quicker response. https://github.com/ollama/ollama/pull/2146

The change includes

1. Adding a new Keep Alive in the Settings > Advanced.
2. The value being empty is default, in which case nothing is sent to chat/generate API
3. If the value is set then the value is fetched from settings and sent to ollama when calling chat/generate API as `keep_alive`

TODO:
- [ ] Validate that the request sent to ollama contains the set value

Related issue #596

Screenshot 2024-02-12 at 11 47 50 PM
Screenshot 2024-02-12 at 11 47 57 PM

### DIFF
--- a/src/lib/apis/ollama/index.ts
+++ b/src/lib/apis/ollama/index.ts
@@ -167,7 +167,7 @@ export const generateTitle = async (token: string = '', model: string, prompt: s
 	return res?.response ?? 'New Chat';
 };
 
-export const generatePrompt = async (token: string = '', model: string, conversation: string) => {
+export const generatePrompt = async (token: string = '', model: string, conversation: string, body: object = {}) => {
 	let error = null;
 
 	if (conversation === '') {
@@ -188,7 +188,8 @@ export const generatePrompt = async (token: string = '', model: string, conversa
 			As USER in the conversation above, your task is to continue the conversation. Remember, Your responses should be crafted as if you're a human conversing in a natural, realistic manner, keeping in mind the context and flow of the dialogue. Please generate a fitting response to the last message in the conversation, or if there is no existing conversation, initiate one as a normal person would.
 			
 			Response:
-			`
+			`,
+			...body
 		})
 	}).catch((err) => {
 		console.log(err);
--- a/src/lib/apis/ollama/index.ts
+++ b/src/lib/apis/ollama/index.ts
@@ -167,7 +167,7 @@ export const generateTitle = async (token: string = '', model: string, prompt: s
 	return res?.response ?? 'New Chat';
 };
 
-export const generatePrompt = async (token: string = '', model: string, conversation: string, body: object = {}) => {
+export const generatePrompt = async (token: string = '', model: string, conversation: string) => {
 	let error = null;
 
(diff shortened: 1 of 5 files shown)


tjbck:
LGTM, Thanks for getting the ball rolling on this!

justinh-rahb:
Nice work @zabirauf, thanks for the PR!

revaraver:
"Keep alive" cannot be found in the current version. Have you removed this function?

yusiwen:
Where is this setting? I'm using 0.3.13 right now, and cannot find this setting, and there is event no "Advanced" tab in settings
:(

jupiterbjy:
@veka-server for OpenWebUI's docker it doesn't respect that flag sadly, still unloading models for me sadly.. wonder if I should set it at bashrc rn

jupiterbjy:
@veka-server @yusiwen actually nevermind location is changed

image

yusiwen:
@jupiterbjy  Thanks, I found it. But still think it's better for setting it per model not globally.

adamoutler:
This should be moved to the Workspace tab, instead of across the entire set of models.   Eg. Maybe I use a generic 70b for mulitple tasks and never want it unloaded, but I have a set of 2b models which are called frequently and need to be unloaded to save RAM.

edurenye:
We should have the option to set `keep_alive` per model

MattariOnline:
> We should have the option to set `keep_alive` per model

I just want to second this; it would be a massive improvement to Open WebUI.

It would be nice, for example, to keep-alive Llama 3.2 3B as an "always available" option, while other models (such as larger models) only get loaded when needed. That way you can use what you need, but if you need a super quick response, there's one waiting and ready.

adamoutler:
Yes.  The Task Model should be able to be kept alive.

vulcan4d:
Moved again? Now I can't find it.  The settings keep moving, might be a good idea to implement a settings search function too.

alienatedsec:
> Moved again?

Potentially, as I can't find it either @vulcan4d

alienatedsec:
image

Under `Advanced Parameters` in `Settings`, these settings when you click your name.

image

LoadingCode233:
Hello, is this feature not working? I've checked the request in my browser and there is no 'keepalive' parameter in the request, but other parameters are present.

SteamNimmersatt:
> Hello, is this feature not working? I've checked the request in my browser and there is no 'keepalive' parameter in the request, but other parameters are present.

It's not working for me either with openwebui 0.5.7 and ollama as backend. Ollama displays "Until: 30m" for the model started by openwebui.

PR #723, opened 2024-02-13
fixed Docker problem some users experienced

jannikstdl:
Some users, including me had this problem when running the Dockerfile since it runs start.sh after the latest changes made to the .sh files:

image

This is a problem with the encoding CRLF, I made a .gitattrobutes file to make shure the file stays in LF.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.sh text eol=lf


tjbck:
LGTM, Thanks!

PR #724, opened 2024-02-13
Snap packaging

mz2:
Adds strictly confined snap packaging of ollama-webui for x86-64 as an alternative to docker & k8s, presently published on the channel `latest/beta` (`sudo snap install ollama-webui --beta`).

This is proposed as a nice alternative to docker (no need to install and configure the nvidia docker runtime for example, systemd service is set up automatically, over-the-air updates, straightforward to access resources and data from user's host system within the limits of the application's confinement) and safer of course than bare installation onto host system.

Installable with:

```bash
sudo snap install ollama --channel beta # or ollama installed any other method, or on another host
sudo snap install ollama-webui --channel beta
```

As you might guess from above I also packaged up [ollama](https://github.com/ollama/ollama/pull/2432) as a snap (this snapped version of ollama-webui does not require it, demonstrating this mostly just to note the ease of installing both).

- strict confinement used with [`network`](https://snapcraft.io/docs/network-interface), [`network-bind`](https://snapcraft.io/docs/network-bind-interface), [`home`](https://snapcraft.io/docs/home-interface), [`removable-media`](https://snapcraft.io/docs/removable-media-interface), [`opengl`](https://snapcraft.io/docs/opengl-interface) interfaces in use, i.e. it can access and serve a port, access home directory and `/media`, and access the GPU (the `opengl` interface also grants access to CUDA etc).
- starts up a systemd service automatically for the service.
- if removable media access is needed (e.g. user prefers storing data under a disk mounted under `/media`), `sudo snap connect ollama:removable-media` achieves that (for security reasons, removable media access not granted without user action).

If this looks interesting, I'm happy to hand over the package on snapcraft.io to a maintainer, and can contribute CI integration to make it easy to keep the snap package up to date whenever you release.

If you want to build this locally, [after installing `snapcraft` and either the multipass or LXD provider for it](https://snapcraft.io/docs/snapcraft-setup) go to the root directory of the repository, and ...:

```bash
snapcraft
```

## Configuration

Offers the following configuration keys and default values, configurable with `sudo snap set ollama-webui key=value`:

- `data-dir`: /var/snap/ollama-webui/common/data
- `enable-signup`: true
- `host`: 0.0.0.0
- `ollama-api-base-url`: http://localhost:11434/api
- `openai-api-base-url`: https://api.openai.com/v1
- `openai-api-key`: (blank)
- `port`: 8080
- `secret-key` (randomized)

### DIFF
--- a/backend/start.sh
+++ b/backend/start.sh
@@ -29,4 +29,4 @@ if [ "$USE_CUDA_DOCKER" = "true" ]; then
   export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/torch/lib:/usr/local/lib/python3.11/site-packages/nvidia/cudnn/lib"
 fi
 
-WEBUI_SECRET_KEY="$WEBUI_SECRET_KEY" exec uvicorn main:app --host 0.0.0.0 --port "$PORT" --forwarded-allow-ips '*'
+WEBUI_SECRET_KEY="$WEBUI_SECRET_KEY" exec uvicorn main:app --host "$HOST" --port "$PORT" --forwarded-allow-ips '*'
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,105 @@
+---
+name: open-webui
+base: core22
+version: '0'  # no released version yet
+summary: ChatGPT-Style Web UI Client for Ollama 🦙 (formerly Ollama WebUI)
+description: |
+  The Open WebUI consists of two primary components:
+  the frontend and the backend (which serves as a reverse proxy,
+  handling static frontend files, and additional features).
+  Both are set up.
+adopt-info: backend
+grade: stable
+confinement: strict
+license: MIT
+architectures:
+  - build-on: [amd64]
+    build-for: [amd64]
+  - build-on: [arm64]
+    build-for: [arm64]
+
+apps:
+  open-webui:
+    command: bin/info.sh
+  listener:
+    command: bin/run-snap.sh
+    daemon: simple
+    install-mode: enable
+    restart-condition: on-failure
+    plugs:
+      - home
+      - removable-media
+      - network
+      - network-bind
+      - opengl
+    environment:
+      ENV: prod
+      SCARF_NO_ANALYTICS: 'true'
+      DO_NOT_TRACK: 'true'
+parts:
+  launcher:
+    plugin: dump
+    source: ./snap/local
+    organize:
+      info.sh: bin/info.sh
+      run-snap.sh: bin/run-snap.sh
+  minilm:
+    plugin: dump
+    source: https://chroma-onnx-models.s3.amazonaws.com/all-MiniLM-L6-v2/onnx.tar.gz
+    organize:
+      onnx: root/.cache/chroma/onnx_models/all-MiniLM-L6-v2/onnx
+  frontend:
+    after: [launcher, minilm]
+    plugin: nil
+    source: .
+    source-type: git
+    override-build: |
+      craftctl default
+      npm ci
+      npm run build
+      cp -r build $CRAFT_PART_INSTALL/build
+    stage-snaps:
+      - node/20/stable
+  backend:
+    after: [frontend]
+    plugin: python
+    source: .
+    source-type: git
+    build-packages:
+      - netcat-openbsd
+      - python3.10
+      - python3.10-venv
+      - python3.10-dev
+      - python3-pip
+      - git
+      - execstack
+    stage-packages:
+      - python3.10
+      - python3.10-venv
+      - pandoc
+      - libavcodec58
+      - libavformat58
+      - libavdevice58
+      - libavfilter7
+    python-requirements:
+      - backend/requirements.txt
+    override-pull: |
+      craftctl default
+      last_committed_tag="$(git describe --tags --abbrev=0)"
+      last_committed_tag_ver="$(echo ${last_committed_tag} | sed 's/v//')"
+      craftctl set version="$(git describe --tags | sed 's/v//')"
+    override-build: |
+      craftctl default
+      pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu --no-cache-dir
+      cp -r backend $CRAFT_PART_INSTALL
+    override-prime: |
+      craftctl default
+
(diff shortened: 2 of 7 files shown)


tjbck:
Amazing stuff, Thanks for the hard work! We're[ planning on changing our name as per Ollama team's request](https://github.com/ollama-webui/ollama-webui/discussions/602), so let's wait until we decide on our new name before we merge this!

mz2:
Thanks! I'll keep improving it meanwhile: I am going to add the support for this package pulling the `ollama` snap as a dependency through the "content" interface, also to allow making it possible to download models via the web frontend.

.gitignore
```diff
@@ -297,4 +297,6 @@ dist
 .yarn/unplugged
 .yarn/build-state.yml
 .yarn/install-state.gz
-.pnp.*
\ No newline at end of file
+.pnp.*
```
justinh-rahb, line 300: No problemo to resolve

backend/start.sh
justinh-rahb, line 1: Makes good sense to make the host configurable for future packaging flexibility in general.

justinh-rahb (commented):
Can't comment on the newly added snap stuff but the modified files look good.

Unrelated but related: I just realized that Ubuntu and Canonical's 20th anniversary are coming up, and now I feel old 😂
I've been using it since 4.10 casually, and 5.04/5.10 seriously. Time for a drink 🥃

snap/snapcraft.yaml
```diff
@@ -0,0 +1,109 @@
...
+    stage-snaps:
+      - node/20/stable
+  backend:
+    after: [frontend]
+    plugin: python
+    source: .
+    source-type: git
+    build-packages:
+      - netcat-openbsd
+      - python3.10
```
tjbck, line 79: Correct me if I'm wrong, but I believe we should update this part to match our docker container image python environment (python3.11).

mz2:
I updated this PR now with latest changes from `main`, including updating the name of the package, and published.

On doing so I hit a security check with the new transient dependency `libctranslate2` (that dynamic library is built with an executable stack -- curious to learn if that will have been intentional or a mistake?)

In my own testing I have been able to use the regular chat functionality with text input just fine, but I hit an issue with the document upload / RAG functionality: on attempting to upload a document I experience a 500 status code from the server but am not clear how I would be able to debug this further short of starting to add debug log statements, track trace output in myself. Am I missing a debug facility I should just turn on?

snap/snapcraft.yaml
mz2: Oh yes, that makes sense.

justinh-rahb:
> I updated this PR now with latest changes from `main`, including updating the name of the package, and published an initial revision to https://snapcraft.io/open-webui on the beta channel (`sudo snap install open-webui --beta`).
> 
> On doing so I hit a security check with the new transient dependency `libctranslate2` (that dynamic library is built with an executable stack -- curious to learn if that will have been intentional or a mistake?)
> 
> In my own testing I have been able to use the regular chat functionality with text input just fine, but I hit an issue with the document upload / RAG functionality: on attempting to upload a document I experience a 500 status code from the server but am not clear how I would be able to debug this further short of starting to add debug log statements, track trace output in myself. Am I missing a debug facility I should just turn on?

That `libctranslate2` needs some more explanation, not sure that I understand. Which package depends on it? If there's something we can do about it, by all means let's get another issue/PR for that going.

As for debugging, I don't think you've missed anything. We really need some more debug log statements, and a DEBUG=true variable to activate it.

backend/start.sh
```diff
@@ -5,6 +5,7 @@ cd "$SCRIPT_DIR" || exit
 
 KEY_FILE=.webui_secret_key
 
+HOST="${HOST:-0.0.0.0}"
```
lainedfles, line 8: In the interest of security (especially without any default transport encryption), shouldn't we limit the listener to 127.0.0.1 by default and require that the user override via `snapctl set host="0.0.0.0"` to listen on all addresses?

snap/snapcraft.yaml
mz2: I've pushed an update to the beta channel built with python 3.11.

mz2:
> Thanks! I'll keep improving it meanwhile: I am going to add the support for this package pulling the ollama snap as a dependency through the "content" interface, also to allow making it possible to download models via the frontend to ollama.

I had second thoughts about this after stating this: it's not necessarily a good idea to do this because a user may a) not care for ollama, or choose to get it from another source, not necessarily the snap (build from source, use a Dockerfile exposed to the host system etc).

mz2:
> That libctranslate2 needs some more explanation, not sure that I understand. Which package depends on it? If there's something we can do about it, by all means let's get another issue/PR for that going.

It's a dependency of one of the ML frameworks included in open-webui. I've understood that an executable call stack in a dynamic library is in most cases an accident (the reason that the snap store automatically validates against it is that it is a security hole), and appears to indeed be an accidental command line flag in that dependency library build since with cleaning up that exec stack bit out in the snapped open-webui, the app works without it set. I can try to dig into that some more at some point, I think it should be fine to keep the processing of that library in place.

> As for debugging, I don't think you've missed anything. We really need some more debug log statements, and a DEBUG=true variable to activate it.

This seemed to have improved since I opened the PR :smile: I got the file upload (and RAG support for documents and websites, both of which I've successfully been using past weeks) feature fixed thanks to some added debug logging that made it clear that the data directory was not getting configured quite right inside the app's sandbox (I was my bad, got fixed by updating the hooks and launcher provided inside the snap).

mz2:
I've updated the branch now against the current `dev` branch, which actually simplified things: the introduction of the configurable `$HOST` is the only change (beside what's under the `snap` directory). If you're happy with this contribution, I'd propose introducing in a follow-up branch a GitHub workflow based automation that would automatically build the snap and publish to the store (at that point it would be best to get the snap ownership to yourselves in the upstream project). I can come chat on Discord about all this if you like.

Also, a fresh update (revision 7) is working its way through the snap store processing that reflects the current HEAD of this branch. To accelerate snapd picking up the update when it's made available in the next < 30mins, you can do `sudo snap refresh open-webui --channel beta` (... or it will be delivered automatically during the next day).

justinh-rahb:
Hi @mz2 I didn't mean to leave this unreviewed for so long. I'd like to chat with Tim first about adopting this under our umbrella, you're more than welcome to join us in Discord!
https://discord.gg/5rJgQTnV4s

PR #725, opened 2024-02-14
fix: no internet connection for whisper if you use docker

jannikstdl:
If you use docker, there is no internetconnection now needed to load the whisper model on first use since it is preloaded in the Docker Image buildtime.

Added env variables in the Dockerfile to change the whisper model and the location where its stored.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,8 +31,8 @@ ENV SCARF_NO_ANALYTICS true
 ENV DO_NOT_TRACK true
 
 #Whisper TTS Settings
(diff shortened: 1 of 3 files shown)


justinh-rahb (approved):
LGTM

justinh-rahb:
As an aside, I'm starting to think that creating two Docker images - one as now with and a 'lite' one without the additional components that can be downloaded during build time - might better serve users' needs. The need to download a 2GB+ image every push may not be ideal for some users, especially if we can download the necessary components on first run and persist them to a mounted volume instead.

tjbck:
LGTM, Thanks!

PR #756, opened 2024-02-16
Correct text for empty OpenAI API URL Field

Arthur2500:
Bildschirmfoto 2024-02-16 um 18 44 23
As you see, the text for no input for the url is wrong. Corrected it to match the heading


tjbck:
Nice catch! LGTM, Thanks!

PR #757, opened 2024-02-16
Kubernetes: WebUI Persistent Volume Claim

braveokafor:
**Summary:**
- Added Kubernetes Persistent Volume Claim (PVC) for Ollama WebUI.
- Added Volume and Volume Mount to attach the PVC to Ollama WebUI deployment.

**Note:**
- This is for #742 .
- Tested on GKE (review attached screenshots).
Screenshot 2024-02-16 at 19 49 30
Screenshot 2024-02-16 at 19 50 18

### DIFF
--- a/kubernetes/helm/values.yaml
+++ b/kubernetes/helm/values.yaml
@@ -5,11 +5,14 @@ ollama:
   image: ollama/ollama:latest
   servicePort: 11434
   resources:
-    limits:
+    requests:
       cpu: "2000m"
       memory: "2Gi"
+    limits:
+      cpu: "4000m"
+      memory: "4Gi"
       nvidia.com/gpu: "0"
-  volumeSize: 1Gi
+  volumeSize: 30Gi
   nodeSelector: {}
   tolerations: []
   service:
@@ -22,16 +25,19 @@ webui:
   image: ghcr.io/ollama-webui/ollama-webui:main
   servicePort: 8080
   resources:
-    limits:
+    requests:
       cpu: "500m"
       memory: "500Mi"
+    limits:
+      cpu: "1000m"
+      memory: "1Gi"
   ingress:
     enabled: true
     annotations:
     # Use appropriate annotations for your Ingress controller, e.g., for NGINX:
       # nginx.ingress.kubernetes.io/rewrite-target: /
     host: ollama.minikube.local
-  volumeSize: 1Gi
+  volumeSize: 2Gi
   nodeSelector: {}
   tolerations: []
   service:
--- a/kubernetes/manifest/base/ollama-statefulset.yaml
+++ b/kubernetes/manifest/base/ollama-statefulset.yaml
@@ -20,9 +20,13 @@ spec:
         ports:
         - containerPort: 11434
         resources:
-          limits:
+          requests:
             cpu: "2000m"
             memory: "2Gi"
+          limits:
+            cpu: "4000m"
+            memory: "4Gi"
+            nvidia.com/gpu: "0"
         volumeMounts:
         - name: ollama-volume
           mountPath: /root/.ollama
@@ -34,4 +38,4 @@ spec:
       accessModes: [ "ReadWriteOnce" ]
       resources:
         requests:
-          storage: 1Gi
+          storage: 30Gi
--- a/kubernetes/manifest/base/webui-deployment.yaml
+++ b/kubernetes/manifest/base/webui-deployment.yaml
@@ -19,9 +19,12 @@ spec:
         ports:
         - containerPort: 8080
         resources:
-          limits:
+          requests:
             cpu: "500m"
             memory: "500Mi"
+          limits:
+            cpu: "1000m"
+            memory: "1Gi"
         env:
         - name: OLLAMA_API_BASE_URL
           value: "http://ollama-service.ollama-namespace.svc.cluster.local:11434/api"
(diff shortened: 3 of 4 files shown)


tjbck:
@jnkstr review would be much appreciated!

justinh-rahb (commented):
I am personally not very familiar with the inner workings of k8s, but my boss is. He says it looks sane and reasonable to him if you're doing network storage 👍

bjornjorgensen:
it works but it dont remember the openai key 
image

Where are the config file where the openai key is saved?

justinh-rahb:
> 
> Where are the config file where the openai key is saved?
>

It's stored as an environment variable `OPENAI_API_KEY`, there's also `OPENAI_API_BASE_URL` if you're using something other than `https://api.openai.com/v1`

bjornjorgensen:
ok so `kubectl create secret generic openai-api-key --from-literal=OPENAI_API_KEY=your_openai_api_key_here -n ollama-webui`
````
apiVersion: apps/v1
kind: Deployment
metadata:
  name: ollama-webui-deployment
  namespace: ollama-webui
spec:
  replicas: 1
  selector:
    matchLabels:
      app: ollama-webui
  template:
    metadata:
      labels:
        app: ollama-webui
    spec:
      containers:
      - name: ollama-webui
        image: ghcr.io/ollama-webui/ollama-webui:main
        ports:
        - containerPort: 8080
        resources:
          limits:
            cpu: "500m"
            memory: "500Mi"
        env:
        - name: OLLAMA_API_BASE_URL
          value: "http://ollama-service.ollama-webui.svc.cluster.local:11434/api"
        - name: OPENAI_API_KEY             
          valueFrom:                       
            secretKeyRef:
              name: openai-api-key        
              key: OPENAI_API_KEY  
        tty: true
        
        volumeMounts:
        - name: webui-volume
          mountPath: /app/backend/data
      volumes:
      - name: webui-volume
        persistentVolumeClaim:
          claimName: ollama-webui-pvc
````
it works but now every user can use my openai key..

**Comment by justinh-rahb** [comment]
The API communication to OpenAI API is proxied by the backend. These variables are indeed global.

**Review by jannikstdl — CHANGES_REQUESTED** [review]
LGTM, but i would change the size of the webui storage to minimum 10Gi since the documents uploaded for raf are also stored under /app/backend/data

Speaking of improvements to the k8s yamls it might be worth to mention that the ollama-statefulset.yaml has also a default storage of 1Gi wich is where the LLMs are stored, i would set this definetly higher. 
And deleting the ressource limits of the ollama-deployment.yaml because it only has 500Mi Memory declared there for the WebUI which is also a bit to low, especially for the RAG ingestion process. We had trouble with these limits (#541), so i would suggest to remothe the limits there sincs k8s takes whats there and needed.

I can test this on monday in production, but this should work.
Also if anyone has more experience in helm than me adding these changes there would be nice.
Thanks!

**Comment by jkh1** [comment]
A bit late but the web-ui pvc change looks good to me. I also second putting in sensible defaults for resources. This may also be something to document in a README inside the kubernetes directory. In my case, the web-ui volume is set to 20 Gi because I expect multiple users to upload documents. For the ollama statefulset, I use 250 Gi of which half is already used with 6 models (not the smallest though). For memory, web-ui has 4 Gi and I had to give ollama 80 Gi to accommodate some of the largeish models. I started with 32 Gi and this was definitely not enough.

**Comment by braveokafor** [comment]
Thanks for the input guys, I just pushed a commit to increase resources. 

I don't use this in production at the moment, so I would appreciate feedback on the minimum resources needed to realistically get started for someone just experimenting (so we don't end up in a CrashLoopBackOff or similar). 

It's quite straightforward to increase the resources from the defaults if necessary.

**Comment by justinh-rahb** [comment]
> It's quite straightforward to increase the resources from the defaults if necessary.

IMO anyone that's deploying in production on k8s oughtta know better than to take the helm chart at face value. It should be expected to have to modify it to suit your particular installation, same for docker-compose files. You don't have to and *probably shouldn't* use **exactly** the one provided 🙂

**Comment by braveokafor** [comment]
@justinh-rahb I agree.

We might also need a new discussion for Kubernetes production readiness stuff.
I'm happy to work on the K8s/ Helm bits

**Review by jannikstdl — APPROVED** [review]
Looks good!

Maybe i can test this next week in prod.

**Comment by bjornjorgensen** [comment]
> Thanks for the input guys, I just pushed a commit to increase resources.
> 
> I don't use this in production at the moment, so I would appreciate feedback on the minimum resources needed to realistically get started for someone just experimenting (so we don't end up in a CrashLoopBackOff or similar).
> 
> It's quite straightforward to increase the resources from the defaults if necessary.

if you are afraid of it will use more then your node has of CPU or mem you can change it like this 
```
resources:
      requests:
        cpu: "100m"  # Minimum CPU required by the container to run
        memory: "100Mi"  # Minimum Memory required by the container to run
      limits:
        cpu: "500m"  # Maximum CPU the container can use
        memory: "500Mi"  # Maximum Memory the container can use
```

bjornjorgensen:
> @justinh-rahb I agree.
> 
> We might also need a new discussion for Kubernetes production readiness stuff. I'm happy to work on the K8s/ Helm bits

yes, one thing that we should frist do is to split it out to two namespaces and two helm repos 
have a look at https://github.com/open-webui/open-webui/pull/768#issuecomment-1950315481

braveokafor:
I feel like two separate namespaces might be a bit overkill, what's the rationale?
I did it [this](https://github.com/braveokafor/ollama-webui-helm/tree/main/charts/ollama-webui-helm/templates) way in my own chart.

tjbck:
Is this ready to be merged?

braveokafor:
@tjbck Yes, the other changes we are discussing will be carried out in separate PRs.

tjbck:
Great, Thanks for the hard work!

PR #771, opened 2024-02-17
feat: windows native start script

tjbck:


domsleee:
Hey, thanks for adding this script 😄 

This line, it is supposed to use the environment variable `PORT` falling back to `8080`, right? It doesn't seem to work for me:
```cmd
SET "PORT=%PORT:8080%"
```

When I try using `PORT` it in powershell:
```powershell
$env:PORT='3000' && .\start_windows.bat
Usage: uvicorn [OPTIONS] APP
Try 'uvicorn --help' for help.

Error: Invalid value for '--port': 'PORT:8080' is not a valid integer.
```

Maybe it should be this instead?
```cmd
IF "%PORT%"=="" SET PORT=8080
```

justinh-rahb:
Very possible @domsleee, we're not native Windows users here 🙃 The project really appreciates people kicking the tires on this.

PR #772, opened 2024-02-17
feat: choose embedding model when using docker

jannikstdl:
**Changes**
- Added the functionallity to change the default embedding model used (all-MiniLM-L6-v2) in the Dockerfile as an ENV variable
- You can now use any sentence-transformer embedding model which can be found here: https://huggingface.co/models?library=sentence-transformers
- default model `all-MiniLM-L6-v2` has low performance nowerdays and supports only english:
image
- `all-MiniLM-L6-v2` is still default, please read the important info down below!
- You can change the model to `intfloat/multilingual-e5-large` which is one of the most powerful embedding models aviable (You can also use the instruct version of the multilingual-e5-large `intfloat/multilingual-e5-large-instruct` which is smaller and almost better as the latest from OpenAI ["text-embedding-large-3"](https://openai.com/blog/new-embedding-models-and-api-updates))
image
- Embedding models are preloaded in the Dockerfile (like the Whisper TTS model) for "no-internet" support.
- deleted some unused imports in the RAG `main.py`

**Improvements**
In my local testing (in german) the output with `intfloat/multilingual-e5-large` as the embedding model are way more accurate to my question given to the LLM.
Also the load times of the RAG were much shorter, i don't know why and whether that is the case for you too. Maybe you can test this and give feedback.

**Open Points**
- Support not only for sentence-transformer models (e.g. bert)
- Store the embedding model in `/app/backend/data/cache`
for now the model dir should be  `~/.cache/torch/sentence_transformers` which is not in the PVC
I didn't find a parameter to change the location here `sentence_transformer_ef = embedding_functions.SentenceTransformerEmbeddingFunction(model_name=SENTENCE_TRANSFORMER_EMBED_MODEL)`
Insted of giving the model name as a string to be downloaded you can also give the param model_name a path. Maybe this would be the way to go.
- there are parameters fot the embedding model and whisper tts model in the preloading to set the device type to `cpu` (our standard for both) ot `cuda` which can lead to better performance when using nvida gpus. We could also set a ENV to change that to the users specific needs
- Maybe implementing an "Auto-re-embed" when the embedding model changes

### **Important Information**
If you have some documents stored under the `/documents` route, changing the embedding model will cause the backend to not be able to read the files. So if you have a lot of docs beeing used for RAG let that by default or re-embed your files.
 I also mentianed this in a Dockerfile comment. 

@tjbck i don't know if
`# wget embedding model weight from alpine (does not exist from slim-buster)
RUN wget "https://chroma-onnx-models.s3.amazonaws.com/all-MiniLM-L6-v2/onnx.tar.gz" -O - | \
    tar -xzf - -C /app
`
in the Dockerfile was ever used, but with this update the `all-MiniLM-L6-v2` is declared in the ENV so it is obsolete imo.
If this is the case, deleting this RUN statement would save some buildtime + image size.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,9 +30,15 @@ ENV WEBUI_SECRET_KEY ""
 ENV SCARF_NO_ANALYTICS true
 ENV DO_NOT_TRACK true
 
-#Whisper TTS Settings
+# whisper TTS Settings
 ENV WHISPER_MODEL="base"
 ENV WHISPER_MODEL_DIR="/app/backend/data/cache/whisper/models"
+
+# any sentence transformer model; models to use can be found at https://huggingface.co/models?library=sentence-transformers
+# Leaderboard: https://huggingface.co/spaces/mteb/leaderboard 
+# for better persormance and multilangauge support use "intfloat/multilingual-e5-large"
+# IMPORTANT: If you change the default model (all-MiniLM-L6-v2) and vice versa, you aren't able to use RAG Chat with your previous documents loaded in the WebUI! You need to re-embed them.
+ENV DOCKER_SENTENCE_TRANSFORMER_EMBED_MODEL="all-MiniLM-L6-v2"
 
 WORKDIR /app/backend
 
@@ -48,7 +54,9 @@ RUN apt-get update \
     && apt-get install -y pandoc netcat-openbsd \
     && rm -rf /var/lib/apt/lists/*
 
-# RUN python -c "from sentence_transformers import SentenceTransformer; model = SentenceTransformer('all-MiniLM-L6-v2')"
+# preload embedding model
+RUN python -c "import os; from chromadb.utils import embedding_functions; sentence_transformer_ef = embedding_functions.SentenceTransformerEmbeddingFunction(model_name=os.environ['DOCKER_SENTENCE_TRANSFORMER_EMBED_MODEL'])"
+# preload tts model
 RUN python -c "import os; from faster_whisper import WhisperModel; WhisperModel(os.environ['WHISPER_MODEL'], device='cpu', compute_type='int8', download_root=os.environ['WHISPER_MODEL_DIR'])"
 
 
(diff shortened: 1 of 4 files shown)


justinh-rahb:
Going to take a deeper dive into the changes when I get a chance later but this will be an awesome feature once merged, thanks!

RWayne93:
Hey this is awesome I was just thinking of trying to add this myself. The current RAG implementation doesn't use any embedding models (at least last time i checked the source) aside from whatever Chromadb does when you upload documents. 

Aside from having an environment variable for docker I think it would be great to pick an embedding model in the settings from the front end web client as well. Similar to how you set what language model you would like to use.

jnkstr:
Do not merge yet. 

This needs special attention to the query collection route. If I do so I get a 500 error in the console.

jnkstr:
@tjbck fixed the error. Just a false naming of the collection name variable in the /query/collection route.

I also merged the conflicts with the new changes regarding to the docs dir.

Dockerfile
```diff
@@ -48,7 +54,9 @@ RUN apt-get update \
     && apt-get install -y pandoc netcat-openbsd \
     && rm -rf /var/lib/apt/lists/*
 
-# RUN python -c "from sentence_transformers import SentenceTransformer; model = SentenceTransformer('all-MiniLM-L6-v2')"
+# preload embedding model
+RUN python -c "import os; from chromadb.utils import embedding_functions; sentence_transformer_ef = embedding_functions.SentenceTransformerEmbeddingFunction(model_name=os.environ['RAG_EMBEDDING_MODEL'])"
```
tjbck, line 58: We might want to save the models to `/data/cache/embedding/models`!

tjbck (changes requested):

Dockerfile
jannikstdl: Ok, added an ENV Variable in the Dockerfile to store the embedding models in this folder, also i added the option to change the device type if possible.

Maybe we should automate the re-imbedd process in future with a script executed in the Dockerfile, because as mentioned changing the embedding model needs to re-imbedd importet docs.

jannikstdl:
Thanks, you are genius!

jiayaoO3O:
Hi, I would like to know if it is possible to specify the environment variable 'RAG_EMBEDDING_MODEL' as 'none' when using docker containers to skip downloading the embedded model. This is because in areas with restricted internet access (such as China), the container fails to start as it is unable to download the model from [huggingface.co](http://huggingface.co/) without using a proxy. The only effect I want is for the UI to connect to ollama without the need to download the embedding model.

PR #789, opened 2024-02-18
feat: delete message

### LINKED ISSUE #652: Feat: delete parts of conversation.
I often face a situation when a discussion diverges because a prompt mistake or necessity of some side questions. This adds unneeded parts to the context that I would rather remove. 

A simple delete X button that would remove a chat bauble all together. I am not sure if would make sense to delete entries in pairs (as prompt and response) or separately.

I think this would greatly help users maintain a shorter and clean context, to avoid context fading, but also for archiving purposes.

dannyl1u:
# Implement Delete Message Functionality

## Description
This pull request introduces functionality allowing users to delete their messages from the chatbot conversation. This feature aims to improve the user experience by enabling the management of conversation flow, particularly useful for removing side questions, clarifications, or typos.

## Motivation
Implementing message deletion addresses the need for clearer conversational context, as highlighted in issue #652. Users can now manage their conversation history more effectively, ensuring that only relevant messages influence the chatbot's responses and overall dialogue quality.

## Implementation Details
**Frontend Changes**:
- Users can delete any message submitted after the initial prompt, a "Delete" (🗑️) button has been added next to each user's prompt in the chat interface, courtesy of @tjbck.
- Although deleted from the visible chat history, these messages are retained within the `history` object with a `deleted=true` flag.
- Messages flagged as `deleted=true` are excluded from the payload in subsequent API calls, ensuring that deleted messages do not influence the conversation flow.
- Upon deletion, the UI immediately updates to reflect the removal of both the user's message and the corresponding chatbot response, maintaining a clean and relevant conversation thread.

## Demo video
https://youtu.be/UklUb_QMrZM

Resolves #652


dannyl1u:
# Implement Delete Message Functionality

## Description
This pull request introduces functionality allowing users to delete their messages from the chatbot conversation. This feature aims to improve the user experience by enabling the management of conversation flow, particularly useful for removing side questions, clarifications, or typos.

## Motivation
Implementing message deletion addresses the need for clearer conversational context, as highlighted in issue #652. Users can now manage their conversation history more effectively, ensuring that only relevant messages influence the chatbot's responses and overall dialogue quality.

## Implementation Details
**Frontend Changes**:
- Users can delete any message submitted after the initial prompt, a "Delete" (🗑️) button has been added next to each user's prompt in the chat interface, courtesy of @tjbck.
- Although deleted from the visible chat history, these messages are retained within the `history` object with a `deleted=true` flag.
- Messages flagged as `deleted=true` are excluded from the payload in subsequent API calls, ensuring that deleted messages do not influence the conversation flow.
- Upon deletion, the UI immediately updates to reflect the removal of both the user's message and the corresponding chatbot response, maintaining a clean and relevant conversation thread.
- `cancelChatCompletion()` is called if user deletes the most recently sent prompt 

## Demo video
https://youtu.be/UklUb_QMrZM

src/lib/components/chat/Messages.svelte
```diff
@@ -221,6 +222,19 @@
 			scrollToBottom();
 		}, 100);
 	};
+
+	const deleteMessagePair = async (messageId) => {
+		history.messages[messageId].deleted = true;
+		history.messages[history.messages[messageId].childrenIds[0]].deleted = true;
```
tjbck, line 228: Thanks for the PR! I believe this should apply to all ancestor messages recursively, not just the first one.

src/lib/components/chat/Messages.svelte
```diff
@@ -221,6 +222,19 @@
...
+
+		const responseId = history.messages[messageId].childrenIds[0];
+		if (history.messages[responseId].childrenIds.length === 0) {
```
tjbck, line 231: this part should probably use `message.done` instead of checking for number of children.

tjbck (changes requested):

src/lib/components/chat/Messages.svelte
dannyl1u: > Thanks for the PR! I believe this should apply to all ancestor messages recursively, not just the first one.

Do you mean we should also be removing all subsequent message/response pairs after the deleted message as well?

src/lib/components/chat/Messages.svelte
dannyl1u: The logic for this is that we should not cancel the chat completion if the deleted message isn't the most recently sent.

example:
User: Prompt1
Llama: Response1
User: Prompt2
Llama: Response2
User: Prompt3
Llama: generating...

In the above scenario, if the user deletes Prompt2 we still want to continue to generate the Response to Prompt3

src/lib/components/chat/Messages.svelte
tjbck: I think got a bit confused as I was leaving the comment, I meant `delete` flag should be set to true for all children messages, right now this won't delete any of the regenerated responses.

tjbck: Also, another thing, it might be better to directly manipulate the history graph object instead of flagging the messages but this should do the job for now!

src/lib/components/chat/Messages.svelte
tjbck: Wouldn't the following do the same job? 

```js
const childrenIds = history.messages[messageId].childrenIds;
if (!history.messages[childrenIds.at(-1)].done) {
	await cancelChatCompletion(localStorage.token, currentRequestId);
}	
```
Also, the second param should be `currentRequestId` not `chatId` 😅 it might be better to remove this part of code entirely, and we can introduce this feature in a separate PR.

dannyl1u: I agree, sounds good, thanks for pointing that out!

src/lib/components/chat/Messages.svelte
dannyl1u: Updated to set `delete=true` flag to all children messages recursively

dannyl1u:
Just realized we should probably ask for user confirmation before deletion. We can reuse the same confirmation UI that's currently being used for deleting entire chats. 

Let me know what you think @tjbck

tjbck:
@dannyl1u Thanks for the hard work, I'll merge this for now! Delete confirmation feature sounds like it would be nice to have, feel free to make a separate PR.

tjbck:
It's not a priority atm, but we might want to refactor this to manipulate the history graph object directly later on.

dannyl1u:
> @dannyl1u Thanks for the hard work, I'll merge this for now! Delete confirmation feature sounds like it would be nice to have, feel free to make a separate PR.

sounds good, thanks!

PR #815, opened 2024-02-20
feat: jwt utils

### LINKED ISSUE #566: feat: jwt expiration setting
It would be a bit more secure if the token has a expiration time. 

I saw in the code that the "create_token" code accepts an expiration but it is not used.

The user gets logged out after a inactive period of time. Maybe you can configure and turn that off in the settings.

tjbck:
Resolves #566


jannikstdl:
Nice! Tested it, maybe this should recreate the token in the localstorage since setting this from the admin panel itself don't change the setting directly. You need to delete the token and login.

PR #838, opened 2024-02-21
feat: stable diffusion integration

tjbck:
Resolves #696

> [!important]
> AUTOMATIC1111/stable-diffusion-webui should be run using `sh webui.sh --api` command with `--api` flag.

image

image

### Changes
image
- Advanced section has been removed and relocated to General.
- Ollama API URL Settings and OpenAI External Setting have been moved to Connections section.

PR #839, opened 2024-02-21
Add release creation workflow

### LINKED ISSUE #255: feat: create releases with ollama-webui version numbers
I would like to package Ollama-WebUI for the NixOS Linux distribution , but missing versions would make it difficult for users to distinguish newer versions from older ones (currently, we only have Git hashes).

Create releases or pseudo-releases with version numbers (as frequent as feasible and with whatever breaking changes, changelog or version numbers as you please).

**Additional context**
NixOS or rather the Nix package manager (available for Mac and other Linux distributions) makes it very easy to get and run the app with a single, short terminal call. It will likely be as follows:
```
nix run nixpkgs#ollama-webui
```
Link to NixOS PR: https://github.com/NixOS/nixpkgs/pull/275448

Btw, thanks for this awesome UI (to me it's the best i found so far for Ollama).

Lanhild:
Fixes #255.

This PR add a `build-release.yml` file to automatically create GitHub releases using the commit hash, in it's short form.

This workflow can then be adapted when Open WebUI will use semantic versioning, to use the `package.json` version number.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -13,11 +13,18 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v2
 
-    - name: Get commit hash
-      id: get_hash
+    - name: Check for changes in package.json
       run: |
-        HASH=$(git rev-parse --short "$GITHUB_SHA")
-        echo "::set-output name=hash::$HASH"
+        git diff --cached --diff-filter=d package.json || {
+          echo "No changes to package.json"
+          exit 1
+        }
+
+    - name: Get version number from package.json
+      id: get_version
+      run: |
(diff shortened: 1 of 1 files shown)


tjbck:
Thanks for the PR! We started using `package.json` for versioning from #838 btw, would this PR create releases if I change the version number from `package.json`?

Lanhild:
@tjbck Just made a new commit.

The workflows checks for diff in the `package.json` file, and uses the version number in the `version` field to tag the release.

You simply need to add in this repository secrets the `TOKEN` GitHub Access Token for the workflow to work.

tjbck:
Great! LGTM, thanks!

PR #848, opened 2024-02-22
Adding Makefile and LLM update script

Patrice-Gaudicheau:
This Pull Request presents a Makefile and LLM Update Script designed to improve the workflow of managing and updating Large Language Models (LLMs) in Docker containers.

### Main characteristics

- **Makefile:** Provides a set of commands for automated management of Docker containers, making it easier to manage common tasks such as starting, stopping, and deleting containers. The Makefile includes the following commands:
    - `install`: Starts containers using `docker-compose up -d`.
    - `remove`: Stop and remove containers and volumes with `docker-compose down -v`.
    - `start`/`stop`: Commands to manage the state of containers.
    - `update`: Pulls the latest changes from the git repository, rebuilds and restarts the containers. This also includes running the new LLM update script to ensure that all installed LLMs are up to date.

- **LLM Update Script (`update_llm.sh`):** Automates the process of retrieving and updating each LLM installed in the Docker container. This script dynamically retrieves the list of installed LLMs and updates them with the latest versions, ensuring that the models are up to date and reducing the risk of version-related issues.

### Motivation

The motivation behind this contribution is to streamline the LLM management process within Dockerized environments.

By automating container updates and management, developers can focus more on development and less on maintenance tasks.

Additionally, it is crucial to ensure that LLMs are kept up to date to take advantage of the latest security improvements and patches.

This contribution is aimed at developers and teams working with LLMs in Docker containers, providing them with tools to improve their efficiency and the maintainability of their projects.


justinh-rahb:
I think this PR idea is great in principle, but I just so happen to be working on my own PR that will address this, in a different way (not using Make for one thing).

Makefile
```diff
@@ -0,0 +1,22 @@
...
+stop:
+	@docker-compose stop
+
+update:
+	# Calls the LLM update script
+	@./update_llm.sh
+	@git pull
+	@docker-compose down
+	# Make sure the ollama-webui container is stopped before rebuilding
+	@docker stop ollama-webui || true
```
tjbck, line 19: We should change the name to `open-webui` here.

update_llm.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# update_llm.sh
```
tjbck, line 2: Perhaps `update_ollama_models.sh` would be more appropriate and descriptive for the name.

tjbck (changes requested):
Thanks for the PR, the rest looks good to me.

jannikstdl:
> I think this PR idea is great in principle, but I just so happen to be working on my own PR (it's just a gist atm) that will address this, in a different way (not using Make for one thing).
> 

Yes but I think the Makefile addresses an other approach since the `make update` command might come in handy.

Makefile
Patrice-Gaudicheau: Good idea, the change has been made

update_llm.sh
Patrice-Gaudicheau: Great suggestion, file name changed

Makefile
```diff
@@ -0,0 +1,25 @@
+install:
```
jannikstdl, line 1: You could add 

```
install-gpu: 
    docker-compose -f docker-compose.yaml -f docker-compose.gpu.yaml up -d --build
```

For GPU support

tjbck:
I'll merge this to dev branch, feel free to make additional PRs if anything needs to be changed, Thanks!

PR #855, opened 2024-02-23
Use semantic versioning

Lanhild:
Instead of having the previous `1.0.0-alpha.101`, we could use [semantic versioning](https://semver.org/) as a way to respect global conventions.

In our case, the current version is `1.0.0-alpha.101`. We could consider this a prerelease, so the semantic version string would be `0.1.0-101`

For more info about semantic versioning:
- https://semver.org/
- https://jubianchi.github.io/semver-check/#/version/0.1.0-101 | Allows you to preview a version string and explain each of it's components

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "open-webui",
-	"version": "1.0.0-alpha.101",
+	"version": "0.1.0-101",
 	"private": true,
 	"scripts": {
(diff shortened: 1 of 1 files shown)


justinh-rahb:
following semver will make a lot of CI/CD tasks easier, I fully support this change.

tjbck:
LGTM, Thanks!

PR #863, opened 2024-02-23
config: allow custom default user role

duhow:
For internal (restricted) websites, allow customize the `DEFAULT_USER_ROLE` with ENV values (set to `user` or `admin` for example).
Since at the moment we need to manually enable the user (`pending` -> `user`), I want to allow self-register for quick access.

### DIFF
--- a/backend/config.py
+++ b/backend/config.py
@@ -187,7 +187,7 @@ DEFAULT_PROMPT_SUGGESTIONS = (
 )
 
 
-DEFAULT_USER_ROLE = "pending"
+DEFAULT_USER_ROLE = os.getenv("DEFAULT_USER_ROLE", "pending")
 USER_PERMISSIONS = {"chat": {"deletion": True}}
 
 


duhow:
Ah, just noticed that latest version does allow changing the default value...

image

justinh-rahb:
Does your modification allow setting that with an envvar now? If so I'd say LGTM, doesn't hurt to have options.

duhow:
Yes, that's the idea, allow customizing value optionally. It defaults to pending if not set.

jannikstdl:
Maybe you cloud ad this as a env to the Dockerfile / as a K8s value too since not to many are looking into the code

duhow:
Sounds good @jannikstdl , I can work on that in future PRs once we address the current ones :)

justinh-rahb:
> Sounds good @jannikstdl , I can work on that in future PRs once we address the current ones :)

We're going to try and clear some of the backlog this weekend ☝️

PR #866, opened 2024-02-23
fix: RAG scan unsupported mime type

asedmammad:
## Description
This fixes an issue with RAG scan that stops loading documents as soon as it reaches a file with unsupported mime type (or any other exceptions).
When the mime type is not supported, the scan operation stops with `'NoneType' object has no attribute 'find'`.

Sample unsupported mime type: [triage_process.graffle](https://github.com/django/django/blob/main/docs/internals/_images/triage_process.graffle)

## Explanation of changes
- Check if `file_content_type` is not None before searching for `text/` and use `TextLoader`
- Move try/except into the for loop so that we continue loading next files even if loading a document fails.
In case of the `triage_process.graffle` we fallback to TextLoader, but it seem to fail with loading, so we continue with the next document.


tjbck:
LGTM, Thanks!

PR #870, opened 2024-02-23
Fix Kubernetes Manifest for Web UI

ching-kuo:
There are currently 2 issues with the Kubernetes manifest:

- open-webui-deployment refers to ollama-webui-pvc but not included in kustomization.yaml, causing issue if using GPU deployment
- ollama-webui-pvc is using the wrong namespace

This commit fixes both issue.

### DIFF
--- a/kubernetes/manifest/base/webui-pvc.yaml
+++ b/kubernetes/manifest/base/webui-pvc.yaml
@@ -4,7 +4,7 @@ metadata:
   labels:
     app: ollama-webui
   name: ollama-webui-pvc
-  namespace: ollama-namespace
+  namespace: open-webui
 spec:
   accessModes: ["ReadWriteOnce"]
   resources:
--- a/kubernetes/manifest/kustomization.yaml
+++ b/kubernetes/manifest/kustomization.yaml
@@ -5,6 +5,7 @@ resources:
 - base/webui-deployment.yaml
 - base/webui-service.yaml
 - base/webui-ingress.yaml
+- base/webui-pvc.yaml
 
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization


tjbck:
@jannikstdl @duhow review wanted!

jannikstdl:
@tjbck LGTM, you can merge this. 


There was indeed an old naming declared as a namespace. This is a important fix!

But I don't really get the point of the [https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/resource/](ressources),at all I think you could delete that but anyway i am not familiar with this kustomize tool and it doesn't hurt to let that in.  
Was this build by @dnviti ?

ching-kuo:
@jannikstdl 
If you don't include the file in `kustomization.yaml`, it will not be applied when running `kubectl apply -k ./kubernetes/manifest` as the instruction in [INSTALLATION.md](https://github.com/open-webui/open-webui/blob/main/INSTALLATION.md?plain=1#L12)

jannikstdl:
Ok get it, it takes the base folder and the GPU patch. 

I use CPU only for now I don't needed that. 

Would in not be better to include 
`base/` as ressource for (maybe) future added yamls?
If you look at the documentation link I posted it should be possible to add entire folders.

ching-kuo:
`Each entry in this list must be a path to a file, or a path (or URL) referring to another kustomization directory`

From my understanding you can only include a kustomization directory, but not a directory with only resources.

Here's what I get if I include a directory

```error: accumulating resources: accumulation err='accumulating resources from 'base/': '<some_path>/base' must resolve to a file': couldn't make target for path '<some_path>/base': unable to find one of 'kustomization.yaml', 'kustomization.yml' or 'Kustomization' in directory '<some_path>/base'```

**Comment by jannikstdl** [comment]
Ok then wo should have an eye on this kustomization.yaml

Thanks for the PR!

**Review by duhow — APPROVED** [review]
The `namespace` should not be defined in any resource at all, then whoever wants to deploy all resources, can define any desired namespace.

```sh
kubectl apply -n llama -k .
```

But for this PR in particular, it does fix the deployment, so go for it.

jannikstdl (approved):

tjbck:
Thanks for reviewing guys!

Issue #878, opened 2024-02-23
infra: Enhance releases workflow and integration


Lanhild:
We've discussed in the past days about ways to refine and enhance the release process for Open WebUI. 

Currently, there are automated GitHub release and Docker image publishing workflows. This is a great first step, but there is the need for something more polished and automated, especially as the project and it's community grows.

I found [Semantic-release](https://github.com/semantic-release/semantic-release). It seems like it is the best-fitting tool for this issue and fits most, if not all, our needs in term of CI.

Below is a transcription of a related Discord conversation on the Open WebUI server:

```
Lanhild — Yesterday at 7:31 PM
where are the release notes though? in a CHANGELOG file?

wildtang3nt — Yesterday at 7:34 PM
oh I meant these when you press the generate button, if there were any PRs in my branch since the last release, they'd also be listed there above the automatic changelog link

Lanhild — Yesterday at 7:35 PM
Well, we could have a script in the package.json that generates release notes in a CHANGELOG.md file, and then adapt the release workflow to take the content of the CHANGELOG since last version

wildtang3nt — Yesterday at 7:36 PM
for sure, there's definitely more than one way to go about it, I'm just spitballing ideas, that was your PR so I'm happy to let you run with it how you think would work best 🙂

Lanhild — Today at 5:57 PM
@wildtang3nt I've tested https://github.com/semantic-release/semantic-release.

From what I understand, it automatically creates a release version based on commits messages. Everything seems configurable, from release notes to release publishing to commit analysis.

See https://semantic-release.gitbook.io/ for their doc.

I think we can't go wrong by using this.

```

I don't have the time to do all the work around this issue, but I can prepare the `package.json` file to fit with [Semantic-release](https://github.com/semantic-release/semantic-release) needs and add the dependency to the project.

h4ckm1n-dev:
I will take a look on it, not sure to have the skill to make it btw

tjbck:
Closing in favour of #2738 2738

PR #914, opened 2024-02-25
fix: add semver to container builds

buroa:
The container builds are missing semver versioning. This should fix it up for you :)

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -52,6 +52,7 @@ jobs:
             type=ref,event=tag
             type=sha,prefix=git-
             type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
           flavor: |
             latest=${{ github.ref == 'refs/heads/main' }}
 


buroa:
@tjbck Easy one for you :)

.github/workflows/docker-build.yaml
```diff
@@ -52,6 +52,7 @@ jobs:
             type=ref,event=tag
             type=sha,prefix=git-
             type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
```
tjbck, line 55: Thanks for the PR! Would `type=semver,pattern={{major}}.{{minor}}.{{patch}}` also work in this case?

onedr0p: That will already happen, no need to add it.

tjbck: Great!

tjbck:
LGTM, Thanks!

justinh-rahb:
Didn't work. Check the currently building Docker image's metadata. We've been struggling with this one for a bit and @Lanhild and I feel we can best solve this by completely overhauling the Release/Build workflows, please see: 
- #878

buroa:
> Didn't work. Check the currently building Docker image's metadata:
> 
> Screenshot 2024-02-25 at 7 27 56 PM
> We've been struggling with this one for a bit and @Lanhild and I feel we can best solve this by completely overhauling the Release/Build workflows, please see:
> 
> * [infra: Enhance releases workflow and integration #878](https://github.com/open-webui/open-webui/issues/878)

It will work once you publish a new release with a new tag.

tjbck:
One way to find out 😄

justinh-rahb:
Not meaning to sound smug, but I *did* try this before in my own branch before. We've got larger issues with our workflow that are known and being considered.

buroa:
@justinh-rahb It looks like these versions are being created by a github action itself; which may be the reason it's not pushing out the builds correctly. Is there a reason it's like this?

chamburr:
When using `GITHUB_TOKEN` to perform tasks, other workflows would not be triggered to "prevent recursive workflow runs". The recommended workaround is to use a Personal Access Token instead. See https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow.

In this case, we would need to replace the token here which is used to create tags:
https://github.com/open-webui/open-webui/blob/6c70d0f770ce4b0acf8d2a33eb7c0f27ffbed631/.github/workflows/build-release.yml#L32

PR #922, opened 2024-02-26
ui: typewriter effect

jannikstdl:
Had this on my local clone. don't know if you like it. Feel free to merge.

https://github.com/open-webui/open-webui/assets/69747628/e28765c0-9e0e-49a2-99fa-77253153885f

Maybe we could ad this whenever title autogeneration is on and the title comes in like in chatgpt.
Something like that:

https://github.com/open-webui/open-webui/assets/69747628/663d67fb-b12d-4bc5-808a-a2fe81dcf2d9

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
 		"katex": "^0.16.9",
 		"marked": "^9.1.0",
 		"svelte-french-toast": "^1.2.0",
+		"svelte-typewriter": "^3.2.3",
 		"tippy.js": "^6.3.7",
 		"uuid": "^9.0.1"
 	}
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -50,7 +50,8 @@
 		"katex": "^0.16.9",
 		"marked": "^9.1.0",
 		"svelte-french-toast": "^1.2.0",
+		"svelte-typewriter": "^3.2.3",
```
tjbck, line 53: We might want to use framework agnostic libraries like `https://www.npmjs.com/package/typewriter-effect` here

tjbck:
I really like the effect but ideally we'd want zero dependency for the effect so I'll close this for now, let's come back to this later. Thanks!

PR #927, opened 2024-02-26
Fix OpenAI integration: allow OpenAI to make chat completion calls with a valid API key

sakkus-ai:
From docker on mac get errors since num_ctx not recognized; remove num_ctx from, and add max_tokens to, body in API calls.

## Pull Request Checklist

- [ ] **Description:** 

---

## Description

Model file calls to the OpenAI API works fine, but chat completions fail withan error related to num_ctx in the body of the call.

backend/openai/main.py has a change to the main route to remove num_ctx from the body; this is replaced with max_tokens.

Chat completions now work.

---

### Fixed

- Bug fix to allow chat completion API calls to OpenAI to work without num_ctx parameter error.

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -52,6 +52,7 @@ jobs:
             type=ref,event=tag
             type=sha,prefix=git-
             type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
           flavor: |
             latest=${{ github.ref == 'refs/heads/main' }}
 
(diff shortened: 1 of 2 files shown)


tjbck:
Correct me if I'm wrong but I believe `num_ctx` and `max_tokens` are not the same?

sakkus-ai:
They don't seem to be.  Looking at the current API reference (https://platform.openai.com/docs/api-reference/chat/create) there is no mention of num_ctx.

Also, I found that num_ctx generates an error.

I could be missing something though.

sakkus-ai:
My suspicion is that num_ctx refers to model context length; perhaps at some point it was necessary to set that in the body?

From the OpenAi docs that may not be necessary or desirable now?

max_tokens is set to allow for a decent number of tokens for a conversation... but I guess my number is a bit arbitrary

tjbck:
If that's the case, I believe it's best if we just remove `num_ctx` from the body instead of replacing with `max_token`!

tjbck:
LGTM, Thanks!

sakkus-ai:
Good news. Thanks very much!

Issue #936, opened 2024-02-27
feat (small ux improvement): Auto-focus on prompt input area when creating a chat or opening an existing chat


incyclum:
**Is your feature request related to a problem? Please describe.**

When I click on "New Chat" or open an existing chat,
I want to type right away in the chat, but I have first to click in the prompt input area.

**Describe the solution you'd like**

Autofocus on the field when opening a new or existing chat.

Also observed that:
- OpenAi ChatGPT UI does it.
- Google Gemini UI does not.
- Microsoft Copilot does not.

**Describe alternatives you've considered**

None.

🙌🫶

incyclum:
Thank you! That was quick! 👏

PR #941, opened 2024-02-27
fix: added focus selector

jannikstdl:
adresses #936 

This currently only focuses the input when switching between chats and reloading the root page.
When i start a chat and then klick on "New Chat" in the Sidebar is somehow doesn't focus the input. @tjbck any ideas on this?


tjbck:
Thanks for getting the ball rolling for this feature!

PR #980, opened 2024-03-01
ui: auth darkmode

### LINKED ISSUE #979: Update /auth styling to be consistent with the rest of the UI
Occasionally I'm accessing the webui from a new device or incognito etc and need to authenticate. The background is white which isn't easy on the eyes, and varies from the color scheme of the rest of the application

Ideally (imo) the signin/signup pages would have the same styling as the rest of the application

I use darkreader but not on all browsers or on mobile

**Additional context**
I'm happy to agree to disagree on this change, but I would like to see it personally. I also can work on a PR to do this myself as I don't imagine that it would be that much effort, even if it is somewhat complicated for some reason I'd be happy to do it

jannikstdl:
Resolves #979 

image

### DIFF
--- a/src/routes/auth/+page.svelte
+++ b/src/routes/auth/+page.svelte
@@ -8,6 +8,7 @@
 
 	let loaded = false;
 	let mode = 'signin';
+	let isSubmitting = false;
 
 	let name = '';
 	let email = '';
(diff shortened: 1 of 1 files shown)


tjbck:
Thanks for the PR and the effort! but I'll close this in favour of the other implementation with https://github.com/open-webui/open-webui/commit/6f4a863562a6c34fce97e3990574ff6a65987a79.

PR #985, opened 2024-03-01
ui: tooltips shadcn svelte

jannikstdl:
**Minimal implementation/draft of svelte shadcn tooltips for proposed change.**

https://github.com/open-webui/open-webui/assets/69747628/f58292f4-0b2a-4d77-b650-d8ae049ccff4

This commit implements a minimal UI change for the proposed feature. The change introduces new Tooltips that allows users to see the action when hovering over some UI elements. This will improve the user experience by making it more declarative.

Git automatically changed filetypes from LF to CRLF, therefore some files show more line changes than there actually (visibly) are.


jannikstdl:
Created a cleaner one #986 because git changed from LF to CRLF which made the changed files look unclear.

PR #992, opened 2024-03-01
show chat in browser title + scroll to bottom

### LINKED ISSUE #957: feat (ux): Focus on latest message when opening an existing chat

Just noticed another small improvement that could be implemented,
after updating to see the amazing prompt autofocus from #936 🎉❤️...

When I open an existing chat conversation,
I have to scroll down to the latest message,
which is kinda annoying when you're continuing a recent conversation.

And I noticed this because now having the prompt autofocus emphasized that seeing a
blinking cursor waiting for you to type but being at the top of the conversation thread
doesn't makes much sense. Imagine opening your WhatsApp chats and being at the top...

Compared to ChatGPT UI where the chat is already focused on the latest message.

Like on ChatGPT, focus on latest message of a chat, so user don't have to scroll down.

When going back and forth between chats in the same work session (in the same Open WebUI tab session),
It may be more useful to stay where the user left instead of latest message.

E.g:

1. Leave "Chat A" somewhere in the middle, e.g. you're reading message 30/50
2. Create a new "Chat B", type right away in the prompt thanks to autofocus :D
3. Go back to "Chat A", instead of being on the latest message (or 1st one) you could stay at 30/50

BUT if I close the tab, log-out, enter a new session, I think opening "Chat A" the next day or so makes more sense focusing on latest message.

Open to discussion or any other ideas :)

Thanks!

jannikstdl:
Resolves #957

https://github.com/open-webui/open-webui/assets/69747628/c9c85e80-5fcc-49a0-bc81-5b0da5a44cfd

### DIFF
--- a/src/routes/(app)/+page.svelte
+++ b/src/routes/(app)/+page.svelte
@@ -69,7 +69,6 @@
 	let tags = [];
 
 	let title = '';
-	let pageTitle = WEBUI_NAME;
 	let prompt = '';
 	let files = [];
 	let messages = [];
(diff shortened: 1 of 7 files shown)


tjbck:
LGTM, Thanks as always!

PR #998, opened 2024-03-02
Fix enable signup env

chamburr:
Convert the sign up environmental variable to a bool, otherwise sign up cannot be disabled.


tjbck:
Good catch, Thanks!

PR #1004, opened 2024-03-03
refac: delete message logic

dannyl1u:
## Pull Request Checklist

- [✓] **Description:** Briefly describe the changes in this pull request.
- [✓] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [✓] **Documentation:** Have you updated relevant documentation?
- [N/A] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?

---

## Description

EDIT (03/04/2024): We decided that the best user experience is to only delete the message/response pair (not all subsequent messages). This pull request manipulates the message `history` object to remove the deleted message/response pair and update the parent/child message relationships.  Refer to below diagram by @tjbck 
image

---

### Changed
- Refactored the message deletion logic by directly manipulating the `history` object, removing message/response pairs and updating parent/child relationships to replace the existing implementation that uses a `isDeleted` flag.

### Removed
- N/A

### DIFF
--- a/src/lib/components/chat/Messages.svelte
+++ b/src/lib/components/chat/Messages.svelte
@@ -228,10 +228,12 @@
 		const messageParentId = messageToDelete.parentId;
 		const messageChildrenIds = messageToDelete.childrenIds ?? [];
 
+		const hasSibling = messageChildrenIds.some(childId => history.messages[childId]?.childrenIds?.length > 0);
+
 		messageChildrenIds.forEach((childId) => {
 			const child = history.messages[childId];
 			if (child && child.childrenIds) {
-				if (child.childrenIds.length == 0) { // if last prompt/response pair
+				if (child.childrenIds.length === 0 && !hasSibling) { // if last prompt/response pair				
 					history.messages[messageParentId].childrenIds = []
 					history.currentId = messageParentId;
 				}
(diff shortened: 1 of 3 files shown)


src/lib/components/chat/Messages.svelte
```diff
@@ -223,32 +223,21 @@
...
-
 	const messageDeleteHandler = async (messageId) => {
-		if (history.messages[messageId]) {
-			history.messages[messageId].deleted = true;
+		const messageParentId = history.messages[messageId]?.parentId;
 
-			for (const childId of history.messages[messageId].childrenIds) {
-				history.messages[childId].deleted = true;
-			}
+		if (messageParentId !== null) {
```
tjbck, line 229: Thanks for the PR, correct me if I'm wrong but shouldn't this part only delete the provided messageId from the list? (e.g. `history.messages[messageParentId].childrenIds.filter((id) => id !== messageId)`)

dannyl1u: `history.messages[messageParentId].childrenIds = []` 
This will delete all subsequent messages in addition to the deleted message itself which is what we want right?

tjbck:
I've refactored the code to make it more readable, let me know if there are any edge cases I've missed!

tjbck:
Encountering issue with the refac, taking a look right now

tjbck:
Reverted the change, at least it seems to work 😅 Thanks for the effort!

PR #1023, opened 2024-03-03
add fake_useragent to get a realistic header for rag/web requests

fbirlik:
## Description

Add FakeUserAgent package for realistic headers in Rag/Web requests

---

### Added

- fake_useragent added to backend/requirements.txt

### Fixed

- Rag/Web requests have random user agents instead of empty user agent


tjbck:
LGTM, Thanks!

Issue #1020, opened 2024-03-03
refac: Multiple document.getElementById accross the codebase


Carlos-err406:
# Bug Report

## Description

**Bug Summary:**
there are many document.getElementById across the code see img.
this has performace and code cleanness issues as well as a safety impact, see `additional information` section below.

**Steps to Reproduce:**
-

**Expected Behavior:**
use a bind:this to the needed elements to save the element into a store so is more accessible, does less queries to the dom and keeps the code cleaner

**Actual Behavior:**
there are many queries to the DOM that could be avoided

## Environment

- **Operating System:** Windows
- **Browser (if applicable):** not applicable

## Reproduction Details

**Confirmation:**

- [x] I have read and followed all the instructions provided in the README.md.
- [x] I have reviewed the troubleshooting.md document.
- [ ] I have included the browser console logs.
- [ ] I have included the Docker container logs.

## Logs and Screenshots
![image](https://github.com/open-webui/open-webui/assets/81443707/913eeb15-9b54-4ab7-958c-2c69007497f5)

**Browser Console Logs:**
[Include relevant browser console logs, if applicable]

**Docker Container Logs:**
[Include relevant Docker container logs, if applicable]

**Screenshots (if applicable):**
[Attach any relevant screenshots to help illustrate the issue]

## Installation Method

[Describe the method you used to install the project, e.g., manual installation, Docker, package manager, etc.]

## Additional Information
Storing an element in a Svelte store offers several advantages over using document.getElementById throughout a codebase:
- Performance: Storing the element in a store allows you to access it directly without needing to repeatedly query the DOM. This can improve performance by reducing the number of DOM queries and updates.
- Safety: Storing the element in a store provides a centralized way to manage and access the element, reducing the risk of errors or inconsistencies that can occur when using document.getElementById in multiple places.
- Code cleanliness: Storing the element in a store can make your codebase cleaner and more maintainable by encapsulating the logic for accessing and manipulating the element in one place.

## Note

If the bug report is incomplete or does not follow the provided instructions, it may not be addressed. Please ensure that you have followed the steps outlined in the README.md and troubleshooting.md documents, and provide all necessary information for us to reproduce and address the issue. Thank you!

Carlos-err406:
i could contribute on this if thats ok

tjbck:
[...]

PR #1025, opened 2024-03-03
rf: less dom queries using bind:this

Carlos-err406:
## Pull Request Checklist

---

## Description

addresses issue #1020 about querying the DOM

refactored the dom queries (document.getElementById) for a more svelte-ish approach using bind:this

---

### Added

-

### Fixed

- centralized the way to access the dom elements in a more efficient and clean manner

### Changed

- the way to query the dom

### Removed

- some of the manual dom queries accross the codebase

### DIFF
--- a/src/lib/components/chat/MessageInput.svelte
+++ b/src/lib/components/chat/MessageInput.svelte
@@ -19,7 +19,7 @@
 
 	export let suggestionPrompts = [];
 	export let autoScroll = true;
-	let chatTextArea:HTMLTextAreaElement
+	let chatTextAreaElement:HTMLTextAreaElement
 	let filesInputElement;
 
 	let promptsElement;
@@ -43,9 +43,9 @@
 	let speechRecognition;
 
 	$: if (prompt) {
-		if (chatTextArea) {
-			chatTextArea.style.height = '';
-			chatTextArea.style.height = Math.min(chatTextArea.scrollHeight, 200) + 'px';
+		if (chatTextAreaElement) {
+			chatTextAreaElement.style.height = '';
+			chatTextAreaElement.style.height = Math.min(chatTextAreaElement.scrollHeight, 200) + 'px';
 		}
 	}
 
@@ -84,7 +84,7 @@
 			if (res) {
 				prompt = res.text;
 				await tick();
-				chatTextArea?.focus();
+				chatTextAreaElement?.focus();
 
 				if (prompt !== '' && $settings?.speechAutoSend === true) {
 					submitPrompt(prompt, user);
@@ -187,7 +187,7 @@
 						prompt = `${prompt}${transcript}`;
 
 						await tick();
-						chatTextArea?.focus();
+						chatTextAreaElement?.focus();
 
 						// Restart the inactivity timeout
 						timeoutId = setTimeout(() => {
@@ -289,7 +289,7 @@
 	};
 
 	onMount(() => {
-		window.setTimeout(() => chatTextArea?.focus(), 0);
+		window.setTimeout(() => chatTextAreaElement?.focus(), 0);
 
 		const dropZone = document.querySelector('body');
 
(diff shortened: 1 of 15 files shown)


src/lib/components/chat/MessageInput.svelte
```diff
@@ -1,7 +1,7 @@
 <script lang="ts">
 	import { toast } from 'svelte-sonner';
 	import { onMount, tick } from 'svelte';
-	import { settings } from '$lib/stores';
```
tjbck, line 4: I'd suggest not to use stores and to keep variables for dom elements in the components.

Carlos-err406: what if the element is queried from outside the component?

tjbck:
Creating stores for each elements only adds additional code without any real tangible benefits, I suggest we only modify the areas that could be addressed with private variables within the component.

src/lib/components/chat/MessageInput.svelte
tjbck: I see your point of possible performance gains but we should keep the code as-is.

Carlos-err406: roger that

Carlos-err406:
> Creating stores for each elements only adds additional code without any real tangible benefits, I suggest we only modify the areas that could be addressed with private variables within the component.

in the case of chatTextArea (wich is like half the dom queries) i think it actually reduces the possible inconsistencies

but, maybe is not a good time for major changes in the code, i'll rollback the store ones then

src/lib/components/chat/Messages/UserMessage.svelte
```diff
@@ -20,18 +20,17 @@
 
 	let edit = false;
 	let editedContent = '';
-
+	let messageEditTextArea: HTMLTextAreaElement;
```
tjbck, line 23: We might want to name it `messageEditTextAreaElement` to denote that it's an element variable!

Carlos-err406: i missed that one, fixing now

src/lib/components/chat/MessageInput.svelte
```diff
@@ -19,7 +19,7 @@
 
 	export let suggestionPrompts = [];
 	export let autoScroll = true;
-
+	let chatTextArea:HTMLTextAreaElement
```
tjbck, line 22: Another one!

src/routes/(app)/+layout.svelte
```diff
@@ -37,7 +37,7 @@
 
 	let ollamaVersion = '';
 	let loaded = false;
-
+	let showShortcutsButton:HTMLButtonElement
```
tjbck, line 40: +1

src/routes/(app)/c/[id]/+page.svelte
```diff
@@ -45,7 +45,7 @@
 	let stopResponseFlag = false;
 	let autoScroll = true;
 	let processing = '';
-
+	let messagesContainerELement: HTMLDivElement;
```
Comment (tjbck) [inline src/routes/(app)/c/[id]/+page.svelte:48]: Typo

Comment (Carlos-err406) [reply src/routes/(app)/c/[id]/+page.svelte]: fixed

src/routes/(app)/+layout.svelte
Carlos-err406: fixed

src/lib/components/chat/MessageInput.svelte
Carlos-err406: fixed

tjbck:
LGTM, Thanks!

justinh-rahb:
Due to the number of related issues, it’s plainly obvious there was no thorough testing on this PR. We value efficiency over aesthetics, ugly vanilla JS that works will always be better than fancy JS that doesn't. Going forward, please ensure that all code changes you introduce have actually been tested as it's clear that things were overlooked here.

PR #1029, opened 2024-03-04
recreate rag collection instead of falling back to stale version

fbirlik:
## Description

store_data_in_vector_db is used to store web pages after retrieval.  Currently, new version of the page is retrieved, split into chunks, but because previous collection exists, new data is dropped.  Queries are executed against previous version instead of latest.

---

### Changed

- check for existence of previous version of the collection and delete it if exists; this ensure new version is stored and used in further queries

### DIFF
--- a/backend/apps/rag/main.py
+++ b/backend/apps/rag/main.py
@@ -108,7 +108,7 @@ class StoreWebForm(CollectionNameForm):
     url: str
 
 
-def store_data_in_vector_db(data, collection_name) -> bool:
+def store_data_in_vector_db(data, collection_name, overwrite: bool = False) -> bool:
     text_splitter = RecursiveCharacterTextSplitter(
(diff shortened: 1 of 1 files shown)


backend/apps/rag/main.py
```diff
@@ -118,6 +118,10 @@ def store_data_in_vector_db(data, collection_name) -> bool:
     metadatas = [doc.metadata for doc in docs]
 
     try:
+        for collection in CHROMA_CLIENT.list_collections():
```
tjbck, line 121: Shouldn't we make it so that it only does this for web contents?

fbirlik: Even when content does not change (file side of things), data is still fetched with loader.load(), split into chunks and it bails at the final step as collection already exists.

Still, calculating vectors is a considerable part of whole operation, so I guess you are right, it might be better to do this overwrite only for web side.

Let me figure out how to update a pull req :)

tjbck:
LGTM, Thanks!

PR #1033, opened 2024-03-04
0.1.109

### LINKED ISSUE #278: Feat: Ability to work with multiple Ollama servers
On my system I have a capable CPU with a large amount of RAM that is able to run quite large models, albeit slowly. I also have an RTX 3060 which is able to run smaller models quite quickly. I can easily have two Docker instances of Ollama running, one for GPU mode and one for CPU only mode. It would be great if we could have an ability to have a single instance of Ollama-Webui with the ability to switch between the two Ollama instances.

Have a settings screen where I can add one, or more, additional Ollama server. Along with the entry screen, have a toggle or drop-down list that would ideally let me set the default Ollama server on a per model level.

An alternative solution, which would be less ideal, would be to simply run two instances of Ollama-Webui, each pointing to a different Ollama container.

### LINKED ISSUE #788: feat: multiple Ollama servers in one webui + load balancing
The #1 downside to Ollama is the fact that it can only process one request at a time, no matter what hardware is available. Right now my solution is to have multiple Ollama instances running, so when one is in use on one graphics card I can use the other instance on the other card.

It would be nice to see open-webui allow multiple Ollama server connections, as I think an environment with multiple users using it simultaneously could really benefit from this. I myself have one machine with two Ollama instances (one per graphics card), and a whole different machine off-site that also has an Ollama instance. Utilizing all three of them would be nice for multiple concurrent users

### LINKED ISSUE #1006: feat: "Cancel" in the Playground does not stop Ollama generation
# Bug Report

## Description

**Bug Summary:**
This is the same problem as #456 but in Playground.

**Steps to Reproduce:**
Start generation in Playground and press Cancel. It does not matter whether you use chat or text completion mode.

**Expected Behavior:**
Ollama should stop immediately.

**Actual Behavior:**
UI stops streaming, but Ollama continues to generate tokens in the background.

## Reproduction Details

**Confirmation:**

- [x] I have reviewed the troubleshooting.md document.
- [ ] I have included the browser console logs.
- [ ] I have included the Docker container logs.

## Additional Information

More info available upon request.

### LINKED ISSUE #1024: Rag/Web requests have empty user_agent header causing rejection from some web sites
# Bug Report

## Description

**Bug Summary:**
Rag/Web requests (#url..) have empty User-Agent header causing some web sites denying the request

**Expected Behavior:**
Either a fix User-Agent or random User-Agent to imitate a real browser

## Note

langchain_community.WebBaseLoader class used to fetch web pages has support for generating fake user agents using fake_useragent package in default use case.  Installing this package is enough to start generating user agents.

Screenshot 2024-03-03 at 2 15 18 PM

### LINKED ISSUE #1031: recreate rag collection instead of falling back to stale version
# Bug Report

## Description

**Bug Summary:**
store_data_in_vector_db is used to store web pages after retrieval. Currently, new version of the page is retrieved, split into chunks, but because previous collection exists, new data is dropped. Queries are executed against previous version instead of latest.

**Steps to Reproduce:**
- open a new chat, fetch a web page using '#' prefix, query the content (preferably content changeable by tester)
- wait until content changes (or update the content manually)
- in a new chat query the same page using '#' prefix
- query results will be based on first request although second request was completed

**Expected Behavior:**
- if a page is fetched again, latest version should be available for query

**Actual Behavior:**
- if a page is fetched again, first retrieved version is stuck for all consequent queries

### LINKED ISSUE #1061: Extra 'S' in 'ASSISTANT' field for GGUF model upload template
https://github.com/open-webui/open-webui/blob/eb51ad14e4caafda1c9fd24c4945044b8776a7a3/src/lib/components/chat/Settings/Models.svelte#L48

Helpfully spotted by a Discord member, **beegie**

Screenshot 2024-03-06 at 10 59 06 AM

### LINKED ISSUE #1064: Option to use Claude 3 and Gemini
Option to use Claude 3 and Gemini

tjbck:
Resolves #1031 #1024 #1006 #788 #278 #1064 #1061


PierrunoYT:
Screenshot 2024-03-10 101727

PR #1077, opened 2024-03-07
fix: numbered lists with an additional line between items are rendered incorrectly

fbirlik:
## Description

update app.css to convert p-tags within li-tags to inline instead of block

---

### Fixed

- numbered lists with an additional newline between items are rendered incorrectly.


tjbck:
LGTM, Thanks!

PR #1165, opened 2024-03-14
refac: Dockerfile

jannikstdl:
I made some changes to the dockerfile, optimizing the codebase for best practices and reducing the size a bit.

Added CUDA Support for the Whisper and embedding models. 

Also added an argument to include Ollama in the Image. 

**Changes**.
- Merged some steps into one
- Commented out (I strongly believe) unused downloads of the "all-MiniLM-L6-v2" default embedding model since it downloads the model in line 75 `RUN python -c "import os; from chromadb. utils import embedding_functions; sentence_transformer_ef = embedding_functions.SentenceTransformerEmbeddingFunction(model_name=os.environ['RAG_EMBEDDING_MODEL'], device=os.environ['RAG_EMBEDDING_MODEL_DEVICE_TYPE'])"`.
- Changed the base image from `node` to `21-bookworm-slim` because 1. deb bookworm is a very stable base and 2. it is a defined version for no automatic latest pulls as this can lead to problems in the future.
- Added `--no-install-recommends` to the apt-get's to reduce image size.
- Added some comments.

image

Note: This should be double checked since this is affecting many users!


justinh-rahb:
Thanks Jannik! I'll get this tested and add my feedback as soon as possible.

jannikstdl:
> @jannikstdl It builds on Mac, and RAG still works! 👌

Thanks! Thats good news :D

justinh-rahb:
> Thanks! Thats good news :D

Nearly a GB smaller in size too, bravo! 👏

Dockerfile
```diff
@@ -1,67 +1,74 @@
 # syntax=docker/dockerfile:1
 
-FROM node:alpine as build
+######## WebUI frontend ########
+FROM node:21-bookworm-slim as build
```
tjbck, line 4: Thanks for the PR! This part shouldn't matter as it's only for the build setp, and is not actually used for deployment. (Check out [multi-stage docker build](https://docs.docker.com/build/building/multi-stage/))

tjbck:
The rest looks good!

Dockerfile
jannikstdl: Ok, but should we not use fixed versions? Like the node 21 release since pulling the latest without testing could lead to problems in the future. Also using smaller images in a build stage image may not affect the final image size but the build speed. 

But node and bookwork slim have almost the same size.

justinh-rahb: I agree with @jannikstdl about pinning the node version, but I agree with @tjbck that I don't see a need to change from alpine to debian in this stage. It should be of minimal difference one way or another, and usually in those cases I err on the side of what's already in play.

jannikstdl: Yes as mentioned Debian and alpine are almost exact the same size so we could use 21 apline.  

So we use what's already tested in long term.

tjbck: Ping me when you update!

jannikstdl: @tjbck ping

jannikstdl:
@justinh-rahb should we not also add `EXPOSE 8080` - our default port - to get this extra option in docker desktop to easily assign the image an external port?
image

justinh-rahb:
@jannikstdl don't see much harm in doing so, all the docs instruct people to set this explicitly but I guess this could be nice for those using management tools such as Docker Desktop or Portainer.

jannikstdl:
I added @yousecjoe to this PR, he mentioned on Discord that he has some ides to optimize the Dockerfile for CUDA and other devices. 

Converted this to draft.

yousecjoe:
Thanks! 

I intend to create the file "Dockerfile-cuda" in your fork.

I will make the file with a combination of your changes and mine. 

We can rename this file once we are happy with the combination of our efforts.

Until it's renamed, you can test it using "docker build -t open-webui -f Dockerfile-cuda".

yousecjoe:
Today, I'm working to reduce the attack surface and minimize the container image size. 

The primary objective is to use a Chainguard image as the base layer because [Chainguard](https://images.chainguard.dev/directory?category=all) images are hardened to significantly reduce the number of CVEs. They are also smaller because they remove everything unnecessary.

yousecjoe:
So far, my effort to use Chainguard's [pytorch-cuda12](https://images.chainguard.dev/directory/image/pytorch-cuda12/versions) image has run into a problem due to conflicts with PyTorch, which is included in the base image and required in the various python `requirements.txt` files that get pulled in from the different projects. The build fails when it tries to add Chromadb because there are two instances of PyTorch, and the pip3 package manager does not know how to resolve this scenario. It is still possible to make this work after some more effort. 

On the other hand, Ollama uses the [nvidia/cuda](https://hub.docker.com/r/nvidia/cuda) image as a base, and that's also what is currently working for me. It's not the most secure or slim solution, but it does work. 

Still, when we enable CUDA, we break CPU embedding. 

Because the nvidia/cuda image works, I will focus on that and move forward with work on supporting both CUDA and CPU embedding in the same Dockerfile. 

After CUDA and CPU embedding work in a single Dockerfile, I can revisit using a more secure and smaller image like that from Chainguard.

justinh-rahb:
> So far, my effort to use Chainguard's [pytorch-cuda12](https://images.chainguard.dev/directory/image/pytorch-cuda12/versions) image has run into a problem due to conflicts with PyTorch, which is included in the base image and required in the various python `requirements.txt` files that get pulled in from the different projects. The build fails when it tries to add Chromadb because there are two instances of PyTorch, and the pip3 package manager does not know how to resolve this scenario. It is still possible to make this work after some more effort.

Had a bit of a feeling you might run into something like this, PyTorch is a bit of a special snowflake that we've had to code around to get working in a universal way that didn't add a couple extra GB to the image size, which is explicitly install the CPU-only version.

Good effort nonetheless, I'm glad to see it explored either way, and yes definitely worth a revisit later down the line if we can sort out this convoluted mess of different platform support for the heavier packages that doesn't piss off one group (the #gpupoor) or the other (the #gpurich).

justinh-rahb:
Thinking about some alternative ways we could do things, just some ideas:

- Have the "default" Dockerfile and backend code setup in such a way that CPU-only is assumed and everything is in place at build to support it out of the box, but if GPU support is detected then it will set this up at first boot.

- Build explicit GPU-supported images with their own image tag that would be built with the support and not require runtime installation at first boot.

jannikstdl:
Made some tests. Can anyone without a CUDA device try to build the normal `Dockerfile `and see if there has been anything changed (functionality, size..)

yousecjoe:
> Thinking about some alternative ways we could do things, just some ideas:
> 
> * Have the "default" Dockerfile and backend code setup in such a way that CPU-only is assumed and everything is in place at build to support it out of the box, but if GPU support is detected then it will set this up at first boot.
> * Build explicit GPU-supported images with their own image tag that would be built with the support and not require runtime installation at first boot.
> 
> Benefit of this approach is it should also work if you build and run directly in a dev environment.

Check out this script for reference: [ollama/scripts/build_docker.sh](https://github.com/ollama/ollama/blob/main/scripts/build_docker.sh)

I have been working towards targeting all architectures in a single Dockerfile and using CICD with a build script. 

The script will build so that CPU is assumed, but it also targets specific CPU architectures, including those with AVX and AVX2. It is worth exploring whether that capability can improve the embedding process, but that specific task is on my backlog. 

When we look at how Ollama solves this issue, we see that they are using a single Dockerfile to produce more than a single image. 

The end result is images targeted at arm64, amd64, and rocm (for AMD GPUs.)

justinh-rahb:
🥷 Sounds like black belt level Docker-fu, I'm all for it if it works!

yousecjoe:
@jannikstdl 

> Made some tests. Can anyone without a CUDA device try to build the normal `Dockerfile `and see if there has been anything changed (functionality, size..)

I'm testing this now. 

Because we hardcode the device type in the environment variable `RAG_EMBEDDING_MODEL_DEVICE_TYPE`, CPU-only can be tested with or without a CUDA device in the system. Nvidia Cuda doesn't come into effect if the device type is hardcoded to "CPU." 

It was wise of you to use the `if else` to check for a change in the environment variable. 

To complete this for both architectures, we need to set this variable automatically through system detection, a parameter from the command line, or a variable in the CICD build script.

yousecjoe:
@jannikstdl  Building your latest commit comes out to: 13.4GB on my system. 

```bash
docker images --format "{{.Repository}}:{{.Tag}} Size: {{.Size}}"
open-webui:latest Size: 13.4GB
```

Before building, I `prune -a` my build cache. 
```bash
docker builder du
Reclaimable:    0B
Total:          0B
```

After building 

```bash
docker builder du
Shared:         13.29GB
Private:        12.25GB
Reclaimable:    25.54GB
Total:          25.54GB
```

justinh-rahb:
> To complete this for both architectures, we need to set this variable automatically through system detection, a parameter from the command line, or a variable in the CICD build script.

👌 perfect

jannikstdl:
> @jannikstdl  Building your latest commit comes out to: 13.4GB on my system. 
> 
> 
> 
> ```bash
> 
> docker images --format "{{.Repository}}:{{.Tag}} Size: {{.Size}}"
> 
> open-webui:latest Size: 13.4GB
> 
> ```
> 
> 
> 
> Before building, I `prune -a` my build cache. 
> 
> ```bash
> 
> docker builder du
> 
> Reclaimable:    0B
> 
> Total:          0B
> 
> ```
> 
> 
> 
> After building 
> 
> 
> 
> ```bash
> 
> docker builder du
> 
> Shared:         13.29GB
> 
> Private:        12.25GB
> 
> Reclaimable:    25.54GB
> 
> Total:          25.54GB
> 
> ```

Ok ty. So it's not only on my machine. I am not very shure at the moment my the Image is bigger if I let the device type set to CPU because it should be the same as before.

jannikstdl:
> 
> - Have the "default" Dockerfile and backend code setup in such a way that CPU-only is assumed and everything is in place at build to support it out of the box, but if GPU support is detected then it will set this up at first bo
> 

Unfortunately there is no way to automatically detect CUDA when building the image. 

Either we create more Images for specific architectures, or we let the user build the image with build args for non cpu or mps hardware. 

The only automation we could integrate in my mind is a script that detects the hardware (like the Ollama install sh if I remember correctly) and builds the image with the specific args.

jannikstdl:
I managed to shrink the image to its reduced size as we had first (around 3,2 GB when using CPU acceleration) 
It was caused by the `pip install -r requerements.txt` running **before** the torch installs.  Since torch and its dependencies are declared in the requirements file and by default it installs bigger (probably GPU) versions of them.

You should now be able to:

`docker build -t openwebui-cpu .`
~3,20 GB

or 

`docker build --build-arg="USE_CUDA=true" -t openwebui-cuda .`
huge ~13 GB

or

`docker build --build-arg="USE_MPS=true" -t openwebui-mps .`
Size around same as CPU

jannikstdl:
Open points:

- CUDA 11 / 12 selection
- compute type tests (int8; float16)

tjbck:
One other idea I had was using Ollama image directly as our base image so that we can ship Ollama built-in. Would be appropriate for this PR as well, let me know what you guys think!

yousecjoe:
> > * Have the "default" Dockerfile and backend code setup in such a way that CPU-only is assumed and everything is in place at build to support it out of the box, but if GPU support is detected then it will set this up at first bo
> 
> Unfortunately there is no way to automatically detect CUDA when building the image.
> 
> Either we create more Images for specific architectures, or we let the user build the image with build args for non cpu or mps hardware.
> 
> The only automation we could integrate in my mind is a script that detects the hardware (like the Ollama install sh if I remember correctly) and builds the image with the specific args.

Correct me if I'm wrong, but Ollama does seem to include a script during the build process that automatically detects CUDA when building their images. 

Line 7 in [ollama/llm/generate/gen_linux.sh](https://github.com/ollama/ollama/blob/main/llm/generate/gen_linux.sh)

The end result is that Ollama creates three images to target specific architectures: amd64, arm64, and rocm.

yousecjoe:
> Open points:
> 
> * CUDA 11 / 12 selection
> * compute type tests (int8; float16)

Ollama targets CUDA 11.3.1. 

This may be the minimum version necessary to run Ollama. 

I have been testing the latest version of CUDA (12.3.2) because I have not yet encountered any issues. 

It does make sense for the broader audience to keep the build argument in sync with what Ollama supports as a bare minimum. 

Is there more context to the line "compute type tests (int8; float16)"?

justinh-rahb:
> One other idea I had was using Ollama image directly as our base image so that we can ship Ollama built-in, zero configuration. Would be appropriate for this PR as well, let me know what you guys think!

Problem there is that at least for macOS this would be a waste of time, you'd lose the hardware support.

yousecjoe:
> One other idea I had was using Ollama image directly as our base image so that we can ship Ollama built-in, zero configuration. Would be appropriate for this PR as well, let me know what you guys think!

My two cents. 

I'm all for pushing the envelope, but in this case combining solutions makes it seem like I'm going the opposite way of micro-services architecture. 

Many people run Ollama for many different solutions. A single Ollama server can be used by multiple projects, so, It would be redundant to have Ollama built into a web UI front end. 

In a perfect world, the LLM backend solution should be isolated from the webui front end. To be more specific, these components should be on separate VLANs with Zero Trust Network Access protecting the APIs to Ollama and any vector databases. 

The more you combine components, the more risk there is for cybersecurity, not to mention you would be tied to any problems Ollama runs into later on down the road.

yousecjoe:
> > One other idea I had was using Ollama image directly as our base image so that we can ship Ollama built-in, zero configuration. Would be appropriate for this PR as well, let me know what you guys think!
> 
> Problem there is that at least for macOS this would be a waste of time, you'd lose the hardware support, so you'd still want to connect to a direct install.
> 
> > Is there more context to the line "compute type tests (int8; float16)"?
> 
> Not all GPUs support all data types.

Yes, that's correct. 

That is also being handled in the script I shared previously. 

Starting at line 60 here [gen_linux.sh](https://github.com/ollama/ollama/blob/main/llm/generate/gen_linux.sh)

It would seem like the Ollama devs have thought of almost everything we can imagine.

justinh-rahb:
@tjbck with that clarification, I don't have any argument against it.

jannikstdl:
> 
> @yousecjoe I also have nothing against adopting Ollama's packaging approach, I just don't fully understand it lol.. until I read it I didn't know most of what they're doing is even possible 😅 Similar to how they've got a `:rocm` tag, could we have a `:main-cuda` and `:main-mps` tag and then those builds only have the code they need/run with appropriate default ENV vars? Or am I still misunderstanding some magic here that lets them have a CUDA-enabled `:main` without the dependency baggage if you don't even have a GPU to run on?


That's basically what I added. You can build the images with the different build arcs and let the gh action which builds the images run the different builds and publish them with the appropriate tag.

yousecjoe:
> > @yousecjoe I also have nothing against adopting Ollama's packaging approach, I just don't fully understand it lol.. until I read it I didn't know most of what they're doing is even possible 😅 Similar to how they've got a `:rocm` tag, could we have a `:main-cuda` and `:main-mps` tag and then those builds only have the code they need/run with appropriate default ENV vars? Or am I still misunderstanding some magic here that lets them have a CUDA-enabled `:main` without the dependency baggage if you don't even have a GPU to run on?
> 
> That's basically what I added. You can build the images with the different build arcs and let the gh action which builds the images run the different builds and publish them with the appropriate tag.
> 
> I also edited the build-images yml.

I just pulled and rebased with your changes; it looks great. I'm going to take a closer look, run some tests, and let you know how it goes. 

If I need to make any changes I'll keep using `Dockerfile-cuda`, then I figure we can remove that file once we close this PR.

yousecjoe:
@jannikstdl, The changes are awesome! I pruned my build cache and verified it was empty before I ran the following command:

```
docker build --platform linux/amd64 -t open-webui --build-arg USE_CUDA=true .
```

Am I using the `--build-arg USE_CUDA=true` correctly? 

I expected this to build the container with support for cuda. When I exec into the container, I check the backend/config.py file and I thought it should show "cuda", but it shows "cpu". Shouldn't this be set to "cuda" inside the container if I built the image with '--build-arg USE_CUDA=true`?

```
root@92ba34ef68f3:/app/backend# cat config.py | grep DEVICE_TYPE
DEVICE_TYPE = os.environ.get(
    "DEVICE_TYPE", "cpu"
root@92ba34ef68f3:/app/backend# 
```

jannikstdl:
> @jannikstdl, The changes are awesome! I pruned my build cache and verified it was empty before I ran the following command:
> 
> ```
> docker build --platform linux/amd64 -t open-webui --build-arg USE_CUDA=true .
> ```
> 
> Am I using the `--build-arg USE_CUDA=true` correctly?

No, as I mentioned earlier the correct way to use build args is `--build-arg="USE_CUDA=true"`
Then it should work.

> I expected this to build the container with support for cuda. When I exec into the container, I check the backend/config.py file and I thought it should show "cuda", but it shows "cpu". Shouldn't this be set to "cuda" inside the container if I built the image with '--build-arg USE_CUDA=true`?
> 
> ```
> root@92ba34ef68f3:/app/backend# cat config.py | grep DEVICE_TYPE
> DEVICE_TYPE = os.environ.get(
>     "DEVICE_TYPE", "cpu"
> root@92ba34ef68f3:/app/backend# 
> ```

Thats just a fallback value. If the ENV "DEVICE_TYPE" is for example set to "CUDA" use this if its not set (mostly if you are starting a dev server on your local machine without docker it is set to default "cpu" to avoid errors.

yousecjoe:
> > @jannikstdl, The changes are awesome! I pruned my build cache and verified it was empty before I ran the following command:
> > 
> > ```
> > docker build --platform linux/amd64 -t open-webui --build-arg USE_CUDA=true .
> > ```
> > 
> > Am I using the `--build-arg USE_CUDA=true` correctly?
> 
> No, as I mentioned earlier the correct way to use build args is `--build-arg="USE_CUDA=true"`
> Then it should work.
> 
> > I expected this to build the container with support for cuda. When I exec into the container, I check the backend/config.py file and I thought it should show "cuda", but it shows "cpu". Shouldn't this be set to "cuda" inside the container if I built the image with '--build-arg USE_CUDA=true`?
> > 
> > ```
> > root@92ba34ef68f3:/app/backend# cat config.py | grep DEVICE_TYPE
> > DEVICE_TYPE = os.environ.get(
> >     "DEVICE_TYPE", "cpu"
> > root@92ba34ef68f3:/app/backend# 
> > ```
> 
> Thats just a fallback value. If the ENV "DEVICE_TYPE" is for example set to "CUDA" use this if its not set (mostly if you are starting a dev server on your local machine without docker it is set to default "cpu" to avoid errors.

Thanks for clarifying, I'll run some more tests.

yousecjoe:
image

It's building now, and it looks like it picked up the build argument. Good work. I'll capture the sizes of both images and share a comparison here.

yousecjoe:
```
docker buildx build --platform linux/amd64 -t open-webui .
```

CPU only build cache: ~3GB

image

yousecjoe:
```
docker build --platform linux/amd64 -t open-webui --build-arg="USE_CUDA=true" .
```

CUDA build cache: ~8GB

image

yousecjoe:
I tested the performance of both. I might need to do something differently, but it was hard to distinguish between Cuda and CPU. 

What do you all use to monitor the performance of the embedding process? 

Could this be a good use of Langfuse? 

I captured screenshots of the System Monitor in Kubuntu while uploading 3 PDFs. 

Can you tell which is Cuda and which is CPU? I expected more difference. 

image


image

yousecjoe:
I had to try and upload a lot more PDFs.

USE_CUDA=true

It seems to still be using only the CPU. 

image

tjbck:
@jannikstdl Is there a way to somehow incorporate Dockerfile-ollama to our main image using docker ARGs?

jannikstdl:
> I had to try and upload a lot more PDFs. I used ~100 pdfs of various lengths.
> 
> USE_CUDA=true
> 
> It seems to be still using only the CPU.
> 

Have you installed the Nvidia CUDA Toolkit on your machine? 
You need that for the Container to access your GPU.

Nvidia GPU
Install the [NVIDIA Container Toolkit](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html#installation).

jannikstdl:
> @jannikstdl Is there a way to somehow incorporate Dockerfile-ollama to our main image using docker ARGs?

Should be, will have a look

jannikstdl:
@tjbck i tested your all in one container. It works but for me on windows running ollama in wsl it only worked if i killed the local ollama running on port 11434. Because it used my local ollama first. Afer i killed the service and started the container again in worked with the in-docker ollama.

Mybe we should mention that in the docs, or we somehow block the extern port 11434, or we let ollama in the container run at a different one. 
Or we just let that how it is since it works with both and the problem that some users can not get the ollama connection (they probably do not have a ollama instance running on their device) would be gone.

yousecjoe:
> > I had to try and upload a lot more PDFs. I used ~100 pdfs of various lengths.
> > USE_CUDA=true
> > It seems to be still using only the CPU.
> 
> Have you installed the Nvidia CUDA Toolkit on your machine? You need that for the Container to access your GPU.
> 
> For my tests is was not super fast but i definitely recognized that the CPU was less loaded.
> 
> Nvidia GPU Install the [NVIDIA Container Toolkit](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html#installation).

I have installed the toolkit previously. I don't have any issues with any of my other containers accessing the GPU. I have the ability to run nvidia-smi and nvcc inside the container. The 100 PDFs never finished uploading because it was only using the CPU and I just restarted the container. I'll look into this further because I suspect mine is not working right, yet. 

I am using your methods to test adding a build argument for Ollama as well

image

jannikstdl:
> @jannikstdl We just did nearly the same thing! 🤣 I'll grab your changes and test them out; thanks!

Yes, you need to take care of the Ollama Base Url since host.docker.internal won't work when using ollama in the same container. 

Also i found a bug, i am on it to fix it.
Will also try if CUDA embedding works for me.

yousecjoe:
> > @jannikstdl We just did nearly the same thing! 🤣 I'll grab your changes and test them out; thanks!
> 
> Yes, you need to take care of the Ollama Base Url since host.docker.internal won't work when using ollama in the same container.
> 
> Also i found a bug, i am on it to fix it. Will also try if CUDA embedding works for me.

Yep, I'm familiar with the base IP requirement. I changed my base URL to "ollama" because it runs on a separate container but is on the same Docker host. 

I see what you did differently. I was trying to run `ollama serve &` before the GPU was accessible. You also grabbed the `INCLUD_OLLAMA` environment variable from the Dockerfile and I was trying to do that but instead wrote a way to check if ollama is installed on the system, and then start it up.

jannikstdl:
> I see what you did differently. I was trying to run `ollama serve &` before the GPU was accessible. You also grabbed the `INCLUDE_OLLAMA` environment variable from the Dockerfile and I was trying to do that but instead wrote a way to check if ollama is installed on the system, and then start it up.

Maybe your default docker runtime is not set to nvidia.
Try `cat /etc/docker/daemon.json` there should be a default docker runtime like this
```
{

 "default-runtime": "nvidia",
  "runtimes": {
     "nvidia": {
         "path": "/usr/bin/nvidia-container-runtime",
         "runtimeArgs": []
     }
 },

}
```
docs https://docs.nvidia.com/dgx/nvidia-container-runtime-upgrade/index.html

but this is not recommended, instead try to run 

`docker run **--gpus all**...` like ollama does too.

Worked for me:

```
(base) PS D:\me> docker run **--gpus all** -p 3333:8080 openwebui-cuda-ollama
No WEBUI_SECRET_KEY provided
Generating WEBUI_SECRET_KEY
Loading WEBUI_SECRET_KEY from .webui_secret_key
INCLUDE_OLLAMA is set to true, starting ollama serve.
Couldn't find '/root/.ollama/id_ed25519'. Generating new private key.
Your new public key is:

ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIJHzj9I9Z8vfheJNiWp1uLpOCT3d5Qwjpog542xC+dRz

time=2024-03-22T10:23:16.671Z level=INFO source=images.go:806 msg="total blobs: 0"
time=2024-03-22T10:23:16.671Z level=INFO source=images.go:813 msg="total unused blobs removed: 0"
time=2024-03-22T10:23:16.671Z level=INFO source=routes.go:1110 msg="Listening on 127.0.0.1:11434 (version 0.1.29)"
time=2024-03-22T10:23:16.672Z level=INFO source=payload_common.go:112 msg="Extracting dynamic libraries to /tmp/ollama550020823/runners ..."
time=2024-03-22T10:23:18.665Z level=INFO source=payload_common.go:139 msg="Dynamic LLM libraries [cpu_avx2 cpu cpu_avx cuda_v11 rocm_v60000]"
time=2024-03-22T10:23:18.665Z level=INFO source=gpu.go:77 msg="Detecting GPU type"
time=2024-03-22T10:23:18.665Z level=INFO source=gpu.go:191 msg="Searching for GPU management library [[libnvidia-ml.so](http://libnvidia-ml.so/)](http://libnvidia-ml.so/)"

time=2024-03-22T10:23:18.665Z level=INFO source=gpu.go:237 msg="Discovered GPU libraries: [/usr/lib/x86_64-linux-gnu/libnvidia-ml.so.1 /usr/lib/wsl/drivers/nv_dispig.inf_amd64_866484083fc526af/libnvidia-ml.so.1]"
time=2024-03-22T10:23:18.683Z level=INFO source=gpu.go:82 msg="Nvidia GPU detected"
time=2024-03-22T10:23:18.683Z level=INFO source=cpu_common.go:11 msg="CPU has AVX2"
time=2024-03-22T10:23:18.689Z level=INFO source=gpu.go:119 msg=**"CUDA Compute Capability detected: 8.6"**
INFO:     Started server process [1]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on [http://0.0.0.0:8080/](http://0.0.0.0:8080/) (Press CTRL+C to quit)
```




> I have installed the toolkit previously. I don't have any issues with any of my other containers accessing the GPU. I have the ability to run nvidia-smi and nvcc inside the container. The 100 PDFs never finished uploading because it was only using the CPU and I just restarted the container. I'll look into this further because I suspect mine is not working right, yet.
> 

Tested it on mine, can confirm this behavior too. 
It used CPU. but if I set the DEVICE_TYPE to "cuda" manually in the code it uses GPU. I hink there is a problem with the ENV. Will investigate further.

jannikstdl:
@yousecjoe Should work now with cuda! Both Ollama and Embedding.

yousecjoe:
> @yousecjoe Should work now with cuda! Both Ollama and Embedding.

I will grab the latest changes and check it out. 

I also maintain a public fork of PrivateGPT, where I use Ollama with GPUs for embedding. I took screenshots of what it looks like when GPUs are enabled and used for embedding. There's little to no hit to the CPU, but there is a dramatic spike in temperature across the GPUs. There is a small, but noticeable bump in usage %. 

These small embedding tests do not significantly strain the 3080 and 1080 Ti combination. 

Utilization from a handful of PDFs:

image

Utilization from a hundred PDFs:

image

jannikstdl:
@yousecjoe Have you tried building with `docker build --gpus all -t openwebui:cuda --build-arg="USE_CUDA=true" --build-arg="INCLUDE_OLLAMA=true" --no-cache .`

jannikstdl:
@justinh-rahb can you test if 

uses the Apple Silicon GPU for the embedding process?

jannikstdl:
@tjbck @justinh-rahb @yousecjoe  If someone could double check CUDA and check MPS is working correctly i could finish this PR if noone has additions.

justinh-rahb:
Nah, manually `docker build` failed to:
```log
139.3 RuntimeError: PyTorch is not linked with support for mps devices
------
Dockerfile:80
--------------------
  79 |
  80 | >>> RUN if [ "$USE_CUDA" = "true" ]; then \
  81 | >>>         pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu117 --no-cache-dir && \
  82 | >>>         pip3 install -r requirements.txt --no-cache-dir; \
  83 | >>>     elif [ "$USE_MPS" = "true" ]; then \
  84 | >>>         pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu --no-cache-dir && \
  85 | >>>         pip3 install -r requirements.txt --no-cache-dir && \
  86 | >>>         python -c "import os; from faster_whisper import WhisperModel; WhisperModel(os.environ['WHISPER_MODEL'], device='cpu', compute_type='int8', download_root=os.environ['WHISPER_MODEL_DIR'])" && \
  87 | >>>         python -c "import os; from chromadb.utils import embedding_functions; sentence_transformer_ef = embedding_functions.SentenceTransformerEmbeddingFunction(model_name=os.environ['RAG_EMBEDDING_MODEL'], device='mps')"; \
  88 | >>>     else \
  89 | >>>         pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu --no-cache-dir && \
  90 | >>>         pip3 install -r requirements.txt --no-cache-dir && \
  91 | >>>         python -c "import os; from faster_whisper import WhisperModel; WhisperModel(os.environ['WHISPER_MODEL'], device='cpu', compute_type='int8', download_root=os.environ['WHISPER_MODEL_DIR'])" && \
  92 | >>>         python -c "import os; from chromadb.utils import embedding_functions; sentence_transformer_ef = embedding_functions.SentenceTransformerEmbeddingFunction(model_name=os.environ['RAG_EMBEDDING_MODEL'], device='cpu')"; \
  93 | >>>     fi
  94 |
--------------------
ERROR: failed to solve: process "/bin/sh -c if [ \"$USE_CUDA\" = \"true\" ]; then         pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu117 --no-cache-dir &&         pip3 install -r requirements.txt --no-cache-dir;     elif [ \"$USE_MPS\" = \"true\" ]; then         pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu --no-cache-dir &&         pip3 install -r requirements.txt --no-cache-dir &&         python -c \"import os; from faster_whisper import WhisperModel; WhisperModel(os.environ['WHISPER_MODEL'], device='cpu', compute_type='int8', download_root=os.environ['WHISPER_MODEL_DIR'])\" &&         python -c \"import os; from chromadb.utils import embedding_functions; sentence_transformer_ef = embedding_functions.SentenceTransformerEmbeddingFunction(model_name=os.environ['RAG_EMBEDDING_MODEL'], device='mps')\";     else         pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu --no-cache-dir &&         pip3 install -r requirements.txt --no-cache-dir &&         python -c \"import os; from faster_whisper import WhisperModel; WhisperModel(os.environ['WHISPER_MODEL'], device='cpu', compute_type='int8', download_root=os.environ['WHISPER_MODEL_DIR'])\" &&         python -c \"import os; from chromadb.utils import embedding_functions; sentence_transformer_ef = embedding_functions.SentenceTransformerEmbeddingFunction(model_name=os.environ['RAG_EMBEDDING_MODEL'], device='cpu')\";     fi" did not complete successfully: exit code: 1
```

jannikstdl:
> Nah, manually `docker build` failed to:
> 
> ```
> 139.3 RuntimeError: PyTorch is not linked with support for mps devices
> ------
> Dockerfile:80
> --------------------
>   79 |
>   80 | >>> RUN if [ "$USE_CUDA" = "true" ]; then \
>   81 | >>>         pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu117 --no-cache-dir && \
>   82 | >>>         pip3 install -r requirements.txt --no-cache-dir; \
>   83 | >>>     elif [ "$USE_MPS" = "true" ]; then \
>   84 | >>>         pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu --no-cache-dir && \
>   85 | >>>         pip3 install -r requirements.txt --no-cache-dir && \
>   86 | >>>         python -c "import os; from faster_whisper import WhisperModel; WhisperModel(os.environ['WHISPER_MODEL'], device='cpu', compute_type='int8', download_root=os.environ['WHISPER_MODEL_DIR'])" && \
>   87 | >>>         python -c "import os; from chromadb.utils import embedding_functions; sentence_transformer_ef = embedding_functions.SentenceTransformerEmbeddingFunction(model_name=os.environ['RAG_EMBEDDING_MODEL'], device='mps')"; \
>   88 | >>>     else \
>   89 | >>>         pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu --no-cache-dir && \
>   90 | >>>         pip3 install -r requirements.txt --no-cache-dir && \
>   91 | >>>         python -c "import os; from faster_whisper import WhisperModel; WhisperModel(os.environ['WHISPER_MODEL'], device='cpu', compute_type='int8', download_root=os.environ['WHISPER_MODEL_DIR'])" && \
>   92 | >>>         python -c "import os; from chromadb.utils import embedding_functions; sentence_transformer_ef = embedding_functions.SentenceTransformerEmbeddingFunction(model_name=os.environ['RAG_EMBEDDING_MODEL'], device='cpu')"; \
>   93 | >>>     fi
>   94 |
> --------------------
> ERROR: failed to solve: process "/bin/sh -c if [ \"$USE_CUDA\" = \"true\" ]; then         pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu117 --no-cache-dir &&         pip3 install -r requirements.txt --no-cache-dir;     elif [ \"$USE_MPS\" = \"true\" ]; then         pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu --no-cache-dir &&         pip3 install -r requirements.txt --no-cache-dir &&         python -c \"import os; from faster_whisper import WhisperModel; WhisperModel(os.environ['WHISPER_MODEL'], device='cpu', compute_type='int8', download_root=os.environ['WHISPER_MODEL_DIR'])\" &&         python -c \"import os; from chromadb.utils import embedding_functions; sentence_transformer_ef = embedding_functions.SentenceTransformerEmbeddingFunction(model_name=os.environ['RAG_EMBEDDING_MODEL'], device='mps')\";     else         pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu --no-cache-dir &&         pip3 install -r requirements.txt --no-cache-dir &&         python -c \"import os; from faster_whisper import WhisperModel; WhisperModel(os.environ['WHISPER_MODEL'], device='cpu', compute_type='int8', download_root=os.environ['WHISPER_MODEL_DIR'])\" &&         python -c \"import os; from chromadb.utils import embedding_functions; sentence_transformer_ef = embedding_functions.SentenceTransformerEmbeddingFunction(model_name=os.environ['RAG_EMBEDDING_MODEL'], device='cpu')\";     fi" did not complete successfully: exit code: 1
> ```

Thanks! I think MPS won't work in Docker for now:
https://github.com/pytorch/pytorch/issues/81224

yousecjoe:
> @justinh-rahb @yousecjoe If someone could double check CUDA and check MPS is working correctly i could finish this PR if noone has additions.
> 
> CUDA acceleration works for me. I can not test MPS / Apple Silicon.

I only have Intel Apple hardware. So, I also cannot test the MPS version.

I can double-check the CUDA support, but I have nothing to add if that works for people who want to use it like I thought I did.  I had a change of mind.

I took a break to work on another open-source front-end project called PrivateGPT. I created a Dockerfile using Chainguard's Python 3.12 image as a base that supports connecting to another container running Ollama for embedding. I reduced the CVEs from ~30 to 3. Running the latest software with the fewest CVEs using a remote Ollama achieved all my goals. 

After working on that project, I returned here realizing I didn't have a use case for the front end to do embedding. My use case is for the web front end to have as few CVEs as possible and use another container running Ollama for all LLM functions, including embedding.

lainedfles:
This is fantastic, thanks!  I can confirm that CUDA is working with Podman on Arch Linux after a few fixes submitted as jannikstdl/open-webui#2. Once this is merged with main, we should probably switch the `print()` with `log.debug()`. Here's the process I used and the trouble I experienced:

#### Build ####
```bash
podman build -t openwebui_cuda --build-arg="USE_CUDA=true" .
```

#### Testing ####
```bash
podman run --gpus all -p 127.0.0.1:3333:8080 --network slirp4netns:allow_host_loopback=true --add-host=ollama.local:10.0.2.2 --env 'OLLAMA_BASE_URL=http://ollama.local:11434' -v openwebui_cuda:/app/backend/data --name openwebui_cuda --replace openwebui_cuda
```

To validate that whisper uses my GPU, I used `nvidia-smi`:

```
$ UVICORN_PID=$(pidof /usr/local/bin/python); date; echo; ps -o command -p $UVICORN_PID; echo; while true; do nvidia-smi --query-accounted-apps=timestamp,gpu_name,pid,gpu_utilization --format=csv,noheader|grep $UVICORN_PID; sleep 0.5; done
Sat Mar 30 08:52:35 PM MDT 2024

COMMAND
/usr/local/bin/python /usr/local/bin/uvicorn main:app --host 0.0.0.0 --port 8080 --forwarded-allow-ips *

2024/03/30 20:52:36.018, NVIDIA GeForce RTX 4090 Laptop GPU, 107583, 0 %
<SNIP>
2024/03/30 20:52:43.280, NVIDIA GeForce RTX 4090 Laptop GPU, 107583, 2 %
2024/03/30 20:52:43.802, NVIDIA GeForce RTX 4090 Laptop GPU, 107583, 2 %
2024/03/30 20:52:44.326, NVIDIA GeForce RTX 4090 Laptop GPU, 107583, 3 %
2024/03/30 20:52:44.851, NVIDIA GeForce RTX 4090 Laptop GPU, 107583, 3 %
2024/03/30 20:52:45.374, NVIDIA GeForce RTX 4090 Laptop GPU, 107583, 3 %
```

#### Problems ####

1. `whisper_device_type` variable logic problem:

    ```
    INFO:     10.0.2.100:44580 - "GET /api/v1/chats/ HTTP/1.1" 200 OK
    audio/wav
    name 'whisper_device_type' is not defined
    INFO:     10.0.2.100:59862 - "POST /audio/api/v1/transcribe HTTP/1.1" 400 Bad Request
    ```

    Resolution:

    ```diff
    diff --git a/backend/apps/audio/main.py b/backend/apps/audio/main.py
    index 2faf07a..9be0589 100644
    --- a/backend/apps/audio/main.py
    +++ b/backend/apps/audio/main.py
    @@ -23,9 +23,13 @@ from utils.misc import calculate_sha256
     
     from config import CACHE_DIR, UPLOAD_DIR, WHISPER_MODEL, WHISPER_MODEL_DIR, DEVICE_TYPE
     
    -if DEVICE_TYPE != "cuda":
    +whisper_device_type = DEVICE_TYPE
    +
    +if whisper_device_type != "cuda":
         whisper_device_type = "cpu"
     
    +print(f"whisper_device_type: {whisper_device_type}")
    +
     
     app = FastAPI()
     app.add_middleware(
    ```
2. `libcudnn_ops_infer.so.8` shared object path issue:

    ```
    INFO:     10.0.2.100:52690 - "GET /api/v1/chats/848d80f6-caac-4fa5-95d4-1c6d69de13a4/tags HTTP/1.1" 200 OK
    Could not load library libcudnn_ops_infer.so.8. Error: libcudnn_ops_infer.so.8: cannot open shared object file: No such file or directory
    ```

    Resolution:

    ```diff
    --- a/backend/start.sh
    +++ b/backend/start.sh
    @@ -28,4 +28,9 @@ if [ "$INCLUDE_OLLAMA" = "true" ]; then
         ollama serve &
     fi
     
    +if [ "$USE_CUDA_DOCKER" = "true" ]; then
    +    echo "CUDA is enabled, appending LD_LIBRARY_PATH to include torch/cudnn & cublas libraries."
    +    export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/torch/lib:/usr/local/lib/python3.11/site-packages/nvidia/cublas/lib"
    +fi
    +
     WEBUI_SECRET_KEY="$WEBUI_SECRET_KEY" exec uvicorn main:app --host 0.0.0.0 --port "$PORT" --forwarded-allow-ips '*'
    ```

3. `libcublas.so.12` dependency missing (and path issue resolved above). It looks like later version of whisper_faster require cu12 (see SYSTRAN/faster-whisper#717):
    ```
    INFO:     10.0.2.100:59554 - "GET /api/v1/chats/848d80f6-caac-4fa5-95d4-1c6d69de13a4/tags HTTP/1.1" 200 OK
    audio/wav
    Library libcublas.so.12 is not found or cannot be loaded
    ```

    Resolution:

    ```diff
    diff --git a/Dockerfile b/Dockerfile
    index 4573de7..d217735 100644
    --- a/Dockerfile
    +++ b/Dockerfile
    @@ -80,6 +80,7 @@ COPY ./backend/requirements.txt ./requirements.txt
     RUN if [ "$USE_CUDA" = "true" ]; then \
             pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu117 --no-cache-dir && \
             pip3 install -r requirements.txt --no-cache-dir; \
    +        pip3 install nvidia-cublas-cu12 nvidia-cudnn-cu12; \
         elif [ "$USE_MPS" = "true" ]; then \
             pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu --no-cache-dir && \
             pip3 install -r requirements.txt --no-cache-dir && \
    @@ -131,4 +132,4 @@ COPY ./backend .
     
     EXPOSE 8080
     
    -CMD [ "bash", "start.sh"]
    \ No newline at end of file
    +CMD [ "bash", "start.sh"]
    ```

jannikstdl:
@lainedfles Thank you very much for the conclusion. I will look into that!

lainedfles:
@jannikstdl I've updated the pull request to merge with the open-webui/dev branch, to handle the CUDA version better and add a new build argument to set the desired embedding model.

```Dockerfile
ARG USE_CUDA_VER=cu121
ARG USE_EMBEDDING_MODEL=all-MiniLM-L6-v2
```

Default CUDA version of 12.1 to avoid using older faster-whisper. Install of Pytorch occurs like this now:

```bash
pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/$USE_CUDA_DOCKER_VER --no-cache-dir
```

I've also re-added  pre-loading of the models when USE_CUDA=true by forcing CPU mode for only these commands. Building online but executing first container run offline is now functional. Seems to be working very well. Most recent test using:

```bash
podman build -t openwebui_cuda --build-arg="USE_CUDA=true" --build-arg="USE_EMBEDDING_MODEL=mixedbread-ai/mxbai-embed-large-v1" .
podman container rm openwebui_cuda; podman volume rm openwebui_cuda;
podman run --gpus all -p 127.0.0.1:3333:8080 --network slirp4netns:allow_host_loopback=true --add-host=ollama.local:10.0.2.2 --env 'OLLAMA_BASE_URL=http://ollama.local:11434' --env 'GLOBAL_LOG_LEVEL=DEBUG' -v openwebui_cuda:/app/backend/data --name openwebui_cuda --replace openwebui_cuda
```

The uvicorn process definitely uses GPU compute resources. Whisper & RAG seems to be working as expected.

jannikstdl:
@justinh-rahb Ready for review!

You should not be able to use these build ARGS:

e.g.
  ```bash
  --build-arg="USE_EMBEDDING_MODEL=intfloat/multilingual-e5-large"
  ```

  For "intfloat/multilingual-e5-large" custom embedding model (default is all-MiniLM-L6-v2), only works with [sentence transforer models](https://huggingface.co/models?library=sentence-transformers). Current [Leaderbord](https://huggingface.co/spaces/mteb/leaderboard) of embedding models.

  ```bash
  --build-arg="USE_OLLAMA=true"
  ```
  For including ollama in the image.

  ```bash
  --build-arg="USE_CUDA=true"
  ```
  To use CUDA exeleration for the embedding and whisper models.

  > [!NOTE]
  > You need to install the [Nvidia CUDA container toolkit](https://docs.nvidia.com/dgx/nvidia-container-runtime-upgrade/) on your machine to be able to set CUDA as the Docker engine. Only works with Linux - use WSL for Windows!

  ```bash
  --build-arg="USE_CUDA_VER=cu117"
  ```
  For CUDA 11 (default is CUDA 12)

**Detailed description for usage in the conversation below, the README or the Dockerfile commentes**

@tjbck also should create 2 more openwebui images with the tag openwebui:with-ollama for included ollama or openwebui :cuda for cuda. As is said below i am not able to test this. This is the fastest way to get this versions running without the need to build the images for yourself.

Also the size of the raw image without any ARGs is lower and the Dockerfile is cleaner and better formatted now.

@yousecjoe @lainedfles Thank you for the effort! You helped to get this up and running!

justinh-rahb:
I've been testing with CUDA and the intfloat e5 large embed model since yesterday already, working great so far but will let you know after I try more stuff out.

lainedfles:
@jannikstdl Is there a good reason why you've omitted pre-loading of the embedding model for CUDA?

```
python -c "import os; from chromadb.utils import embedding_functions; sentence_transformer_ef = embedding_functions.SentenceTransformerEmbeddingFunction(model_name=os.environ['RAG_EMBEDDING_MODEL'], device='cpu')"; \
```

jannikstdl:
> @jannikstdl Is there a good reason why you've omitted pre-loading of the embedding model for CUDA?
> 
> 
> 
> ```
> 
> python -c "import os; from chromadb.utils import embedding_functions; sentence_transformer_ef = embedding_functions.SentenceTransformerEmbeddingFunction(model_name=os.environ['RAG_EMBEDDING_MODEL'], device='cpu')"; \
> 
> ```

Yes I tested it with and without. For me it did not save time. 

If this preloaded the whisper and embedding models for cpu and I first used it with cuda it seems that is was loading it again. I guess because the parameter device_type changed from cpu to cuda and it loads the model (or a optimized version for cuda) again.

Hope you know what I mean. Let me know if you have other experiences.

lainedfles:
> Ok, i could add the preload line. It doesn't hurt. For offline support that would be better, did not think about that.

Thank you!

> As far as external connections i can not tell, have you find anything to disable that metadata pull?

I believe so! See #1419. Note that the problem isn't directly related to any changes within this pull request but I'll elaborate a bit here. The problem seems to be that the Chroma call to sentence_transformers downloads a snapshot format since Chroma doesn't support [passing any arguments](https://github.com/chroma-core/chroma/blob/1d6ceb6a3fe15c33094c2d9a8fa30ff2dfa64379/chromadb/utils/embedding_functions.py#L72) except for `model_name` & `device`.

This results in the filesystem path using blobs and symbolic links. Presently the code calls Chroma with the model shortname. This triggers a huggingface pull every time it's executed.

Instead I've written a dedicated function to handle this on our side by calling huggingface_hub directly to gain the full filesystem path (and update the model). We then pass this path through Chroma and it uses the files directly avoiding the network connection.

Sending arguments like `local_dir` to the huggingface_hub library can provide better control but neither Chroma nor sentence_transformers support the necessary arguments. It looks like a future release of Chroma will support passing kwargs but sentence_transformers will need changes as well.

lainedfles:
Your branch builds as expected and RAG works but whisper inference is broken because *libcudnn_ops_infer.so.8* cannot be found:

```
INFO:apps.audio.main:file.content_type: audio/wav
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): huggingface.co:443
DEBUG:urllib3.connectionpool:https://huggingface.co:443 "GET /api/models/Systran/faster-whisper-base/revision/main HTTP/1.1" 200 1837
INFO:faster_whisper:Processing audio with duration 00:02.940
Could not load library libcudnn_ops_infer.so.8. Error: libcudnn_ops_infer.so.8: cannot open shared object file: No such file or directory
```

The fix: jannikstdl/open-webui#4

tjbck:
Ready to be merged?

.github/workflows/docker-build.yaml
```diff
@@ -53,14 +52,31 @@ jobs:
...
-      - name: Build and push Docker image
+      - name: Build and push Docker image with Ollama
         uses: docker/build-push-action@v5
         with:
           context: .
           push: true
           platforms: linux/amd64,linux/arm64
-          tags: ${{ steps.meta.outputs.tags }}
-          labels: ${{ steps.meta.outputs.labels }}
+          tags: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:with-ollama
```
justinh-rahb, line 81: Drop the `with-`, just `:ollama`, it's cleaner 👌

cleaner

tjbck:
Seems like there's a conflict with our dev branch, let me know once it's been resolved!

jannikstdl:
> Seems like there's a conflict with our dev branch, let me know once it's been resolved!

Resolved now :)

jannikstdl:
@lainedfles why thumbs down? Do you have any additions?

lainedfles:
> @lainedfles why thumbs down? Do you have any additions?

Looks like a bad merge, this reverts d001d7afb12e686c2b9bd2549b9044539a443cb0

jannikstdl:
Ok, that's true, it should be added now that we know what's wrong, we can address it ;) @tjbck

tjbck:
@lainedfles Just noticed, should be added back now!

lainedfles:
Thanks! I've been using a franken-merge for a few days now with this and a few other PRs and I love it.

ghost (commented):
Resolved, thanks!

G4Zz0L1:
Just for info, if you want to make a :rocm image for gpu (not cpu like now), I tested it with the following import, and it seems to work fine.
Maybe with an env `USE_ROCM`, like the cuda one.

`pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/rocm5.7`

justinh-rahb:
Looking into it, thanks @G4Zz0L1 !

PR #1196, opened 2024-03-18
fix: auto scroll fails to stop with small scroll amounts

fbirlik:
## Description

Reduce scroll threshold required to disable auto scroll from 50px to 5px

---

### Fixed

- auto scroll fails to stop with small scroll amounts

### DIFF
--- a/src/routes/(app)/+page.svelte
+++ b/src/routes/(app)/+page.svelte
@@ -813,7 +813,7 @@
 			on:scroll={(e) => {
 				autoScroll =
 					messagesContainerElement.scrollHeight - messagesContainerElement.scrollTop <=
-					messagesContainerElement.clientHeight + 50;
+					messagesContainerElement.clientHeight + 5;
 			}}
 		>
 			<div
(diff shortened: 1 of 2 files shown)


tjbck:
Having the threshold set to 5px seems too lax in my opinion, the threshold value should at least equal to the height of our webui font size.

fbirlik:
Currently, in auto-scroll mode, scroll is sent to bottom with each received token in stream mode, so any move below threshold is reset.  As far as I understand this is the reason for the jumpy behavior.  You need to make at least 50px scroll jump between two received tokens currently.  If tok/sec is high, it is too hard or nearly impossible to overcome this.

Eg. my logitech mouse makes roughly half a line jumps with wheel in mac and I can make finer adjustments with the touchpad.  We should be able to overcome the threshold with minimal effort.

Having said that, px threshold here should only work as an indication of intent in my opinion.

From practical perspective now it feels similar to chatgpt in desktop and phone.  I checked that my font size is 16px and setting the threshold to 15px still feels a bit clunky.

tjbck:
Sure, it mostly looks good to me so let's get back to this if we hear any issue reports from the community!

PR #1218, opened 2024-03-19
fixing typo MessageInput

gibru:
Fixed typo in message input box:

open-webui


tjbck:
Good catch, this also needs to be applied to all translation file btw 😅

justinh-rahb:
> Good catch, this also needs to be applied to all translation file btw 😅

I've got a patch ready if this one doesn't move quickly.

gibru:
@justinh-rahb 

since you've got the patch ready, go ahead. I'm still working out the technical side of collaborating on git so you'll be faster ;)

Issue #1081, opened 2024-03-07
feat: smarter ollama load balancing


tjbck:
right now it just uses `random.choices`, refer to my comment here: https://github.com/open-webui/open-webui/blob/8ed5759d0e9424f87d01fe3f8013116c4ba2004f/backend/apps/ollama/main.py#L39

asedmammad:
@tjbck I can work on this one. I am considering implementing a configuration parameter that allows users to select from various strategies.

tjbck:
@asedmammad Feel free to create a draft PR! I'll also get actively involved in this one!

lewismacnow:
> @tjbck I can work on this one. I am considering implementing a configuration parameter that allows users to select from various strategies.

I'd be very thankful for an ability to restrict models and/or Users to a specific connection (i.e. prevent/stop load balancing in open-webui in some situations).
I drew some examples of these situations in #1527

longregen:
I'd like to bump this since now Ollama has support for multiple loaded models at the same time: https://github.com/ollama/ollama/pull/3418

bkev:
I think it would be really useful to have a preference for a certain node when it's online if that's possible? For example, openweb UI running on a low powered server with Ollama, but a higher powered node, when powered up and available is preferred for to better performance.

haydonryan:
> I think it would be really useful to have a preference for a certain node when it's online if that's possible? For example, openweb UI running on a low powered server with Ollama, but a higher powered node, when powered up and available is preferred for to better performance.

100% came here to post this.  I have an epyc server that stays on 24/7 and gives ok performance.  I'd much rather my desktop (with a 3090) serve the model when it's on.  At the moment I have found a workaround to copy the model name to "llama3.1:latest-desktop" so it appears as something different in the list and manually select that, but would love something more automated. 

Keep up the great work with this!

bkev:
> > I think it would be really useful to have a preference for a certain node when it's online if that's possible? For example, openweb UI running on a low powered server with Ollama, but a higher powered node, when powered up and available is preferred for to better performance.
> 
> 100% came here to post this. I have an epyc server that stays on 24/7 and gives ok performance. I'd much rather my desktop (with a 3090) serve the model when it's on. At the moment I have found a workaround to copy the model name to "llama3.1:latest-desktop" so it appears as something different in the list and manually select that, but would love something more automated.
> 
> Keep up the great work with openwebui!

[...]

PR #1276, opened 2024-03-24
Better ollama load balancing

asedmammad:
## Pull Request Checklist

---

## Description

This introduces a load balancer scheme in ollama app of the backend, resolving issue #1081. The implementation allows setting different strategies in the `.env` file, with support for RoundRobin and WeightedRoundRobin currently available.

---

### Added

- Implemented load balancer scheme for ollama backend

- Added RoundRobin and WeightedRoundRobin strategies

- Configured strategies in `.env.example` file for easy customization


asedmammad:
We might consider adding the ability to change the strategy, weight values and other parameters directly through the UI. This would allow users to easily configure the load balancer settings without needing to modify the `.env` file.

hchasens:
Does this take into account task affinity (model affinity) for model types? 

Ollama requires time to load in new models, so it makes sense to send requests of the same model to the same server. In this way you can create and manage task affinity. For example, if you have two requests going to phi3 and two going to llama3, you're going to want to sort them by type so that both phi3 requests end up at the same server. If a server has to regularly switch between models that'll create a lot of unnecessary overhead (especially if you have limited VRAM or a large model library). You might be able to reduce a bit of overhead by using a session persistence (would only work for low user counts), but you ideally want a model aware load balancer that can dynamically address a set of model requests of q types to a bank of n servers (even better if you can sort the models by size and group smaller models together on a single server). 

Apologies if you've already solved this. I haven't looked over the whole PR, but the problem is an interesting one. I'm busy for the next two weeks but if this hasn't been address yet (or by then) I'd be happy to look into creating an algorithm with you to address this.

asedmammad:
@hchasens 
I was not available in the past few weeks. Your suggestion seems a great idea, and could be a valuable addition to the load balancer. The PR in it's current state does not implement that. We could definitely add it as another load balancing algorithm alongside the ones currently available in the PR.
If you're still available and willing, I'd be happy to collaborate with you on implementing this feature.

hchasens:
@asedmammad I've also been unavailable the last few weeks. I'm still interested in working on this though.

PR #1299, opened 2024-03-25
K8s fixes

jannikstdl:
I configured a new Open WebUI instance on my K8S node and i noticed some points.

**Changed**

- Old naming of the Open WebUI PVC

**Added**

- Sometimes the ENV allocated in the webui-deployment.yaml was overwritten by the `OLLAMA_BASE_URL `check in `backend/config.py` and allocated `http://host.docker.internal:11434` which wont work on K8S. Added a check to set the right base URL `http://ollama-service.open-webui.svc.cluster.local:11434` when running on K8S

### DIFF
--- a/backend/config.py
+++ b/backend/config.py
@@ -218,8 +218,8 @@ if OLLAMA_BASE_URL == "" and OLLAMA_API_BASE_URL != "":
         else OLLAMA_API_BASE_URL
     )
 
-if ENV == "prod" and RUNNING_ON_K8S == "":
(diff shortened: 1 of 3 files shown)


jannikstdl:
If you notice something let me know!

slash-proc:
nice! you beat me to the punch with fixing the default OLLAMA_BASE_URL :D

tjbck:
LGTM, Thanks!

PR #1329, opened 2024-03-28
fix: Helm Chart storageClassName typo

ConnorsApps:
## Pull Request Checklist

- [X] **Documentation:** Have you updated relevant documentation?

---

## Description
Setting the persistent volume class gets ignored by kubernetes due to a typo. See the [k8s docs](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reserving-a-persistentvolume). The only reason this worked previously was that it was using the cluster's default storage class.

---

### Fixed
- Helm chart PVC storage class value

### Changed

- Helm chart PVC storage class no longer using cluster default


tjbck:
Good catch! LGTM, Thanks!

PR #1340, opened 2024-03-28
allow version to work with -rc tag

joequant:
## Pull Request Checklist

- [ X] **Description:** Briefly describe the changes in this pull request.
- [ X] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ X] **Documentation:** Have you updated relevant documentation?
- [ X] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?

---

## Description

fix version check if there version has an -rc tag

---

### Fixed

- version fails if there is an -rc in the version string

### DIFF
--- a/backend/apps/ollama/main.py
+++ b/backend/apps/ollama/main.py
@@ -203,7 +203,7 @@ async def get_ollama_versions(url_idx: Optional[int] = None):
 
         if len(responses) > 0:
             lowest_version = min(
-                responses, key=lambda x: tuple(map(int, x["version"].split(".")))
+                responses, key=lambda x: tuple(map(int, x["version"].replace("-rc", ".").split(".")))
             )
 
             return {"version": lowest_version["version"]}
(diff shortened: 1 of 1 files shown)


tjbck:
LGMT! Just to make sure we won't be breaking anything, could you share a screenshot of how the version with 'rc' tags look like? Thanks!

slash-proc:
There seem to be a few rc tags in the ollama repository. [The newest tag is 0.1.32-rc1 for instance](https://github.com/ollama/ollama/tree/v0.1.32-rc1). 

They don't appear in the releases nor on docker hub. I don't seem the harm but I'm curious what error exactly requires this fix.

slash-proc:
On second thought, the only issue I see is when the comparison is made within the same minor release but where one is the normal release and the other is an release candidate. When comparing the release version with an rc version using `x['version'].replace('-rc', '.')` you'd get 0.1.32 `(0, 1, 32)` coming out after 0.1.32-rc1 `(0, 1, 32, 1)`'

``` python
>>> (0, 1, 32) > (0, 1, 32,1)
False
´´´

tjbck:
@slash-proc yeah I believe it's best if we just drop the '-rc*' from the version, thanks for looking into this!

slash-proc:
and a fix for that issue was pushed right as I posted that lol

Issue #483, opened 2024-01-15
feat: OAuth/OIDC 


K-J-VV:
**Is your feature request related to a problem? Please describe.**
Add feature to enable account creation via OAuth/OIDC

**Describe the solution you'd like**
Integrating ollama-webui with auth providers such as Authentik, Keycloak, etc

**Describe alternatives you've considered**
N/A

**Additional context**
Being able to create a local account as an admin upon initialization should remain, but afterwards the admin account should be able to add the relevant information to change the default login page to redirect to a self-hosted OAuth/OIDC provider

mafrasiabi:
It would be very good to have integration with AzureAD too.

josh:
Ever since the new user accounts were rolled out, I've been wanting some kind of way to delegate auth as well.

In my specific case, my ollama-webui is behind a [Tailscale](https://tailscale.com) VPN. They have a pretty lightweight set of [trusted identity headers](https://tailscale.com/kb/1312/serve#identity-headers) when proxied to. It seems like a few other SSO providers like [Authelia can be configured to pass along trusted headers](https://caddyserver.com/docs/caddyfile/directives/forward_auth#examples).

From the ollama-webui side, you'd configure some envvar like `OLLAMA_AUTH_TRUSTED_HEADER=X-Remote-User` to enable this delegated auth mode. The Python backend would look for this header and automatically log in or create an account for the user given in the header value.

I wondering if this approach would cover your auth use cases as well? If so, I might take a stab at a PR implementation.

explorigin:
Just a heads up. I'm working on integrating https://fastapi-users.github.io/fastapi-users/latest/ into ollama-webui. That will give us OAUTH2 for these clients: https://frankie567.github.io/httpx-oauth/oauth2/#provided-clients. Give me a few weeks.

elvis-cai:
thanks @explorigin 🙇 it will be a killer feature in the open llm UI market. 
I am keen to have this feature for demo in the coming weeks, if you don't mind, Could I know if there's any estimation for it available?

explorigin:
Sorry, no estimate. I started on it but work/life got crazy. If you need it for work then work on it. I've started with fastapi-users library but that has some considerations:
- It uses different REST-API endpoints that are difficult to customize so changing some of these will require a bit of front-end work as well.
- It uses a slightly different DB schema that will need reconciling
[...]

PR #1347, opened 2024-03-28
feat: allow authenticating with a trusted email header

cheahjs:
## Pull Request Checklist

---

## Description

Adds an environment variable `WEBUI_AUTH_TRUSTED_EMAIL_HEADER` that can be specified to be used as the source of a trusted email header that should be used to automatically register and sign in.

Automatic sign in is achieved by adding the boolean `trusted_header_auth` to the result of calling `/api/config`. If this is set to true, the web app immediately triggers a sign in.

On the backend, if the header is set, a few changes are made:
1. `/signin` will always use the value of header specified by `WEBUI_AUTH_TRUSTED_EMAIL_HEADER` as the email address of the user, and we will ignore the password field
    * If the user is not registered, the backend will automatically register the user, with the name set as the email address from the header, and the password set as a random UUID string (to avoid modifying any more of the authentication logic)
2. A user calling `/update/password` will return an error, since a password is useless when the header is set.

It is a security hole to have `WEBUI_AUTH_TRUSTED_EMAIL_HEADER` set and an authenticating reverse proxy that sets the header is not the only way to reach the backend, as a user would be able to spoof any email address in the header and gain access to that email's account.

Styling of the automatic sign in page, and the error message when hitting `open-webui` without the necessary header.
image

This is tangentially related to the request https://github.com/open-webui/open-webui/issues/483, but this delegates the OAuth bit to a reverse proxy and doesn't handle OAuth inside of open-webui.

---

### Added

- 🔒 **Trusted Email Header**: `WEBUI_AUTH_TRUSTED_EMAIL_HEADER` can be used to authenticate users when used with an authenticating reverse proxy.


tjbck:
is there an easy way to test this feature?

tjbck:
I'll merge this to our dev branch for now and test more, Thanks!

MarlNox:
Is this feature implemented in the recent releases, or still in dev?

justinh-rahb:
@MarlNox been implemented for quite a while now, yes.

PR #1352, opened 2024-03-29
fix the select model bar

zishengwu:
## Pull Request Checklist

- [ ] **Description:** Fix the model selection area and separate it from the dialogue area
- [ ] **Changelog:** Adjusted the layout of chat page 

---

## Description

Fix the model selection area and separate it from the chat area, so that the model selection area will always be displayed without the need for manual sliding on top

Issue can be found in: https://github.com/open-webui/open-webui/issues/1350

---

### Added

- None

### Fixed

- None

### Changed

- Adjusted the layout of chat page (/src/routes/(app)/c/[id]/+page.svelte)

### Removed

- None

### DIFF
--- a/src/routes/(app)/c/[id]/+page.svelte
+++ b/src/routes/(app)/c/[id]/+page.svelte
@@ -855,6 +855,14 @@
 		/>
 		<div class="flex flex-col flex-auto">
 			<div
+				class="{$settings?.fullScreenMode ?? null
+					? 'max-w-full'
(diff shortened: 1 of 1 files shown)


tjbck:
The location of model selector was intentionally added at the top of the chat page but I see your point. I'll close this for now but I'll come up with a more streamlined UI to switch between the models soon and add you as a contributor, Thanks!

Issue #797, opened 2024-02-19
feat: static share url to chat


7a6ac0:
At present, open-webui can share chat messages to [openwebui](https://openwebui.com/). Is it possible to add an option to consolidate chat messages into a single static URL for sharing with others (similar to chatgpt)? By visiting this static URL, users would be able to view all the chat messages.

justinh-rahb:
I think I'd like this option as well.

PR #1306, opened 2024-03-26
Feature/support auth by api key

vaayne:
## Pull Request Checklist

- [x] **Description:** Briefly describe the changes in this pull request.
- [ ] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ ] **Documentation:** Have you updated relevant documentation?
- [x] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?

---

## Description

Support creation of API keys for authentication purposes, allowing users to utilize them with our API of [LiteLLM OpenAI Proxy Service](https://litellm.vercel.app/docs/proxy/quick_start)

image

### Example Usage
```python
from openai import OpenAI
from rich import print

BASE_URL = "http://localhost:18080/litellm/api/v1"

TOKEN = "sk-89ca1254569345f5b00d3d9b80b2d174"
MODEL = "bedrock-claude-3-haiku"

oai = OpenAI(
    api_key=TOKEN,
    base_url=BASE_URL,
)

resp = oai.chat.completions.create(
    stream=True,
    messages=[{"role": "user", "content": "What is the meaning of life?"}],
    model=MODEL,
)

for message in resp:
    content = message.choices[0].delta.content
    if not content:
        continue
    print(content, end="")
print()
```

---

### Added

- create API Key in frontend
- auth by api key

### Fixed

- [List any fixes or corrections]

### Changed

- [List any changes or updates]

### Removed

- [List any removed features or files]

PR #1364, opened 2024-03-31
feat: add local sharing of chats

cheahjs:
## Pull Request Checklist

---

## TODOs

- [x] Implement database migrations
- [x] Update shared chat
- [x] Delete shared chat link
- [x] Delete shared chat when the original chat gets deleted
- [x] Test clean install
- [x] Test backward compatibility

## Description

Adds the ability to share a chat locally with other authenticated users on the same instance.

This isn't a complete implementation (deletion isn't implemented at the moment), but the design this implements is:

1. Each chat can have a `share_id` that points to a chat ID that was the snapshot of the chat when it was shared.
   * It being a snapshot means that any updates to the original chat will not be reflected in the shared chat
2. The shared chat is a standard chat DB entry, except with the `user_id` set to `shared` to distinguish it from regular chats.
3. The frontend gets a new route, `/s/[id]`, that is read-only and disables any of the buttons that would modify state.
 
Things that are unaddressed:

1. If the original chat is deleted, the shared chat is not deleted.
    * This applies to all mechanisms of deletion, such as deleting by chat id, or deleting all of a user's chats.
2. There is no mechanism in the frontend to update an already shared chat
3. There is no mechanism in the frontend to delete a shared chat
5. Do we want to make it possible for unauthenticated users to view shared chats, ie a public sharing option? I think this use case is currently addressed by the share to openwebui.com option, but people might have a stronger feeling towards data residency?
6. A potential feature is to allow a user to continue the shared chat.
    * For example, on Poe  image

https://github.com/open-webui/open-webui/assets/818368/0fa1afe6-05c8-495e-937d-180953239d47

May address https://github.com/open-webui/open-webui/issues/797

---

### Added

- 📩 **Local Sharing**: Share your chats with other users on your Open WebUI instance


jannikstdl:
Very nice! For point 4 I think it's okay to let this for authenticated users only. 

Maybe it would be a nice addition in future having the option to change that.

tjbck:
This is amazing! I'll add reviews in a bit!

backend/apps/web/models/chats.py
```diff
@@ -20,6 +20,7 @@ class Chat(Model):
     title = CharField()
     chat = TextField()  # Save Chat JSON as Text
     timestamp = DateField()
+    share_id = CharField(null=True, unique=True)
```
tjbck, line 23: Fine with current implementation, but I'm thinking it might be bettter to have `public` or `shared` boolean flag and reuse `chat_id`?

src/routes/(app)/s/[id]/+page.svelte
Comment (tjbck) [inline src/routes/(app)/s/[id]/+page.svelte:1]: I'll redesigned this part soon, ideally we'd want to separate this page from the main layout to look more something like: https://openwebui.com/c/hub/5b08456b-9481-4a0e-9c91-3933f4da82a6

backend/apps/web/models/chats.py
tjbck: We might also want to add migration script somewhere that runs before the app starts to ensure backward compatibility. Ideally it should automatically detect added fields and reflect changes in our `webui.db`.

tjbck: Related: https://github.com/open-webui/open-webui/pull/1306

src/routes/(app)/s/[id]/+page.svelte
Comment (tjbck) [reply src/routes/(app)/s/[id]/+page.svelte:1]: No worries! I'll take care of the design side of things, I appreciate the effort 🙌

7a6ac0:
If you want to implement a snapshot that any updates to the original chat will not be reflected in the shared chat. I think it can add a new DB Schema that is the same as Chat. 

### SharedChat
- id
- chat_id
- chat

tjbck:
Implemented pretty much everything! If anyone could help with the testing (clean install, backward compatibility), I'd greatly appreciate it!

tjbck:
Just verified it works both with clean install and existing ones, Thanks for the PR!

PR #1386, opened 2024-04-01
feat: default profile image with user initials

dannyl1u:
## Pull Request Checklist

---

## Description

Change the default user profile image to an icon showing their initials, similar to ChatGPT's UI. I believe this looks much better than the current default gravatar image. 

---
image
image

### Added

- New default user profile image set as initial avatar

### Changed

- Default profile image/gravatar, clicking on "Use Gravatar" will set the profile image the initial avatar instead of the old user.png gravatar

### Removed

- Current default user.png gravatar


tjbck:
Great PR! Testing with non-alphanumeric names would be required here.

dannyl1u:
TODO:
- [ ] emoji parsing
- [ ] fit initials of names consisting of multiple names into avatar
image
image

dannyl1u:
names consisting of multiple names will be abbreviated to first letter of first name, first letter of last name
e.g. Edgar Allan Poe -> EP

dannyl1u:
Non-english names work as well
image

dannyl1u:
Tested with non-alphanumeric characters, trailing whitespaces, multiple-word names, PR is ready for review.

lainedfles:
This is awesome, great work! I've tested it with a few different usernames including `nul` characters which it handles well. One problem I encountered was image corruption when canvas spoofing is enabled like in the case of [Firefox privacy.resistFingerprinting](https://support.mozilla.org/en-US/kb/firefox-protection-against-fingerprinting).

image

I've submitted dannyl1u/open-webui#1 in attempt to provide a workaround by testing a 1x1 pixel:

> Browsers and plugins that spoof canvas data produce corrupt images. In attempt to mitigate:
> 
> * Add canvasPixelTest() to test a single pixel and test the RGB values
> * Test canvasPixelTest() inside generateInitialsImage() and use default `/user.png` if failure detected
> * Call canvasPixelTest() directly within settings to avoid setting an invalid image
> * Use toast.error() with 10 second autoClose

dannyl1u:
@lainedfles 
Thanks for the PR! I tried enabling fingerprint protection on Firefox and couldn't reproduce the issue, is there a specific plugin and/or browser you used that produced the corrupt image?

lainedfles:
> @lainedfles Thanks for the PR! I tried enabling fingerprint protection on Firefox and couldn't reproduce the issue, is there a specific plugin and/or browser you used that produced the corrupt image?

I've been able to reproduce it with:

- [LibreWolf](https://librewolf.net/) (which enables it by default)
- [Firefox](https://www.mozilla.org/en-US/firefox/new/)
- [Waterfox](https://www.waterfox.net/)

To enable RFP in Firefox/Waterfox you must set the `privacy.resistFingerprinting` property to *true* in `about:config` manually like:
image
After which the browser must be closed and re-opened. Then upon the initialization of the JavaScript canvas object, Firefox will open a dialog pop-up requesting access to the canvas:
image
Here you must select block else access is allowed! I can make a video if needed.

I've only tested using Linux but I don't think it's an OS problem but an HTML 5 spoofing problem.

dannyl1u:
@lainedfles thank you, was able to reproduce the issue, great catch! I'll review the PR

dannyl1u:
image

The toast.error() gets partially hidden by the toast.success() upon sign-up, so I've decided to remove the toast.error() for signups for now.

lainedfles:
> The toast.error() gets partially hidden by the toast.success() upon sign-up, so I've decided to remove the toast.error() for signups for now.

I've submitted another PR for you to consider 😄 open-webui#2

dannyl1u:
@lainedfles 
Looks great, thanks! I've added some changes, I think toast.info() is more appropriate in this case, also updated the toast message text. Let me know if you have any feedback 👍 
image

lainedfles:
I love it, thank you! Can't wait for the merge.

lainedfles:
@dannyl1u Hey, I've realized some additional problems resolved by dannyl1u/open-webui#3. Hopefully this is the last PR I'll trouble you with.

I've realized that I'd added the wrong toast property `autoClose` which is used by the wrong library. Instead, use the correct property for **svelte-sonner** `duration`. Also, these changes entirely remove the [Gravatar](https://gravatar.com/) functionality. I don't think we want to replace it entirely as many uses use it, instead, I've added a new button named "Use Initials". Finally, add en-US translation for i18l.

> Fix: Restore Gravatar functionality, add button initials, toast duration
>
> - Restore Gravatar functionality
> - Add new button for "Use Initials"
> - Set both buttons to use text-left class
> - Update toast property autoClose to duration (wrong library, my bad!)
> - Update toast messages to clarify that this isn't "Gravatar" but "avatar"
> - Add i18n text to en-US/translation.json

It looks like this:
image

dannyl1u:
I agree with the changes, thanks as always!

tjbck:
Thanks for the effort guys! Currently looking into ways to migrate most of the image generation code to the backend, refac coming soon!

tjbck:
Kept the current image generation implementation as-is (we might want to get back to this later and migrate the code to our backend for better reliability), removed unnecessary notification on signup and modified the styling a bit to blend in nicely with the rest of the components. Great work! Thanks!

PR #1403, opened 2024-04-03
Added boto3 to  requirements.txt

helmanofer:
## Description

In the case of using bedrock with litellm import boto3 exception is raised

---

### Added

- boto3 to  requirements.txt


tjbck:
LGTM, Thanks!

Issue #1302, opened 2024-03-26
Why does this connect to a remote ip?


demesm:
What is the purpose of this connection to a remote ip?

<img width="699" alt="image" src="https://github.com/open-webui/open-webui/assets/23346713/17ffe2a6-db44-496b-92b8-8beaccee68ab">

Darin755:
This is fairly concerning. This project has the potential to allow people to actually have privacy while using AI but some of the design choices do not seem to focus on privacy.

tjbck:
Related #807, To reiterate, WebUI does not make any external connections in operation and everything stays on your machine. We would also love to know what part of the code is making this connection, if I had to make an educated guess based on the precedented issue reports, it seems like one of our dependency libraries upstream is making the connection. Feel free to continue your investigation and keep us updated!

lainedfles:
I've intended to communication my experience in this regard eventually and this ticket is a good opportunity. To date, I've almost exclusively used the Docker container with Podman (self-built and official download). I think I've adequately identified the source of all external network connections I've encountered, which all appear to be entirely legitimate. I've also not yet found anything suspicious in the source code of this project or dependencies (granted I've not performed a full audit 😀 but my first pull request was quickly implemented to disable Chroma telemetry #618).

Note that the [Dockerfile explicitly attempts to preload models for Whisper and Chroma embedding (via sentence_transformers) during the build process](https://github.com/open-webui/open-webui/blob/ac294a74e7bc5f080970c049c656506e36f9407b/Dockerfile#L67). In my experience, the containers function entirely offline with a startup delay (this is a topic for another discussion and IMHO a ticket is warranted). You can now use the `GLOBAL_LOG_LEVEL="DEBUG"` environment variable to get a deeper glimpse.

Here are my notes:

### Connections

- _huggingface.co_: sentence_transformer (RAG)

Connection initiated [here](https://github.com/UKPLab/sentence-transformers/blob/1e35d8c159ea4fcb9253d5d680480ba475cf81b8/sentence_transformers/SentenceTransformer.py#L107):
```
Mar 26 01:30:58 hostname open-webui[305428]: INFO:sentence_transformers.SentenceTransformer:Load pretrained SentenceTransformer: all-MiniLM-L6-v2
Mar 26 01:30:58 hostname open-webui[305428]: DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): huggingface.co:443
```
[...]

PR #1419, opened 2024-04-04
feat: improve embedding model update & resolve network dependency

ghost:
## Pull Request Checklist

---

## Description

Improve embedding model update & resolve network dependency. This permits uvicorn to start without network access and eliminates the network dependency. Huggingface will not be contacted unless manual update is initiated from the GUI/API or `RAG_EMBEDDING_MODEL_AUTO_UPDATE` is set to `True`.

The `huggingface_hub` function `snapshot_download` is called directly to allow for control of the `local_files_only` *kwarg*. Also, clean up RAG main.py to remove unused commented code and associated sentence_transformers direct import. The `embedding_model_get_path()` is used to obtain the full filesystem path to the snapshot which, when passed to Chroma, doesn't attempt to download or update automatically.

Relates to #1302 and may fix issues like #1122.

image

---

### Added

- Add config variable `RAG_EMBEDDING_MODEL_AUTO_UPDATE` to control update behavior
- Add RAG utils `embedding_model_get_path()` function to output the filesystem path in addition to update of the model using `huggingface_hub`
- Add GUI setting to execute manual update process

### Fixed

- Fix Huggingface network dependency on startup

### Changed

- Update and utilize existing RAG functions in main: `get_embedding_model()` & `update_embedding_model()`

### Removed

- Removed unused commented code & associated import


justinh-rahb:
One caution I'd have for updating a model.. it may render previously embedded documents unusable without resetting the vector storage and re-uploading documents.

justinh-rahb:
> As the code is now, Chroma will attempt to download/update every time it's started.

That might explain why it breaks for people sometimes.. 🤔

lainedfles:
I think you're correct. I've provided a little more background in this [comment](https://github.com/open-webui/open-webui/pull/1165#issuecomment-2037827127):

> > As far as external connections i can not tell, have you find anything to disable that metadata pull?
> 
> I believe so! See #1419. Note that the problem isn't directly related to any changes within this pull request but I'll elaborate a bit here. The problem seems to be that the Chroma call to sentence_transformers downloads a snapshot format since Chroma doesn't support [passing any arguments](https://github.com/chroma-core/chroma/blob/1d6ceb6a3fe15c33094c2d9a8fa30ff2dfa64379/chromadb/utils/embedding_functions.py#L72) except for `model_name` & `device`.
> 
> This results in the filesystem path using blobs and symbolic links. Presently the code calls Chroma with the model shortname. This triggers a huggingface pull every time it's executed.
> 
> Instead I've written a dedicated function to handle this on our side by calling huggingface_hub directly to gain the full filesystem path (and update the model). We then pass this path through Chroma and it uses the files directly avoiding the network connection.
> 
> Sending arguments like `local_dir` to the huggingface_hub library can provide better control but neither Chroma nor sentence_transformers support the necessary arguments. It looks like a future release of Chroma will support passing kwargs but sentence_transformers will need changes as well.

tjbck:
@lainedfles We might also want to test this with clean install, but great work! Merging this to dev for now, but we should test this feature more extensively. Thanks for the hard work!

lainedfles:
@tjbck Excellent refactor! Clean install seems to work as expected as well as changing the model.

One potential issue is with the removal of:
```
    if app.state.RAG_EMBEDDING_MODEL_PATH == old_model_path:
        status = False
```
The intention was to notify the user if any change had been made and also act on it in the future to automatically reset the vector DB an re-scan all documents (someday LOL). The refactored version always returns success if the model doesn't require an update or connection to hugginface fails. The huggingface_hub module only seems to produce an exception if the module is changed without a working connection to huggingface. These scenarios might result in users taking unnecessary action to re-import documents.

It seems that present ChromaDB/Sentence_Transformer/Huggingface_Hub functions don't return much as to a change status when handling the embedding models. Perhaps we could compare `get_embedding_model_path(form_data.embedding_model, False)` with `get_embedding_model_path(form_data.embedding_model, True)` instead during the update and return a string instead of bool or a custom HTTP response code that can be handled by the client? At least we'd know if the model is updated or not.

PR #1435, opened 2024-04-05
feat: trigger docker build after release

cheahjs:
## Pull Request Checklist

---

## Description

Fixes the Docker workflow not building tagged images.

This was because releases are programatically created via GitHub Actions, and it is designed specifically such that any events originating from an action cannot trigger another action, as such the release action creating a new tag wasn't triggering the tagged Docker build (only the `main` branch build would have been triggered on push).

This fixes the inability to trigger by allowing the Docker workflow to be triggered by `workflow_dispatch` events, and the release workflow will manually trigger a tagged Docker build once it's done creating the release.

Test example:

* Tagged Docker build: https://github.com/cheahjs/open-webui-fork/actions/runs/8574795311/job/23502387401
    * image
* Release workflow triggering tagged Docker build: https://github.com/cheahjs/open-webui-fork/actions/runs/8574789343/job/23502369111
* Tagged Docker image: https://github.com/cheahjs/open-webui-fork/pkgs/container/open-webui-fork/200277278?tag=v0.1.118-dockertest

This still isn't ideal, since Docker builds are triggered twice, and you end up with different Docker images for what is effectively the same underlying Git commit (`main`, `git-<hash>`, `<tag>` should all ideally point to the same image).

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -57,3 +57,14 @@ jobs:
         path: .
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Trigger Docker build workflow
+      uses: actions/github-script@v7
+      with:
+        script: |
+          github.rest.actions.createWorkflowDispatch({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            workflow_id: 'docker-build.yaml',
+            ref: 'v${{ steps.get_version.outputs.version }}',
+          })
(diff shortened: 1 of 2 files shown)


.github/workflows/build-release.yml
```diff
@@ -57,3 +57,14 @@ jobs:
         path: .
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Trigger Docker build workflow
```
tjbck, line 61: Thanks for the PR! Quick q: would this only run for releases in our `main` branch?

tjbck: Great! Thanks for the clarification!

tjbck:
LGTM, Thanks as always!

PR #1476, opened 2024-04-09
fix: support batching chromadb

buroa:
Fixes a error where submitting large documents results in: https://github.com/chroma-core/chroma/issues/1049


tjbck:
Great! LGTM, Thanks!

PR #1495, opened 2024-04-10
feat: parallelize docker build

cheahjs:
## Pull Request Checklist

---

## Description

Parallelizes the Docker image builds. All 6 possible builds (matrix of standard, cuda, ollama, on x86_64 and arm64) are run in parallel on separate build agents.

Example build: https://github.com/cheahjs/open-webui-fork/actions/runs/8631540693

There are a few more drive by changes that I can back out if necessary:

1. In addition to the static `cuda` and `ollama` tags, build them as suffixes to existing tags (eg `v0.1.118` would get `v0.1.118`, `0.1.118-cuda`, and `0.1.118-ollama`).
2. Setup image layer caching, with image layers being cached into GHA's cache.
    * Effectiveness of this might be minimal, the layers are big enough that we can blow through GitHub's 2GB cache limit.
3. `cuda` and `ollama` tags are built outside of the `main` branch as well.
4. Run the frontend JS builds on the build platform. As the final output are static HTML/JS bundles, we can avoid virtualization and compile it on native architecture.
5. Reordered the dependency install, fetch apt packages before python deps, since I believe that we're more likely to update Python dependencies than OS dependencies, so don't waste the apt cache.

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -134,6 +134,7 @@ jobs:
             type=sha,prefix=git-
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
+            type=raw,enable=${{ github.ref == 'refs/heads/main' }},prefix=,suffix=,value=cuda
           flavor: |
             latest=${{ github.ref == 'refs/heads/main' }}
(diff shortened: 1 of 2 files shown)


tjbck:
Thanks as always!

PR #1532, opened 2024-04-13
fix(helm): remove trailing slash

### LINKED ISSUE #1478: Helm chart adds extra / to end of ollama URL, which is invalid
# Bug Report

## Description

Bug Summary:
This is a follow up from #1422. That got us closer to fixing the issue, but it's still present because a `/` was left behind. This is just to ask that that last `/` be removed. It seems unlike most browsers, the http libraries open-webui uses doesn't like having two `/` in a row in a URL.

Fix should be a quick removal of the trailing `/` from line 10 of helpers.tpl.

Helm chart adds extra / to end of ollama URL, which is invalid. Can by fixed by removing /from line 10 of the _helpers function

Steps to Reproduce:
Install the helm chart and log in and check ollama's status in the settings. Just trim off the / and then refresh and it works.

Expected Behavior:
Should just work without having to make this fix.

Actual Behavior:
Can't connect to ollama and thus no models work, nor can they be downloaded.

## Environment

Linux (kubernetes on Ubuntu 20.04)
Microsoft Edge

## Reproduction Details

**Confirmation:**

- [x] I am on the latest version of both Open WebUI and Ollama.

## Installation Method

Helm

dariothornhill:
## Pull Request Checklist

---

## Description

The [_helper file](https://github.com/open-webui/open-webui/blob/2da7dd67ead27db0f04dd85a77f66fb04a9168be/kubernetes/helm/templates/_helpers.tpl#L13) introduces a trailing `/` to the ollama service url which causes requests to be written with `//` and results in 404 responses from the ollama service.

This results in ollama models being unusable. 

Removing the training slash here resolves the issue.

---

### Fixed

- Removed trailing `/` from ollama url in the helm chart template helper (#1478)

FYI closes #1478

### DIFF
--- a/kubernetes/helm/templates/_helpers.tpl
+++ b/kubernetes/helm/templates/_helpers.tpl
@@ -10,7 +10,7 @@ ollama
 {{- if .Values.ollama.externalHost }}
 {{- printf .Values.ollama.externalHost }}
 {{- else }}
-{{- printf "http://%s.%s.svc.cluster.local:%d/" (include "ollama.name" .) (.Release.Namespace) (.Values.ollama.service.port | int) }}
+{{- printf "http://%s.%s.svc.cluster.local:%d" (include "ollama.name" .) (.Release.Namespace) (.Values.ollama.service.port | int) }}
 {{- end }}
 {{- end }}
 


tjbck:
Thanks!

PR #1533, opened 2024-04-13
refac: non-privileged Dockerfile

ghost:
## Pull Request Checklist

---

## Description

Follow Docker best practices: update Dockerfile to run as a non-root user. This reduces the potential attack surface and follows the Docker security recommendation. See:

https://docs.docker.com/engine/security/#conclusions
https://docs.docker.com/develop/develop-images/instructions/#user

Basic testing succeeds, standard and builds with `--build-arg="USE_CUDA=true"` `--build-arg"USE_OLLAMA=true"` function as expected.

> [!IMPORTANT]
> This changes the effective numeric user ID and group ID of the container user. All existing data volumes must be updated accordingly, for example: `docker run --rm -u root -v open-webui:/app/backend/data -it open-webui chown -Rv open-webui:open-webui /app/backend/`

---

### Added

- Add `ARG`s for new system **_open-webui_** user/group UID & GID.
- Use `USER` instruction where possible to complete the build as limited user.
- Set `ENV PATH=` and execute `chown` & `COPY` with `--chown` as required.

### Fixed

- Security improvement.

### Changed

- Move `LD_LIBRARY_PATH=` required for CUDA to Dockerfile using `ENV LD_LIBRARY_PATH=`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,22 +10,37 @@ ARG USE_CUDA_VER=cu121
 # for better performance and multilangauge support use "intfloat/multilingual-e5-large" (~2.5GB) or "intfloat/multilingual-e5-base" (~1.5GB)
 # IMPORTANT: If you change the default model (all-MiniLM-L6-v2) and vice versa, you aren't able to use RAG Chat with your previous documents loaded in the WebUI! You need to re-embed them.
 ARG USE_EMBEDDING_MODEL=all-MiniLM-L6-v2
+# Limited system user UID
+ARG USE_UID=998
+# Limited system user GID
+ARG USE_GID=998
+
 
 ######## WebUI frontend ########
 FROM --platform=$BUILDPLATFORM node:21-alpine3.19 as build
 
+# Use args
+ARG USE_UID
+ARG USE_GID
+
+# User/Group
+RUN addgroup -S open-webui -g $USE_GID && adduser -S open-webui -G open-webui -u $USE_UID
+USER open-webui:open-webui
+
 WORKDIR /app
 
-COPY package.json package-lock.json ./
+COPY --chown=$USE_UID:$USE_GID package.json package-lock.json ./
 RUN npm ci
 
-COPY . .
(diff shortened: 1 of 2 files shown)


tjbck:
More testing from other people wanted here!

jannikstdl:
Yes that would be good for K8s as well since the default setting on most providers does not let you run a container in root mode.

smichard:
I've conducted a series of tests on the provided Dockerfile with the following observations:

1. **Container Build**: Successful build of the container image using the Dockerfile.
2. **Execution on Linux**: The container runs as expected on a Linux host, in line with the documentation.
3. **Execution on Kubernetes Platforms**:
   - **Google Kubernetes Engine (GKE) and OpenShift**:
     - **With `WEBUI_SECRET_KEY` set**: Encountered an error indicating that the `uvicorn` module is not found.
     - **Without `WEBUI_SECRET_KEY` set**: Faced a permissions error preventing the execution of the `start.sh` script.
webui_secret_provided
no_webui_secret_provided

justinh-rahb:
@smichard It appears if it's running in rootless mode that the user account is not given the correct `$PATH` variable to find `uvicorn` that would have been installed as a system Python package in the container image.

justinh-rahb:
> This actually highlights a new incompatibility with `uv` introduced per #1577. It doesn't work properly with userspace. We'd need to run `uv` with root & `--system` and shuffle paths around to account for differences in library paths, or create the required directories (and set proper permission & ownership) before running `uv` as the limited user which will likely require access to system paths like `/usr/local/bin`, or wait until upstream provides a solution: [astral-sh/uv#1584](https://github.com/astral-sh/uv/issues/1584).

Well darnit, I knew there was going to be a catch. Seeing as how we aren't working with rootless _yet_, seems no harm **right now**, but ya we'll need to figure it out.

lainedfles:
Actually, I don't want to pursue this right now. Thanks!

PR #1539, opened 2024-04-13
feat: add {{prompt:start:length}} and {{prompt:end:length}} to title gen

cheahjs:
## Pull Request Checklist

---

## Description

Adds support for specifying a subset of the start and end of prompt for title generation. For example, the following title generation prompt uses the first and last 4096 characters of the prompt:

```
Create a concise, 3-5 word phrase with an emoji as a header for the following query.
Do not use the word title.
Do not use any formatting.
Examples of titles:
😢 Sad Story
🎂 How To Bake A Cake
✉️ Email Draft
💻 Programming Help

Here is the first 4096 characters of the query:
{{prompt:start:4096}}

Here is the last 4096 characters of the query:
{{prompt:end:4096}}
```

Rationale for this is when using models with mismatched context windows, which breaks when the shorter window model is used for generating the title. For example, in my use case, I'm using Gemini 1.5 Pro as the main model, and using the cheaper Claude 3 Haiku model for title generation, but any chats exceeding Claude's context window would fail. This makes the assumption that most instructions relevant to title generation will be either at the start or end of the prompt.

---

### Added

- ✨ **Enhanced Title Generation**: Added support for using subsets of the prompt for title generation, allowing for more flexibility and control.


tjbck:
LGTM, Thanks!

varenc:
Regarding this example title gen prompt:

```
Here is the first 4096 characters of the query:
{{prompt:start:4096}}

Here is the last 4096 characters of the query:
{{prompt:end:4096}}
```

If the `prompt` in question was, say, 5000 characters, wouldn't this prompt essentially just duplicate the number of characters in the query? It'd send 8192 characters, instead of 5000, most of which are duplicated. This seems a bit wasteful and needlessly more costly for hosted models.

Unfortunately if you want to include both the first and the last chars I don't think there's a clean solution to this without extra logic to change behavior based on the prompt's length. 

The simplest solution I can think of is to use a richer template language for the title gen prompt like [ejs](https://github.com/mde/ejs). With that you could do something like this:

```ejs
Create a concise, 3-5 word phrase with an emoji as a header for the following query.
....
<%  if (prompt.length > 8192) { %>
Here is the first 4096 characters of the query:
<%= prompt.slice(0, 4096) %>

Here is the last 4096 characters of the query:
<%= prompt.slice(-4096) %>

<% } else { %>
Here is the query:
<%= prompt %>
<% } %>
```

There's other reasons for using a template language like ejs. Rolling your own template parsing like this example works fine in practice, but it could get tripped up by some weird adversarial examples. For example:

```
let template="{{prompt:end:5}} {{prompt:end:10}} {{prompt:start:5}} {{prompt:start:10}}"
let prompt="abcdefghijklmnopqrstuvwxyz"
console.log(promptTemplate(template, prompt));
```
outputs: `vwxyz {{prompt:end:10}} abcde {{prompt:start:10}}` which seems wrong.

But if you set template to:
```
let template="{{prompt:end:5}} xxxxxxxxx {{prompt:end:10}} {{prompt:start:5}} xxxxxxxxxxxx {{prompt:start:10}}"
```
It then correctly outputs `vwxyz xxxxxxxxx qrstuvwxyz abcde xxxxxxxxxxxx abcdefghij`

I haven't tracked down exactly why but I think it has something to do with the regex search running on a template string that's changing during the while loop and the regex doesn't look for characters behind an index it think it already processed. This exact example isn't a big deal, but I think it hints at some of the benefits of using a templating system.

PR #1569, opened 2024-04-16
feat: TTS support for openedai-speech external api

ghost:
## Pull Request Checklist

---

## Description

Add _**optional**_ support for the [openedai-speech](https://github.com/matatonic/openedai-speech) TTS speech engine API. The *ENABLE* & *BASE_URL* variables must be set to activate integration. *SPEECH_VOICES* can be set to limit or add voices to match the backend openedai-speech configuration. The default voice is `alloy` matching the OpenAI implementation.

Seems to be working well. I've only tested using the [minimal Docker container](https://github.com/matatonic/openedai-speech/blob/main/Dockerfile.min) without CUDA & GPU: `--env 'ENABLE_OPENEDAI_SPEECH=true' --env 'OPENEDAI_SPEECH_BASE_URL=http://127.0.0.1:8000/v1'`.

Based on work by @jmtatsch. Thanks!

Edit: Oops, I meant to tag @lee-b.

---

### Added

- Update `config.py` to add the environment variables: **ENABLE_OPENEDAI_SPEECH** (default: false), **OPENEDAI_SPEECH_BASE_URL** (default: ''), and **OPENEDAI_SPEECH_VOICES** (default: `alloy;echo;fable;onyx;nova;shimmer`).
- New backend functions for frontend features: `get_openedai_audio_voices()` & `get_openedai_audio_enabled()`.
- New frontend functions: `synthesizeOpenedAISpeech()`, `getOpenedAISpeechEnabled()`, and `getOpenedAISpeechVoices()`
- GUI Option to select "Opened AI Speech" (openedai) TTS engine with minimal settings support and is toggled by **ENABLE_OPENEDAI_SPEECH**.

### Fixed

- Fix to: _Replace quotations to avoid corrupt TTS (looking at you tinyllama)_.

### Changed

- Share the existing OpenAI `speech()` backend function with `/opened/audio/speech` path alias.


lainedfles:
> Would suggest that this is more general than OpenedAI-Speech, and the variables shouldn't contain OPENEDAI, but should simply be something like OPENAI_TTS_BASE_URL to override the more general OPENAI_BASE_URL.

I'd selected OPENEDAI for speech because the generic OPENAI_TTS_BASE_URL override prevents other users from accessing the vanilla OpenAI service. I think there's still a place for this functionality but I didn't want to include it here since my intention is to support only OpenedAI.

> For background, OpenedAI-Speech emulates **_only_** the speech parts of OpenAI's API, and so you need the option to support that as separate endpoints for TTS/STT (or just for speech in general).

Yes, this is exactly my intention! I think we need to support the capability to use both services simultaneously, per-user.

> LocalAI supports both text gen and STT / TTS, though, so would be more natural as an easy replacement. However, the ability to select a different endpoint for TTS/STT is still welcome, as OpenedAI's TTS/STT support is a bit more sophisticated.

I'm largely unfamiliar with LocalAI, I'll have to check it out.
 
> So really I think we just want:
> 
> OPENAI_BASE_URL (for text gen, and anything else that isn't overridden, like TTS / STT) OPENAI_TTS_BASE_URL (for when you want to specify a particular server for TTS) OPENAI_STT_BASE_URL (for when you want to specify a particular server for STT) OPENAI_SPEECH_VOICES (if we don't want to implement querying the voices right now)

I agree that this would be beneficial but I'd hope to leave it out of this PR unless there's a consensus otherwise?

> Would also suggest that we just use these TTS/STT vars if set and non-blank, rather than having separate ENABLE var(s).

I'd actually set it up like this when I began but ended up adding the ENABLED var for a justification in my mind that I can't recall. I think I was trying to simplify the UI integration with potential to leave the other configuration persistent in the future (in a database perhaps?). I'll revisit. Thanks for you input!

justinh-rahb:
I can see good points on both sides, I think it best we get a tie-breaker vote for how to proceed perhaps?

tjbck:
Let's follow the existing convention. `TTS_OPENAI_API_BASE_URL` would be appropriate here.

lainedfles:
> Let's follow the existing convention. `TTS_OPENAI_API_BASE_URL` would be appropriate here. (e.g. check out my custom openai url implementation for RAG embedding model)

That's what I'd hoped to avoid with this PR. I don't personally use Open AI but want to support "Opened" AI for TTS while allowing other users to use the vanilla implementation unchanged should they choose. I think it's probably best that we  close this PR to allow someone else to author a different solution.

PR #1602, opened 2024-04-18
feat: (pending closure) separate openai speech API provider

lee-b:
WORK IN PROGRESS

FILING THIS PR MORE FOR AWARENESS/TRACKING RIGHT NOW - SUGGESTIONS WELCOME

In particular, I should:

- merge the docker-customized stuff into the repo's own docker files and startup scripts
- probably try to implement authenticated remote openai support (though I don't care for it personally) or just not have stub code in there (is_openedai_speech) that hints at future support, for now.

## Description

Add support for a separate OpenAI-compatible speech API base URL.

This allows the use of Ollama or an OpenAI-compatible service such as Llama.CPP for text gen (which may be faster / easier for text generation), and another API provider like LocalAI or OpenedAI-Speech for the text-to-speech and speech to text functionality.

---

### Added

- Add support for a separate OpenAI-compatible speech API providers, using an OPENAI_AUDIO_BASE_URL environment variable.
- docker-compose.customized.yaml which sets the new env vars
- run_customized.sh which simplifies running openai with the above customized config

### Fixed

- Allows use of llama.cpp or another text-only OpenAI endpoint / model whilst using a separate/dedicated speech API provider

### Changed

- Supports a new OPENAI_AUDIO_BASE_URL, which will be used rather than OPENAI_BASE_URL for speech services, when set.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -298,3 +298,6 @@ dist
 .yarn/build-state.yml
 .yarn/install-state.gz
 .pnp.*
+
+ollama-data/
+
--- a/.gitignore
+++ b/.gitignore
@@ -301,3 +301,4 @@ dist
 
 ollama-data/
 
+.vscode/
--- a/backend/apps/openai/main.py
+++ b/backend/apps/openai/main.py
@@ -100,18 +100,18 @@ async def update_openai_key(form_data: KeysUpdateForm, user=Depends(get_admin_us
 
 @app.post("/audio/speech")
 async def speech(request: Request, user=Depends(get_verified_user)):
-    audio_base_url = os.environ.get("OPENAI_AUDIO_BASE_URL", "")
-    if audio_base_url:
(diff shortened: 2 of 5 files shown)


justinh-rahb:
There was another PR that I believe was most of the way there to doing this already before it was closed and the fork branch deleted by the author. Shame, it probably only needed a bit more work. It was decided in discussion of that that we should follow this convention for the environment variable:

> Let's follow the existing convention. TTS_OPENAI_API_BASE_URL would be appropriate here. (e.g. check out my custom openai url implementation for RAG embedding model)

tjbck:
Feel free to continue working on this PR but I'll be working on this feature myself following all the existing conventions.

tjbck:
image


OpenAI-compatible url support has been added to our dev branch, let us know if you encounter any issues!

PR #1653, opened 2024-04-21
fix: litellm as subprocess

tjbck:
image


justinh-rahb:
Beware of Judas in your midst.

PR #1693, opened 2024-04-22
feat: hybrid search with reranking

buroa:
This adds three features:

1. Hybrid search. Ensemble `BM25` + `ChromaDB` together. I use the documents returned from the ChromaDB collection to build the BM25 retriever.
2. Reranking support. Using native `sentence_transformers` we can rerank the results from the hybrid search to get better relevant documents.
3. Relevance filter. Best used with a reranking model, you can set the threshold to automatically filter out non-relevant documents.

I updated the UI to allow setting the reranking model and the relevance filter via API calls.

Lastly, `query_embeddings_function` returns a lambda function that is responsible for providing the correct embeddings for whatever embedding model / engine the user chose. I use this inside the custom ChromaDB retriever when searching and replaced it where relevant as well.

Please let me know if I need to explain anything here. I hope it makes sense :)


justinh-rahb:
I need to test this more to really know if one or the other way is qualitatively better, but functionally everything appears to be working as intended and didn't break the original implemention. LGTM!

buroa:
> I need to test this more to really know if one or the other way is qualitatively better, but functionally everything appears to be working as intended and didn't break the original implemention. LGTM!

Thanks for testing. It all depends on what data you are using. I have a few CSV-type datasets and this does much better with `needle in the haystack` type queries.

mario-nguyen:
> This adds two things:
> 
> 1. Hybrid search. Ensemble `BM25` + `ChromaDB` together. I use the documents returned from the ChromaDB collection to build the BM25 retriever.
> 2. Reranking support. Using native `sentence_transformers` we can rerank the results from the hybrid search to get better relevant documents.
> 
> I updated the UI to allow setting the reranking model, just like the embedding model.
> 
> `query_embeddings_function` returns a lambda function that is responsible for providing the correct embeddings for whatever embedding model / engine the user chose. I use this inside the custom ChromaDB retriever when searching and replaced it where relevant as well.
> 
> Please let me know if I need to explain anything here. I hope it makes sense :)

Could you share the data and test results from the pull request? This seems like a really promising solution, and I'd like to understand its impact better. Thank you for your contribution!

buroa:
> > This adds two things:
> > 
> > 1. Hybrid search. Ensemble `BM25` + `ChromaDB` together. I use the documents returned from the ChromaDB collection to build the BM25 retriever.
> > 2. Reranking support. Using native `sentence_transformers` we can rerank the results from the hybrid search to get better relevant documents.
> > 
> > I updated the UI to allow setting the reranking model, just like the embedding model.
> > `query_embeddings_function` returns a lambda function that is responsible for providing the correct embeddings for whatever embedding model / engine the user chose. I use this inside the custom ChromaDB retriever when searching and replaced it where relevant as well.
> > Please let me know if I need to explain anything here. I hope it makes sense :)
> 
> Could you share the data and test results from the pull request? This seems like a really promising solution, and I'd like to understand its impact better. Thank you for your contribution!

I cannot share my data as it's proprietary.

que-nguyen:
> I cannot share my data as it's proprietary.

Apologize for the confusion. I was referring to dummy data,  not proprietary information.

Dockerfile
```diff
@@ -8,8 +8,9 @@ ARG USE_CUDA_VER=cu121
 # any sentence transformer model; models to use can be found at https://huggingface.co/models?library=sentence-transformers
 # Leaderboard: https://huggingface.co/spaces/mteb/leaderboard 
 # for better performance and multilangauge support use "intfloat/multilingual-e5-large" (~2.5GB) or "intfloat/multilingual-e5-base" (~1.5GB)
-# IMPORTANT: If you change the default model (sentence-transformers/all-MiniLM-L6-v2) and vice versa, you aren't able to use RAG Chat with your previous documents loaded in the WebUI! You need to re-embed them.
+# IMPORTANT: If you change the embedding model (sentence-transformers/all-MiniLM-L6-v2) and vice versa, you aren't able to use RAG Chat with your previous documents loaded in the WebUI! You need to re-embed them.
 ARG USE_EMBEDDING_MODEL=sentence-transformers/all-MiniLM-L6-v2
+ARG USE_RERANKING_MODEL=""
```
tjbck, line 13: We might want to set a default reranking model as well if it's not too big!

Dockerfile
```diff
@@ -62,8 +65,11 @@ ENV WHISPER_MODEL="base" \
 
 ## RAG Embedding model settings ##
 ENV RAG_EMBEDDING_MODEL="$USE_EMBEDDING_MODEL_DOCKER" \
-    RAG_EMBEDDING_MODEL_DIR="/app/backend/data/cache/embedding/models" \
+    RAG_RERANKING_MODEL="$USE_RERANKING_MODEL_DOCKER" \
     SENTENCE_TRANSFORMERS_HOME="/app/backend/data/cache/embedding/models"
+
+## Hugging Face download cache ##
+ENV HF_HOME="/app/backend/data/cache/embedding/models"
```
tjbck, line 72: Is this being used?

Dockerfile
buroa: They are pretty large and compute intensive so I left it out.

Dockerfile
tjbck: Thanks for the clarification!

tjbck:
LGTM, Thanks for the hard work 💪

Dockerfile
buroa: @arsaboo It really depends on your data. I am using embedding BAAI/bge-m3 with a reranking BAAI/bge-reranker-v2-m3 with good results.

Issue #1162, opened 2024-03-14
feat: relative path support


hahchenchen:
# Bug Report
webui docker images do not support relative path. 
## Description
for xample, i want to start webui at localhost:8080/webui/, does the image parameter support the relative path configuration?

hahchenchen:
@tjbck

blackshot:
i'm having the same problem, i can't deploy it behind a reverse proxy under specific path ONLY because static files doesn't have the path prefix
![image](https://github.com/open-webui/open-webui/assets/3688208/129ef7f8-d1d2-46ae-8c27-42946a4bef4d)

CyberIngeniero:
I also have the same problem.

tjbck:
If not officially supported, but I can see that being useful for a lot of people. Feel free to make a PR!

justinh-rahb:
This can be accomplished through reverse-proxy and request rewrite rules as well.

blackshot:
> This can be accomplished through reverse-proxy and request rewrite rules as well.

nope, i deployed behind a reverse proxy but static files needs to be compiled with basepath too

hahchenchen:
> If not officially supported, but I can see that being useful for a lot of people. Feel free to make a PR!

no, the proxy is not useful. the static files path is hard code.

blackshot:
https://kit.svelte.dev/docs/configuration#paths

hahchenchen:
> This can be accomplished through reverse-proxy and request rewrite rules as well.

no, the proxy is not useful. the static files path is hard code.

hass-demacia:
Use caddy2, Caddyfile config
```
        rewrite /_app/* /open-webui{http.request.uri}

        handle_path /open-webui/* {
                reverse_proxy http://localhost:8080
        }
```

It seems browser cant fetch javascript static files. But get a runtime error:
```
app.78c246b0.js:1 Error: Not found: /open-webui/
    at ce (start.867272e0.js:1:14918)
    at re (start.867272e0.js:1:8155)
    at Object.goto (start.867272e0.js:1:18073)
    at Module.Nt (start.867272e0.js:3:1691)
    at open-webui/:66:11
```

behackl:
I've successfully deployed an instance behind a reverse proxy using the branch attached to #1701. I wouldn't say that I am particularly happy with the way how I solved it, but perhaps this is useful to whoever decides to work on a proper implementation. :smile:

lmorin-inria:
[...]

PR #1701, opened 2024-04-23
(feat) allow deployment under relative base path

behackl:
## Pull Request Checklist

---

## Description

This introduces some changes mainly to the frontend (plus the construction of `manifest.json` in the backend) that allows setting a relative base path via the environment variable (or, in the context of the docker build process, the build argument) `WEBUI_BASE_PATH`.

Please note: I am not familiar with svelte at all, and it feels like there might be a much simpler solution than what I did here -- however, I've used this branch to deploy a working instance behind a standard reverse proxy configuration (in this case Apache):

```
ProxyPreserveHost On
ProxyPass /chat http://example.com:8077
ProxyPassReverse /chat http://example.com:8077
```

such that the interface is available at `https://original.example.com/chat`.

Related to #1162.

Unfortunately I won't have time to put much more effort towards this, maybe this can be useful to whoever comes up with a proper implementation of this feature eventually. :-)

---

### Added

- Docker build argument `WEBUI_BASE_PATH` which controls the relative base path used in the built frontend (plus in `manifest.json`).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,8 @@ ARG USE_CUDA_VER=cu121
 # IMPORTANT: If you change the default model (all-MiniLM-L6-v2) and vice versa, you aren't able to use RAG Chat with your previous documents loaded in the WebUI! You need to re-embed them.
 ARG USE_EMBEDDING_MODEL=all-MiniLM-L6-v2
 
+ARG WEBUI_BASE_PATH=""
+
 ######## WebUI frontend ########
 FROM --platform=$BUILDPLATFORM node:21-alpine3.19 as build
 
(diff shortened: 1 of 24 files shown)


behackl:
Thanks!

> WEBUI_BASE_PATH can be changed at docker build time only, not at deployment time. Is it correct?

Indeed, this solution only works when building the docker image.

A version that works during deployment time too would require restructuring the docker build process a bit; currently `node` is only available in the first build stage (and not in the final container). The frontend would need to be rebuilt whenever the container starts up.

tjbck:
Closing due to conflicts. Unless there's a more elegant solution to this, we should just add docs on how to host our webui with a relative base path using other reverse proxy tools like nginx. I'm not a huge fan of this current solution for adding relative base path support natively. I appreciate the effort you put in regardless 🙌

behackl:
I guess it would still be necessary to change the frontend-generated URLs from absolute to relative ones. Except if you were thinking of some sort of automatic URL rewriting by the reverse proxy; I've found these sort of solutions to be somewhat unreliable though.

In any case, I'm sure someone will be able to come up with a clean solution. :-)

PR #1722, opened 2024-04-24
feat: add ability to pause model for current chat

Joe0:
## Description

Added the ability to pause models in chats. This makes it far less annoying to switch between models in a single chat.

image

---

### Added

 - Added the ability to pause models in chats.

### DIFF
--- a/src/lib/components/chat/ModelSelector.svelte
+++ b/src/lib/components/chat/ModelSelector.svelte
@@ -10,34 +10,87 @@
 
 	const i18n = getContext('i18n');
 
+	type Model = {
+		name: string;
+		paused: boolean;
+	};
+
 	export let selectedModels = [''];
+	export let renderedModels: Model[] = [{ name: '', paused: false }];
+
+	const updatePreferences = (preferences: string[]) => {
+		if (
+			preferences.join(',') ===
+			renderedModels
+				.filter((models) => !models.paused)
+				.map((it) => it.name)
+				.join(',')
+		) {
+			return;
+		}
+
+		const newRenderedModels = preferences.map((model) => ({
+			name: model,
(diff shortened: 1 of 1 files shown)


justinh-rahb:
Thank you for your contribution with the pause functionality for models in chats. I've reviewed the changes, and I appreciate the effort put into enhancing our model management capabilities.

I'd like to understand more about the specific use cases this feature is targeting. From my perspective, if a model is not needed temporarily, it could simply be removed from the selected list for a chat and added back when necessary. Could you please elaborate on the advantages of pausing a model in the select list over removing it? Are there particular scenarios or issues that this feature aims to mitigate?

Understanding the real-world applications and the problems this solution is intended to solve would help us all better appreciate its utility and make an informed decision on integrating this feature.

Joe0:
Thanks for the quick response.

Oftentimes I'm running a model locally and want that to always run, but when I need better/more accurate responses I use a model like gpt-4-turbo which has costs associated with it. When going back and forth, it's extremely annoying to constantly add and remove gpt-4-turbo, as it's almost at the end of the list.

Other times, I have several models I want to use in the chat, but only want to use them some of the time. Constantly adding and removing them for individual messages is quite burdensome currently, and this makes it way easier to determine which models have a chance to "speak". The motivation here isn't necessarily monetary, but time-wise, some of the models are slow and I want a quick response.

If you think of it in terms of humans talking in a group, I use this feature similar to asking a specific person or subset of people a question mid conversation because they're better equipped to answer.

tjbck:
Ideally I'd like to not add any additional buttons to the navbar, I'll see what could be done with this. Thanks for the PR.

tjbck:
Added '@' implementation to our dev branch! Let me know if you encounter any issues with it!

PR #1756, opened 2024-04-25
feat: toggle hybrid search

buroa:
This is an update to my previous PR which adds a toggle for hybrid searching. Once toggled, it will unlock the reranking model field as well as the relevance score threshold. Even if you turn it on, you do not have to use reranking. It's only used if the reranking model is updated and set.

Also fixed a small issue where querying a collection would result in the scores out of order. Embeddings use a distance while reranking scores (0-1). This ensures the collections are ranked in reverse order when using hybrid + reranking, so they appear .99, .98. .97...; while embeddings (no reranking) are 300, 301, 302... respectively (this *has not changed* even with this PR).

Screenshot 2024-04-25 at 5 32 02 PM

Ref: #1693


tjbck:
LGTM, Thanks!

PR #1773, opened 2024-04-26
Fixed OpenRouter heart beats breaking streaming

Rmaan:
## Pull Request Checklist

- [X] **Documentation:** Have you updated relevant documentation?
  No need
  No

---

## Description

Open WebUI doesn't work with Open Router as advertised in the Readme. When you try to do a chat completion, you can see errors on the front-end like:
image
Underlying reason is OpenRouter, when used in streaming mode sends heart beats in the stream like `: OPENROUTER PROCESSING` but front-end is not ignoring them.

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format) all events that starts with a colon should be ignored. I simply did that.

---

### Fixed

- Fixed OpenRouter heart beats breaking streaming

### DIFF
--- a/src/lib/apis/streaming/index.ts
+++ b/src/lib/apis/streaming/index.ts
@@ -31,7 +31,7 @@ async function* openAIStreamToIterator(
 				console.log(line);
 				if (line === 'data: [DONE]') {
 					yield { done: true, value: '' };
-				} if (line.startsWith(':')) {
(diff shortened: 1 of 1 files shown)


tjbck:
@cheahjs

tjbck:
Thanks!

Issue #1397, opened 2024-04-02
feat: external DB support


tjbck:

Lanhild:
PostgreSQL would be a good candidate IMO

lainedfles:
> PostgreSQL would be a good candidate IMO

I agree but I think the best consideration is to support ORM, something like [SQLAlchemy ORM](https://docs.sqlalchemy.org/en/20/orm/index.html#sqlalchemy-orm). As long as we adhere to **Object Relational Mapper** compatible statements, many major databases could be supported! SQLAlchemy presently supports a variety of [dialects](https://docs.sqlalchemy.org/en/20/dialects/index.html#dialects) including [PostgreSQL](https://docs.sqlalchemy.org/en/20/dialects/postgresql.html), [MySQL and MariaDB](https://docs.sqlalchemy.org/en/20/dialects/mysql.html), [SQLite](https://docs.sqlalchemy.org/en/20/dialects/sqlite.html), [Oracle](https://docs.sqlalchemy.org/en/20/dialects/oracle.html), and [Microsoft SQL Server](https://docs.sqlalchemy.org/en/20/dialects/mssql.html).

justinh-rahb:
@lainedfles I am in full agreement that we need a more flexible solution to data storage, but I get real nervous about ORMs, heard too many horror stories of absolutely unhinged monstrosities of queries full of JOINS being emitted by them. They can make writing the code easier, but it's usually at the cost of efficiency.

buroa:
Postgres would be the best use case here.

TheBongStack:
Any update on this one? Where is the data being handled to be stored in database? at /rag/utils.py ?

sfn-git:
> Any update on this one? Where is the data being handled to be stored in database? at /rag/utils.py ?
Not sure whether you are talking about storing the information relating to connecting to the database but I do not believe this path has any relation to that.

 I am currently looking at creating relational diagram based off of the models located at backend\apps\web\models. I also believe [SQLAlchemy ORM](https://docs.sqlalchemy.org/en/20/orm/index.html#sqlalchemy-orm) is the best option to implement this from a relational database perspective. 

Would there also be considerations for something like Mongo in the future?

Lanhild:
> Would there also be considerations for something like Mongo in the future?
@sfn-git 

IMO, MongoDB isn't appropriate for the practical use case here as Open WebUI database is well defined

PR #1791, opened 2024-04-27
feat: add support for using postgres for the backend DB

cheahjs:
## Pull Request Checklist

---

## Description

Addresses the application DB portion of https://github.com/open-webui/open-webui/issues/1397

Adds a new env var `DATABASE_URL` that can be used to point to a SQLite database, Postgres database, or MySQL database. I have only tested with SQLite and Postgres. The URL schema can be found at https://docs.peewee-orm.com/en/latest/peewee/playhouse.html#connect

As part of this work, a few schema migrations were done.

1. `DateField` and `DateTimeField` were migrated to `BigIntegerField`
  * This is because we've been storing unix epoch timestamps into these fields, which SQLite doesn't care, but Postgres and MySQL do not accept integers in `date` or `datetime` fields.
  * There's also additional logic in the migration scripts to avoid having to actually migrate from `date`/`datetime` to `integer`, as this requires casting and custom logic per database/
2. Some `CharField`s were migrated to `TextFields`. `CharField` has a default max length of 255, but SQLite doesn't enforce this length limit.

This **does not** migrate an existing SQLite DB onto a remote DB.

To test locally with Postgres,

1. Start a postgres instance `docker run --name webui-postgres -e POSTGRES_PASSWORD=mysecretpassword -p:5433:5432 --rm postgres`
4. Run the backend with a Postgres DATABASE_URL: `DATABASE_URL=postgres://postgres:mysecretpassword@localhost:5433/postgres GLOBAL_LOG_LEVEL=DEBUG ./dev.sh`

---

### Added

- **🛢️ External Database Support**: Connect to a custom SQLite or Postgres database using the `DATABASE_URL` environment variable.

### DIFF
--- a/backend/apps/litellm/main.py
+++ b/backend/apps/litellm/main.py
@@ -21,6 +21,8 @@ from utils.utils import get_verified_user, get_current_user, get_admin_user
 from config import SRC_LOG_LEVELS, ENV
 from constants import MESSAGES
 
+import os
+
 log = logging.getLogger(__name__)
 log.setLevel(SRC_LOG_LEVELS["LITELLM"])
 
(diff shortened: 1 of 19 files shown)


tjbck:
Amazing stuff 🙌 LGTM, Thanks!

PR #1780, opened 2024-04-26
fix: harden openai streaming parsing

cheahjs:
## Pull Request Checklist

- [x] **Description:** Briefly describe the changes in this pull request.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ ] **Documentation:** Have you updated relevant documentation?
- [ ] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?

---

## Description

Fixes https://github.com/open-webui/open-webui/issues/1759. Uses optional chaining to prevent errors when a delta is not present, and also catches any errors parsing the JSON.

This is complementary to https://github.com/open-webui/open-webui/pull/1773, since without #1773 this fix would still log errors.

---

### Fixed

- Compatibility with OpenAI endpoints that return unexpected values

PR #1795, opened 2024-04-27
fix: handle carriage returns in OpenAI streams

cheahjs:
## Pull Request Checklist

---

## Description

Trim trailing carriage returns from OpenAI responses. Carriage returns won't break after https://github.com/open-webui/open-webui/pull/1780 was merged, but this will prevent errors from being logged.

We should probably switch to using something that is more spec-compliant, like https://github.com/openai/openai-node/blob/master/src/streaming.ts potentially (licensed under Apache)


tjbck:
LGTM, Thanks!

PR #1799, opened 2024-04-27
fix: docker gpus option "all" support in run-compose.sh

karaketir16:
## Description

`docker --gpus=all` is a valid and mostly used command. regex updated to match this.

---

### Changelog Entry

updated regex in run-compose.sh

### Added

`./run-compose --enable-gpu[count=all]` now supported


tjbck:
LGTM, Thanks!

PR #1821, opened 2024-04-28
🎖️Added translations for Playground and Archived Chats

Yanyutin753:
## Description

🎖️Added translations for Playground and Archived Chats
---

### Additional Information

only added translations for Playground and Archived Chats by using use [google](https://translate.google.com)


tjbck:
Thanks!

PR #1829, opened 2024-04-28
fix: fluid streaming was "pausing" when tab was not visible

cheahjs:
## Pull Request Checklist

---

## Description

Check if the the page is visible (https://developer.mozilla.org/docs/Web/API/Page_Visibility_API) before sleeping. When the page is not visible, Javascript timers are throttled to once a second, so if the fluid streaming option was turned on, it would barely make any progress when not visible.

---

### Fixed

- The fluid streaming response option will better handle when the page is not visible.


tjbck:
LGTM!

Issue #1837, opened 2024-04-28
feat: Increase context length setting for new models


notasquid1938:
Now that models such as this one https://ollama.com/library/dolphin-llama3:256k have created contexts lengths up to 256k, can the context length slider be adjusted to not max out at 16k? Also would it be possible to make context length settable by model? Is there a downside to just using the highest context value even for models that don't support it?

NiLon:
> Is there a downside to just using the highest context value even for models that don't support it?

Short answer would be that yes, it's bad to use bigger context value than the model is trained for. The model starts to generate nonsense. There are some techniques to reach longer context windows by playing with the index positions but these are generally things for the model to do rather than setting it to unsupported values. Like for example the model you are speaking of likely already has these tricks.

notasquid1938:
Went ahead and made a pr with this change #1843 just waiting on approval

notasquid1938:
Looks like another model just dropped:
https://huggingface.co/gradientai/Llama-3-8B-Instruct-Gradient-1048k
The context limit needs to be increased to 1048576 (Or maybe add an option for the user to set it as high as they want to avoid constantly increasing it?)

NiLon:
This setting might be fundamentally flawed to start with. It is inherently model-specific. Setting it too high consumes more memory and makes the model output nonsense because it wasn't trained for it. Having such global value might not be at all correct thing to do.

https://github.com/ggerganov/llama.cpp/discussions/3111

notasquid1938:
Maybe adding all the parameters as an option for modelfiles instead of a global setting could serve as a case by case solution? The only issue I see is not being able to adjust non-modelfile models.

GrayXu:
So when using an external API, these settings would not take effect?

max5800:
> So when using an external API, these settings would not take effect?

do you got an answer for that?
Is there an option to set context windows per model which are connected via api?

PR #1843, opened 2024-04-29
Increase-Context-Length

notasquid1938:
Changed max context length slider to go up to 256K now that certain models support that.

I originally posted this as a feature request #1837 and then realized it's just changing two numbers.

### DIFF
--- a/src/lib/components/chat/Settings/Advanced/AdvancedParams.svelte
+++ b/src/lib/components/chat/Settings/Advanced/AdvancedParams.svelte
@@ -493,7 +493,7 @@
 						id="steps-range"
 						type="range"
 						min="1"
-						max="16000"
+						max="256000"
 						step="1"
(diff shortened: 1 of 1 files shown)


tjbck:
Closing this PR but updated the token length limit to 1024000!

Silentoplayz:
> Closing this PR but updated the token length limit to 1024000!

Go big or go home! 🤣

notasquid1938:
> Closing this PR but updated the token length limit to 1024000!

Looks like another model just dropped:
 https://huggingface.co/gradientai/Llama-3-8B-Instruct-Gradient-1048k
 The context limit needs to be increased to 1048576 (Or maybe add an option for the user to set it as high as they want to avoid constantly increasing it?)

tjbck:
Removed the limit for the number input field!

PR #1858, opened 2024-04-29
fix: various rag api calls and ui cleanup

buroa:
1. Cleaned up the query params to be on the same row (if hybrid search is set)
2. Fixed the api calls to `rag/api/v1/query/collection` which were broken
3. Hybrid search without a reranker uses embeddings with a cosine similarity, so their score is always 0.0 - 1.0; just like a reranking model. So always reverse the filter.

Screenshot 2024-04-29 at 12 18 24 PM


tjbck:
LGTM, Thanks!

PR #1861, opened 2024-04-29
fix: tailwind rounded class for message save buttons

jannikstdl:


tjbck:
Closing as it's already fixed :)

PR #1963, opened 2024-05-04
add arabic language

Hamad3bdulla:


src/lib/i18n/locales/languages.json
justinh-rahb, line 1: Should be below the default option.

justinh-rahb:
Thanks for this additional language PR! In keeping with convention, we organize the list alphabetically by the code, with the exception of the default `en-US` being at the top.

src/lib/i18n/locales/languages.json
bo3bdo: i add it below en-US ? 
I see English is the default

src/lib/i18n/locales/languages.json
```diff
@@ -1,4 +1,8 @@
 [
+	{
+		"code": "ar-BH",
+		"title": "العربية"
```
justinh-rahb, line 4:
```json
	{
		"code": "ar-BH",
		"title": "العربية (Arabic)"
```

bo3bdo:
oh ok i update and add new one to dev

PR #1977, opened 2024-05-05
✨fix an issue where images downloaded in safari browser on iphone are text files

Yanyutin753:
## Pull Request Checklist

---

## Description

✨fix an issue where images downloaded in safari browser on iphone are text files

---

### Added

- Not specifically cited Mime-types, i created a MimeTypes to get MimeType from file name suffixes. 

### Fixed

- fix an issue where images downloaded in safari browser on iphone are text files

---

### Additional Information

- Through the self-test, now i can download the corresponding file

### DIFF
--- a/backend/apps/images/main.py
+++ b/backend/apps/images/main.py
@@ -323,9 +323,11 @@ def save_b64_image(b64_str):
             header, encoded = b64_str.split(",", 1)
             mime_type = header.split(";")[0].split(":")[1]
             img_data = base64.b64decode(encoded)
-            image_format = ('.webp' if mime_type == 'image/webp' 
-                            else mimetypes.guess_extension(mime_type) 
-                            or '.png')
+            image_format = (
+                ".webp"
+                if mime_type == "image/webp"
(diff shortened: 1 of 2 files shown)


Yanyutin753:
sorry, i seem to have submitted it to `dev`. I don't think i understand what that means!

Yanyutin753:
> Thanks for the PR! This should be fixed on dev, added you as a co-author as well. Let me know if the issue persists!

thanks!

Yanyutin753:
i find the bug in the `mimetypes.guess_extension(mime_type) ` 
sometime, if the type of my image is `image/webp` , this code will work incorrectly !
it will return none

so that i change the code like this
```
image_format = (
    ".webp"
    if mime_type == "image/webp"
    else mimetypes.guess_extension(mime_type) or ".png"
)
```
The second bug is that when I download an image using ios safari, it doesn't recognize my image type, so it saves it as a text file and can't be saved properly. I'm not sure if this is a situation that everyone uses

PR #2003, opened 2024-05-06
Restore password via email

juliojesusvizcaino:
## Pull Request Checklist

---

## Description

Allow users to restore their password via email.
---

### Added

- Allow users to restore their password via email. 

---

### Additional Information

- Show a link to the users "Forgot your password?". Write the email and click on "Send Reset Email". It makes a POST request to `/api/v1/auths/request-password-reset`
- The endpoint returns, and a background task runs for email sending.
  - It checks if the user exists.
  - It creates a JWT identifying the user. It includes the start of the hash (sha256) of the current password, so it can't be used twice to change the user password.
  - It renders a basic HTML template with a link to the `frontend/auth/reset-password?token=<the jwt>`.
  - It sends the email.
- The user receives the email and clicks on the `Reset Password` link.
- They are redirected to `frontend/auth/reset-password?token=<the jwt>`.
- They enter the new password and `Reset Password`. It makes a POST request to `/api/v1/auths/reset-password`.
- In the endpoint, the token is checked and the password is changed.
- The user is redirected to the login page.

For this to work, configure some environment variables:

```sh
ENABLE_MAIL="True"
MAIL_USERNAME="user@example.com"
MAIL_PASSWORD=""
MAIL_FROM="user@example.com"
MAIL_PORT="587"
MAIL_SERVER="smtp.gmail.com"
MAIL_FROM_NAME="Desired Name"
MAIL_STARTTLS="True"
MAIL_SSL_TLS="False"

FRONTEND_URL="http://localhost:5173"
```


justinh-rahb:
Thanks for this PR! I think this would be useful on its own, but it may be even better to make the email sending code more generalized so that we can use it for other purposes, such as account invitations or other types of transactional mail.

juliojesusvizcaino:
@justinh-rahb I just refactored it a bit, so the mail config is accessible from the `config.py` file. This way, we could send emails just by using it:

```py
    message = MessageSchema(
        subject="Password Reset Request",
        recipients=[user.email],
        body=content,
        subtype=MessageType.html,
    )

    fm = FastMail(MAIL_CONF)
    await fm.send_message(message)
```

We could also create some wrapper, or maybe define `FastMail` as a dependency to inject in fastapi:

```py
class MailSender:
    def __init__(self, background_task: BackgroundTasks) -> None:
        if config.MAIL_CONFIG is None:
            raise Exception("No Mail Configured")

        self.fm = FastMail(config.MAIL_CONFIG)
        self.background_task = background_task

    def __call__(self, message: MessageSchema, template_name: str | None = None):
        self.background_task.add_task(self.fm.send_message, message, template_name)

# later in the app
@app.route('...')
async def do_things(mail_sender: Annotated[MailSender, Depends()]):
    message = MessageSchema(
        subject="Password Reset Request",
        recipients=[user.email],
        body=content,
        subtype=MessageType.html,
    )
    mail_sender(message)
    return {'message': 'whatever'}
```

I like how it is now because of its versatility, what do you think?

justinh-rahb:
> I like how it is now because of its versatility, what do you think?

Agreed, this is much more flexible now and will provide a useful hook we can use for other similar purposes. Great work!

juliojesusvizcaino:
Added some documentation: https://github.com/open-webui/docs/pull/65

juliojesusvizcaino:
I'm unsure how to test this with Cypress, should I mock the backend calls?

justinh-rahb:
> I'm unsure how to test this with Cypress, should I mock the backend calls?

Cypress test coverage is far from 100% right now. I'd be comfortable merging this for the time being and waiting for a separate PR for test coverage later. And merging now will mean someone else might take that on even, saving you the work.

juliojesusvizcaino:
That sounds nice! This is the first time I create a pull request in a public project, do you know what would be the next step for this to be merged?

backend/config.py
```diff
@@ -9,6 +9,7 @@
 
 from pathlib import Path
 import json
+from fastapi_mail import ConnectionConfig
```
tjbck, line 12: Native implementation will be preferred here.

.env.example
```diff
@@ -14,4 +14,16 @@ ANONYMIZED_TELEMETRY=false
 
 # Use locally bundled version of the LiteLLM cost map json
 # to avoid repetitive startup connections
-LITELLM_LOCAL_MODEL_COST_MAP="True"
\ No newline at end of file
+LITELLM_LOCAL_MODEL_COST_MAP="True"
+
+ENABLE_MAIL="True"
```
tjbck, line 19: `ENABLE_EMAIL`

src/routes/auth/reset-password/+page.svelte
tjbck, line 1: This should be included as a component from the same `auth` route.

backend/apps/web/routers/auths.py
```diff
@@ -373,3 +371,50 @@ async def get_api_key(user=Depends(get_current_user)):
...
+
+if MAIL_CONFIG and FRONTEND_URL:
+
+    class ForgetPassword(BaseModel):
+        email: EmailStr
+
+    class ForgetPasswordResponse(BaseModel):
+        message: str
+
+    @router.post("/request-password-reset", response_model=ForgetPasswordResponse)
```
tjbck, line 388: `/password-reset/request`

backend/apps/web/routers/auths.py
```diff
@@ -373,3 +371,50 @@ async def get_api_key(user=Depends(get_current_user)):
...
+        }
+
+    class ResetPassword(BaseModel):
+        token: str
+        new_password: str
+
+    class ResetPasswordResponse(BaseModel):
+        message: str
+
+    @router.post("/reset-password", response_model=ResetPasswordResponse)
```
tjbck, line 409: `/password-reset/request/verify`

tjbck:
The rest LGTM, Thanks for this PR!

.env.example
tjbck: all occurrence for `MAIL` should be renamed to `EMAIL`

src/routes/auth/reset-password/+page.svelte
juliojesusvizcaino: We need a way to make the reset password component shareable, so the link in the emails opens the "new password" flow.

If we use the same component, we need a way to open that component, via a query param? `.../auth?reset-password=true`

backend/config.py
juliojesusvizcaino: Ok, I can use `aiosmtplib` instead, the same used by `fastapi_mail`. I installed the `fastapi_mail` library because we would end up with a simpler but similar implementation that we would need to maintain.

We can use the `smtplib` module in the standard library, or install `aiosmtplib` to make it asynchronous. If we use the second one, we end up with an additional dependency anyway.

tjbck: Let's use `smtplib` for now and we can come back to this later.

.env.example
```diff
@@ -14,4 +14,16 @@ ANONYMIZED_TELEMETRY=false
...
+EMAIL_USERNAME="user@example.com"
+EMAIL_PASSWORD=""
+EMAIL_FROM="user@example.com"
+EMAIL_PORT="587"
+EMAIL_SERVER="smtp.gmail.com"
+EMAIL_FROM_NAME="Desired Name"
+EMAIL_STARTTLS="True"
+EMAIL_SSL_TLS="False"
+
+FRONTEND_URL="http://localhost:5173"
```
tjbck, line 29: `WEBUI_URL` env var already exists!

tjbck:
Closing due to conflicts, I strongly encourage you to open a new PR! 🙌

PR #2032, opened 2024-05-07
Added new translation strings + more missing i18n keys

jannikstdl:
- Added german translation to the newly added keys
- Added one more unused i18n key i found (Help Tooltip)


tjbck:
LGTM, Thanks!

PR #2033, opened 2024-05-07
Update chat timestamp locale

jannikstdl:
- Added localisation support for the new chat timestamp.

Note: Since this is a i18n variable key, i added the possible keys as comment, so the are recognizes by the parser and don't get automatically removed


tjbck:
Thanks!

PR #2039, opened 2024-05-07
chore: Small Python requirement fixes

akx:
## Description

This PR fixes the `backend/requirements.txt` file a bit:

* `youtube-transcript-api` wasn't pinned like the rest of the versions are; it is now.
* some requirements with `[extras]` were duplicated; they don't need to be.
* `uuid` has been a standard library module for 18 years as far as I can see; there should be no need to install a version from 2006 from PyPI.

---

### Fixed

- Backend requirement versions slightly corrected.


tjbck:
Good catch, Thanks!

PR #2040, opened 2024-05-07
fix: do not use hardware ID in document ID generation

akx:
## Description

[`uuid.uuid1()`](https://docs.python.org/3.12/library/uuid.html#index-5) uses the machine's hardware ID (among other things) for UUID generation. I see no reason to use partially deterministic IDs here; every other use of UUIDs in the project seems to be using the random `uuid4()` function.

---

### Fixed

- Vector database document IDs no longer use deterministic UUIDv1 identifiers.


tjbck:
LGTM, Thanks!

PR #368, opened 2024-01-03
feat: ✨ bun support, backend lint, frontend & backend CI

ThatOneCalculator:
- Adds bun support
- Changes default commands to use bun
- Adds formatting and linting for backend
- Moves frontend CI to Bun
- Adds backend CI

PR #2041, opened 2024-05-07
chore/fix: re-enable linting for backend

akx:
## Description

This PR re-enables linting for the backend (disabled 4 months ago in https://github.com/open-webui/open-webui/pull/368 because "too many issues") using [Ruff](https://github.com/astral-sh/ruff), and fixes the issues it finds. Each commit should be fine to review in isolation, if this PR otherwise feels too large. Most of the cumulative diff is reordering imports.

I deliberately disabled some rules for the time being, though, with TODO notes that they really ought to be fixed; bare `try:except:`s for one are a bad habit, since they catch e.g. `KeyboardInterrupt`s, so chances are the backend will ignore attempts of stopping it if the signal lands in a suitable place 😅 
Conversely, more rules could be added later.

It also switches formatting from `black` to `ruff` (they're 99.x% compatible, but `ruff` is a lot faster and formats some things in a neater way). The formatting actually found a couple of stray commas that inadvertently turned some logging statements into no-op 1-tuple constructions... 😄 

For CI, I made sure that the pinned versions from `requirements.txt` are used for Ruff – before this, CI could install a different version of Black than what `requirements.txt` said.

## Testing

Since there is no test suite, all I really could do is test the app runs fine with these changes.

I ran under `coverage` (`coverage run -m uvicorn main:app ...`), which gave me a coverage of about 49%.

## Dependencies

`black` was dropped in favor of `ruff`; technically, Ruff should be a development-time dependency only, but there is no `requirements-dev.txt` or similar and Black was in the runtime deps, so I didn't change that.

---

### Changed

- Python code is now linted and formatted using Ruff.

---

### DIFF
--- a/.github/workflows/format-backend.yaml
+++ b/.github/workflows/format-backend.yaml
@@ -28,12 +28,12 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install black
+        run: grep '^ruff' backend/requirements.txt | xargs python -m pip install
 
       - name: Format backend
         run: npm run format:backend
+        env:
+          RUFF_OUTPUT_FORMAT: github
 
       - name: Check for changes after format
         run: git diff --exit-code
--- a/.github/workflows/lint-backend.yml
+++ b/.github/workflows/lint-backend.yml
@@ -20,8 +20,8 @@ jobs:
       - name: Use Bun
         uses: oven-sh/setup-bun@v1
       - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install pylint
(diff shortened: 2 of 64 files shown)


akx:
@cheahjs Please let me know if there's any interest in getting this merged (I see there's adjacent work in https://github.com/open-webui/open-webui/pull/2182) so I know whether or not to keep rebasing this.

akx:
Rebased on `dev`. cc @cheahjs @tjbck

akx:
@tjbck Rebased once more on `dev`.

justinh-rahb:
@tjbck have we reached a decision on this?

thlehmann-ionos:
@tjbck can you please comment on the state of linting in Open WebUI?

PR #2055, opened 2024-05-07
fix: admin dashboard styles refactor

albertso:
## Pull Request Checklist

---

## Description

open-webui--refactor-admin-styles01a

I have rewritten the Admin page to improve code readability and mantainability, and to facilitate inclusion of more tabs in the future. Visually and functionally the page remains the same, except for some style fixes. In the video I demonstrate how the row action buttons are now visible where they weren't, and also how the ui responds the same as before, even when the markup has changed.

---

### Fixed

- Fixed scrollbar not showing and not being able to select actions at the end of the row.
- Fixed user actions hidden from admin, previously line 270 `{#if user.role !== 'admin'}` is reversed.
- Fixed the way how some tailwind classes were being applied.

### Changed

- Readability improved, line count reduced
- Prepared markup to accept for more tabs.
- Localized all strings
- Moved all styles to the style tag at the bottom of the file

---

### Additional Information

The main improvement here relies on the use of tailwind's `@apply` directive, which allows to apply tailwind styles in css:

```css
section { @apply flex flex-col }
```

There are also improvements in how some conditional stylings were being applied, where instead of swapping a bunch of tailwind classes in code, now the code just changes a single class that controls the styles, which are separated from the markup.

```html
<button
  class="flex items-center gap-2 text-xs px-3 py-0.5 rounded-lg {user.role ===
    'admin' &&
    'text-sky-600 dark:text-sky-200 bg-sky-200/30'} {user.role ===
    'user' && // ...">
```
One way to do it, seems would be this:
```html
<button class="is-{user.role}" />
<style> button:global(.is-pending){ @apply /* ... */ } </style>
```

But this doesn't work entirely, because doing it this way the classes are not known to the svelte compiler. We can use the `:global()` directive, but the tailwind processor cannot create the `dark:` classes for the dark theme. The only way this seems to work is to inform svelte of all the classes, like this:

```html
<button
  class:is-user={user.role=='user'}
  class:is-admin={user.role=='admin'}
  class:is-pending={user.role=='pending'} />
<style> button.is-pending{ @apply text-gray-600 dark:text-gray200 } </style>
```

---
I modified this file because I'm attempting to implement some kind of plugins/extensions functionality to open-webui, even if it's only to see how far can I get. I was creating a page of its own in the `/plugins` route, but then realized this should be part of the admin page.

### DIFF
--- a/src/routes/(app)/admin/+page.svelte
+++ b/src/routes/(app)/admin/+page.svelte
@@ -3,8 +3,8 @@
 	import { WEBUI_NAME, config, user } from '$lib/stores';
 	import { goto } from '$app/navigation';
 	import { onMount, getContext } from 'svelte';
-  import type { Writable } from 'svelte/store';
-  import type { i18n as i18nType } from 'i18next'
+  	import type { Writable } from 'svelte/store';
+  	import type { i18n as i18nType } from 'i18next'
 
 	import dayjs from 'dayjs';
 	import relativeTime from 'dayjs/plugin/relativeTime';
@@ -33,8 +33,8 @@
 	let showUserChatsModal = false;
 	let showEditUserModal = false;
 	
-  enum View { Overview, Plugins };
-  let view:any = View.Overview;
+  	enum View { Overview, Plugins };
+  	let view:any = View.Overview;
 
 	onMount(async ()=>{
 		if ($user?.role!=='admin') await goto('/');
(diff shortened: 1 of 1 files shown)


tjbck:
Thanks for the PR! I'll cherrypick some of the changes you made and add you as a co-author! For your future PRs, I know there isn't any styling guides at the moment, but if you could follow existing conventions that would be great. Notably, using `@apply` directive isn't something I'm keen on using. Regardless, I appreciate the effort!

albertso:
Alright

PR #2079, opened 2024-05-08
feat: add Finnish localization 🇫🇮

akx:
## Description

This PR adds a Finnish localization.

The only term I couldn't think of a great localization for is `token`, but that can be fixed later.

---

### Added

- Added a Finnish localization.

---

### Additional Information

- I'm a native Finnish speaker.


tjbck:
Thanks!

PR #2182, opened 2024-05-11
chore: address eslint errors

cheahjs:
## Pull Request Checklist

---

## Description

1. Address all eslint errors that are flagged when running `npm run lint:frontend`
2. Add types for `i18next` usage across the codebase
3. Add `eslint-plugin-unused-imports` to remove unused imports


src/lib/apis/audio/index.ts
```diff
@@ -12,19 +19,20 @@ export const getAudioConfig = async (token: string) => {
...
+			error = err.detail ?? err;
 			return null;
 		});
 
 	if (error) {
 		throw error;
 	}
 
-	return res;
+	return res!;
```
akx, line 35: Having to add `!`s like this smells like a bug...

akx: TBH the then/catch/err/throw/... async handling here seems like it can be replaced with simple `async`/`await`s unless I'm missing something. See https://github.com/open-webui/open-webui/pull/2227

akx: (That said, I'll have to reopen #2227 as a new PR anyway since you can't push to a closed PR, so that doesn't really affect this PR 😅. However, the type-related comments from that PR might apply to this one too.)

akx: See: #2233. It might be a good idea to start building types for the API on top of that, if it gets eventually merged?

PR #2227, opened 2024-05-13
fix/chore: (wip) DRY API error handling

akx:
## Description

This starts to clean up the copy-pasted frontend API access code.

---

### Changelog Entry

TBD

---

### Additional Information

- See https://github.com/open-webui/open-webui/pull/2182#discussion_r1598129838

### DIFF
--- a/src/lib/apis/audio/index.ts
+++ b/src/lib/apis/audio/index.ts
@@ -1,30 +1,8 @@
 import { AUDIO_API_BASE_URL } from '$lib/constants';
+import { doFormRequest, doGetRequest, doJSONRequest } from '$lib/apis/helpers';
 
 export const getAudioConfig = async (token: string) => {
-	let error = null;
-
-	const res = await fetch(`${AUDIO_API_BASE_URL}/config`, {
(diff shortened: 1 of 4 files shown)


tjbck:
Thanks for the PR but not a huge fan for this approach, I'll be closing this. I appreciate the effort!

akx:
@tjbck Can you please elaborate why? Right now there's tens of duplicates of the same error-handling code with slight differences (some do `detail`, some don't, etc.). This ties into #2182, where the current code needs explicit `!`s added to opt out of type safety.

tjbck:
I'm sure you have your preferences with how js/ts projects should be implemented, and I do think adding type to a certain extent is helpful and important, but I also personally don't think having 100% type safety isn't the most important part and I don't want to change my workflow and how I code just to have 100% coverage. Thanks for your understanding!

akx:
@tjbck Thanks for the clarification – that's fine! I can trivially add a default (`T = any`) to the `<T>` type constraints, so you don't need to change your workflow there.

PR #2291, opened 2024-05-15
Skip email validation for header-based auth

sd109:
## Pull Request Checklist

- [X] **Target branch:** Pull requests should target the `dev` branch.
- [X] **Testing:** Have you written and run sufficient tests for the changes?
- [X] **Code Review:** Have you self-reviewed your code and addressed any coding standard issues?

---

## Description

When authenticating via HTTP headers, I think email format validation should be skipped. Using header-based auth implies that user credentials have been validated elsewhere by the application passing the headers and I would also like to be able to provide arbitrary usernames via headers rather than being restricted to email addresses.

---

### Added

- Ability to provide non-email address usernames when using header-based authentication.

### DIFF
--- a/backend/apps/web/routers/auths.py
+++ b/backend/apps/web/routers/auths.py
@@ -168,7 +168,9 @@ async def signup(request: Request, form_data: SignupForm):
             status.HTTP_403_FORBIDDEN, detail=ERROR_MESSAGES.ACCESS_PROHIBITED
         )
 
-    if not validate_email_format(form_data.email.lower()):
+    if not WEBUI_AUTH_TRUSTED_EMAIL_HEADER and not validate_email_format(
+        form_data.email.lower()
+    ):
         raise HTTPException(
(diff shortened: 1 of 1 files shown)


tjbck:
Thanks for the PR! but closing as this will break #2003 if we allow non email usernames. I might come back to this later but for now this won't be supported.

sd109:
@tjbck Isn't this orthogonal to resetting passwords via email? Or are you concerned with use cases where some users authenticate with proxy headers and others sign up for local accounts with an email? If a user authenticates via HTTP headers then they will never even see the sign up page.

Based on #2003, could we just change the condition in this PR to be
```
if (EMAIL_ENABLED or not WEBUI_AUTH_TRUSTED_EMAIL_HEADER) and not validate_email_format(...))
```
to make the two changes compatible?

This looks like a very promising project and I'd love to make use of it but my use case is a non-starter without the ability to completely delegate auth to the proxy headers for all users and bypass all email-related features.

tjbck:
we might want to introduce different env var `VALIDATE_EMAIL` here then, defaulting to `True`.

PR #1667, opened 2024-04-22
Add azure openai

algorithmlover2016:
## Pull Request Checklist

- [ ] **Description:** Briefly describe the changes in this pull request.
- [ ] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ ] **Documentation:** Have you updated relevant documentation?
- [ ] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?

---

## Description

[Insert a brief description of the changes made in this pull request]

---

### Added

- [List any new features or additions]

### Fixed

- [List any fixes or corrections]

### Changed

- [List any changes or updates]

### Removed

- [List any removed features or files]

PR #2308, opened 2024-05-16
Adds support for Azure OpenAI

frandominguezl:
## Pull Request Checklist

---

## Description

Adds native support for Azure OpenAI, in a similar way that it's implemented for OpenAI, rather than relying in LiteLLM.

---

### Added

- Support for Azure OpenAI


justinh-rahb:
Builds on Mac, but I haven't got an Azure API to test with.

jannikstdl:
This should be already possible with litellm, isn't it?

tjbck:
Ideally we'd want to integrate azure support to current existing openai codebase. I appreciate the effort!

Related #1667

rcarmo:
Can we get a resolution here? #1667 is also closed, and this would be an extremely useful feature.

justinh-rahb:
> Can we get a resolution here? #1667 is also closed, and this would be an extremely useful feature.

Closed for same reason: it was implemented as an entirely separate app within the backend when what Tim desires is for the existing openai app to be extended to support azure endpoints instead.

frandominguezl:
Sorry for coming back to this, but it would be good to implement this by default. Could you please let me know how do you want to do this using the existing OpenAI codebase so I can work on it?

Thanks in advance @tjbck

PR #2316, opened 2024-05-16
modifying build to run application as non-root user

codeInTheShell:
## Pull Request Checklist

---

## Description

The default build runs as root and might cause issues in restrictive work environments. This PR changes that open-webui runs with a user called openwebui with the default user and group id 1000, that also owns all files under /app

---

### Added

- [application user]

### Fixed

- [n/a]

### Changed

- [pip libraries are installed in virtual user environment that is owned by the application user and not root]

### Removed

- [removed all references to root]

### Security

- [see description]

### Breaking Changes

- [n/a]

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,8 +27,8 @@ RUN npm run build
 FROM python:3.11-slim-bookworm as base
 
 # Use args
-ARG USE_CUDA
 ARG USE_OLLAMA
+ARG USE_CUDA
 ARG USE_CUDA_VER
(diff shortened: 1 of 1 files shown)


cheahjs (commented):
This is a breaking change for all existing users as `/app/backend/data` is chmod 0755 by default and owned by root:root, any user that upgrades to this image will break.

Given the number of support requests we get from users not understanding Docker, I would not feel comfortable shipping a breaking change.

PR #2322, opened 2024-05-17
feat: optional UID/GID docker build args

cheahjs:
## Pull Request Checklist

---

## Description

Adds two new build args  `UID` and `GID` to set the user and group ID of the files in container and to run as that user. This is added with a note that non-root configurations are untested.

To build as non-root: `docker build --build-arg UID=1000 --build-arg GID=1000 --tag=open-webui:nonroot .`

This is a no-op change in the default root configuration, which is why `$HOME` is still set to `/root`. We can't switch to nonroot by default, as this breaks existing users (and we have enough support requests around Docker usage as it is already, and we encourage users to setup Watchtower for automatic updates). We also don't build nonroot containers in CI because we're already running into total concurrency quotas on GitHub Actions. Adding a nonroot variant to the build matrix increases the number of tags from 3 to 6, and the total number of images from 6 to 12.

Test docker build: https://github.com/cheahjs/open-webui-fork/actions/runs/9121520626

Addresses https://github.com/open-webui/open-webui/discussions/1698 and https://github.com/open-webui/open-webui/discussions/1021

Prior attempts: https://github.com/open-webui/open-webui/pull/1533 https://github.com/open-webui/open-webui/pull/2316

Tested locally with base, bundled ollama, and CUDA.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,9 @@ ARG USE_CUDA_VER=cu121
 # IMPORTANT: If you change the embedding model (sentence-transformers/all-MiniLM-L6-v2) and vice versa, you aren't able to use RAG Chat with your previous documents loaded in the WebUI! You need to re-embed them.
 ARG USE_EMBEDDING_MODEL=sentence-transformers/all-MiniLM-L6-v2
 ARG USE_RERANKING_MODEL=""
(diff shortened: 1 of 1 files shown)


tjbck:
LGTM!

PR #2325, opened 2024-05-17
feat: python code execution

### LINKED ISSUE #1321: feat: code block execution
Executing code allow the tool to be used as an ideation and copilot in software development for brainstorming purposes. It also allows user to instead of asking question to LLM and hopping for the correct answer, to actually create code and run it to get correct answer. 

Eventually having something like code interpreter as shared in #851 will be great but I think that may require much more changes to get it right e.g. some form of function calling pattern that works across multiple models to pick the code interpreter and some form of chaining to respond based on those results.

This can be a good stepping stone to eventually build that.

Have a 'Run Code' in the Code block which sends the code over to a new API in backend e.g. `/code/api/v1/run`, which executes that code in a somewhat secure manner and responds back with response. To start with we can support only Python code.

**Additional context**

Here is very basic prototype I whipped up: https://github.com/zabirauf/open-webui/pull/1/files
Want to get thoughts before I work on it further.

tjbck:
Somewhat resolves #1321

Running entirely on the browser side and external libraries are not supported yet.


santosh0787:
can you provide documentation on how to execute the code?

Nordln:
For anyone else also looking for some docs on this feature, reading up on https://pyodide.org/en/stable/usage/wasm-constraints.html has been handy in getting a handle on it. Also, editing any prompt reply with the following lets you run a call to pyodide python's built-in help
```
```python
help(); 
```
A nice feature. Love to see how it evolves.

carlo088:
"external libraries are not supported yet"

Any updates on the possibility of integrating external libraries when execution python snippets? Thanks

PR #2352, opened 2024-05-17
Standardize fetch API

elliottlawrence:
## Pull Request Checklist

---

## Description

There are quite a few places in the frontend code that use a similar pattern of calling `fetch()` and then dealing with error handling. This should probably be standardized with a helper function.

---

### Changed

- Refactored `fetch` calls to have standardized error handling

### Breaking Changes

---

### Additional Information

I tried to be somewhat cautious about changing existing callsites of `fetch`. Most of them follow the same pattern but some are slightly different and I left those as-is. I figured this would be a good start and the rest can be cleaned up later.

I did a sanity check that the UI still seems to work, but there are quite a few changes here so I haven't tested every API call.

### DIFF
--- a/src/lib/apis/utils/index.ts
+++ b/src/lib/apis/utils/index.ts
@@ -14,11 +14,7 @@ export const fetchApi = async (
 		})
 		.catch((err) => {
 			console.log(err);
-			if ('detail' in err) {
-				error = err.detail;
-			} else {
-				error = 'Server connection failed';
-			}
+			error = err;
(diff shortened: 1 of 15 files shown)


src/lib/apis/utils/index.ts
elliottlawrence, line 1: This is where I added the new helper function.

elliottlawrence:
@tjbck - Any reason this PR was closed?

PR #2377, opened 2024-05-18
fix of 'show username' feat now working when chat bubbles active

jannikstdl:
### Fixed

- Fixes "show username" feat not working when "Chat Bubbles UI" is aktive. User now gets notified by a toast and "show username" will be deactivated automatically when both active. 
- Added one missing i18n key to a toast in the same file
- Fixed missing padding for the UserMessage. I noticed that when I deactivated Chat Bubbles UI the space between the ProfileImage and the Username was missing


tjbck:
Cherry-picked changes :), Thanks!

PR #2384, opened 2024-05-19
Update Serbian translation

kostich:
## Description

I've used the `main` branch as the base initially so I've missed some strings in the `dev` branch. Rebased on `dev` and updated the translation.

---

### Changed

- Updated the Serbian translation


tjbck:
Thanks!

PR #2416, opened 2024-05-20
⭐Organizes almost all languages and completes those without translations

Yanyutin753:
## Pull Request Checklist

- [x] **Target branch:** Pull requests should target the `dev` branch.
- [x] **Description:** Briefly describe the changes in this pull request.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ ] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [ ] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [x] **Testing:** Have you written and run sufficient tests for the changes?
- [ ] **Code Review:** Have you self-reviewed your code and addressed any coding standard issues?

---

## Description

⭐Organizes almost all languages and completes those without translations

PR #2454, opened 2024-05-21
Fixed Arabic Translation

AlyMobarak:
# Pull Request Checklist

- **Label:** fix

# Changelog Entry

### Description

- As an Arabic user, I've noticed a couple of Arabic mistakes scattered all over. Also, a couple of translations weren't even done in Arabic. So I tried my best as an Arabic native to fix them all.

### Added

- Added new Arabic translations.

### Fixed

- Corrected several Arabic translations by fixing grammar or word choice.


Yanyutin753:
In fact, this branch can be integrated into #2416, because in fact, many languages are not very perfect, I am a Chinese, some other languages use translation, translation is not necessarily accurate, but also we have to pr together, so that the page is better for family and friends to share

Yanyutin753:
#2416 @AlyMobarak I have merged your branch together, now we are co-authors, let us pr together, OK?

AlyMobarak:
Alright sounds good ig

AlyMobarak:
Anytime!

PR #2576, opened 2024-05-26
Update languages.json - Addede Danish

alex404hm:
I have added the danish languest


tjbck:
Looks like you forgot to include the actual translation.json file.

PR #2597, opened 2024-05-27
Fix errors and improve the Vietnamese translation

que-nguyen:
Fix errors and improve the Vietnamese translation


tjbck:
Thanks!

PR #2004, opened 2024-05-06
feat: add user toggleable web search

cheahjs:
## Pull Request Checklist

- [x] **Description:** Briefly describe the changes in this pull request.
- [ ] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ ] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [ ] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [x] **Testing:** Have you written and run sufficient tests for the changes?
- [ ] **Code Review:** Have you self-reviewed your code and addressed any coding standard issues?

---

## Description

Adds the ability to perform web searches via the RAG API (`rag/api/v1/websearch`) using the following search providers:

1. [SearXNG](https://searx.space/), configured with `SEARXNG_QUERY_URL`, eg `SEARXNG_QUERY_URL=https://search.projectsegfau.lt/search?q=<query>`
2. [Google Programmatic Search Engine](https://programmablesearchengine.google.com/about/), configured with `GOOGLE_PSE_API_KEY` and `GOOGLE_PSE_ENGINE_ID`
3. [Brave Search](https://brave.com/search/api/), configured with `BRAVE_SEARCH_API_KEY`
4. [serpstack](https://serpstack.com/) - Google proxy, configured with `SERPSTACK_API_KEY` (and an optional `SERPSTACK_HTTPS=false`, since the free tier doesn't allow for HTTPS connections)
5. [serper](https://serper.dev) - Google proxy, configured with `SERPER_API_KEY`.

Users can configure how many of the top search results to crawl with `RAG_WEB_SEARCH_RESULT_COUNT`, and `RAG_WEB_SEARCH_CONCURRENT_REQUESTS` controls the number of concurrent requests made to crawl the search results.

Users can toggle web search on or off on the UI, which causes the frontend to use a prompt to generate a search query, calling the RAG API to search for that query, and then injecting the results of that as a RAG document.

Implements #586

---

### Added

- **🔍Web Search for RAG**: You can now perform web searches when chatting using providers like SearXNG, Google PSE, Brave Search, serpstack, and serper.

PR #2599, opened 2024-05-27
feat: websearch

tjbck:
#2004


tjbck:
Thanks for all the hard work @cheahjs 🙌

Issue #2644, opened 2024-05-29
enh: Optimize the style of the ResponseMessage


jasinliu:
**Is your feature request related to a problem? Please describe.**
As title.  The space between paras is too large.

Before:
![image](https://github.com/open-webui/open-webui/assets/85552719/67ba57e2-bad8-4927-97ef-d284a17197a8)

After:
![image](https://github.com/open-webui/open-webui/assets/85552719/f6daea45-d1c6-430b-b475-8ee738c94754)

tjbck:
Help wanted! PR welcome!

tjbck:
Thanks!

PR #2645, opened 2024-05-29
enh: Update the style of ResponseMessage to change the space between paras

jiajunly:
#2644


tjbck:
More testing wanted here!

PR #2799, opened 2024-06-04
Enable use of http_proxy environment variable in openai and ollama calls

mindspawn:
# Changelog Entry

### Description

ClientSession method in aiohttp module requires trust_env=True to use http_proxy from the environment. Without this, open-webui breaks in a proxy only environment.

Related Documentation: https://docs.aiohttp.org/en/stable/client_reference.html

### Added

- Enable use of http_proxy for openai and ollama API calls

### Changed

- Add trust_env True to all ClientSession calls

### Fixed

- #2798


tjbck:
Thanks!

mindspawn:
> Thanks!

Thank you for your fantastic work :)

PR #2818, opened 2024-06-04
fix: Update PR template

Lanhild:
Simply update `label` with `prefix`.

`Label` could easily be confused with issues/PR labels from GitHub.


tjbck:
LGTM, Thanks!

PR #2866, opened 2024-06-05
Update CONTRIBUTING.md to include proper first step

bartowski1182:
Should point to opening a discussion rather than an issue


tjbck:
Thanks!

PR #2870, opened 2024-06-06
Fix proxy not being used for /api/version/updates

mindspawn:
When checking for version updates, proxy was not being used.


tjbck:
Thanks!

PR #2923, opened 2024-06-08
Support Outlook Message File Format

mindspawn:
Enable support for Outlook .msg format.
This change introduces support for the Outlook .msg format in open-webui, enabling users to directly drag and drop single emails, multiple emails, or entire email threads from Outlook into the interface, facilitating seamless interaction.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -133,8 +133,7 @@ RUN pip3 install uv && \
     fi; \
     chown -R $UID:$GID /app/backend/data/
 
-# Dependency for email message extraction
-RUN pip3 install extract_msg
+
 
 # copy embedding weight from build
 # RUN mkdir -p /root/.cache/chroma/onnx_models/all-MiniLM-L6-v2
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -133,7 +133,8 @@ RUN pip3 install uv && \
     fi; \
     chown -R $UID:$GID /app/backend/data/
 
-
+# Dependency for email message extraction
```
tjbck, line 136: Correct me if I'm wrong but I believe this part is not required.

justinh-rahb: Indeed, already in `requirements.txt`

mindspawn: Wow you guys are fast :) yes that's true. Good catch. My bad.
Will fix along with another formatting issue.

tjbck:
LGTM, Thanks!

Dockerfile
justinh-rahb: > Wow you guys are fast :)

monsters-inc-watching-you
😂

mindspawn: ha ha you guys are crazy.. you took care of it before me :) .. I thought I was going crazy as to why there were no diffs between my fork and the dev branch :)

PR #2937, opened 2024-06-08
Enable case independent environment variable testing in start.sh

mindspawn:
This matches the env var testing done in our Python code.
Hopefully fixes the issue that many folks were having with cuda and whisper not working together. At least it did for me :)


tjbck:
Thanks!

PR #3033, opened 2024-06-11
Fixed the issue where a single URL error disrupts the Web Search

que-nguyen:
To address the unresolved issue in the LangChain library where a single URL error disrupts the data loading process, the lazy_load method in the WebBaseLoader class has been modified. The enhanced method now handles exceptions appropriately, logging errors and continuing with the remaining URLs.

fixed-bug


tjbck:
LGTM Thanks!

PR #3041, opened 2024-06-11
feat: custom HOST env variable on windows

cocktailpeanut:
### Description

Currently you can customize the HOST with an environment variable in non-windows environments https://github.com/open-webui/open-webui/blob/main/backend/start.sh#L9 but this is not implemented on Windows. On windows it's using a hardcoded value of 0.0.0.0 https://github.com/open-webui/open-webui/blob/main/backend/start_windows.bat#L32 

Ironically it is Windows where we need the ability to use custom HOST because 0.0.0.0 sometimes has trouble on windows and need to use 127.0.0.1

### Added

Just replicated the custom HOST feature in start.sh into start_windows.bat


tjbck:
Thanks!

Issue #2839, opened 2024-06-05
bug: Allow Web Search RAG to Continue on Invalid Hostname Resolution


0zheermao0:
# Bug Report

## Description

**Bug Summary:**
When using the Web Search RAG feature, if the search engine returns a hostname that the server cannot resolve, all search results are blocked from proceeding to the next step of RAG processing.

**Steps to Reproduce:**
1.Use the Web Search RAG feature with a URL that points to a hostname that cannot be resolved by the DNS.
2.Observe that once the unresolvable hostname is encountered, all other valid search results are also blocked and do not proceed to the next step.

**Expected Behavior:**
If a hostname cannot be resolved, that specific URL should be ignored, and the Web Search RAG should continue processing the remaining valid URLs without interruption.

**Actual Behavior:**
The Web Search RAG stops processing all URLs and answer directly through the models when encountering an unresolvable hostname.

## Environment

- **Open WebUI Version:** v0.2.4
- **Ollama (if applicable):** 0.1.39

- **Operating System:** Windows 11, wsl2, Debian
- **Browser (if applicable):** [e.g., Chrome 100.0, Firefox 98.0]

## Reproduction Details

**Confirmation:**

- [x] I have read and followed all the instructions provided in the README.md.
- [x] I am on the latest version of both Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.

## Logs and Screenshots

**Command Console Logs:**
```
ERROR:apps.rag.main:[Errno -2] Name or service not known
Traceback (most recent call last):
  File "/mnt/d/github_project/nlp/open-webui/backend/apps/rag/main.py", line 827, in store_web_search
    loader = get_web_loader(urls)
  File "/mnt/d/github_project/nlp/open-webui/backend/apps/rag/main.py", line 694, in get_web_loader
    if not validate_url(url):
  File "/mnt/d/github_project/nlp/open-webui/backend/apps/rag/main.py", line 723, in validate_url
    return all(validate_url(u) for u in url)
  File "/mnt/d/github_project/nlp/open-webui/backend/apps/rag/main.py", line 723, in <genexpr>
    return all(validate_url(u) for u in url)
  File "/mnt/d/github_project/nlp/open-webui/backend/apps/rag/main.py", line 712, in validate_url
    ipv4_addresses, ipv6_addresses = resolve_hostname(parsed_url.hostname)
[...]

PR #3049, opened 2024-06-12
Refactor URL validation function

que-nguyen:
- The check for private IP addresses often did not yield the expected results, especially with errors like: `[Errno -2] Name or service not known`. (related  [#2839](https://github.com/open-webui/open-webui/issues/2839))
- Removed the check for private IP addresses in the URL validation process.
- Simplified the `validate_url` function to focus on validating the URL format and checking the existence of the URL using a HEAD request. Checking with HEAD requests is generally faster and more accurate as it validates whether the web resource actually exists.

### DIFF
--- a/backend/apps/rag/main.py
+++ b/backend/apps/rag/main.py
@@ -8,6 +8,7 @@ from fastapi import (
     Form,
 )
 from fastapi.middleware.cors import CORSMiddleware
+import requests
 import os, shutil, logging, re
 from datetime import datetime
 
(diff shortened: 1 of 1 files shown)


tjbck:
Seems safe but I'm not 100% certain 😂

tjbck:
One way to find out, LGTM!

justinh-rahb:
It's also the root cause of this issue:
- #3071

tjbck:
Reverting.

PR #3073, opened 2024-06-12
Set searxng language to auto and enable safesearch (moderate).

que-nguyen:
Configure searxng with language param set to auto and add "safesearch": 1 (moderate) for safer web results.


tjbck:
LGTM!

Issue #3074, opened 2024-06-12
Open WebUI Fails to Load When Accessed Over HTTPS


rbrinson:
# Bug Report

## Description

I have both Ollama and Open WebUI installed as containers in Docker on a server running Debian Linux 12. I have Open WebUI exposed to the Internet through Cloudflare Tunnels. I have a hostname setup on Cloudflare Tunnels Dashboard that provides a connection running over HTTPS with a valid SSL certificate and points to my Nginx Proxy Manager instance, which resolves the request to the server IP address and port where Open WebUI is running at my house.

This morning, I saw that there were image updates for both Ollama and Open WebUI. So, I updated my stack in Portainer. I navigated to my Cloudflare Open WebUI domain and logged in. I was greeted with the normal Changelog modal. When I tried to dismiss it, it wouldn't go away. So, I refreshed the page only to get a blank screen. Multiple refreshes did not help.

**Bug Summary:**
The Open WebUI application is failing to fully load, thus the user is presented with a blank screen.

**Steps to Reproduce:**

1. Navigate to the HTTPS url for Open WebUI v.0.3.3
2. Log in

**Expected Behavior:**
I expect to see a Changelog modal, and after dismissing the Changelog, I should be logged into Open WebUI able to begin interacting with models through Ollama.

**Actual Behavior:**

1. Try to dismiss the Changelog modal
2. Modal doesn't go away; hard refresh site
3. Blank screen

## Environment

- **Open WebUI Version:** 0.3.3
- **Ollama (if applicable):** 0.1.43

- **Operating System:** Debian Linux 12 (server)
- **Browser (if applicable):** Tested on Firefox 127.0 and Chrome 126.0.6478.55 on updated Arch Linux (client)

## Reproduction Details
This only occurs when the user is accessing Open WebUI over HTTPS. On my local network, if I navigate to the IP address and port for Open WebUI, all is fine, except that it is running over HTTP, not HTTPS. On Firefox, when I connect to the HTTPS url, there is a lock in the address bar. If I click on that lock and then click on Connection Secure, there is a button that says "Disable Protection for Now". Clicking on that allows Open WebUI to load, though there are now lines struck through the lock symbol and the https portion of the url.

The culprit seems to be [Mixed Content](https://support.mozilla.org/en-US/kb/mixed-content-blocking-firefox). This is confirmed in the Firefox console logs:
[...]

Issue #3069, opened 2024-06-12
Mixed Content w/ HTTPS


Fusseldieb:
Since one of the last updates to 3.x, OpenWebUI no longer loads. The DevTools console throws:
```
Mixed Content: The page at 'https://example.com/' was loaded over HTTPS, but requested an insecure resource 'http://example.com/api/v1/tools/'. This request has been blocked; the content must be served over HTTPS.

TypeError: Failed to fetch
    at window.fetch ...
```

The fetch function in question makes a request to an API endpoint that has a constant in it which is hardcoded as **http**:

https://github.com/open-webui/open-webui/blob/0917fa6f4ac015810211d85eb41da5d335762952/src/lib/constants.ts#L7

This causes issues, as the page is trying to load a http resource over a https page, essentially.

A potential solution would be getting the protocol to correctly prepend http or https, depending on the instance. Maybe `location.protocol`?

rbrinson:
I just posted about the same issue (https://github.com/open-webui/open-webui/issues/3074). So, me too!

PR #3075, opened 2024-06-12
fix: Mixed Content w/ HTTPS

Altair-Bueno:
Fixes Mixed Content w/ HTTPS #3069 #3074. Removes dangling constant

---

# Changelog Entry

### Fixed

- Mixed Content w/ HTTPS


tjbck:
This is for dev env.

Altair-Bueno:
> This is for dev env.

Still weird you hardcode the port and force the use of http... even if it is only for dev

PR #3104, opened 2024-06-12
fix: ollama long response timeout

TheTerrasque:
Changes:
* Added configurable timeout to `aiohttp.ClientSession` call to ollama, to work around default timeout of 5 minutes being too short in some situations 
* Set the new default timeout to 15 minutes (900 seconds)
* Described new environment variable in troubleshooting.md under title _"Error on Slow Reponses for Ollama"_


TheTerrasque:
messed up source branch, will re-create

PR #3112, opened 2024-06-13
Domain whitelisting for web search results

que-nguyen:
Note: Meta search engines like SearxNG do not support filtering results by domain natively, so this filtering is applied post-search.
Added a filter to restrict search results to specified domains.
Updated the processing of search results to apply the whitelist filter before returning the final results.
Ensured that only results from allowed domains (if specified in the whitelist) are included in the output.


programminginhackerway:
> SearxNG

Thanks, does it work well? I found that SearxNG only supports the high priority option and it can't handle the unwanted websites in results.

programminginhackerway:
> Note: Meta search engines like SearxNG do not support filtering results by domain natively, so this filtering is applied post-search. Added a filter to restrict search results to specified domains. Updated the processing of search results to apply the whitelist filter before returning the final results. Ensured that only results from allowed domains (if specified in the whitelist) are included in the output.



> > Meta search engines like SearxNG do not support filtering results by domain natively, so this filtering is applied post-search.
> 
> Thanks, does it work well? I found that SearxNG only supports the high priority option and it can't handle the unwanted websites in results.

backend/apps/rag/main.py
```diff
@@ -112,6 +112,7 @@
     YOUTUBE_LOADER_LANGUAGE,
     ENABLE_RAG_WEB_SEARCH,
     RAG_WEB_SEARCH_ENGINE,
+    RAG_WEB_SEARCH_WHITE_LIST_DOMAINS,
```
tjbck, line 115: `RAG_WEB_SEARCH_DOMAIN_FILTER_LIST`

backend/config.py
```diff
@@ -895,6 +895,15 @@ class BannerModel(BaseModel):
     os.getenv("RAG_WEB_SEARCH_ENGINE", ""),
 )
 
+RAG_WEB_SEARCH_WHITE_LIST_DOMAINS = PersistentConfig(
+    "RAG_WEB_SEARCH_WHITE_LIST_DOMAINS",
+    "rag.rag_web_search_white_list_domains",
```
tjbck, line 900: `rag.web.search.domain.filter_list`

backend/config.py
```diff
@@ -895,6 +895,15 @@ class BannerModel(BaseModel):
...
+    [
+        # "example.com",
```
tjbck, line 902: The default option should be more explicit.

backend/apps/rag/search/main.py
```diff
@@ -1,8 +1,18 @@
 from typing import Optional
-
+from urllib.parse import urlparse
 from pydantic import BaseModel
 
 
+def filter_by_whitelist(results, whitelist):
```
tjbck, line 6: `get_filtered_results` with `filter_list` param

backend/apps/rag/search/brave.py
```diff
@@ -1,15 +1,15 @@
...
-from apps.rag.search.main import SearchResult
+from apps.rag.search.main import SearchResult, filter_by_whitelist
 from config import SRC_LOG_LEVELS
 
 log = logging.getLogger(__name__)
 log.setLevel(SRC_LOG_LEVELS["RAG"])
 
 
-def search_brave(api_key: str, query: str, count: int) -> list[SearchResult]:
+def search_brave(api_key: str, query: str, whitelist:List[str], count: int) -> list[SearchResult]:
```
tjbck, line 12: `filter_list` should be optional here.

tjbck:
The rest LGTM!

tjbck:
LGTM, Thanks!

PR #3212, opened 2024-06-16
Add host.docker.internal overide for docker dev

robbie-cahill:
Just adding the `host.docker.internal` override to the docker command for the dev branch.

This got me on Linux where docker does not natively support `host.docker.internal` and so open webui could not connect with my local ollama. Adding the override fixes this.


tjbck:
Thanks!

PR #3293, opened 2024-06-19
Refine and update Vietnamese translations in translation.json

que-nguyen:
Update: Refine and update Vietnamese translations in translation.json. 
Fix: Rename 'whitelist' to 'filter_list' in searxng.py function.


tjbck:
Thanks!

PR #3351, opened 2024-06-21
Minor doc improvement

Robinsane:
Indendation made it look as if GPU support was only meant for Ollama on different server


tjbck:
PRs a

Issue #2337, opened 2024-05-17
bug: hanging connection causing blank screen


Zambito1:
# Bug Report

## Description

**Bug Summary:**
If the Open WebUI backend hangs indefinitely, the UI will show a blank screen with just the keybinding help button in the bottom right.

![image](https://github.com/open-webui/open-webui/assets/7004857/e1b08129-39a4-4323-ba36-18ae6aa5c980)

**Steps to Reproduce:**

I noticed this because I run Open WebUI on my desktop, and Ollama on another machine. If I connected to a VPN on my desktop, LAN connections from my desktop would hang indefinitely. When I tried to boot up Open WebUI, I would just see the screen above.

The easiest way to reproduce this is to run `nc -lp 11434` and try to use that as the Ollama server. You will see something like:

```
$ nc -lp 11434
GET /api/tags HTTP/1.1
Host: host.docker.internal:11434
Accept: */*
Accept-Encoding: gzip, deflate
User-Agent: Python/3.11 aiohttp/3.9.5
```

As the Open WebUI backend tries to connect to Ollama, but `nc` will never respond to the request, nor will it close the connection. You will see the above screen when you open the web interface.

I was able to work around my VPN issue by splitting my connection to only route traffic that I want to access over the VPN with the virtual interface.

**Expected Behavior:**

You can see the web interface, and ideally there will be some sort of timeout to show a connection error.

**Actual Behavior:**

You are hit with a wall of nothing.

## Environment

- **Open WebUI Version:** v0.1.124
- **Ollama (if applicable):** NA

- **Operating System:** NA
- **Browser (if applicable):** NA

## Reproduction Details

**Confirmation:**

- [x] I have read and followed all the instructions provided in the README.md.
- [x] I am on the latest version of both Open WebUI and Ollama.
- [ ] I have included the browser console logs.
- [x] I have included the Docker container logs.

## Logs and Screenshots

**Browser Console Logs:**
NA, the connection issue is on the backend. There is nothing that fails in the UI logs.

**Docker Container Logs:**
This is what happens when you load the index page:
```
INFO:     172.17.0.1:58940 - "GET /api/config HTTP/1.1" 200 OK
INFO:     172.17.0.1:58940 - "GET /api/v1/auths/ HTTP/1.1" 200 OK
INFO:     172.17.0.1:58940 - "GET /api/changelog HTTP/1.1" 200 OK
INFO:apps.ollama.main:get_all_models()
```
[...]

PR #3479, opened 2024-06-28
feat: add OLLAMA_API_TIMEOUT config variable

gartnera:
Address https://github.com/open-webui/open-webui/issues/2337 by allowing the user to manually configure a ollama request timeout. If the ollama server is is powered off, the UI will just hang until the `/ollama/api/version` request times out. My ollama server is not always running, but I still want to use the openai api.

I think this should be a configuration item because while `OLLAMA_API_TIMEOUT=.1` may work for me, others my run their ollama on remote servers.

# Changelog Entry

### Added

- Add `OLLAMA_API_TIMEOUT` config variable

### DIFF
--- a/backend/apps/ollama/main.py
+++ b/backend/apps/ollama/main.py
@@ -47,6 +47,7 @@ from config import (
     SRC_LOG_LEVELS,
     OLLAMA_BASE_URLS,
     ENABLE_OLLAMA_API,
+    OLLAMA_API_TIMEOUT,
     AIOHTTP_CLIENT_TIMEOUT,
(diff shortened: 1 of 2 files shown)


tjbck:
The recommended method will be disabling the Ollama connection entirely when not in use, Thanks!

gartnera:
Ok if anyone else comes across this, feel free to use my docker image: `ghcr.io/gartnera/open-webui:git-6f28c9b`

Just set the timeout via `-e OLLAMA_API_TIMEOUT=.1` on `docker run`

PR #3327, opened 2024-06-20
BREAKING CHANGE/sqlalchemy instead of peewee

jonathan-rohde:
# Pull Request Checklist

### Note to first-time contributors: Please open a discussion post in [Discussions](https://github.com/open-webui/open-webui/discussions) and describe your changes before submitting a pull request.

**Before submitting, make sure you've checked the following:**

- [x] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [x] **Description:** Provide a concise description of the changes made in this pull request.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ ] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [ ] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [x] **Testing:** Have you written and run sufficient tests for validating the changes?
- [x] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [x] **Prefix:** To cleary categorize this pull request, prefix the pull request title, using one of the following:
  - **BREAKING CHANGE**: Significant changes that may affect compatibility
  - **build**: Changes that affect the build system or external dependencies
  - **ci**: Changes to our continuous integration processes or workflows
  - **chore**: Refactor, cleanup, or other non-functional code changes
  - **docs**: Documentation update or addition
  - **feat**: Introduces a new feature or enhancement to the codebase
  - **fix**: Bug fix or error correction
  - **i18n**: Internationalization or localization changes
  - **perf**: Performance improvement
  - **refactor**: Code restructuring for better maintainability, readability, or scalability
  - **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc.)
  - **test**: Adding missing tests or correcting existing tests
  - **WIP**: Work in progress, a temporary label for incomplete or ongoing work

# Changelog Entry

### Description

Replacing peewee with SQLAlchemy. The combination of FastAPI with Peewee is not supported and we should avoid using both libraries in parallel.
The change replaces the whole ORM model with SQLAlchemy and uses Alembic for schema migrations.
The initial database migration script is checking the database schema for table existence and skips table creation in case the table already exists.

### Changed

- Replace Peewee with SQLAlchemy
- Replace peewee-migrate with alembic

### Breaking Change

[...]

PR #3595, opened 2024-07-02
feat: db migration

tjbck:
Extension of #3327


tjbck:
@jonathan-rohde Encountering this issue:
image

tjbck:
@jonathan-rohde One thing we should test before merging this to dev/main is verifying update to 0.3.8 work without any problem from other version like 0.2.0 when we didn't have supports for functions. I'd greatly appreciate if you could verify this!

jonathan-rohde:
@tjbck This will be most likely not working. The initial database migration of alembic is not throwing any errors during migration, if a table already exists. The existing table will just be ignored.

However if between the older version (e.g. 0.2.0) a peewee-migration was adding new columns to existing tables, this change is not reflected in the alembic migration. One feature I have in mind which will cause such a conflict: oauth login.

That was the reason why I flagged it as "breaking change"
In theory it should be possible to not create just one initial migration but recreate all the steps that peewee migration had. But it will be very tedious.

tjbck:
If restoring peewee migrations to alembic is all we need, that's what we'll have to do 😅

tjbck:
Perhaps we could add back in the old peewee migration code, and conditionally execute the peewee migration code only if `migratehistory` table is detected then bring the database up to the latest version, then remove (or rename) the`migratehistory` table to indicate the database has been successfully migrated?

Peter-De-Ath:
I have been trying this out on an existing install from latest dev branch using postgres database.

Generally it seems okay, but I am experiencing "random"  lockups, I haven't pinpointed taking any particular action to trigger this behavior yet

Error Message on backend.
sqlalchemy.exc.TimeoutError: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00 (Background on this error at: https://sqlalche.me/e/20/3o7r)

This doesn't seem to recover without a restart of owui

tjbck (commented):
@jonathan-rohde https://github.com/open-webui/open-webui/pull/3595#issuecomment-2207168583 seems like it might be related to [this change](https://github.com/open-webui/open-webui/pull/3595/commits/da403f3e3cf9ce700da2fdb477e0bdfc4794d37d), I'm curious to know what the reason was behind this decision for opting-in for session factory instead of context manager. Just want to make sure that I won't be breaking anything if I were to update the code to use:

```py
def get_db():
    db = SessionLocal()
    try:
        yield db
    finally:
        db.close()
```

I appreciate the help as always!

jonathan-rohde:
The session factory has the advantage, that within one common context, for example a thread, you get the same session and you are able to chain the operations within one transaction.
You will have more trouble doing it with a context manager. Although at the moment there are hardly any combined actions within one transaction in the peewee implementation.

PR #3859, opened 2024-07-14
i18n: Update Catalan Translation

aleixdorca:
# Changelog Entry

### Description

i18n: Updated Catalan Translation with new strings


tjbck:
Thanks!

PR #4046, opened 2024-07-22
perf: optimize query for chat list

### LINKED ISSUE #4035: [bug/perf] large db, slow query to `/chats/`
# Bug Report

## Description

**Bug Summary:**
i've been having this issue where the `/chats/` endpoint takes like 30+ seconds to respond with the chats list. 

i did some poking, and queried the database using `sqlite3`. the `select * from chats` query takes forever to complete, with a huge amount of encoded file data dumped. 

i'm assuming this is image/file data. When i upload an image (in this case, multiple images) the size of the database increases, and the `/chats/` endpoint takes longer to resolve.

image

Deleting all chats resolves the slow downs. 

the query used by the backend loads the entire table for a single user, which in my case, contained hundreds of images. 
```sql
SELECT chat.id AS chat_id, chat.user_id AS chat_user_id, chat.title AS chat_title, chat.chat AS chat_chat, chat.created_at AS chat_created_at, chat.updated_at AS chat_updated_at, chat.share_id AS chat_share_id, chat.archived AS chat_archived FROM chat WHERE chat.user_id = ? AND chat.archived = 0
```

result from sqlite3 query: 
image

**Steps to Reproduce:**
1. Upload large files (images) to the chat
2. repeat step one until your database is massive
3. open devtools/network
4. refresh open-webui, and see the duration of the `/chats` request. 

**Expected Behavior:**
the query should be quick. In a separate environment, with ~100 chats, the query time is like 1/100th of a second. 

**Actual Behavior:**
the query took 30 seconds. 

## Environment

- **Open WebUI Version:** v0.3.10
- **Operating System:** ubuntu 24 -> docker

# possible solutions
(just thinking out loud)
- paginate chat list 
- load files elsewhere at query time and inject them into the chat/completion call, to avoid database bloat ( unrelated but would help with db bloat)
- exclude the chat contents from the query/create a new query specifically for the chat list which does not include the entire chat contents; as this data isnt (?) required to display the sidebar

thearyadev:
# Changelog Entry

## Description
Performance issue #4035 has additional details. 
This PR optimizes the query made to the database by selecting only the required columns. 
The images which are stored in the database chat history make the slow down apparent when querying that endpoint, which happens on every reload. 

This change will allow for significantly improved performance with large database sizes. 

This PR doesn't solve the root issue, as the issue is caused by excessive database size. 

## Added

- a function, similar to the existing database query function used in the `/chats` endpoint, which only queries the fields it needs

## Changed

- changed `/chats` and `/chats/list` endpoints to use the new function.
- the function also had a bug which has now been resolved: 
`skip: int` was passed as a positional argument to the function parameter `include_archived: bool`
```python
    # argument #2 should not be "skip"
    return Chats.get_chat_list_by_user_id(user.id, skip, limit) 
```
changed to 
```python
    return Chats.get_chat_title_id_list_by_user_id(user.id, skip=skip, limit=limit)
```
using keyword arguments to safely pass the params. 
## Screenshots or Videos
**test done with 500mb database, artificially bloated with images**
before optimization

https://github.com/user-attachments/assets/16419820-adbe-4b8b-85ec-391dbfdfb129

after optimization

https://github.com/user-attachments/assets/4f0bdd5b-7ab1-438e-bff4-64fa5d530e0c

### DIFF
--- a/backend/apps/webui/models/chats.py
+++ b/backend/apps/webui/models/chats.py
@@ -244,13 +244,14 @@ class ChatTable:
                 .all()
             )
             return [ChatModel.model_validate(chat) for chat in all_chats]
+
     def get_chat_title_id_list_by_user_id(
         self,
         user_id: str,
         include_archived: bool = False,
         skip: int = 0,
         limit: int = 50,
-        ) -> List[ChatTitleIdResponse]:
+    ) -> List[ChatTitleIdResponse]:
         with get_db() as db:
             query = db.query(Chat).filter_by(user_id=user_id)
             if not include_archived:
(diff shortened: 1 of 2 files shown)


Yanyutin753:
Amazing, In my test, this makes open-webUi much faster and more powerful！！！

Yanyutin753:
Let me venture to imagine that if we can add some optimistic loading, this would make open-webui more powerful!

tjbck:
Thanks

Issue #4049, opened 2024-07-23
enh: select model in dropdown with arrow keys


thearyadev:
ability to use the Up/Down arrow keys on the keyboard to navigate the model selection list, triggered by the Enter key. 
_Originally posted by @silentoplayz in https://github.com/open-webui/open-webui/issues/3992#issuecomment-2244170597_

silentoplayz:
> _Originally posted by @silentoplayz in [#3992 (comment)](https://github.com/open-webui/open-webui/pull/3992#issuecomment-2244170597)_

For clarification, and because I used an LLM in attempt to make my original comment clearer, I will mention that I'd like to use the up/down arrow keys on my keyboard to choose which model I want to select with the Enter key on my keyboard. The quoted part before I edited my comment isn't fully clear on this, and appears to almost ask for something else now that I give it another glance.

PR #3962, opened 2024-07-17
feat: enter selects first model in list

thearyadev:
# Pull Request Checklist
**Before submitting, make sure you've checked the following:**

- [x] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [x] **Description:** Provide a concise description of the changes made in this pull request.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [x] **Testing:** Have you written and run sufficient tests for validating the changes?
- [x] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [x] **Prefix:** To cleary categorize this pull request, prefix the pull request title, using one of the 
# Changelog Entry

### Description
Add ability to select model in model selection dialog using the enter key.
Feature detailed in #3960. 

### Added
- Event key listener `on:keydown` on `src/lib/components/ModelSelector/Selector.svelte` `input` tag to check for the keydown event including the `enter` key to trigger a model selection.

### Additional Information
- Manually tested edge cases:
  - 0 search results does not cause an error
  - selection via `enter` keydown is only triggered if the input is in focus
- other ideas
  - indicator to show the user `enter` selects the top result

### Screenshots or Videos

https://github.com/user-attachments/assets/e53e2861-93db-4143-a84b-0bf6e872df71

PR #4059, opened 2024-07-23
feat: select model with arrow keys

thearyadev:
# Changelog Entry

## Description
implements enhancement outlined in #4049
in the model selection dialog, the user can press enter to select the model currently selected in the list. (#3962)
this PR adds the ability to move the selected model using the arrow keys, and to select it with the enter key. 

## Changed

- refactored `on:keydown` and related components to keep track of the currently selected model via arrow keys. 

# Additional Information

- models that are out of view will be scrolled to when they are selected with arrow keys.
- if the user clicks out of the modal, or selects a model, or opens a new chat, the `pseusoSelectedIndex` is reset. 
  - this was done to avoid the state of this from lingering across selections. For example, if the user selects a model, then opens a new chat, the pseudoSelectedIndex could be off screen. This could be solved by resetting the state utilizing a store, but that could have possible side effects so i've decided to keep it contained within the component.
- holding the arrow keys down does not cause any unexpected behaviour 
# Screenshots or Videos
https://github.com/user-attachments/assets/4ccc2225-9b1e-4fdd-a724-3b3b8b8670a2

### DIFF
--- a/src/lib/components/chat/ModelSelector/Selector.svelte
+++ b/src/lib/components/chat/ModelSelector/Selector.svelte
@@ -17,7 +17,6 @@
 	import { getModels } from '$lib/apis';
 
 	import Tooltip from '$lib/components/common/Tooltip.svelte';
-	import WebParams from '$lib/components/documents/Settings/WebParams.svelte';
 
 	const i18n = getContext('i18n');
 	const dispatch = createEventDispatcher();
--- a/src/lib/components/chat/ModelSelector/Selector.svelte
+++ b/src/lib/components/chat/ModelSelector/Selector.svelte
@@ -17,6 +17,7 @@
 	import { getModels } from '$lib/apis';
 
 	import Tooltip from '$lib/components/common/Tooltip.svelte';
+	import WebParams from '$lib/components/documents/Settings/WebParams.svelte';
 
 	const i18n = getContext('i18n');
 	const dispatch = createEventDispatcher();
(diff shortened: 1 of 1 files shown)


Peter-De-Ath (approved):

silentoplayz:
Looks good to me!

src/lib/components/chat/ModelSelector/Selector.svelte
```diff
@@ -43,6 +44,8 @@
 	let searchValue = '';
 	let ollamaVersion = null;
 
+	let pseudoSelectedIndex = 0;
```
tjbck, line 47: `selectedModelIdx`

tjbck:
The rest LGTM!

src/lib/components/chat/ModelSelector/Selector.svelte
```diff
@@ -17,6 +17,7 @@
 	import { getModels } from '$lib/apis';
 
 	import Tooltip from '$lib/components/common/Tooltip.svelte';
+	import WebParams from '$lib/components/documents/Settings/WebParams.svelte';
```
tjbck, line 20: Mistake?

thearyadev: oops. probably my auto import.

silentoplayz (approved):

tjbck:
LGTM, Thanks!

Issue #4100, opened 2024-07-24
enh: improved model selector search using fuzzy search library


thearyadev:
# Describe the solution you'd like
improved fuzzy search for model selection using a fuzzy search package

# Describe alternatives you've considered 
- filtering non alpha numeric chars 

# Additional context 
the current implementation creates poor search results, specifically for model names which include any non alphanumeric, and is substring based. a fuzzy search algorithm would allow more dynamic searching capabilities, like short forms or incomplete queries that would otherwise have no results. 

existing behaviour (current dev release -> `git-413a903`):

https://github.com/user-attachments/assets/9cefc198-7924-4633-8aab-e256728875a9

proposed change using a fuzzy search library: 
 (i used [QuickScore](https://github.com/fwextensions/quick-score) in this demo, but any fuzzy search library could be used. I tested [fuse.js](https://github.com/krisk/fuse) and the results were very poor, granted i spent a total of 10 seconds trying to figure it out)

https://github.com/user-attachments/assets/ea8c88fb-63ab-4a34-af2f-f5dd696865ec

This can also incorporate multiple keys of an object, which means additonal data can be very easily considered. For example, the search could also consider model `id` or the model `owner`. In the demo i only used a single field.

PR #4170, opened 2024-07-29
feat: fuzzy model search

thearyadev:
# Changelog Entry

## Description

this pr implements #4100 

fuzzy finding is very powerful, especially when querying data with special characters, or multiple fields. ~~The substring lookup has been replaced with fuzzy search using QuickScore~~

## Added

- ~~dependency: quick-score: 0.2.0~~
- fuzzy search algorithm. Fork of `quick-score`, simplified. 

## Changed

- removed substring filter

---

# Additional Information

- probably works on all languages
**chinese**
``` 
模型: [ "小型模型", "高级模型", "GPT模型" ]  -> Model: [ "Small model", "Advanced model", "GPT model" ]
小型: [ "小型模型" ] -> Small: [ "small model" ]
高级: [ "高级模型" ] -> Advanced: [ "Advanced Model" ]
```

# Screenshots or Videos

## old search

https://github.com/user-attachments/assets/c04275eb-dc8a-4c42-b351-754e98bc3157

## new search

https://github.com/user-attachments/assets/500987b5-9d85-4df2-9708-6a670b3fdcd6

## also works with tags

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,6 +30,7 @@
 				"marked": "^9.1.0",
 				"mermaid": "^10.9.1",
 				"pyodide": "^0.26.1",
+				"quick-score": "^0.2.0",
 				"socket.io-client": "^4.2.0",
 				"sortablejs": "^1.15.2",
 				"svelte-sonner": "^0.3.19",
@@ -7600,6 +7601,11 @@
 			"resolved": "https://registry.npmjs.org/queue-tick/-/queue-tick-1.0.1.tgz",
 			"integrity": "sha512-kJt5qhMxoszgU/62PLP1CJytzd2NKetjSRnyuj31fDd3Rlcz3fzlFdFLD1SItunPwyqEOkca6GbV612BWfaBag==",
 			"dev": true
+		},
+		"node_modules/quick-score": {
+			"version": "0.2.0",
+			"resolved": "https://registry.npmjs.org/quick-score/-/quick-score-0.2.0.tgz",
+			"integrity": "sha512-y9PW9rl7BhjTykkGYi4wh0FsXWrWFv3IJVxtAI9FMargSOhqzRN3OjO2itpw63PoSu1FY07i8B2aCwpolQucOQ=="
 		},
 		"node_modules/quick-temp": {
 			"version": "0.1.8",
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,7 +30,6 @@
 				"marked": "^9.1.0",
 				"mermaid": "^10.9.1",
 				"pyodide": "^0.26.1",
-				"quick-score": "^0.2.0",
 				"socket.io-client": "^4.2.0",
 				"sortablejs": "^1.15.2",
 				"svelte-sonner": "^0.3.19",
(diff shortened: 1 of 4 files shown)


silentoplayz:
MEOW, but also LGTM! 😸

thearyadev:
im going to verify this isn't a dumpster fire for some different languages before ready.

justinh-rahb:
One point I'd like to address is that introducing new dependencies solely for time-saving purposes may not always be the best approach. For a straightforward and self-contained function, such as a search algorithm, we could choose to re-implement just the necessary methods or copy the logic internally. This way, we won't have to be concerned about potential issues that could arise with that dependency down the line. As long as the algorithm fulfills its intended purpose, it’s unlikely to require significant modifications in the future.

thearyadev:
> One point I'd like to address is that introducing new dependencies solely for time-saving purposes may not always be the best approach. For a straightforward and self-contained function, such as a search algorithm, we could choose to re-implement just the necessary methods or copy the logic internally. This way, we won't have to be concerned about potential issues that could arise with that dependency down the line. As long as the algorithm fulfills its intended purpose, it’s unlikely to require significant modifications in the future.

I understand the concern about introducing a new dependency for a small enhancement like this. 

a fuzzy implementation would still bring the most dynamic search experience, in comparison to other search methods. 

That being said, you're concerned about the about the unnecessary overhead of adding another dependency, perhaps we could explore bundling quick-score or even creating a refactored fork with only the necessary features and bundling that to satisfy this use case? This way, we can still leverage the existing implementation without having to implement a fuzzy search algorithm, and avoid adding a million deps.

justinh-rahb:
> > One point I'd like to address is that introducing new dependencies solely for time-saving purposes may not always be the best approach. For a straightforward and self-contained function, such as a search algorithm, we could choose to re-implement just the necessary methods or copy the logic internally. This way, we won't have to be concerned about potential issues that could arise with that dependency down the line. As long as the algorithm fulfills its intended purpose, it’s unlikely to require significant modifications in the future.
> 
> a fuzzy implementation would still bring the most dynamic search experience in comparison to other search methods.
> 
> That being said, you're concerned about the about the unnecessary overhead of adding another dependency, perhaps we could explore bundling quick-score or even creating a refactored fork with only the necessary features and bundling that to satisfy this use case? This way, we can still leverage the existing implementation without having to implement a fuzzy search algorithm and avoid adding a million deps.

Copy-pasta works for me, the code seems pretty static so I don't think we'll be missing out on any hot new features 😆

thearyadev:
made some changes to the code to make it more typescript compliant, and to remove some features that aren't strictly required for this use case. should make the code easier to maintain.

justinh-rahb:
I'll give the PR a test and we'll see how she goes

tjbck:
Let's refactor this with https://www.npmjs.com/package/fuse.js

tjbck:
Closing in favour of https://github.com/open-webui/open-webui/pull/4273/commits/6ac40552b485ab1977f6e2b8a233859e3bf2b6af

thearyadev:
> Closing in favour of https://github.com/open-webui/open-webui/pull/4273/commits/6ac40552b485ab1977f6e2b8a233859e3bf2b6af

In my testing I found that fuse.js had weird results, specifically with special characters. 

I didn't spend much time turning dials so it's likely possible to get better results. Either way, should be better than substring search.

Issue #2078, opened 2024-05-08
enhancement: chat history pagination (infinite scroll)


bannert1337:
# Bug Report

## Description

**Bug Summary:**
When importing and loading a large number of chat sessions into Open WebUI, the application experiences significant performance degradation, causing the interface to freeze temporarily. This issue seems related to the side-panel attempting to load all chat sessions simultaneously.

**Steps to Reproduce:**
1. Export a large set of ChatGPT chat sessions.
2. Import these sessions into Open WebUI.
3. Access the Open WebUI to observe the loading process.
4. Notice that the interface freezes for a duration, particularly when the side-panel attempts to load all chats.

**Expected Behavior:**
The interface should remain responsive and load chat sessions smoothly without causing the entire application to freeze, regardless of the number of chats being loaded.

**Actual Behavior:**
The interface freezes temporarily when trying to load a large number of chat sessions at once, significantly affecting user experience and accessibility.

## Environment

- **Open WebUI Version:** 0.1.123
- **Ollama (if applicable):** 0.1.32

- **Operating System:** Debian 12
- **Browser (if applicable):** Floorp 11.12.2 (64-Bit)

## Reproduction Details

**Confirmation:**

- [X] I have read and followed all the instructions provided in the README.md.
- [X] I am on the latest version of both Open WebUI and Ollama.
- [ ] I have included the browser console logs.
- [ ] I have included the Docker container logs.

## Additional Information

One possible improvement could be the implementation of pagination or lazy loading for the chat sessions to avoid overwhelming the browser's capabilities by attempting to load extensive data simultaneously.

## Note

If the bug report is incomplete or does not follow the provided instructions, it may not be addressed. Please ensure that you have followed the steps outlined in the README.md and troubleshooting.md documents, and provide all necessary information for us to reproduce and address the issue. Thank you!

lukestanley:
This should probably be split into separate issues. Importing ChatGPT data happens via the frontend right now and it doesn't show it's status at all, you'd need to open the browser console to see the upload status. Every conversation has it's own POST request. Pagination of the conversation list is a separate issue from the import issues. @tjbck

bannert1337:
The import was successful. After import, the side panel loads all chats at once, therefore freezing for some seconds.

PR #4266, opened 2024-08-01
feat: Sidebar infinite scroll (pagination)

thearyadev:
# Changelog Entry

## Description
This pr adds scroll pagination to the sidebar chat list.

This will improve first-load performance, and increase ui responsiveness in cases where there are many items in the chat history. 

#2078 

## Added

- sidebar now infinitely scrolls

## Changed
- chat list endpoint uses query params `skip` and `limit`
- database query function uses `skip` as offset and `limit` as limit
- default values for `skip` and `limit` in the backend are changed to `0` and `-1` to ensure that this change does not impact other users of this endpoint, as the frontend api function has been changed to include these default values as well. limit -1 with no offset would work as if those arguments were not passed.
- all rehydration calls which update the `chats` store now use the `pageSkip` and `pageLimit`  to determine how much of the chat list actually needs to be rehydrated. 
   - for example, creating a new chat causes the sidebar to rehydrate (to get the new chat in the list), at this point, we only need to re fetch the chats that are visible. So if the user has 40 chats loaded, and they create a new chat, only 40 chats are loaded.
- instances where the full chat is required to be loaded (eg: search, import), pagination is disabled, and all components that query the backend will load all the chats (since all the chats are visible)
- **feature: sidebar uses scroll event to load more chats into the sidebar**

---

# Additional Information
- this has a very large number of edge cases, which ive made a valiant attempt at resolving. 
- additional independent testing recommended. 
- this may become very unresponsive in poor network conditions. Loading the chats still takes some time. this is a tradeoff, we move initial load time to the scroll load time across the full duration of the scroll.
- threshold for loading a new set: 60% of the scroll height
- number of items loaded on each poll: dynamically evaluated. eq `Math.round(window.innerHeight / 35) + 5` where 35 is the approximate height of a single chat in pixels.
- new issues:
  - if the user zooms out after page load to the point with the chat list `div` is smaller than the viewport height, the `div` will no longer be scrollable, so the user can only see the chats that were loaded already. 
  - ~~viewing the chat list filtered by tag, when the user deletes a tag from a chat, the chat list goes back to showing all chats. this doesn't happen in `0.3.11` I spent an hour trying to fix this with no progress.~~ resolved
 
# progress
- [x] basic implementation
- [x] fix screen size edge cases
- [x] fix rehydration usage
- [x] experiment with different `pageLimit` values in different network conditions and screen sizes
- [x] fix tag filter view switch hydration state (when filtered by tag, then rehydrating the chat list, pagination is disabled)
- [x] add loading indicator at the bottom of the sidebar
- [x] refactor repeated code into a function
- [x] remove debugging code / prep for review

# Screenshots and Videos

<details>
<summary>demo</summary>

## basic functionality
https://github.com/user-attachments/assets/513b601c-b60c-40e3-94ed-3e18be4a9f44

## apply filter loads all chats and disabled pagination

https://github.com/user-attachments/assets/ea5c3586-f9f2-4ac6-9354-2570b0f68b9f

## search bar focus disables pagination and loads all chats

https://github.com/user-attachments/assets/39bf4a89-adec-4435-8b34-e0fca388ad46

## deleting or archiving a chat only rehydrates up to the current page

https://github.com/user-attachments/assets/35dbac2e-3faa-46b5-b52f-21fc5b477e80

## attempt chat list refetch while pagination disabled (fetches all chats)
https://github.com/user-attachments/assets/5d1fd869-00fe-41a8-ac94-d8a41c1d8b03

## new chat, title gen, and subsequent messages only rehydrates up to the current page

https://github.com/user-attachments/assets/209b99a0-bb6e-4bbf-9066-c35702bb73b9

## determine page size based on `window.innerHeight`

https://github.com/user-attachments/assets/477e3681-5fc7-479c-aecf-b06708e523db

## mobile experience

https://github.com/user-attachments/assets/1db0c4b4-94ee-4ba6-a46a-e96ecb204b5f

## slow network sim

https://github.com/user-attachments/assets/c34b59af-d8ca-4a9e-9983-42dec3bf5326

</details>


tjbck:
LFGGGG 🚀 love this PR!

src/lib/components/layout/Sidebar.svelte
```diff
@@ -185,7 +194,9 @@
 				await tick();
 				goto('/');
 			}
-			await chats.set(await getChatList(localStorage.token));
+			await chats.set(
+				await getChatList(localStorage.token, 0, $pageSkip * $pageLimit || $pageLimit)
```
thearyadev, line 198: missed bool eval order

erickgamer787:
I loved

src/lib/components/layout/Sidebar.svelte
```diff
@@ -49,6 +52,12 @@
 	let showDropdown = false;
 
 	let filteredChatList = [];
+	let paginationScrollThreashold = 0.6;
+	let nextPageLoading = false;
+	let tagView = false;
+	let chatPagniationComplete = false;
+
+	pageLimit.set(20);
```
thearyadev, line 60: could cause issues on large screens, or zoomed out. 

only 20 chats load, so if the list of chats is smaller than the screen height, there is no scroll event that can be made to then get the next 

maybe get screen height from the window and determine the pageLimit from that. Could also improve performance on mobile, only a few chats need to be loaded in comparison to a larger screen.

src/lib/components/layout/Sidebar.svelte
```diff
@@ -477,7 +514,33 @@
...
+						// extend existing chats
+						const nextPageChats = await getChatList(
+							localStorage.token,
+							$pageSkip * $pageLimit,
+							$pageLimit
+						);
+						// once the bottom of the list has been reached (no results) there is no need to continue querying
+						chatPagniationComplete = nextPageChats.length === 0;
+						await chats.set([...$chats, ...nextPageChats]);
+						nextPageLoading = false;
```
thearyadev, line 540: could add a loading indicator to the bottom of the chat list. in case the user scrolls faster than their network can respond, it will inform the user that additional chats are still loading.

src/lib/components/layout/Sidebar.svelte
```diff
@@ -410,7 +433,15 @@
 						class="w-full rounded-r-xl py-1.5 pl-2.5 pr-4 text-sm bg-transparent dark:text-gray-300 outline-none"
 						placeholder={$i18n.t('Search')}
 						bind:value={search}
-						on:focus={() => {
+						on:focus={async () => {
+							// loading all chats. disable pagination on scrol.
+							scrollPaginationEnabled.set(false);
```
thearyadev, line 438: refacto: pagination disable as function

PR #4273, opened 2024-08-01
0.3.11

tjbck:


michaelpoluektov:
https://github.com/open-webui/open-webui/blob/dev/backend/main.py#L1116
https://github.com/open-webui/open-webui/blob/dev/backend/main.py#L1288

`chat_completed` is redefined: is that normal?

tjbck:
> https://github.com/open-webui/open-webui/blob/dev/backend/main.py#L1116 https://github.com/open-webui/open-webui/blob/dev/backend/main.py#L1288
> 
> `chat_completed` is redefined: is that normal?

Good catch, resolved!

PR #4237, opened 2024-07-31
refactor: Simplify functions

michaelpoluektov:
# Pull Request Checklist

Motivation: there are no docs on functions, so if people will keep using code as documentation might as well make the code more readable.

Also this makes it simpler to add more features/custom named params to functions.

## Testing:

- Tested with all combinations of function outputting/not outputting stream and OWUI expecting/not expecting stream
- Did not test with every single possible case (Iterator, Generator, starting/not starting with "data:" etc.): testing welcome.

**Before submitting, make sure you've checked the following:**

- [x] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [x] **Description:** Provide a concise description of the changes made in this pull request.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [x] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [x] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [ ] **Testing:** Have you written and run sufficient tests for validating the changes?
- [x] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [x] **Prefix:** To cleary categorize this pull request, prefix the pull request title, using one of the following:
  - **BREAKING CHANGE**: Significant changes that may affect compatibility
  - **build**: Changes that affect the build system or external dependencies
  - **ci**: Changes to our continuous integration processes or workflows
  - **chore**: Refactor, cleanup, or other non-functional code changes
  - **docs**: Documentation update or addition
  - **feat**: Introduces a new feature or enhancement to the codebase
  - **fix**: Bug fix or error correction
  - **i18n**: Internationalization or localization changes
  - **perf**: Performance improvement
  - **refactor**: Code restructuring for better maintainability, readability, or scalability
  - **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc.)
  - **test**: Adding missing tests or correcting existing tests
  - **WIP**: Work in progress, a temporary label for incomplete or ongoing work

# Changelog Entry

### Description
Refactor everything around function calling.
- Remove unneeded `async` in `get_event_emitter`
- Remove unused imports
- Split `generate_function_chat_completion` into multiple smaller functions
- DRY changes
- Use dict.pop when appropriate

No functional changes expected: everything should work exactly as before.

PR #4295, opened 2024-08-02
refactor: Refactor OpenAI API to use helper functions, silence LSP/linter warnings

michaelpoluektov:
# Pull Request Checklist

Sorry for the large amount of refactor PR's. I understand it's a pain to review but it's a lot easier to work on a project when you can rely on your LSP to hint at possible errors and not get inundated with warnings and whatnot.

This addresses a few things:
- Helper functions introduced in  #4237 are now re-used for OpenAI
- A few files were auto-formatted
- Silence LSP warnings (potentially unbound variables, inconsistencies in type annotations, including `get_all_models`)
- Silence linter warnings (bare `except:`, unused imports, `== None` instead of `is None`, unused variables, function re-definition with the same name)

# Changelog Entry

### Description

- Refactor everything around OpenAI API

### DIFF
--- a/backend/apps/openai/main.py
+++ b/backend/apps/openai/main.py
@@ -30,7 +30,7 @@ from config import (
     MODEL_FILTER_LIST,
     AppConfig,
 )
-from typing import List, Optional
+from typing import List, Optional, Literal, overload
 
 
 import hashlib
@@ -262,12 +262,22 @@ async def get_all_models_raw() -> list:
     return responses
 
 
-async def get_all_models() -> dict[str, list]:
+@overload
+async def get_all_models(raw: Literal[True]) -> list: ...
+
+
+@overload
+async def get_all_models(raw: Literal[False] = False) -> dict[str, list]: ...
+
+
+async def get_all_models(raw=False) -> dict[str, list] | list:
     log.info("get_all_models()")
     if is_openai_api_disabled():
-        return {"data": []}
+        return [] if raw else {"data": []}
 
     responses = await get_all_models_raw()
+    if raw:
+        return responses
 
     def extract_data(response):
         if response and "data" in response:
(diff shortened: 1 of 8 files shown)


tjbck:
LGTM, Thanks!

backend/apps/openai/main.py
```diff
@@ -234,72 +228,66 @@ def merge_models_lists(model_lists):
...
 
 
-async def get_all_models(raw: bool = False):
+def is_openai_api_disabled():
+    api_keys = app.state.config.OPENAI_API_KEYS
+    no_keys = len(api_keys) == 1 and api_keys[0] == ""
+    return no_keys or not app.state.config.ENABLE_OPENAI_API
+
+
+async def get_all_models_raw() -> list:
```
tjbck, line 237: Not a huge fan of having a separate `get_all_models_raw` function, perhaps it's better to make this inner function for `get_all_models` to enable `raw=True` param?

michaelpoluektov: LSP's don't like it when the output type of a function depends on an argument: at least PyRight isn't smart enough to figure out that `get_all_models()` returns a list if and only if `raw` is set to `True`, and starts yapping about how you can't use a `list` as a `dict` (when raw is False)

Only other way to make the warnings go away is to call `assert isinstance` after each call to `get_all_models()` or implement some monstrosity with generics.

My hope is that eventually we can add a static type checker to core backend components. Makes it a lot easier to catch errors early.

tjbck: I wouldn't want to change the current function just for LSP's sake. Perhaps`Union[dict, list]` could work?

michaelpoluektov:
The LSP can infer `Union[list, dict]` but when you use the result it
doesn't know if its list or dict. If you really don't like it I'll change
it back once I get back home.

backend/apps/openai/main.py
michaelpoluektov: Found a solution with `@overload`.

tjbck:
LGTM, Thanks!

PR #4485, opened 2024-08-08
PWA theme colour match dark/light themes

jtsorlinis:
# Changelog Entry

### Description

- When using open-webui as a PWA the bottom and top bars are always white, which doesn't look great if using dark modes.

### Added

- PWA theme colour should now be light or dark depending on the selected theme.

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -1,4 +1,4 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
 	<head>
 		<meta charset="utf-8" />
--- a/src/app.html
+++ b/src/app.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
 	<head>
 		<meta charset="utf-8" />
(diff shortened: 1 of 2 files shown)


src/app.html
```diff
@@ -38,15 +40,21 @@
...
 					document.documentElement.classList.add('her');
 				} else {
 					document.documentElement.classList.add('dark');
 				}
 
+				if (
+					localStorage?.theme.includes('dark') ||
+					(localStorage.theme === 'system' && isSystemDark)
+				) {
+					window.themeColor.content = '#0d0d0d';
```
tjbck, line 55: Shouldn't this also be added to the listener?

erickgodev:
Look, I was going to implement this but I had problems with the OLED and the HUR making the iPhone not recognize the current theme, oh you have to edit it in the settings too because if you change the theme it will only appear when you reload the app

parthpat12:
The issue still persists on iOS 17.5.1. I’m on latest v0.3.15 and the white status bars at top still show. Can we reopen this issue since it’s not fixed? The text on the bottom is also getting covered by swipe bar.
Screenshot 2024-08-26 at 3 18 50 PM

jtsorlinis:
Yeah I'm happy to make the suggested changes but someone closed the PR already

tjbck:
Just reopened, but I'd encourage you to rebase first!

src/app.html
jtsorlinis: I've added it to the listener!

jtsorlinis:
I've added it to the listener, this won't work perfectly for black vs dark grey but having it as dark grey still looks a lot better than bright white

tjbck:
Is this PR tested? Could you provide us with a video demo?

PR #4674, opened 2024-08-16
sec: Sanitize 11labs voice id to address semgrep security issue: tainted-path-traversal-stdlib-fastapi

crizCraig:
# Changelog Entry

### Description

* Voice id's are passed to `/speech` without sanitization. They should be open AI names or elevenlabs id's, so the PR calls `get_elevenlabs_voices()` to ensure the passed ID is one of those. Else an HTTP 400 is returned. 

### Added

* I cached `get_elevenlabs_voices()` with `@lru_cache` to avoid waiting 1 second for every `/speech` call on just that. 

### Changed

* I changed the return of get_elevenlabs_voices to a dict as the number of voices are getting to the point you'd want O(1) lookup: https://api.elevenlabs.io/v1/voices

### Security

* Addressed tainted-path-traversal-stdlib-fastapi within the audio api

### Testing

* I tested TTS from elevenlabs, openai, and the default in-browser synthesize locally with the change

### How to test

* Set elevenlabs env vars
```
    AUDIO_TTS_ENGINE=elevenlabs
    AUDIO_TTS_API_KEY=sk_...  # Your Elevenlabs API key
    AUDIO_TTS_VOICE=EXAVITQu4vr4xnSDxMaL  # Sarah from https://api.elevenlabs.io/v1/voices
    AUDIO_TTS_MODEL=eleven_multilingual_v2
```
* Hit the call button and vibe!
image
image

### DIFF
--- a/backend/apps/audio/main.py
+++ b/backend/apps/audio/main.py
@@ -254,6 +254,13 @@ async def speech(request: Request, user=Depends(get_verified_user)):
             raise HTTPException(status_code=400, detail="Invalid JSON payload")
 
         voice_id = payload.get("voice", "")
+
+        if voice_id not in get_available_voices():
+            raise HTTPException(
+                status_code=400,
+                detail="Invalid voice id",
+            )
+
         url = f"https://api.elevenlabs.io/v1/text-to-speech/{voice_id}"
 
         headers = {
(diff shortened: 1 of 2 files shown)


tjbck:
LGTM, Thanks!

PR #4714, opened 2024-08-18
sec: Allow setting CORS origin 

crizCraig:
# Changelog Entry

### Description

- My semgrep scan pickedup CORS * as a medium level issue, but for open webui, I'd consider it higher since you could have malicious web apps grabbing chat history.

### Added

- CORS_ALLOW_ORIGIN env var with semicolon delimited allowed origins
- Checks to make sure origins are valid urls or * as frontend will silently fail with CORS errors otherwise

### Changed

- Changed all fastapi cors middleware declarations to use the new env var, e.g.

```python
 app.add_middleware(
    CORSMiddleware,
    allow_origins=CORS_ALLOW_ORIGIN,  # used to be ["*"]
...
```

The default value is still `["*"]` so current users should be unaffected

### Security

- Allows setting CORS origin in `.env` for production deployments

### How it was tested

- I tried all functionality with and without setting `CORS_ALLOW_ORIGIN`
- I tried valid and invalid CORS to make sure setting had an effect
  - `http://localhost:5173,http://localhost:8080`
  - `notavalidurl`
- I tested `apps/images`, `apps/audio`, `apps/openai`, and `main.py` in `backend`
- I DID NOT test `apps/ollama`, `apps/rag`, or `apps/webui` as I'm not sure how to trigger those, but the change is identical to the other apps. The middleware is configured on import here, but not exercised in manual tests like the above

### Screenshots or Videos

- Image generation works!
image

### Additional notes

I'll add documentation about the env var in the deploy section

### DIFF
--- a/backend/apps/audio/main.py
+++ b/backend/apps/audio/main.py
@@ -38,6 +38,7 @@ from config import (
     AUDIO_TTS_MODEL,
     AUDIO_TTS_VOICE,
     AppConfig,
+    CORS_ALLOW_ORIGIN,
 )
 from constants import ERROR_MESSAGES
 from utils.utils import (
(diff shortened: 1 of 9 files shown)


tjbck:
LGTM, Thanks!

PR #4800, opened 2024-08-22

aleixdorca:
# Changelog Entry

### Description

Updated the catalan translation.


tjbck:
Thanks!

PR #4813, opened 2024-08-22
feat: show rag status when using models with knowledge collections

jannikstdl:
# Changelog Entry

### Added

- **📊 Status Bar for Knowledge Collections**: Displays a status bar during vector retrieval to inform users of loading progress and improve UX.

### Description

- Displays a status bar when the user is using a model with Knowledge Collections. Similar to ChatGPT, this improves UX by showing the user what is currently happening, as large collections often have longer loading times during retrieval. This makes the user aware that the longer loading time is due to vector retrieval and not the language model.

### Screenshots or Videos

https://github.com/user-attachments/assets/22b6ec69-7334-4861-9df8-7c677a5a669d

PS: had this in my fork, maybe this is useful..


tjbck:
Can we rebase first?

tjbck:
Thanks!

PR #4971, opened 2024-08-28
Lighthouse seo score and error fixes

sebdanielsson:
# Changelog Entry

### Description

No errors are reported for the SEO tests and it is now scored at 66, up from 50. Only way to improve it is to allow search engines to index which is probably unwanted for most people.

### Fixed

- Added robots.txt file with Disallow All addresses a Lighthouse SEO error while still preventing content indexing by search engines.
- A meta description has been added, addressing a Lighthouse recommendation and positively impacting the SEO score.

### Misc

I updated the description to only "Open WebUI" as per your recommendation in the previous PR☺️


tjbck:
Thanks!

PR #4976, opened 2024-08-28
i18n: Update Catalan Translation

aleixdorca:
# Changelog Entry

### Description

Updated the Catalan translation.


tjbck:
Thanks!

Issue #668, opened 2024-02-07
feat: LDAP User management


this-josh:
**Is your feature request related to a problem? Please describe.**
Users should be managed via LDAP, similar to [Portainer](https://docs.portainer.io/admin/settings/authentication/ldap)

**Describe the solution you'd like**
LDAP user management implemented

**Describe alternatives you've considered**
- OAUTH
- Microsoft Active directory

I believe LDAP is the best starting point

**Additional context**

One key feature of `ollama-webui` is its ability to have users, but without LDAP this doesn't have same network level utility.

jnkstr:
Yes agree, this would be a use case for us as well.

tjbck:
Related: https://github.com/ollama-webui/ollama-webui/issues/483

SurvivaLlama:
I can help test this.

eingemaischt:
This would be great for us as well - an auth via Header, set by an reverse proxy, would be a great alternative. Then we could add different auth mechanisms (OpenID, LDAP, Kerberos etc) without implementing each of them in webui itself...

justinh-rahb:
> This would be great for us as well - an auth via Header, set by an reverse proxy, would be a great alternative. Then we could add different auth mechanisms (OpenID, LDAP, Kerberos etc) without implementing each of them in webui itself...

Excellent news, this has been implemented:
- #1347

Manfredzimmermann:
How can I connect Open-WebUI with Microsoft Active directory? I'm looking for an excellent solution. A fixed module in Open-WebUI for Active Directory (LDAP) would be a dream

hostingnuggets:
Me too! +100 from me ;-)

peske:
+1

sry9681:
+1

lduplaga:
Would be awesome to have this feature.

seal61:
+1

lduplaga:
I solved my case with Microsoft authentication by this 

https://docs.openwebui.com/tutorial/sso#microsoft

seal61:
> I solved my case with Microsoft authentication by this
> 
> https://docs.openwebui.com/tutorial/sso#microsoft

does not help with local active directory / ldap server sadly, but might be useful for some.

WilsonZiweiWang:
My implementation: https://github.com/WilsonZiweiWang/open-webui/commit/bfe556010bed6e6e4a0a63a9ef1e27fe20e3ca17
[...]

PR #5056, opened 2024-08-31
feat: Integrate LDAP authentication

WilsonZiweiWang:
# Pull Request Checklist

# Changelog Entry

### Description

Integrate LDAP authentication

### Added

Frontend:
Added a button to the login page following the existing style for oauth options to show LDAP login form if LDAP is enabled.
Added UI for LDAP configs through `admin settings` -> `Connection`

Backend:
Added endpoints to handle LDAP authentication requests and configs.

Configs:
Added a config to toggle LDAP sign-in method.
Added configs for LDAP server including host, port, Application DN, etc. You can also specify the values in the env file.

---

### Additional Information

The initial implementation was based on my use case with my LDAP server so it might not work out of the box.

Related issue: https://github.com/open-webui/open-webui/issues/668


WilsonZiweiWang:
@Peter-De-Ath
To address this [comment](https://github.com/open-webui/open-webui/issues/668#issuecomment-2322627904) of yours. The way it authenticates doesn't work in my use case. I think a reasonable workaround would be to make the `DN` used for authentication fully customizable through the admin settings, where we control the components that make up the `DN`.

justinh-rahb:
> @Peter-De-Ath To address this [comment](https://github.com/open-webui/open-webui/issues/668#issuecomment-2322627904) of yours. The way it authenticates doesn't work in my use case. I think a reasonable workaround would be to make the `DN` used for authentication fully customizable through the admin settings, where we control the components that make up the `DN`.

This is more or less a requirement with any LDAP authentication plugin, since there's a HUGE variety in how domains and their OUs are organized. Most that I've seen will give very granular options for setting search filters, bind DNs, etc.

As a proof of concept this PR is working (with minor adjustment to account for _my_ LDAP server), but for production it needs to be more configurable.

WilsonZiweiWang:
> This is more or less a requirement with any LDAP authentication plugin, since there's a HUGE variety in how domains and their OUs are organized. Most that I've seen will give very granular options for setting search filters, bind DNs, etc.
> 
> As a proof of concept this PR is working (with minor adjustment to account for _my_ LDAP server), but for production it needs to be more configurable.

Thanks for the info!

WilsonZiweiWang:
@justinh-rahb
Hi, to reply your comment from https://github.com/open-webui/open-webui/issues/668#issuecomment-2322733312
> This won't interact well with WEBUI_LOGIN=false, the email/username and password fields are hidden

I separated the forms, and now the LDAP form can show whether or not the sign-in is enabled.

> Could we not use the email field for login? We already get it from the LDAP server to create the account

Do you mean disable email/password login when LDAP is enabled?

> Would prefer if LDAP is enabled that a button be shown in a similar style to the OIDC login buttons, hidden otherwise

I added the button following the existing style.

> May want to be able to configure which field is used for the username

The attribute for the username is now configurable.

justinh-rahb:
> > Could we not use the email field for login? We already get it from the LDAP server to create the account
> 
> Do you mean disable email/password login when LDAP is enabled?
> 
> The attribute for the username is now configurable.

With the configurable username field, this shouldn't be a problem I suppose, one could just use the email field then 👌

I'll build and test this when I have some time.

src/lib/apis/auths/index.ts
```diff
@@ -110,6 +110,150 @@ export const getSessionUser = async (token: string) => {
 	return res;
 };
 
+export const ldapUserSignIn = async (user: string, password: string) => {
+	let error = null;
+
+	const res = await fetch(`${WEBUI_API_BASE_URL}/auths/ldap`, {
```
gaby, line 116: Use `joined_url = urllib.parse.urljoin(base_url, url_to_join)` to join the Base URL with the ldap path.

src/lib/apis/auths/index.ts
```diff
@@ -110,6 +110,150 @@ export const getSessionUser = async (token: string) => {
...
+		throw error;
+	}
+
+	return res;
+};
+
+export const getLdapConfig = async (token: string = '') => {
+	let error = null;
+
+	const res = await fetch(`${WEBUI_API_BASE_URL}/auths/admin/config/ldap`, {
```
gaby, line 148: Use `joined_url = urllib.parse.urljoin(base_url, url_to_join)`

src/lib/apis/auths/index.ts
```diff
@@ -110,6 +110,150 @@ export const getSessionUser = async (token: string) => {
...
+		throw error;
+	}
+
+	return res;
+}
+
+export const updateLdapConfig = async (token: string = '', enable_ldap: boolean) => {
+	let error = null;
+
+	const res = await fetch(`${WEBUI_API_BASE_URL}/auths/admin/config/ldap`, {
```
gaby, line 175: Use `joined_url = urllib.parse.urljoin(base_url, url_to_join)`

src/lib/apis/auths/index.ts
```diff
@@ -110,6 +110,150 @@ export const getSessionUser = async (token: string) => {
...
+		throw error;
+	}
+
+	return res;
+}
+
+export const getLdapServer = async (token: string = '') => {
+	let error = null;
+
+	const res = await fetch(`${WEBUI_API_BASE_URL}/auths/admin/config/ldap/server`, {
```
gaby, line 205: Use `joined_url = urllib.parse.urljoin(base_url, url_to_join)`

src/lib/apis/auths/index.ts
```diff
@@ -110,6 +110,150 @@ export const getSessionUser = async (token: string) => {
...
+		throw error;
+	}
+
+	return res;
+}
+
+export const updateLdapServer = async (token: string = '', body: object) => {
+	let error = null;
+
+	const res = await fetch(`${WEBUI_API_BASE_URL}/auths/admin/config/ldap/server`, {
```
gaby, line 232: Use `joined_url = urllib.parse.urljoin(base_url, url_to_join)`

backend/open_webui/apps/webui/routers/auths.py
```diff
@@ -107,6 +121,90 @@ async def update_password(
...
+    USE_SSL = request.app.state.config.USE_SSL
+    CA_CERT_FILE = request.app.state.config.CA_CERT_FILE
+
+    if not ENABLE_LDAP:
+        raise HTTPException(400, detail="LDAP authentication is not enabled")
+    
+    try:
+        tls = Tls(validate=CERT_REQUIRED, version=PROTOCOL_TLS, ca_certs_file=CA_CERT_FILE, ciphers='ALL')
+    except Exception as e:
+        log.error(f"An error occurred on TSL: {str(e)}")
```
gaby, line 148: Typo `TSL`, should be `TLS`

backend/open_webui/apps/webui/routers/auths.py
```diff
@@ -107,6 +121,90 @@ async def update_password(
...
+                token = create_token(
+                    data={"id": user.id},
+                    expires_delta=parse_duration(request.app.state.config.JWT_EXPIRES_IN),
+                )
+
+                # Set the cookie token
+                response.set_cookie(
+                    key="token",
+                    value=token,
+                    httponly=True,  # Ensures the cookie is not accessible via JavaScript
```
gaby, line 190: Ideally we should add `secure=True`  here, but I believe that requires `open-webui` to support HTTPS natively

backend/open_webui/apps/webui/routers/auths.py
```diff
@@ -389,6 +487,84 @@ async def update_admin_config(
...
+        "search_filters": request.app.state.config.LDAP_SEARCH_FILTERS,
+        "use_tls": request.app.state.config.USE_SSL,
+        "certificate_path": request.app.state.config.CA_CERT_FILE
+    }
+
+@router.post("/admin/config/ldap/server")
+async def update_ldap_server(
+    request: Request, form_data: LdapServerConfig, user=Depends(get_admin_user)
+):
+    requried_fields = ['label','host', 'attribute_for_username', 'app_dn','app_dn_password','search_base', 'use_tls']
```
gaby, line 523: Spaces missing here between items

backend/open_webui/config.py
```diff
@@ -1435,3 +1435,73 @@ class BannerModel(BaseModel):
...
 )
+
+####################################
+# LDAP
+####################################
+
+ENABLE_LDAP = PersistentConfig(
+    "ENABLE_LDAP",
+    "ldap.enable",
+    os.environ.get("ENABLE_LDAP", "True").lower() == "true",
```
gaby, line 1446: Should default be `"True"` here?

backend/open_webui/apps/webui/routers/auths.py
```diff
@@ -107,6 +121,90 @@ async def update_password(
...
+    LDAP_APP_DN = request.app.state.config.LDAP_APP_DN
+    LDAP_APP_PASSWORD = request.app.state.config.LDAP_APP_PASSWORD
+    USE_SSL = request.app.state.config.USE_SSL
+    CA_CERT_FILE = request.app.state.config.CA_CERT_FILE
+
+    if not ENABLE_LDAP:
+        raise HTTPException(400, detail="LDAP authentication is not enabled")
+    
+    try:
+        tls = Tls(validate=CERT_REQUIRED, version=PROTOCOL_TLS, ca_certs_file=CA_CERT_FILE, ciphers='ALL')
```
gaby, line 146: The ciphers param should be configurable via ENV variables/Web UI.

stondino00:
Any update on getting this merged?

gaby:
@WilsonZiweiWang Can you take a look at my comments, thanks

src/lib/apis/auths/index.ts
WilsonZiweiWang: This is TS and the method you mentioned is for Python. Plus, this is how other functions join the URL parts and I was just following the existing style.

backend/open_webui/apps/webui/routers/auths.py
WilsonZiweiWang: `Ciphers` is now configurable.

gaby:
@WilsonZiweiWang Is this still draft?

WilsonZiweiWang:
> Is this still draft?

Will mark this ready after I finish the checklist items.

stondino00:
> > Is this still draft?
> 
> Will mark this ready after I finish the checklist items.

Will this be in the next update @WilsonZiweiWang?  Thanks so much.

lee-b:
Looking at the proposed changes:

+app.state.config.USE_SSL = USE_SSL
+app.state.config.CA_CERT_FILE = CA_CERT_FILE

I have a few observations:

- These settings appear to be global (unprefixed), whereas the other config entries are LDAP_ prefixed.  That could be very bad, becoming mixed with the settings for other parts of open-webui.

- **As such, I think that this code should not be merged as-is, unless the project owners agree that they're planning to globally enable or disable SSL from one flag, and planning to globally set one specific CA as trusted.**

- This is opt-in security. Security should be enabled by default and opt-out, rather than opt-in, so "USE_SSL" (instead of something like "DISABLE_SSL", or rather, "LDAP_DISABLE_X509") would be bad practice here.

- System-wide configuration should generally be used for trusted X509 certs, rather than expecting custom configuration for every application to make it work.  Note that this is in keeping with the whole point of centralised authentication like LDAP, which is to centralise security configuration and concerns, and reduce the risk of misconfiguration on one particular service or system.

- Nitpick: SSL is outdated; it's TLS these days.  What is really meant, however, is X509 public key cryptography.


So, I think would be best to change the above to something like:

# default : False
LDAP_SERVER_DISABLE_X509

# optional, uses system-wide CAs by default
LDAP_CA_CERT_FILE ...

This way, LDAP encryption would "just work", if x509 is already configured correctly on the server side and on the client machine.

gaby:
@lee-b That's assuming you are installing open-webui on your hosts. A lot of people use the Docker container, thus these configs are needed.

Agree that the variables should be prefixed with `LDAP_` since they could confused someone to think that Open-WebUI provides HTTPS listener, which it doesnt.

SSL/TLS for LDAP should be enabled by default. But configurable.

lee-b:
> @lee-b That's assuming you are installing open-webui on your hosts. A lot of people use the Docker container, thus these configs are needed.
> 
> Agree that the variables should be prefixed with `LDAP_` since they could confused someone to think that Open-WebUI provides HTTPS listener, which it doesnt.
> 
> SSL/TLS for LDAP should be enabled by default. But configurable.

Yep, a configurable CA cert file makes sense as a container environment variable. Good point there.

Xyz00777:
hi an admin of a research company here, we would like to use it but need ldap support, it would be really aweseom if it would come in near time because handling of many users would be a pain without it....

WilsonZiweiWang:
Thanks for the feedback, all.

@lee-b 

>  so "USE_SSL" (instead of something like "DISABLE_SSL", or rather, "LDAP_DISABLE_X509") would be bad practice here

> SSL is outdated; it's TLS these days.

I used `USE_SSL` because, in the `ldap3` lib, the parameter in the LDAP server object is called `use_ssl`. See this [line](https://github.com/open-webui/open-webui/pull/5056/files#diff-26c1f42ed2225c0ee15cdc3aec215d94d4b1c8606a242754c73de2192a1ec5fbR181).  I use the term TLS now.

>  Security should be enabled by default and opt-out

This setting is enabled by default already. See [here](https://github.com/open-webui/open-webui/pull/5056/files#diff-cd27af7c865ad1597c964f178a82e65988b1961428ce0bb0f455a88766d2343cR1560)

> These settings appear to be global (unprefixed), whereas the other config entries are LDAP_ prefixed.

These configs are now `LDAP_` prefixed.

stondino00:
This didn't appear to make it into 0.3.33.  Hoping for 0.3.34.

stondino00:
Can this be merged?

morgan55555:
Tested this locally, seems that LDAP auth cannot work without WEBUI_AUTH=true (auth will be completely disabled) and ENABLE_SIGNUP=true (Will throw "You do not have permission to access this resource. Please contact your administrator for assistance." error, if this is a first time login).
But enabling WEBUI_AUTH and ENABLE_SIGNUP allow for all users to create non-LDAP accounts, and seems that there is no option to allow only LDAP auth.
Maybe I'm missing something? Any suggestions?

WilsonZiweiWang:
@morgan55555 

Thanks for testing the feature. The cause is indeed the calling of `signup()` in the `/ldap` route. It makes sense to me to have an option to only allow LDAP authentication while disabling all possible ways of signing and the normal sign-up. It might be better to have another config for it so we have better control over the flow.

> Hotfixed locally with
`request.app.state.config.ENABLE_LOGIN_FORM and not request.app.state.config.ENABLE_LDAP`

I assume you replaced `backend/open_webui/apps/webui/routers/auths.py` lines 336 and 337 with this? I'd prefer to use a new config for the fix.

morgan55555:
> I assume you replaced `backend/open_webui/apps/webui/routers/auths.py` lines 336 and 337 with this? I'd prefer to use a new config for the fix.

Yes, but only for a hotfix, this solution as I said is potentially vulnerable to POST injection.
Seems that ENABLE_LOGIN_FORM=false is enough for only allow LDAP authentication, same as OAuth, cause this not only hiding actual login form, this is also disabling signup.
So, only a ENABLE_LOGIN_FORM bypass is needed for signup function, by an optional argument (or, maybe, break this function in two).

tjbck:
Is this PR ready to be reviewed/merged?

morgan55555:
> Is this PR ready to be reviewed/merged?

Seems that it's not currently ready, 'cause it doesn't work with only LDAP enabled.
It's breaking with "ENABLE_LOGIN_FORM=false" arg.
But seems, that it is only issue.

Fixed it locally with this workaround in auths.py:

`
            if not user:
                await signup(
                    request,
                    response,
                    SignupForm(email=mail, password=str(form_data.password), name=cn),
                    True
                )
                user = Auths.authenticate_user(mail, password=str(form_data.password))
`

`
@router.post("/signup", response_model=SessionUserResponse)
async def signup(request: Request, response: Response, form_data: SignupForm, skipCheck: bool = False):
    if WEBUI_AUTH:
        if (
            not request.app.state.config.ENABLE_SIGNUP
            or not request.app.state.config.ENABLE_LOGIN_FORM and not skipCheck
        ):
`

stondino00:
So does this workaround need updated in this PR @WilsonZiweiWang?

jk-f5:
Just an FYI in case anyone doesn't know this. [Dex](https://github.com/dexidp/dex) is an OIDC identity service that can be configured to authenticate using an LDAP auth backend (and many others).

It has the added benefit, that once configured, then you have a generalized OIDC provider you can use with other apps that don't support LDAP auth.

tjbck:
@morgan55555 if I merge this PR to dev, will you be able to make a subsequent PR to address the issue you described?

WilsonZiweiWang:
> @morgan55555 if I merge this PR to dev, will you be able to make a subsequent PR to address the issue you described?

I’d appreciate that. I don’t think I’ll be able to work on it anytime soon, as I’ve been quite busy with work lately. I will rebase it again to resolve the conflicts and I will leave it to you @tjbck. Thanks!

morgan55555:
> > @morgan55555 if I merge this PR to dev, will you be able to make a subsequent PR to address the issue you described?
> 
> I’d appreciate that. I don’t think I’ll be able to work on it anytime soon, as I’ve been quite busy with work lately. I will rebase it again to resolve the conflicts and I will leave it to you @tjbck. Thanks!

@WilsonZiweiWang @tjbck

https://github.com/open-webui/open-webui/pull/6743
Code with some fixes, need to be tested.
Big thanks for @WilsonZiweiWang, based on your code.

tjbck:
Closing in favour of https://github.com/open-webui/open-webui/pull/6743, Thanks everyone their effort!

tjbck:
LDAP support merged to dev. More testing wanted from the community here!

PR #5209, opened 2024-09-06
i18n: Update Catalan translation.json

aleixdorca:
# Changelog Entry

### Description

i18n: Updated Catalan translation


tjbck:
Thanks!

PR #5275, opened 2024-09-08
fix: Add padding to compensate for iPhone nav bar

sebdanielsson:
# Changelog Entry

### Description

This PR adds padding on devices that uses a bottom navigation bar that overlays the app. For desktop and phones that does not utilize a floating navigation bar nothing is changed. However for iPhone this fix will now add padding to compensate for this.

### Fixed

- Added bottom padding to compensate for iPhone navigation bar

---

### Screenshots or Videos

Before:

image

After:

image

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -4,7 +4,10 @@
 		<meta charset="utf-8" />
 		<link rel="icon" href="%sveltekit.assets%/favicon.png" />
 		<link rel="manifest" href="%sveltekit.assets%/manifest.json" crossorigin="use-credentials" />
-		<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1" />
+		<meta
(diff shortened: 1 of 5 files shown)


tjbck:
LGTM, Thanks!

parthpat12:
Added padding is too high. In PWA with keyboard, you can see lot of empty space. 
IMG_2128

sebdanielsson:
> Added padding is too high. In PWA with keyboard, you can see lot of empty space. 
> 
> IMG_2128
> 
> 

The left and right padding didn't change with this PR. But I agree that it's too much. Maybe open a PR and fix it? I might be able to take a look if I have some sparetime this weekend.

Regarding the bottom padding I'll try to have the extra padding only when the bottom navigation bar is present.

PR #5276, opened 2024-09-08
fix: Set theme colors to fix white top bar for PWA

sebdanielsson:
# Changelog Entry

### Description

This PR adds the theme-color property so the background above and under the app matches the apps background. An example is provided at the bottom of the PR.

### Fixed

- Add theme-color property to match the apps theme in PWA and when scrolling above and under the app on some browsers.

### Screenshots or Videos

Before:

image

image

After:

image

image

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -8,6 +8,7 @@
 			name="viewport"
 			content="width=device-width, initial-scale=1, maximum-scale=1, viewport-fit=cover"
 		/>
+		<meta name="theme-color" content="#171717" />
 		<meta name="robots" content="noindex,nofollow" />
 		<meta name="description" content="Open WebUI" />
 		<link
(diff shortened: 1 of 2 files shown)


silentoplayz:
Related - https://github.com/open-webui/open-webui/pull/5275

Also, LGTM!

ingmferrer:
Related - #4485

tjbck:
I believe this won't update when you toggle between the two themes if I'm not mistaken.

sebdanielsson:
> I believe this won't update when you toggle between the two themes if I'm not mistaken, could anyone verify this?

Nice catch, that's correct. I'll work on this.👍

sebdanielsson:
> I believe this won't update when you toggle between the two themes if I'm not mistaken, could anyone verify this?

Should be working great now. So now the theme-color meta tag is behaving like:

oled-dark = #000000
dark = #171717
light = #ffffff
her = #983724
system - dark = #171717
system - light = #ffffff

The theme color is set to the correct color both on page load if a theme is present in local storage, and updated dynamically if the user changes the theme in the OS and is using the system theme. Feel free to test it out and give some feedback if something is not behaving as expected👍

tjbck:
Thanks!

PR #5281, opened 2024-09-08
feat: persist functions self.valves to database during function create

ndrsfel:
# Changelog Entry

### Description

- Added a feature that adds `self.valves` during `Pipe.__init__()` to `function.valves` entry in the database and empty if not defined during init.
- This is especially useful, if you want to persist environment variable as valve values and follow best practice with functions valves.

---

### Additional Information

Replicate with minimal example

```python
import os
from typing import List, Union, Generator, Iterator

class Pipe:
    class Valves(BaseModel):
        API_KEY: str = Field(default="default-value")

    def __init__(self):
        self.type = "manifold"
        self.id = "my-id"
        self.name = "my-prefix/"
        self.valves = self.Valves(
            **{"API_KEY": os.getenv("API_KEY", "")}
        )

    def pipes(self) -> List[dict]:
        print(f"{self.valves=}") # this used to be "API_KEY=default-value" previously instead of API_KEY from env
        return {"id": "my-id", "name": "my-name"}
    
    def pipe(self, body: dict) -> Union[str, Generator, Iterator]:
        return body
```


tjbck:
Valves are initialised every time, best practice should be setting the default value as such:

`API_KEY: str = Field(default=os.getenv("API_KEY", "default-value"))`

Issue #5323, opened 2024-09-10
Bug: the default sbert model was trained on tweets and shouldn't be used for RAG


thiswillbeyourgithub:
# Bug Report

## Expected Behavior:
The default value for RAG should make sense: the chunk size and overlap should be compatible with the default model.

## Actual Behavior:

The default model is `sentence-transformers/all-MiniLM-L6v2`. Despite what its reputation seems to be, IMO it's a pretty terrible model for today's standard.

To quote from [their hf page](https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2) : 
- `Our model is intended to be used as a sentence and short paragraph encoder.`
- `The sequence length was limited to 128 tokens.`
- `By default, input text longer than 256 word pieces is truncated.`
- Also it's english only.

[I have](https://github.com/simonw/llm-sentence-transformers/issues/8#issue-1880545810) [been saying](https://github.com/langchain-ai/langchain/issues/2442#issuecomment-1601119941) [in multiple places](https://github.com/langchain-ai/langchain/issues/2026#issuecomment-1601331113) now that this shouldn't be used by default ever, as it was trained on basically short tweets and **silently clips any text above their limit**.

Now there are several ways to go about this:
1. You can simply drastically reduce the default chunk size. I don't think it's a good idea but at least it makes sense for the model.
2. if you want to keep an english only model I can suggest using [bge-small-en-v1.5](https://huggingface.co/BAAI/bge-small-en-v1.5). It's 130Mb, compared to the 90Mb of all-MiniLM-L6. I think it has 512 token limit (and a more compact tokenizer) but still an improvement. Otherwise there's also [another one from jina.ai](https://huggingface.co/jinaai/jina-embeddings-v2-base-en)
3. If you want to switch to a decent multilingual model there's [bge-m3](https://huggingface.co/BAAI/bge-m3/tree/main) but it's 2Go, or [jina's colbert model just out of the over](https://huggingface.co/jinaai/jina-colbert-v2/tree/main) that's 1Go. Each has 8192 token length support and more modern tokenizers.

FWIW my prefered suggestion would be to use a multilingual model (they also are the only one with that long context AFAIK), jina's colbert might be a sane default as it's not so big.

AFAIK all those models can be used by the sentence-transformers package or as hugging face embeddings. As you can see [here for example](https://www.sbert.net/examples/applications/computing-embeddings/README.html?highlight=bge)

[...]

Issue #5291, opened 2024-09-09
RAG not able to summarize uploaded file


tkg61:
# Bug Report

## Installation Method

Helm install of 0.3.21

**Confirmation:**

- [x] I have read and followed all the instructions provided in the README.md.
- [x] I am on the latest version of both Open WebUI and Ollama.
- [ ] I have included the browser console logs.
- [ ] I have included the Docker container logs.
- [x] I have provided the exact steps to reproduce the bug in the "Steps to Reproduce" section below.

## Expected Behavior:

When i ask for a summary of a file, it gives me a summary

## Actual Behavior:

It says it doesn't know what i am talking about 

## Description

**Bug Summary:**

the system is not recognizing that a file is uploaded and says it doesn't know what to do

## Reproduction Details

**Steps to Reproduce:**
Run Openwebui like normal
upload any file
ask it to summarize the file

## Logs and Screenshots

**Browser Console Logs:**
[Include relevant browser console logs, if applicable]

**Docker Container Logs:**
[Include relevant Docker container logs, if applicable]

**Screenshots/Screen Recordings (if applicable):**

![image](https://github.com/user-attachments/assets/11f7514e-929d-478b-89c5-c8d2dcff15f7)

## Additional Information

When i ask it a specific question it can answer it but the summarize query should work: 

![image](https://github.com/user-attachments/assets/237b25f6-3077-49fe-a0d1-2dc9e6a2bd49)

## Note

If the bug report is incomplete or does not follow the provided instructions, it may not be addressed. Please ensure that you have followed the steps outlined in the README.md and troubleshooting.md documents, and provide all necessary information for us to reproduce and address the issue. Thank you!

peuportier:
HI @tkg61  What model are you using for embedding ? can you provide us more detail about the configuration you are using concerning the embedding ? Did you have install tika , what your chunk parameter ... Please go on admin settings and Document to let us know. Thanks you

thiswillbeyourgithub:
Pretty sure it has something to do with https://github.com/open-webui/open-webui/discussions/5202#discussioncomment-10578585

tkg61:
I am using all the defaults in Openwebui (model, content extraction, rag template, etc). All of it is unmodified.

tkg61:
I did a little more experimentation. It seems like keeping the defaults but just adding in Tika did not solve anything.

[...]

Issue #5034, opened 2024-08-30
If empty [context] RAG should crash


thiswillbeyourgithub:
# Bug Report

## Installation Method

Docker compose

## Environment

- **Open WebUI Version:** 0.3.16

- **Operating System:** ubuntu 22
- **Browser (if applicable):** Brave

**Confirmation:**

- [ x ] I have read and followed all the instructions provided in the README.md.
- [ x ] I am on the latest version of both Open WebUI and Ollama.
- [ x ] I have included the browser console logs.
- [ x ] I have included the Docker container logs.
- [ x ] I have provided the exact steps to reproduce the bug in the "Steps to Reproduce" section below.

## Expected Behavior:

The expected behavior is that if the context is empty an error should happen.

## Actual Behavior:

No error happens and thr LLM just hallucinates answer.

## Description

**Bug Summary:**

It seems my RAG setup is broken. No matter what type of document I send, which browser I use etc looking at the traffic via langfuse shows me that the [context] is replaced by an empty string. The purpose of this issue is to add an exception when the context.strip() is empty.

## Reproduction Details

**Steps to Reproduce:**
Irrelevant.

Issue #4775, opened 2024-08-21
RAG on uploaded files / web search can only work for the first chat


foxworld306:
# Bug Report

## Installation Method

[Describe the method you used to install the project, e.g., git clone, Docker, pip, etc.]

## Environment

- **Open WebUI Version: 0.3.13
- **Ollama (if applicable): 0.3.5
- ** Web search tool: Searxng:latest

- **Operating System: Ubuntu 22.04
- **Browser (if applicable): Chrome, Edge

**Confirmation:**

- [x ] I have read and followed all the instructions provided in the README.md.
- [x ] I am on the latest version of both Open WebUI and Ollama.
- [ ] I have included the browser console logs.
- [ x] I have included the Docker container logs.
- [x ] I have provided the exact steps to reproduce the bug in the "Steps to Reproduce" section below.

## Expected Behavior:

- Continuous chat on file content
- Using web search multiple times in a conversation session

## Actual Behavior:

I found that in the Open Webui environment I deployed, it is not possible to continuously have RAG conversations with uploaded documents. It only works for the first conversation in the New Chat session I opened. From the second conversation onwards, it seems that I have never uploaded any documents. But when I look at the logs of Open Webui, I can see that the backend has already provided document retrieval results. It seems that starting from the second conversation, the retrieval results were not provided to LLM as expected, only LLM's knowledge was used to answer.
This issue also affects my online search operations. I must perform the search action in the first conversation of each New Chat session, otherwise starting from the second conversation, even if I can see Searxing returning the search results, I won't be able to get the correct answer.

I have already tried:

- Place the document in workspace ->documents
- Using different LLM models (Qwen2 and GLM4)
- reset vector storage
- Using or not using Hybrid search,
- Different Embedding Models
- Different Rerank models
- Using different web search methods, including searxng, google pse, duckduckgo, tavily.

All the above operations cannot solve this problem.

Can you please help to check?

## Logs and Screenshots
![image](https://github.com/user-attachments/assets/64128dcd-5e25-4a49-889c-299d412bbc64)
![image](https://github.com/user-attachments/assets/3b928152-c107-4aa2-a6ac-2532d3838e2b)
![image](https://github.com/user-attachments/assets/81fffbac-af4d-4047-ba2e-e0e9b2b9ccb2)

## Additional Information

[Include any additional details that may help in understanding and reproducing the issue. This could include specific configurations, error messages, or anything else relevant to the bug.]

## Note

[...]

PR #5378, opened 2024-09-12
fix: RAG and Web Search + RAG enhancements

thiswillbeyourgithub:
- **fix: log exception when issues of collection querying**
- **enh: use non hybrid search as fallback if hybrid search failed**
- **fix: much improved RAG template**
- **logs: crash if rag_template would be wrong**
- **fix: handle case where [query] happens in the RAG context**
- **fix: add check that the context for RAG is not empty if the threshold is 0**
- **log: added a debug log if detecting a potential prompt injection attack**
- **added a few type hints**

# Pull Request Checklist

- Linked the related discussions below + the fix is quite urgent IMO

- [ X ] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [ X ] **Description:** Provide a concise description of the changes made in this pull request.
- [ X ] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ X ] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [ X ] **Testing:** Have you written and run sufficient tests for validating the changes?
- [ X ] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [ X ] **Prefix:** To cleary categorize this pull request, prefix the pull request title, using one of the following:

---

# Body
*I wrote that first as a discussion message then it got out of hand so I did as best as I could to adhere to your format*

## Disclaimer
I'm extremely happy with Open-WebUI and very enthusiastic about its future and the future of pipelines. Hence, I was quite disappointed by some aspects of the code. I try to explain all of this below but want to clarify that I don't judge any of the maintainers: I can't even begin to imagine what you must have on your plate with a repo of 40k stars. But I still think I found some practices that will most certainly lose you more time than you could gain by coding quickly. I am totally open for discussion on any of the points I raised, I could totally be wrong. And part of all that might be attributed to misunderstandings about what RAG is supposed to do for the user, in which case understanding it here would help clarify the documentation. I will still continue using Open-WebUI and provide accounts for all my community, but want to voice my concerns. To reiterate: my current mindset is the one of a mostly happy user, concerned about some decisions and practices, wanting to do sort of a soft wake up call while remaining respectful of the well being of the maintainers.

## Context
RAG and web search were broken for at least weeks. Using langfuse I found out that the `[context]` in the RAG template was replaced by an empty string. I also found out that nothing in the code stops that even though I see no reason not to do that.

## The issue
- I've located my issue: it happens only when using hybrid search.
    - I was certain I had tried disabling it, but as I said, the lack of checks / logs makes the debugging process tedious and harder.
- The issue is in `backend/open_webui/apps/rag/utils.py`, in `query_collection_with_hybrid_search`
    - The query happens in a try block, but the exception is ignored.
    - In my case the exception was `Torch not compiled with CUDA enabled`. I don't know why this happened exactly but I's not the purpose of this PR.
- As a direct consequence: the context was never filled.
- The RAG template replacement happens without checking if that's alright.
- And finally the RAG template does not even tell the LLM to warn the user if the context seems broken or empty.
    - It even specifically tells the LLM to avoid mentioning it's using the context. This means that if you ask an open ended question to the LLM, hoping it would use the document, then it will completely hallucinate instead of being grounded by the document. This also makes it harder to get the LLM to confirm that the context was valid because it avoids mentioning it.

## Additional related issues in this PR
- if "[query]" is present in the context, the template replacement will be wrong as it will replace it too. I used uuid (standard python lib) to generate a seed in that case.
- There's still an issue with injection that will happen if `<context>` or `</context>` is present inside the context. I don't know exactly how you want to deal with that. I added a warning anyway. This is NOT to take lightly as pipelines will end up exposing more and more APIs and attackers will always have the edge.

## Other remarks
- At least the function "get_rag_context" is not typed (arguments nor outputs)
- In a general way, at least for the rag utils file : only 3 functions out of 14 have an output type, making it all the more harder for the community to dive on specific issue like I'm doing right now.
    - If you want to add very fast optional type checking I can strongly recommend [beartype](https://github.com/beartype/beartype). In one of my project I've been using it by decorating pretty much all my functions with a beartype decorator that either crash or warns in case of inconsistencies (depending on an env variable). It's been a huge headache saver and I think you should try it. [Here's my code as an example](https://github.com/thiswillbeyourgithub/WDoc/blob/main/WDoc/utils/typechecker.py). Without the env parameter, the decorator directly returns the function so there is **no** overhead. If you set the env parameter then the overhead is minimal, thanks to beartype.
- As I modified the RAG template, it has to be modified in the docs repository too.
- When debugging, I simply added asserts and expected it to trigger red banners, but was surprised to see that it didn't. I strongly suggest adding an env variable or a UI parameter to allow at least admin accounts to see debugging banners instead of only showing fatal errors. I would not be surprised if there were more bugs that users didn't even know about.

## Commits in this PR
- Properly log exceptions occurring during querying (both hybrid and not)
- If all collections failed to be queried in hybrid search: fallback to using non hybrid search
- Improved the RAG template, especially important is the mention that if the information is not coming from the context, the LLM should say so.
- Added sanity checks for the RAG template replacement.
- Fixes cases where "[query]" is present in the context
- If the relevancy factor is 0, the context should never be empty. I added a check for this.
- Added a prompt injection warning of "<context>" and "</context>" are present in the context. This is NOT to take lightly as pipelines will end up exposing more and more APIs and attackers will always have the edge.
- Added a few type hints.

### What this PR fixes
- My issue at #5034
- The ensuing discussion at #5050

### It probably fixes too:
- #5002 and #5022 by @qiulang
- #5291 by @tkg61
- #3363 by @hugobox with @changchiyou and @Rom0067 and @anonymousmaharaj and @devingfx

### It might also fix
- #3518 by @Robinsane
- #4775 by @foxworld306 and #4779 with @GridexX
- #3868 with @gaetanquentin and @BennisonDevadoss
- #176 of @bhaidar and @dmvieira 

## Closing remarks
I'll be blunt but honest: I think that the relative lack of asserts, the fact that exceptions are not shown to the admin, the relative lack of type hints, the apparent lack of test coverage, will progressively increase the burden of developing Open-WebUI as well as make it harder for the community to provide helpful issues, as well as step in and help. I think RAG might have been silently broken for weeks, but maybe many other things are broken without us knowing about it.

Open-WebUI and Pipelines are ambitious, and you need all the help modern developpement practices can bring you!

I now have reasons to doubt that the rest of the codebase is properly checking things, or that anything that breaks is logged.

I think it would be awesome if you added my suggestion regarding showing errors to the admin, then progressively added more asserts as you naturally refactor the code here and there. This might be the most efficient way to go about this as I understand that increasing the test coverage for example requires a significant effort. Also beartype typechecking during the developpement phase or as an opt in for users might be very helpful. (But do use their latest release candidate instead of the latest release.)

Then again, thank you again for making Open-WebUI, it's a great improvement for me and my community. Have a nice day.

PS: While I have your attention, could you share your thoughts about #5323 ? I explained in there that the default RAG model was trained on tweets and silently clips after a few tokens but Open-WebUI's default is still to use that with a 1500 chunk size.

---

# Changelog Entry

### Description

RAG and web search were broken for me and I think many others. The fix was to add a fallback to the non hybrid search. While I was at it I saw numerous things I decided to patch, all still related to this issue.

### Added

- If all collections failed to be queried in hybrid search: fallback to using non hybrid search
- Added sanity checks for the RAG template replacement.
- Added a few type hints.

### Fixed

- Properly log exceptions occurring during querying (both hybrid and not)
- Improved the RAG template, especially important is the mention that if the information is not coming from the context, the LLM should say so.
- Fixes cases where "[query]" is present in the context
- If the relevancy factor is 0, the context should never be empty. I added a check for this.

### Security

- Added a prompt injection warning of "<context>" and "</context>" are present in the context. This is NOT to take lightly as pipelines will end up exposing more and more APIs and attackers will always have the edge.

### DIFF
--- a/backend/open_webui/apps/rag/utils.py
+++ b/backend/open_webui/apps/rag/utils.py
@@ -150,8 +150,8 @@ def query_collection(
                     embedding_function=embedding_function,
                 )
                 results.append(result)
-            except Exception:
-                pass
+            except Exception as e:
+                log.exception(f"Error when querying the collection: {e}")
         else:
             pass
 
@@ -178,8 +178,11 @@ def query_collection_with_hybrid_search(
                 r=r,
             )
             results.append(result)
-        except Exception:
-            pass
+        except Exception as e:
+            log.exception(
+                "Error when querying the collection with "
+                f"hybrid_search: {e}"
+            )
     return merge_and_sort_query_results(results, k=k, reverse=True)
 
 
--- a/backend/open_webui/apps/rag/utils.py
+++ b/backend/open_webui/apps/rag/utils.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import uuid
 from typing import Optional, Union
 
 import requests
@@ -197,8 +198,15 @@ def rag_template(template: str, context: str, query: str):
         f"RAG template contains an unexpected number of '[context]' : {count}"
     )
     assert "[context]" in template, "RAG template does not contain '[context]'"
-    template = template.replace("[context]", context)
-    template = template.replace("[query]", query)
+
+    if "[query]" in context:
+        query_placeholder = str(uuid.uuid4())
+        template = template.replace("[QUERY]", query_placeholder)
+        template = template.replace("[context]", context)
+        template = template.replace(query_placeholder, query)
+    else:
+        template = template.replace("[context]", context)
+        template = template.replace("[query]", query)
     return template
 
 
--- a/backend/open_webui/apps/rag/utils.py
+++ b/backend/open_webui/apps/rag/utils.py
@@ -198,6 +198,12 @@ def rag_template(template: str, context: str, query: str):
         f"RAG template contains an unexpected number of '[context]' : {count}"
     )
     assert "[context]" in template, "RAG template does not contain '[context]'"
+    if "<context>" in context and "</context>" in context:
+        log.debug(
+            "WARNING: Potential prompt injection attack: the RAG "
+            "context contains '<context>' and '</context>'. This might be "
+            "nothing, or the user might be trying to hack something."
+        )
 
     if "[query]" in context:
         query_placeholder = str(uuid.uuid4())
--- a/backend/open_webui/apps/rag/utils.py
+++ b/backend/open_webui/apps/rag/utils.py
@@ -48,7 +48,7 @@ def query_doc_with_hybrid_search(
     k: int,
     reranking_function,
     r: float,
-):
+) -> dict:
     try:
         collection = CHROMA_CLIENT.get_collection(name=collection_name)
         documents = collection.get()  # get all documents
@@ -93,7 +93,7 @@ def query_doc_with_hybrid_search(
         raise e
 
 
-def merge_and_sort_query_results(query_results, k, reverse=False):
+def merge_and_sort_query_results(query_results: list[dict], k: int, reverse: bool = False) -> list[dict]:
     # Initialize lists to store combined data
     combined_distances = []
     combined_documents = []
@@ -139,7 +139,7 @@ def query_collection(
     query: str,
     embedding_function,
     k: int,
-):
+) -> dict:
     results = []
     for collection_name in collection_names:
         if collection_name:
@@ -166,7 +166,7 @@ def query_collection_with_hybrid_search(
     k: int,
     reranking_function,
     r: float,
-):
+) -> dict:
     results = []
     failed = 0
     for collection_name in collection_names:
--- a/backend/open_webui/apps/rag/utils.py
+++ b/backend/open_webui/apps/rag/utils.py
@@ -167,6 +167,7 @@ def query_collection_with_hybrid_search(
     r: float,
 ):
     results = []
+    failed = 0
     for collection_name in collection_names:
         try:
             result = query_doc_with_hybrid_search(
@@ -183,6 +184,10 @@ def query_collection_with_hybrid_search(
                 "Error when querying the collection with "
                 f"hybrid_search: {e}"
             )
+            failed += 1
+    if failed == len(collection_names):
+        raise Exception("Hybrid search failed for all collections. Using "
+                        "Non hybrid search as fallback.")
     return merge_and_sort_query_results(results, k=k, reverse=True)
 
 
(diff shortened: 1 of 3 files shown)


backend/open_webui/apps/rag/utils.py
```diff
@@ -262,19 +289,25 @@ def get_rag_context(
...
+                            collection_names=collection_names,
+                            query=query,
+                            embedding_function=embedding_function,
+                            k=k,
+                            reranking_function=reranking_function,
+                            r=r,
+                        )
+                    except Exception as e:
+                        log.debug("Error when using hybrid search, using"
+                                    " non hybrid search as fallback.")
```
dmvieira, line 308: Idk if log debug is better when sending exception
```suggestion
                        log.exception(e)
```

thiswillbeyourgithub: Thanks for taking a look.

I decided to use debug because what's happening is
1. `query_doc_with_hybrid_search` crashes
2. we log it with `log.exception` at line 184
3. Then, only if all the collection failed
4. We raise an exception
5. That is caught at line 307
6. And shown with log.debug.

The purpose of this 2nd exception is to tell `get_rag_context` to fallback to using non hybrid search. It's not *really* and exception, especially compared to the exceptions caught at line 184.

Hence my reasoning was that the fallback is caused by an exception but we warn that we're using the fallback with a debug.

In all cases I'm not too hung up on this and can change it if anyone has a strong feeling about this.

tjbck:
Thanks!

qiulang:
@thiswillbeyourgithub I use the latest dev code, which contains your fix to see whether RAG works now.

I don't know whether I use it correctly or not, I find there is no difference and RAG still fails to work in my case. I have to use # to specify the exact document then RAG can give me the answer, otherwise RAG's answer is just nonsense.

I also try to add and remove documents to my model's **Knowledge**, but it makes no difference.

BTW, I also see this comments to https://github.com/open-webui/open-webui/discussions/5460, how you will see them.

PR #5511, opened 2024-09-19
fix: restore `__file__` variable for imported toolkits and functions

EtiennePerot:
Commit cf86ba778605bf0e7494817e83dbf940c0086277 changed the way toolkits and functions were imported to use `exec`, whereas they previously were written to files and `import`ed. The use of `exec` means that module-global variables such as `__file__` are no longer defined.

This change breaks [this code execution tool for Open WebUI](https://github.com/EtiennePerot/open-webui-code-execution) (issue: https://github.com/EtiennePerot/open-webui-code-execution/issues/11), as the module needs to re-execute its own code in a subprocess in order to properly sandbox itself. This is done using `__file__` in order to know where the module's code is located.

This PR creates a temporary in-memory file that contains the imported toolkit or function's code and exists only during the import process. Then it injects the path to this in-memory file as the `__file__` variable in the `exec` context. This restores the ability for the toolkit or function being imported to rely on `__file__`.

# Pull Request Checklist

- [x] **First-time contributors**: Created a discussion about the change: https://github.com/open-webui/open-webui/discussions/5510
- [x] **Target branch:** `dev` branch.
- [x] **Description:** See above.
- [x] **Changelog:** See below.
- [x] **Documentation:** No relevant documentation to update.
- [x] **Dependencies:** None.
- [x] **Testing:** Tested manually. Fairly trivial change.
- [x] **Code review:** Done.
- [x] **Prefix:** `fix`.

# Changelog Entry

### Fixed

- Restored toolkits and functions' ability to use the `__file__` variable to refer to their own code.

### Screenshots or Videos

Without this fix, tools and functions do not have the `__file__` variable defined.

### DIFF
--- a/backend/open_webui/apps/webui/utils.py
+++ b/backend/open_webui/apps/webui/utils.py
@@ -84,7 +84,13 @@ def load_toolkit_module_by_id(toolkit_id, content=None):
     module = types.ModuleType(module_name)
     sys.modules[module_name] = module
 
+    # Create a temporary in-memory file and use it to define `__file__` so
+    # that it works as expected from the module's perspective.
+    temp_fd = os.memfd_create(f"tmp:{module_name}")
     try:
+        os.write(temp_fd, content.encode("utf-8"))
+        module.__dict__["__file__"] = f"/proc/{os.getpid()}/fd/{temp_fd}"
+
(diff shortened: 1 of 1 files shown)


tjbck:
LGTM, Thanks!

Issue #5544, opened 2024-09-20
Oauth error after upgrading from 0.3.16 to any newer version


itaybar:
I am using openweb ui as docker image on k8s  env and usign an OPENID provider (ADFS), it all worked well in v0.3.16, but when I am upgraded to version 0.3.18 or 0.3.22 (didn't try any other versions), the login failed with error `invalid_request MSIS9692: Recived invalid OAuth request. the basic Authorization header must have ClientID and ClientSecret UrlEncoded then seperated by ':'`.

Off course I didnt changed anything except the Openweb ui version.

I didn't saw anything in the code changed in this areas but maybe I am missing somthing

itaybar:
Maybe the changing the authlib version 1.3.1 to 1.3.2 
![image](https://github.com/user-attachments/assets/4d161a51-f7da-4d20-87fb-09a3aae7a436)
which talks about changing something the headers

Legalcomplex:
@itaybar curious: how did you setup ADFS as a OPENID provider and configure OIDC on Open WebUI

PR #5546, opened 2024-09-20
downgrade authlib to 1.3.1

itaybar:
#5544 fixing issue that crated due to bug in authlib 1.3.2 that prevent using any client id that have ":" in it


tjbck:
Is this PR tested?

itaybar:
Tested and worked perfectly

tjbck:
Great, thanks!

PR #5578, opened 2024-09-21

aleixdorca:
# Changelog Entry

### Description

Updated again, with new strings.


tjbck:
Thanks!

PR #5599, opened 2024-09-22
0.3.24

### LINKED ISSUE #3620: enh/issue: performance optimisation
The performance of long chat threads is really bad. I don't know if it is a browser client side issue or an issue on the server side.

A performance fix would be great, but even the ability to do a shallow clone where only the last context window or so of the chat is cloned would help.

I am resorting to starting a fresh chat with memory items added and trying to use the prompt to pick up where the last chat left off.

**Additional context**
I have tried a few different browsers, but the best fix so far has been to start a new chat and try my best to get it back on roughly the same track.

tjbck:
Resolves #3620


itaybar (commented):
You can use the redis dict ttl functionality instead of the periodic deletion if you are in redis mode (this would also prevent wierd situations that two instances tries to delete the same key (which throw exception i guess), and for the local dictionary there is no need for that just like it didn’t exist until now

tjbck:
I'm aware, we won't be using an external dependency.

PR #5614, opened 2024-09-22
Add Python 3.11 venv and Open-WebUI setup scripts for Windows and Unix-like systems

mikeldev0:
This change adds two new setup scripts to facilitate the creation of a virtual environment with Python 3.11 and installation of Open-WebUI on both Windows and Unix-like systems.

The first script, written in bash, checks if Python 3.11 is installed, creates a venv environment, activates it, installs Open-WebUI, and runs the server.

The second script, written in batch for Windows, performs similar tasks using PowerShell to interact with the system.

With this change, users can easily set up their project environments with Python 3.11 and Open-WebUI, regardless of their operating system.

**Changes made:**

* Added bash script for Unix-like systems to create venv environment with Python 3.11 and install Open-WebUI
* Added batch script for Windows to create venv environment with Python 3.11 and install Open-WebUI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -307,3 +307,4 @@ dist
 cypress/videos
 cypress/screenshots
 .vscode/settings.json
+myenv
--- a/run-pip.bat
+++ b/run-pip.bat
@@ -1,0 +1,28 @@
+@echo off
+
+:: Check if Python 3.11 is installed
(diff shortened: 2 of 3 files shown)


tjbck:
Closing this but feel free to add this to our docs!

PR #5632, opened 2024-09-23
enh: open PDF citations on the associated page

jannikstdl:
### Description

- If you use PDFs, the file is now opened on the right page of the text snippet citation on supported browsers.
- Styling changes to clarify to the user that the link is clickable.
- Additional "tippyOptions" property to add specific properties to the Tooltip (needed to remove the off-mount delay for scrolling containers).
- "page" is now a i18n key
- German locales

### Added

- **📄 PDF Citation Support & UI Enhancements**: PDFs now open directly at the referenced page in supported browsers, improving navigation. Additionally, clickable links in the citation modal have updated styling to make them more visually clear and user-friendly.


tjbck:
Thanks!

Issue #4116, opened 2024-07-25
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory


windowshopr:
# Bug Report

## Description
**Bug Summary:**
Latest git repository as of July, 2024, running `docker compose up --pull always --build` in PowerShell of Windows 10, running Docker Desktop, results in the following traceback:

```
115.4 ✓ 2861 modules transformed.
121.5 rendering chunks...
143.4
143.4 <--- Last few GCs --->
143.4
143.4 [16:0x7fe77847d030]   104087 ms: Mark-Compact (reduce) 1948.6 (1994.4) -> 1948.6 (1994.9) MB, 1332.58 / 0.00 ms  (+ 45.0 ms in 0 steps since start of marking, biggest step 0.0 ms, walltime since start of marking 1404 ms) (average mu = 0.249, current mu = [16:0x7fe77847d030]   106348 ms: Mark-Compact (reduce) 1948.9 (1994.9) -> 1948.8 (1995.1) MB, 2017.23 / 0.00 ms  (average mu = 0.167, current mu = 0.108) allocation failure; GC in old space requested
143.4
143.4
143.4 <--- JS stacktrace --->
143.4
143.4 FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
143.4 ----- Native stack trace -----
143.4
144.4 npm ERR! path /app
144.4 npm ERR! command failed
144.4 npm ERR! signal SIGABRT
144.4 npm ERR! command sh -c npm run pyodide:fetch && vite build
144.4
144.4 npm ERR! A complete log of this run can be found in: /root/.npm/_logs/2024-07-25T17_53_55_234Z-debug-0.log
------
failed to solve: process "/bin/sh -c npm run build" did not complete successfully: exit code: 1
```

**Steps to Reproduce:**
`git clone https://github.com/open-webui/open-webui.git`
`cd` into the directory
`docker compose up --pull always --build` with Docker Desktop running

**Expected Behavior:**
Container should build and run without error.

**Actual Behavior:**
[Describe what actually happened.]

## Environment

- **Open WebUI Version:** 0.3.10
- **Ollama (if applicable):** 0.3.0 Windows Preview

- **Operating System:** Windows 10
- **Browser (if applicable):** Chrome

I tried increasing the VMEMM limit from 2Gb to 4Gb in the .yaml file, however same issue.

justinh-rahb:
[...]

PR #5639, opened 2024-09-23
build: Allow node to use more RAM during build

sebdanielsson:
# Changelog Entry

### Description

Allow node to use more RAM during the npm run build step. I'm developing on an M1 Macbook Air with 16 GB of RAM. I've configured Orbstack which I use instead of Docker Desktop to let Docker use 12 GB of RAM. When building the Dockerfile Node only use around 2096 MB of RAM during the npm build step. After some troubleshooting I increased the limit to 4096 MB and the build completed successfully.

Not sure how this value is calculated or if it's fixed for everyone but you can check it with:

```sh
docker run --rm node:21-alpine3.19 node -e "console.log(v8.getHeapStatistics().heap_size_limit / 1024 / 1024 + ' MB');"
```

### Changed

- Increase the amount of RAM node can use during build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,7 @@ RUN npm ci
 
 COPY . .
 ENV APP_BUILD_HASH=${BUILD_HASH}
+ENV NODE_OPTIONS=--max-old-space-size=4096
 RUN npm run build
 
 ######## WebUI backend ########


sebdanielsson:
This change would have probably solved #4116.

tjbck:
Unsure if we have to modify our Dockerfile here, perhaps adding this info to our docs would be a better fit?

sebdanielsson:
> Unsure if we have to modify our Dockerfile here, perhaps adding this info to our docs would be a better fit?

Out of curiosity, what value do you get when running:

```sh
docker run --rm node:21-alpine3.19 node -e "console.log(v8.getHeapStatistics().heap_size_limit / 1024 / 1024 + ' MB');"
```

Just want to know how this value is set, whether it's dynamically set depending on the dev environment or statically set.

Yeah I could all this to the documentation too.

Peter-De-Ath:
> > Unsure if we have to modify our Dockerfile here, perhaps adding this info to our docs would be a better fit?
> 
> Out of curiosity, what value do you get when running:
> 
> ```shell
> docker run --rm node:21-alpine3.19 node -e "console.log(v8.getHeapStatistics().heap_size_limit / 1024 / 1024 + ' MB');"
> ```
> 
> Just want to know how this value is set, whether it's dynamically set depending on the dev environment or statically set.
> 
> Yeah I could all this to the documentation too.

Running on linux host with 48GB RAM

Result: 2096 MB

sebdanielsson:
> > > Unsure if we have to modify our Dockerfile here, perhaps adding this info to our docs would be a better fit?
> > 
> > 
> > Out of curiosity, what value do you get when running:
> > ```shell
> > docker run --rm node:21-alpine3.19 node -e "console.log(v8.getHeapStatistics().heap_size_limit / 1024 / 1024 + ' MB');"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Just want to know how this value is set, whether it's dynamically set depending on the dev environment or statically set.
> > Yeah I could all this to the documentation too.
> 
> Running on linux host with 48GB RAM
> 
> Result: 2096 MB

Interesting. Are you able to complete a Docker build with that amount?

tjbck:
I'm generally okay with merging this to dev if there won't be any issues with the ghrc image building process, were you able to test this part?

sebdanielsson:
> I'm generally okay with merging this to dev if there won't be any issues with the ghrc image building process, were you able to test this part?

Yeah it build fine on GitHub Actions. But speaking of CI/CD, I'm not fully sure if this will affect more powerful computers negatively (like CI/CD) because of this value being hardcoded as an environment variable. So as you said, might be more safe to explain in the docs?

tjbck:
Agreed, let's put it in our docs for now. Thanks!

PR #5735, opened 2024-09-26
Fix: default configuration in README.md

aamsur-933:
I just installed ollama in my own server and interested with this project. I got a little bit confused with the readme.

I just though that the Nvidia GPU Support command is not related with the command on its above, so i started create this PR.

Big thanks, great project!


aamsur-933:
So let's the confusion spread to others ...

PR #5829, opened 2024-09-30
fix:  performance issues on large collections 

jannikstdl:
# Pull Request Checklist

- Fixes `embedding_function `to be called every single time when the `query_doc `for loop is executed.
- Fixes load times in large collection wich causes the retrieval to sometimes take up to minutes


tjbck:
Thanks!

Issue #3527, opened 2024-06-29
feat: knowledge/projects (revamped documents section)


tjbck:
#3521 #3231 #2044 #3110 #3464 #1096

- [x] Refactor Collection handling -> Rename it to project
- [x] Deprecate current tags in favour of project groupping
- [x] Each project should be able to handle multiple files (save all embeddings to same collection)
- [x] Should be team-agnostic and should allow inter-team project use case
- [x] Project notes (#3231)
- [ ] Provide clear guidance on API usage
 - Upload file via Files API
 - Process and link file to project(s) using file id
 - Chat with Files using project id(s) in chat completion payload

joelmnz:
In addition to my suggestion in #3231 , have seen what Claude has done with their "agent workspaces" (a space/doc an agent can read & write to), that is a great idea, what if we could get our agents/chats to have read-write memory/notes. I have seen the concept of "memory" in Open WebUI but haven't found docs or how to use it. Could that be incorporated, maybe as a separate ticket but considered here.

If this is done already or in some other way please let me know.

antonio-lamberti:
Hi! Does anybody now if the upload docs (as in the workspace - documents) is available in the API? #5886

antonio-lamberti:
> @antonio-lamberti Existing documents section has been deprecated in dev, but detailed knowledge API usage tutorial will be provided shortly!

Thanks! So does this mean that right now, the upload via API feature it's not working, right? I saw other people [complaining ](https://github.com/open-webui/open-webui/discussions/4692) the same...

Also, I would know the API usage tutorial will be updated when the new version is released, right?

Thanks for the great job!!

okaeiz:
Several issues has been closed referencing this one as the original post. I am experiencing extremely slog RAG on UI. I upload a relatively large Markdown file and it takes too long to get uploaded and saved to the vector database. Is this the normal behavior? I am not experiencing this when working with PDFs.

jnko266:
@okaeiz I am experiencing the exact same - my OWUI RAG only has 1 CSV file and 1 (quite small) Markdown file. Lookups take 60+ seconds per question. Anyone has any ideas?

shc443:
@jnko266 Same issue! did you figure it out why?? =[

PR #5861, opened 2024-10-01
feat: knowledge/projects

tjbck:
#3527


src/lib/components/workspace/Knowledge.svelte
```diff
@@ -0,0 +1,175 @@
...
+	import Pencil from '../icons/Pencil.svelte';
+	import DeleteConfirmDialog from '../common/ConfirmDialog.svelte';
+	import ItemMenu from './Knowledge/ItemMenu.svelte';
+
+	let query = '';
+	let selectedItem = null;
+	let showDeleteConfirm = false;
+
+	let filteredItems;
+	$: filteredItems = $knowledge.filter((item) => query === '' || item.name.includes(query));
```
jannikstdl, line 28: Knowledge Search is case sensitive. This would be a fix + i added the description field for search

```
	$: filteredItems = $knowledge.filter(
		(item) =>
			query === '' ||
			item.name.toLowerCase().includes(query.toLowerCase()) ||
			item.description.toLowerCase().includes(query.toLowerCase())
	);

```

PR #5864, opened 2024-10-01
build: Remove bun

sebdanielsson:
# Changelog Entry

### Description

The bun lock file is outdated. If there are still compatibility issues, maybe we should keep it in a separate branch until npm can be fully replaced with bun?

### Removed

- Removed outdated bun.lockb


tjbck:
Thanks!

PR #5918, opened 2024-10-05
refac: remove docker warnings

nowheresly:
# Changelog Entry

Remove docker warnings to ensure compatibility with future versions of Docker

### Description

- The official syntax for Dockerfile is described [here](https://docs.docker.com/reference/dockerfile/#env)
- ENV must use '='
- AS must be in upper case


tjbck:
Please rebase and target our dev branch.

Nowheresly:
done

PR #5920, opened 2024-10-05
fix: ensure Dockerfile and github actions use the same nodejs version

nowheresly:
nodejs version from github actions and the Dockerfile should be the same

# Changelog Entry

Use nodejs 21. Ensure github actions and docker use the same nodejs version

### DIFF
--- a/.github/workflows/format-build-frontend.yaml
+++ b/.github/workflows/format-build-frontend.yaml
@@ -21,7 +21,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: '20' # Or specify any other version you want to use
+          node-version: '21' # Or specify any other version you want to use
 
       - name: Install Dependencies
         run: npm install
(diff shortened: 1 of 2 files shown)


Nowheresly:
I could change this PR to stick with nodejs 21 if you prefer

Nowheresly:
In order to run with node 22, cypress has to be updated as explained [here](https://docs.cypress.io/guides/references/changelog#13-14-0)

tjbck:
Rationale behind this PR?

Nowheresly:
> Rationale behind this PR?

Building and running the test with node 20 in github actions while using node 21 to build in the docker image may lead to discrepancies since the tests may succeed while the production docker image may fail because they don't use the same node version

tjbck:
Makes sense, Thanks!

Nowheresly:
the cypress integration tests seem to have issues with node 21.
Maybe should we rollback to node 20
(after all node 20 is LTS, not node 21. And node 22 is not yet LTS)

PR #5926, opened 2024-10-05
Update to node 22

nowheresly:
https://github.com/cypress-io/github-action/pull/1189

# Changelog Entry

Use nodejs 22


Nowheresly:
it seems that cypress has remove support for nodejs 21[here](https://github.com/cypress-io/github-action/pull/1189)

Nowheresly:
it seems to work now using node 22

tjbck:
Thanks!

PR #5955, opened 2024-10-07
feat: add code execution status to chat messages.

EtiennePerot:
This PR adds `code_executions` as an array of code execution statuses to chat messages. The intent of this data is to be displayed in a similar manner as citations: at the bottom of the message, with buttons that open a modal for more info. As a point of reference, here is how similar data looks in ChatGPT when it executes code:

**ChatGPT code execution status**:

Screenshot from 2024-09-10 23-08-47

**ChatGPT code execution modal**:

Screenshot from 2024-09-10 23-09-06

**Why not use the existing "citations" modals?** Code execution data doesn't fit well in that, because it fundamentally differs in its formatting. Code execution status includes the code that was run (which benefits from being syntax-highlighted), and the output and generated files. This differs from citations which are just list of document names and links.

Additionally, code execution is a process, whereas citations are only emitted once. This is why code execution data uses an ID-based approach, where each code execution instance is identified by a unique ID and can be updated by emitting a new `code_execution` message with the same ID. This allows the code execution status to be updated as code runs.

# Pull Request Checklist

- [x] **Description:** See above.
- [x] **Changelog:** OK.
- [x] **Documentation:** No documentation to update.
- [x] **Testing:** Manually tested to produce buttons similar to citations at the bottom of messages.
- [x] **Code review:** OK.
- [x] **Prefix:** `feat`

# Changelog Entry

 ### Description

Citation messages can now be used to convey code execution status.

### Added

* New type of citation to display code execution status.

### Screenshots or Videos

Code execution video

### DIFF
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -175,10 +175,30 @@
 					message.statusHistory = [data];
 				}
 			} else if (type === 'citation') {
-				if (message?.citations) {
-					message.citations.push(data);
+				if (data?.type === 'code_execution') {
+					// Code execution; update existing code execution by UUID,
+					// otherwise append.
+					if (!message?.code_executions) {
+						message.code_executions = [];
+					}
+					let is_update = false;
+					for (let i = 0; i < message.code_executions.length; i++) {
+						if (message.code_executions[i].uuid === data.uuid) {
+							message.code_executions[i] = data;
+							is_update = true;
+							break;
+						}
+					}
+					if (!is_update) {
+						message.code_executions.push(data);
+					}
 				} else {
-					message.citations = [data];
+					// Regular citation.
+					if (message?.citations) {
+						message.citations.push(data);
+					} else {
+						message.citations = [data];
+					}
 				}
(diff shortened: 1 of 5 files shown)


src/lib/components/chat/Chat.svelte
```diff
@@ -179,6 +179,21 @@
 				} else {
 					message.citations = [data];
 				}
+			} else if (type === 'code_execution') {
```
tjbck, line 182: Perhaps it's better to use the existing 'citations' event type and add a new 'code_execution' sub data type here.

EtiennePerot: Done.

EtiennePerot:
PR updated to include the modal dialog for code execution status. Also added video of the feature.

tjbck:
LGTM, Thanks!

tjbck:
@EtiennePerot did some refac (notably uuid -> id, but mostly styling changes), testing wanted with dev here!

EtiennePerot:
Looks great and works great, thanks! Tested from `dev`.

PR #5977, opened 2024-10-07
perf: Introduce offline mode variable for Air-gapped environment and disable version check 

saravanan30erd:
# Pull Request Checklist

Fix: https://github.com/open-webui/open-webui/discussions/5759

### Description

Open WebUI having performance issues in offline (air-gapped) environment. When visiting the site, the API ( /api/version/updates ) will check the version (current & latest) information through the Internet. If its air-gapped environment then It takes around more than 2 mins to load the website (basically after the connection timeout, it uses current version as latest version)

### Added

Introduced offline mode variable for Air-gapped environment and disabled version check from external source.


gaby:
@saravanan30erd @tjbck Please re-open this. This is much needed

saravanan30erd:
@gaby will do it.

saravanan30erd:
@tjbck I have opened PR https://github.com/open-webui/open-webui/pull/6003 after rebase. Please check it, thanks.

PR #6003, opened 2024-10-08
add offline_mode and disable version check

saravanan30erd:
Fix: https://github.com/open-webui/open-webui/discussions/5759 https://github.com/open-webui/open-webui/pull/5977

### Description

Open WebUI having performance issues in offline (air-gapped) environment. When visiting the site, the API ( /api/version/updates ) will check the version (current & latest) information through the Internet. If its air-gapped environment then It takes around more than 2 mins to load the website (basically after the connection timeout, it uses current version as latest version)

### Added

Introduced offline mode variable for Air-gapped environment and disabled version check from external source.


tjbck:
Thanks!

PR #6104, opened 2024-10-10
fix: ollama embed support list[str] or str on input

Peter-De-Ath:
### Fixed
/ollama/api/embed

{"detail":[{"type":"list_type","loc":["body","input"],"msg":"Input should be a valid list","input":"some text"}]}

should support both list of string and a string.

always returns `{"embeddings: [[]]}` as direct to ollama


tjbck:
Thanks!

PR #6122, opened 2024-10-11
perf: Improve Speed of Document Insertion in Knowledge Bases

RobinBially:
## Description
This pull request optimizes the process of inserting documents into knowledge bases by eliminating redundant embedding operations. Previously, embedding was performed twice: once for the file and again for the collection. Now, the vectors from the file embedding are reused, reducing the overall processing time. This optimization has been implemented for Qdrant at this stage.

### Changed
- Optimized document insertion process for knowledge bases
- Reduced embedding operations from two to one by reusing file vectors
- Implemented optimization for Qdrant

### Performance
- Improved speed of document insertion in knowledge bases

## Additional Information
This change significantly enhances the efficiency of document insertion, particularly for large files or when dealing with a high volume of documents. The optimization is currently limited to Qdrant, but may be extended to other vector databases in future updates.

## Checklist
- [x] Verified that the pull request targets the dev branch
- [x] Provided a concise description of the changes
- [x] Added a changelog entry following the Keep a Changelog format
- [ ] Updated relevant documentation (if applicable)
- [x] No new dependencies added
- [x] Performed self-review of the code
- [x] Ensured adherence to the project's coding standards

## Testing
- [x] Written and run tests to validate the changes
- [x] Verified improved performance with various document sizes and quantities

## Additional Notes
-

PR #6137, opened 2024-10-12
perf: Improve Speed of Document Insertion in Knowledge Bases

RobinBially:
reopening #6122 

- added support for chroma and milvus
- refactoring
- rebased on dev

### DIFF
--- a/src/lib/components/chat/Messages/UserMessage.svelte
+++ b/src/lib/components/chat/Messages/UserMessage.svelte
@@ -16,6 +16,7 @@
 	import Tooltip from '$lib/components/common/Tooltip.svelte';
 	import FileItem from '$lib/components/common/FileItem.svelte';
 	import Markdown from './Markdown.svelte';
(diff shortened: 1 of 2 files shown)


RobinBially:
rebased again.

Is there anything else preventing this from being merged? @tjbck

tjbck:
Generally unsure if I'd want to merge at this point, later down the line I might add more configurable embedding options for knowledge collections (e.g. different chunking approaches for different collections).

RobinBially:
This is what I'm also doing right now. I'm planning to integrate collections with multimodal content. You can still have that functionality. Additionally, this is not a small performance gain. When using larger embedding models or if you want to use the reindexing feature that I'm planning to submit, this would be the difference between, for example, 1 hour and 2 hours of embedding time.

tjbck:
PR generally looks good, I just need to work on other stuff first 🙌

PR #6157, opened 2024-10-13
Increase asyncio timeout to 30 sec for openai

pimdo:
Fix for 

https://github.com/open-webui/open-webui/discussions/6069


tjbck:
This will cause issues with the UI being unresponsive when the server is unreachable. If you have any other fixes feel free to make a PR.

pimdo:
think unreachable server can be solved by an initial `/ping` or `/health` check.
for the eager loading of models, maybe can lazy load models, don't know much about the client side though.

I will use older version as I noticed another issue now:
 horizontal scroll issue: need to scroll left and right to see the history or chat details

PR #6293, opened 2024-10-20
fix: don't overwrite source.name with an URL, keep name so user can decide what should be visible

RocketRider:
I am setting up an external RAG system with URLs. But I don't want teh full URL shown in the Chat message.
In my opinion it is better not to overwrite the name. So the user can set it to the URL or a name, but when it is overwritten we can't set a custom name.
image

In addition, when removing the overwrite it will be handled as in the detail dialog.

### DIFF
--- a/src/lib/components/chat/Messages/Citations.svelte
+++ b/src/lib/components/chat/Messages/Citations.svelte
@@ -15,11 +15,6 @@
 				source = { ...source, name: metadata.name };
 			}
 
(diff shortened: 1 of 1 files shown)


RocketRider:
What is the intention of the code i tried to get removed?
If someone wants name and source to be the same he can specify it twice. But if the source always overwrite the name, there is no way to specify the name anymore. 
Would it be ok if we would only set the name to the source if the name is not set?

PR #6307, opened 2024-10-21
improve url handling in citations

RocketRider:
Make the use of urls more explicit in the handling of citations, this fixes the overwrite of the name when an id is an url.


tjbck:
Please target dev branch, and fill out the form.

pfedasullivan:
@RocketRider Just wanted to let you know that I've ran into the same issue you are trying to fix and glad you have this in the works. If it can get included into the dev branch that would be most appreciated.

RocketRider:
Thanks for the comments.
https://github.com/open-webui/open-webui/pull/6324

PR #6324, opened 2024-10-22
feat: Allow URL in citations to be different from the displayed name

RocketRider:
I have a RAG based Backend which hosts the source documents on a different server. When calling it through the OpenAI Interface, I am returning the citations but currently I was not able to set the displayed name independent of the URL.
I want the displayed name to be something like "myPDF.pdf" but it should link in the modal view to "http://test.de/myPDF.pdf". In order to do that, I had to remove the overwrite of the source.name field. It was getting overwritten as soon as the id is an URL. I now added a separate field url, that is set from the id, if it is an url.

Before:
image

After:
image

The documentation does currently not at all mention this capability of providing citations through the OpenAI Interface.

# Changelog Entry

### Description

- Allow an external URL in citations, but still display the filename to the user.

### DIFF
--- a/src/lib/components/chat/Messages/Citations.svelte
+++ b/src/lib/components/chat/Messages/Citations.svelte
@@ -54,7 +54,7 @@
 				}
 
 				if (id.startsWith('http://') || id.startsWith('https://')) {
-					source = { name: id };
+					source = { ..source, url: id };
 				}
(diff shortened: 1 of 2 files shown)


tjbck:
LGTM, Thanks!

PR #6327, opened 2024-10-22

aleixdorca:
# Changelog Entry

i18n: Updated the catalan translation file.


tjbck:
Thanks!

PR #6339, opened 2024-10-22
fix: get userinfo from endpoint, not only from token

Cyb4Black:
as was suggested by @alvarolopez in #6262

# Pull Request Checklist

- [ ] ~~**Testing:** Have you written and run sufficient tests for validating the changes?~~
  - again there are no oauth test classes, did some quick tests
  - will do real world testing with keycloak tomorrow

# Changelog Entry

### Description

- OpenID Connect userinfo claims may not be present in ID token as the specification does not enforce it, therefore Open WebUI cannot be used with providers not including them.

### Fixed

- Use OpenID Conect userinfo from endpoints to get userinfo claims, not only relying on ID token containing them.
  - only do extra call if actually needed

### DIFF
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -139,6 +139,11 @@ class OAuthManager:
             log.warning(f"OAuth callback error: {e}")
             raise HTTPException(400, detail=ERROR_MESSAGES.INVALID_CRED)
         user_data: UserInfo = token["userinfo"]
+        if not user_data:
+            user_data: UserInfo = await client.userinfo(token=token)
+        if not user_data:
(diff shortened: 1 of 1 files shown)


tjbck:
Thanks!

alvarolopez:
I am afraid this still does not work as expected. The following code will not set the `user_data` with the correct `userinfo`, as it will contain the ID token claims (check https://github.com/lepture/authlib/blob/64c3eed442a86a7ff6bdc8f63e119cf201213eed/authlib/integrations/starlette_client/apps.py#L83)

        if not user_data:
            user_data: UserInfo = await client.userinfo(token=token)

I think that the `userdata` should be obtained always from the `userinfo` endpoint.

PR #6399, opened 2024-10-25
Update api/chat stream to default to true

Natedorr:
PR from discussion:

https://github.com/open-webui/open-webui/discussions/6388

update the GenerateChatCompletionForm stream to be defaulted to true.

This defaults the /api/chat/ endpoint to default to streaming being true


tjbck:
LGTM, Thanks!

PR #6432, opened 2024-10-26
fix: Prevent crash while setting Strict-Transport-Security security header

diwakar-s-maurya:
Issue discussion: https://github.com/open-webui/open-webui/discussions/6431

# Changelog Entry

### Fixed

- Prevent crash while setting Strict-Transport-Security header through HSTS environment variable.


tjbck:
Good catch, Thanks!

PR #6468, opened 2024-10-27
feat: support authentication for chromadb

diwakar-s-maurya:
# Changelog Entry

### Added

- Support authentication mechanism for chromadb as per [official docs](https://docs.trychroma.com/deployment/docker#authentication-with-docker)


tjbck:
Thanks!

Issue #5729, opened 2024-09-26
Chat history UI does not show summary of o1 model chat


codehopper-uk:
# Bug Report

## Installation Method

Helm install
```
helm list -n open-webui 
NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART                  APP VERSION
open-webui      open-webui      8               2024-09-26 10:57:03.380747635 +0000 UTC deployed        open-webui-3.1.18      0.3.23     

```
## Environment

- **Open WebUI Version:** [v0.3.23] according to helm release, but About on the UI says v0.3.30 (latest)
- **Ollama (if applicable):** [e.g., v0.2.0, v0.1.32-rc1]

- **Operating System:** image: ghcr.io/open-webui/open-webui
- **Browser (if applicable):** Chrome Version 128.0.6613.138 (Official Build) (x86_64)

**Confirmation:**

- [ ] I have read and followed all the instructions provided in the README.md.
- [ ] I am on the latest version of both Open WebUI and Ollama.
- [ ] I have included the browser console logs.
- [ ] I have included the Docker container logs.
- [ ] I have provided the exact steps to reproduce the bug in the "Steps to Reproduce" section below.

## Expected Behavior:

Chat history for o1 model queries to show summary of chat, as it does for other models eg 4o

## Actual Behavior:

Chat history for o1 model with Chat Streaming off results in unlabeled/textless chat history. 

## Description

**Bug Summary:**
Chat history for o1 model with Chat Streaming off results in unlabeled/textless chat history. Chat Streaming is required to be turned off otherwise o1 models do not work as per issue [5490](https://github.com/open-webui/open-webui/issues/5490#issuecomment-2361248888)
## Reproduction Details

**Steps to Reproduce:**
- Open New chat
- Open controls on top left > Advanced parameters > Turn off Stream Chat Response
- Select any o1 model (mini or preview for example)
- enter a query eg 2+2
- the model responds as expected
- the chat history does not update (as it does for other models that follow the above steps, eg 4o latest)

## Logs and Screenshots

**Screenshots/Screen Recordings (if applicable):**

1. o1 model query with streaming off (notice unlabeled/empty chat history summary on left)
<img width="1427" alt="image" src="https://github.com/user-attachments/assets/5d7ed69e-91b1-458e-bec6-d1671ac6fa09">
2. 4o latest model query with streaming off
[...]

PR #6524, opened 2024-10-28
fix: Remove max_completion_tokens restriction in title generation handler

Ryan526:
# Changelog Entry

### Description

- This commit removes the `"max_completion_tokens": 50` line in `backend/open_webui/main.py` at line 1574. The restriction was causing issues with o1 models, where reasoning tokens were consuming significantly more and preventing valid titles from being generated. By setting the value to `pass` (removing the cap), the token limit is no longer enforced, preventing models from hitting the token ceiling before completion. This fix is needed to allow title generation when using o1 models.

### Changed

- Updated line 1574 in backend/open_webui/main.py

### Fixed

- #5729

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1571,7 +1571,7 @@ Artificial Intelligence in Healthcare
             {"max_tokens": 50}
             if app.state.MODELS[task_model_id]["owned_by"] == "ollama"
             else {
-                "max_completion_tokens": 50,
+                pass
             }
         ),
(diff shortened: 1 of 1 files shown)


tjbck:
We would want this only for o1 models, feel free to reopen!

Ryan526:
> We would want this only for o1 models, feel free to reopen!

kinda realized it might be pointless as I didn't know you could easily set your task model in the admin panel....

image

PR #6537, opened 2024-10-29
i18n Add new language - CZ, add missing translations to RO

t0ny-d3v:
### Added

- cs_CZ- new translation

### Changed

- ro_RO add missing translations


tjbck:
Thanks!

PR #6645, opened 2024-11-01
feat: Intial support for pgvector as backing vector database

jk-f5:
# Changelog

## Description
The PR introduces support for using pgvector as a backing vector datastore.

### Notes
- One can bring their own PGVECTOR_DB_URL or choose to use the primary app database connection (assuming it is postgres with pgvector support).
- Currently the vector size is set to 1536. Any embedding models producing smaller vector sizes are padded with zeros. (Setting a size on a vector column is required in order to index it)
  - Another option is to create additional vector columns of common vector sizes.

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -20,6 +20,7 @@ from open_webui.env import (
     WEBUI_FAVICON_URL,
     WEBUI_NAME,
     log,
+    DATABASE_URL,
 )
 from pydantic import BaseModel
 from sqlalchemy import JSON, Column, DateTime, Integer, func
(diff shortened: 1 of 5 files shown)


.devcontainer/devcontainer.json
tjbck, line 1: Unwanted file here.

docker-compose.devcontainer.yaml
tjbck, line 1: Unwanted,

backend/open_webui/config.py
```diff
@@ -931,6 +932,9 @@ class BannerModel(BaseModel):
 
 VECTOR_DB = os.environ.get("VECTOR_DB", "chroma")
 
+if VECTOR_DB == 'pgvector' and not DATABASE_URL.startswith("postgres"):
```
tjbck, line 935: We should have a separate `PGVECTOR_DB_URL` env var, which takes in DATABASE_URL as it's default value if PGVECTOR_DB_URL env var is "" or None.

tjbck:
Thanks!

Issue #6748, opened 2024-11-06
New cloned customized models are not saving their System Prompt the first time


danielj23:
# Bug Report
 
## Installation Method
docker *latest

## Environment

[v0.3.35](https://github.com/open-webui/open-webui/releases/tag/v0.3.35)

Windows 10

**Confirmation:**

- [ x] I have read and followed all the instructions provided in the README.md.
- [ x] I am on the latest version of both Open WebUI and Ollama.
- [ ] I have included the browser console logs.
- [ ] I have included the Docker container logs.
- [x ] I have provided the exact steps to reproduce the bug in the "Steps to Reproduce" section below.

## Expected Behavior:

A model I am cloning should save the System Prompt correctly for the new clone

## Actual Behavior:

The actively cloned model is not saving the System Prompt.

## Description

**Bug Summary:**
I cloned an existing model and changed the system prompt, and after clicking Save & Update, the System Prompt was not updated for the new cloned model.

## Reproduction Details

**Steps to Reproduce:**
Clone an existing customized model.
During the edit of the new clone, change the name which adds (Clone) by default.   This step may not be required and should be tested.
Change the System Prompt of the new cloned model.
Click Save & Update
Re-open that cloned model
The System Prompt is the original cloned System Prompt.

**Additional Notes:**

I am getting around this by immediately clicking Save once I have started the clone.  Then I open the edit again and start my editing.  Clicking Save & Update then is all fine.

mafzaal:
[PR](https://github.com/open-webui/open-webui/pull/6772)

tjbck:
Should be fixed in dev, testing wanted here!

PR #6772, opened 2024-11-07
**Fix** Issue 6748 - New cloned customized models are not saving their System Prompt the first time

mafzaal:
Fix for [Issue 6748](https://github.com/open-webui/open-webui/issues/6748)


tjbck:
Appreciate the PR, model editor is being overhauled entirely atm.

PR #6888, opened 2024-11-12
fix/feat: Revert DOMPurify in citations HTML

michaelpoluektov:
Reverts open-webui/open-webui#6879

Is there a reason to use `DOMPurify` here? I had used it originally for security reasons just in case, but now that we've got artifacts that render unsanitised HTML I don't see why sanitising it in citations is necessary.

Or is there another mechanism that prevents artifacts from running malicious code, in which case, can we re-use that?

`DOMPurify` with the default configuration breaks things like `iframe` which can be useful for citations

### DIFF
--- a/src/lib/components/chat/Messages/CitationsModal.svelte
+++ b/src/lib/components/chat/Messages/CitationsModal.svelte
@@ -1,6 +1,4 @@
 <script lang="ts">
-	import DOMPurify from 'dompurify';
-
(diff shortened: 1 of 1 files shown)


tjbck:
Artifacts are sandboxed with iframe, as long as we sandbox this with iframe as well I see no issue. Feel free to reopen!

PR #7041, opened 2024-11-19
feat: Ability to configure the use of private API keys in an environment

antpyykk-kone:
# Pull Request Checklist

### Note to first-time contributors: Please open a discussion post in [Discussions](https://github.com/open-webui/open-webui/discussions) and describe your changes before submitting a pull request.

**Before submitting, make sure you've checked the following:**

- [x] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [x] **Description:** Provide a concise description of the changes made in this pull request.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ ] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [x] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [ ] **Testing:** Have you written and run sufficient tests for validating the changes?
- [x] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [x] **Prefix:** To cleary categorize this pull request, prefix the pull request title, using one of the following:
  - **BREAKING CHANGE**: Significant changes that may affect compatibility
  - **build**: Changes that affect the build system or external dependencies
  - **ci**: Changes to our continuous integration processes or workflows
  - **chore**: Refactor, cleanup, or other non-functional code changes
  - **docs**: Documentation update or addition
  - **feat**: Introduces a new feature or enhancement to the codebase
  - **fix**: Bug fix or error correction
  - **i18n**: Internationalization or localization changes
  - **perf**: Performance improvement
  - **refactor**: Code restructuring for better maintainability, readability, or scalability
  - **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc.)
  - **test**: Adding missing tests or correcting existing tests
  - **WIP**: Work in progress, a temporary label for incomplete or ongoing work

# Changelog Entry

### Description

- Introduce a configuration option to disable private API key use and generation in an environment. When disabled, only JWT can be used to authenticate, helping to prevent potential misuse. By default private API keys are still enabled.

### Added

- A new configuration option `ENABLE_API_KEY_AUTH` to control private API key use in an environment.

### Changed

- Prevent authentication using API key, if `ENABLE_API_KEY_AUTH` config is disabled
[...]

PR #7322, opened 2024-11-25
Feat: API Keys allow limited endpoints vs all endpoints (configurable)

DmitriyAlergant:
By default, only allow API keys authentication to /chat/completions and /models endpoints (LLM invocation). Admin can configure the app to allow API Keys for All Endpoints (default: false) - but this has security implications that are likely unwanted in many contexts. See https://github.com/open-webui/open-webui/discussions/7018 for justification.

Also adding **[ ] Code formatting** step to the default GitHub PR template, based on recent experience of a new contributor (failed formatting CIs on backend and frontend) - it would be better to be informed of the formatting requirements via this checklist rather then failing CIs.

# Pull Request Checklist

- [X] **Testing:** Have you written and run sufficient tests for validating the changes?
- [X] **Prefix:** To cleary categorize this pull request, prefix the pull request title, using one of the following:

# Changelog Entry

### Description

- Limit API Key authentication to only /api/chat/completions and /api/models endpoints by default, unless a config option was enabled to allow all endpoints to be accessed

### Added

- Configurable option "API Key allow access to all endpoints"

### Security

- Limit API Key authentication to only /api/chat/completions and /api/models endpoints by default, unless a config option was enabled to allow all endpoints to be accessed

---

### Additional Information

- https://github.com/open-webui/open-webui/discussions/7018


DmitriyAlergant-T1A:
Note to maintainer: ideally that would have been done via the new Groups functionality as a group-level permission, but I have not had capacity to implement it this way at this time.  Implementing it as a general config was straightforward, as it basically repeated the recently merged https://github.com/open-webui/open-webui/pull/7041

DmitriyAlergant-T1A:
The proposed default (false/disabled) technically makes it a somewhat “breaking” change if someone actually relied on API keys for triggering other endpoints; though it would be trivial for them to fix;  

This default will also allow more users to have a more secure environment, so in my opinion it should default to disabled.

DmitriyAlergant-T1A:
Note to @tjbck: ideally I strongly believe that API Key authentication must **always** be restricted to LLM invocation endpoints, and the configurable option is not even necessary. But my previous PR for that (a one-liner) was rejected...

If we are having it configurable, this could have been done via the new Groups functionality as a group-level permission; I have not had capacity to implement it this way at this time. Implementing it as a general config was straightforward, and it basically repeated the recently merged PR https://github.com/open-webui/open-webui/pull/7041

The proposed default (false/disabled) technically makes it a somewhat “breaking” change if someone currently relied on API keys for triggering other endpoints (not only for invoking the completion), though it would be trivial for them to re-enable current behavior by just toggling the setting. Default 'false/disabled' will get users to have a more secure environment, so in my opinion it should default to disabled (unless you agree with even harder line of always restricting endpoints)

tjbck:
All of our endpoints are properly scoped (user/admin), This PR won't be merged. Appreciate the effort.

DmitriyAlergant-T1A:
@tjbck please hear me out.   Yes, all endpoints are segregated User/Admin. However: 

1. API keys do not expire (unlike JWTs), and also do not revalidate OAuth authentication status. A user may have already left the company, but still retain his API key that would still work.

2. API keys are often used with 3rd party software tools (and in codebases/deployed software) and are for this reason more prone to leakage due to the usage pattern - either by the users(admins) mistake, or by malicious 3rd party tools.

There is a **vast difference in risk profile** (what **happens** on exploitation) between different API endpoints. 
Please take a look at the comparison table below.

CleanShot 2024-11-25 at 20 04 55@2x


Given the API key has inherently larger risk of being exploited (see above why), the more sensitive endpoints deserve additional level of protection - like disallowing their access by the API key.

tjbck:
Sounds reasonable, thanks for the detailed explanation. I'll get to this PR when I have the capacity!

DmitriyAlergant-T1A:
Thanks for the consideration!  

Now that you saw the risk profile concerns, you may actually review my original suggestion,  which was that API keys should ONLY work for /chat/completion and /models, full stop - not even configurable.  Whoever needs to integrate to other endpoints, they can do /signin to get a JWT...  That would dramatically simplify the change down to a one-liner. 

But I do not insist. As far as my use-case is concerned, we're OK with a configuration toggle.

tjbck:
Addressed in dev!

PR #7515, opened 2024-12-01
feat: PWA share to OWUI

Peter-De-Ath:
### Description

When Open WebUI is installed as a PWA

- Allows sharing documents from file managers
- Allows sharing images
- Allows sharing Youtube Links
- Allows sharing Website URLS
- Allows sharing text

#### Limitations
Only supported on Android using Chrome
Uploading only 1 file at a time

### Added

Service Worker
Sharing Images/Documents/Text

### Screenshots or Videos

https://github.com/user-attachments/assets/edceb61b-21f7-44ea-b6b8-8dd0ac07a449


src/service-worker.js
tjbck, line 1: Might want to use an alternative for this, I'll try to come up with something here!

grebenyuksv:
You are online despite the airplane mode, right? 🤔 I'm getting one of these two results when reloading the page after going offline, and I can't find the code that would be supposed to achieve my goal: make the browser cache all static resources and continue talking to OpenAI API after I shut down the backend. I suspect that's just not supported. Could you confirm it?

image
image

PR #7646, opened 2024-12-05
added option to disable websocket emiting

itaybar:
Added option to disable usage websocket and user-count websocket

This feature includes adding those to the config in order that the UI will not emit on every event to the backend

 #7633


tjbck:
Closing due to unwanted changes, and conflicts. You might want to check if the changed code will work without an issue with the latest dev.

Issue #7583, opened 2024-12-03
Bug: Signout doesn't terminate SSO sessions


sreinwald:
# Bug Report

## Installation Method

Docker compose 

## Environment

- **Open WebUI Version:** v0.4.7
- **Ollama (if applicable):** v0.4.7
- **Operating System:** Ubuntu 24.04.1 LTS
- **Browser (if applicable):** Tested in Firefox v133.0 and Google Chrome 131.0.6778.86

**Confirmation:**

- [x] I have read and followed all the instructions provided in the README.md.
- [x] I am on the latest version of both Open WebUI and Ollama.
- Not applicable: I have included the browser console logs.
- Not applicable: I have included the Docker container logs.
- [x] I have provided the exact steps to reproduce the bug in the "Steps to Reproduce" section below.

## Expected Behavior

When a user clicks on the logout button, it should fully terminate their session both within Open WebUI and with the SSO provider, in our case Keycloak. 
The user should have to re-authenticate before being able to access the webui after logging out.

## Actual Behavior

Clicking the logout button only deletes the cookie for Open Webui, but does not end the session in Keycloak. 

As a result, users can still access the webui (and any other SSO-enabled service) without having to re-authenticate by simply clicking the button to sign in with the IDP again. 

## Description

**Bug Summary:**
Logout functionality is incomplete because when using SSO, logging out does not properly terminate the user's SSO session.

## Reproduction Details

**Steps to Reproduce:**

1. Log into Open WebUI using a Keycloak identity.
2. Click on the logout button within the application.
3. While the webui cookie is deleted, but the Keycloak session remains active.
4. Clicking the button to log in with OIDC immediately redirects the user to the authenticated state of the webui without having to re-authenticate

## Logs and Screenshots

**Screenshots/Screen Recordings (if applicable):**
- https://www.youtube.com/watch?v=4bB6qe2ahDE

## Additional Information

In my opinion, this is potentially a security issue because when a user logs out of an application, they expect all SSO sessions to terminate. The session not being terminated properly means anyone with physical access to the user's device could use the still existing session to access all other SSO-enabled services without having to re-authenticate. 

I've taken a brief look at the code, and it seems like the logout function does not attempt to call an end session endpoint at all, which is why I suspect this is not related to Keycloak specifically but likely affects every OAuth2/OIDC deployment. 

[...]

PR #7678, opened 2024-12-07
feat: implement OAuth logout functionality for keyclock to terminate sso session

ZaibanAli:
# Pull Request Checklist

- [X] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [X] **Testing:** Have you written and run sufficient tests for validating the changes?
- [X] **Prefix:** To cleary categorize this pull request, prefix the pull request title, using one of the following:

# Changelog Entry

### Description

- Signout doesn't terminate SSO sessions [ISSUE_LINK](https://github.com/open-webui/open-webui/issues/7583)
- Clicking the logout button only deletes the cookie for Open Webui, but does not end the session in Keycloak.
- As a result, users can still access the webui (and any other SSO-enabled service) without having to re-authenticate by simply clicking the button to sign in with the IDP again.

### Added

- Adding id_token to the cookie when the oidc provider signup is True. This is required to cleanup the session when signing out.

### Changed

- Updated the `/signout` endpoint to include logic for cleaning up the oidc session (keycloak) using the `id_token`. (backend/open_webui/apps/webui/routers/auths.py)
- Enhanced the OAuth callback functionality to set an `id_token` cookie when using the OAUTH oidc provider. (backend/open_webui/utils/oauth.py)

### Deprecated

- [List any deprecated functionality or features that have been removed]

### Removed

- [List any removed features, files, or functionalities]

### Fixed

- Resolved an issue where the `signout` function did not correctly handle Keycloak logouts due to missing `id_token`.

### Security

- [List any new or updated security-related changes, including vulnerability fixes]

### Breaking Changes

---

### Additional Information

- [ISSUE_7583](https://github.com/open-webui/open-webui/issues/7583)

### Screenshots or Videos

- [Demonstration Video](https://drive.qaiv.com/s/8tkkEHQXYZ7i6Jm)

### DIFF
--- a/backend/open_webui/apps/webui/routers/auths.py
+++ b/backend/open_webui/apps/webui/routers/auths.py
@@ -29,7 +29,11 @@ from open_webui.env import (
     SRC_LOG_LEVELS,
 )
 from fastapi import APIRouter, Depends, HTTPException, Request, status
-from fastapi.responses import Response
+from fastapi.responses import RedirectResponse, Response
+from open_webui.config import (
+    OAUTH_PROVIDER_NAME,
+    OAUTH_LOGOUT_URL,
+)
 from pydantic import BaseModel
 from open_webui.utils.misc import parse_duration, validate_email_format
 from open_webui.utils.utils import (
@@ -498,8 +502,17 @@ async def signup(request: Request, response: Response, form_data: SignupForm):
 
 
 @router.get("/signout")
-async def signout(response: Response):
+async def signout(request: Request, response: Response):
     response.delete_cookie("token")
+    
+    if OAUTH_PROVIDER_NAME.value == "keycloak" and OAUTH_LOGOUT_URL:
+        id_token = request.cookies.get("id_token", None)
+        if id_token:
+            logout_url = f"{OAUTH_LOGOUT_URL}?id_token_hint={id_token}"
+            response.delete_cookie("id_token")
+            return RedirectResponse(url=logout_url)
+
+    # Fall back to the default signout
     return {"status": True}
 
 
--- a/backend/open_webui/apps/webui/routers/auths.py
+++ b/backend/open_webui/apps/webui/routers/auths.py
@@ -33,6 +33,7 @@ from fastapi import APIRouter, Depends, HTTPException, Request, status
 from fastapi.responses import RedirectResponse, Response
 from open_webui.config import (
     OPENID_PROVIDER_URL,
+    ENABLE_OAUTH_SIGNUP,
 )
 from pydantic import BaseModel
 from open_webui.utils.misc import parse_duration, validate_email_format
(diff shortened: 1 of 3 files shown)


sreinwald:
First of all: thanks for taking on the issue, I was going to tackle it today as well.

Some notes:
I believe adding new ENV variables is not necessary. [OIDC RP-Initiated Standard](https://openid.net/specs/openid-connect-rpinitiated-1_0.html) is to publish the `end_session_endpoint` at the OIDC provider's `/.well-known/openid-configuration` API endpoint, which we get from `OPENID_PROVIDER_URL`. 

It's probably cleaner to check if the OIDC provider supports the OIDC RP-Initiated Standard by supplying the `end_session_endpoint` and then call that, rather than have the user supply two new ENV variables, and do individual checks for each OIDC provider, as your implementation only works for Keycloak right now.

ZaibanAli:
@sreinwald thanks for this information, i didn't knew about this. I agree this makes more sense since this is more elegant and can be used for other providers.

Any idea how to tackle the issue with the specific query parameter i.e. `id_token_hint` required by logout endpoint of keycloak?

ZaibanAli:
@sreinwald I have now removed the additional variable, can you please review again.

backend/open_webui/apps/webui/routers/auths.py
```diff
@@ -3,6 +3,7 @@
 import time
 import datetime
 import logging
+import httpx
```
tjbck, line 6: should be implemented with `aiohttp`

ZaibanAli: I have updated the code to use aiohttp instead of httpx.

tjbck:
Looks great, Thanks!

TheDropZone:
@ZaibanAli 
So, I've deployed this OAuth functionality alongside a keycloak instance, and am running into some issues with the cookie deletion when open-webui and keycloak live on different domains.
- openwebui01-domain.com
- kc01-domain.com

The cookies (token, oauth_id_token) live and are set on the open-webui domain (openwebui01-domain.com), but the signout url from keycloak (and the openid_config) is on kc01-domain.com. So, it seems that when the signout endpoint requests to delete the cookies (living on openwebui01) but then redirects to kc01-domain for signout, the cookies don't get deleted (due to cookie domain mismatch). Because I'm redirected to the kc01-domain for the signout, the cookies dont exist at that domain to be deleted, and token and oauth_id_token live forever.

Has anyone else seen this issue, or have ideas on the proper way to handle this?

mig5:
Why is this only possible if `ENABLE_OAUTH_SIGNUP` is set to true?

If I've pre-configured some users in Open WebUI with the same attributes that are in my OIDC backend for those users, I should be able to handle SSO logout for those users when logging out of Open WebUI, even if `ENABLE_OAUTH_SIGNUP` is false (I don't want other OIDC users to self-register in Open WebUI).

PR #7780, opened 2024-12-11
Disable Polling Transport When WebSockets Are Enabled and Implement Cleanup Locking Mechanism

jk-f5:
## Overview
  
This merge request introduces changes to optimize WebSocket usage and ensure proper operation in distributed environments.

1. **Disable the "polling" transport when WebSocket support is enabled** to reduce unnecessary overhead and prevent connectivity issues in multi-replica deployments without sticky sessions.
2. **Implement a Redis-based locking mechanism** to synchronize the `periodic_usage_pool_cleanup` task across multiple instances in a distributed deployment.

### 1. Disable Polling Transport When WebSockets Are Enabled

**Rationale:**

By default, Socket.IO uses both `polling` and `websocket` transports. In environments with multiple application instances (replicas) behind a load balancer without sticky sessions, the `polling` transport can lead to connectivity issues:

- **400 Errors Without Sticky Sessions:**
  - The `polling` transport relies on HTTP long-polling requests that must consistently reach the same server instance. Without sticky sessions, requests from the same client may be routed to different instances, causing the server to not recognize the session and respond with HTTP 400 errors.
  - This happens because the initial HTTP request that establishes the Socket.IO session and the subsequent polling requests are handled by different servers, leading to mismatches in session IDs and authentication tokens.
  - As a result, clients experience failed connections and errors in the application.

- **Solution:**
  - By disabling the `polling` transport and using only `websocket`, which operates over a persistent TCP connection, we avoid the need for sticky sessions.
  - WebSocket connections are maintained over the same TCP connection, ensuring that all communication is directed to the same server instance.
  - Most modern load balancers support WebSocket connections and can route them correctly without sticky sessions.

**Benefits:**

- **Improved Stability:** Eliminates 400 errors caused by misrouted polling requests in non-sticky environments.
- **Better Performance:** Reduces overhead associated with the fallback polling mechanism, leading to more efficient resource utilization.
- **Simplified Deployment:** Removes the requirement for sticky sessions or additional load balancer configuration, making it easier to deploy in scalable environments.

### 2. Implement Redis-Based Locking for `periodic_usage_pool_cleanup`

**Rationale:**

In a distributed deployment with multiple instances of the application, running the `periodic_usage_pool_cleanup` task concurrently can lead to race conditions and inconsistent state in the `USAGE_POOL`. By introducing a Redis-based lock:

- **Ensures Single Instance Execution:**
  - Only one instance acquires the lock and performs the cleanup at any given time.
  - Prevents conflicts and potential data corruption due to simultaneous access.

- **Compatibility with Distributed Systems:**
  - Redis serves as a centralized coordination point accessible by all instances.
  - The lock mechanism leverages Redis's atomic operations to manage concurrency.

### DIFF
--- a/backend/open_webui/apps/socket/main.py
+++ b/backend/open_webui/apps/socket/main.py
@@ -13,7 +13,7 @@ from open_webui.env import (
     WEBSOCKET_REDIS_URL,
 )
 from open_webui.utils.auth import decode_token
-from open_webui.apps.socket.utils import RedisDict
+from open_webui.apps.socket.utils import RedisDict, RedisLock
 
 from open_webui.env import (
     GLOBAL_LOG_LEVEL,
@@ -32,7 +32,7 @@ if WEBSOCKET_MANAGER == "redis":
         cors_allowed_origins=[],
         async_mode="asgi",
         transports=(
-            ["polling", "websocket"] if ENABLE_WEBSOCKET_SUPPORT else ["polling"]
+            ["websocket"] if ENABLE_WEBSOCKET_SUPPORT else ["polling"]
         ),
         allow_upgrades=ENABLE_WEBSOCKET_SUPPORT,
         always_connect=True,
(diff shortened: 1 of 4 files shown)


tjbck:
Closing due to conflicts, feel free to reopen!

PR #7802, opened 2024-12-12
Fix: Openai-edge-tts speaking markdown not plain text

ALIENvsROBOT:
Openai-edge-tts is always speaking with markdown which results ("hashtag hashtag hashtag... [AI response]") to fix I used mardown2 and re libraries to convert check if the input is markdown or not. if its markdown it will convert to plain text. (needs testing)


tjbck:
This should be done from the frontend side, feel free to reopen!

PR #8130, opened 2024-12-27
refac

tjbck:


DmitriyAlergant-T1A:
Thanks, that is lightning fast !

BTW I realized the culprit for the curruption by i18n was the ":" symbol not "|".      
But thanks for rolling back i18n.

PR #8147, opened 2024-12-27
feat: add background saving for chat messages

cr-zhichen:
In `middleware.py`, a background database saving functionality has been added to address the issue discussed in #8088, where the output speed decreases due to database latency.


tjbck:
This will break the existing async chat functionality, I'd be open to merging this but we'll need to make this explicitly opt-in and must be explicitly enabled by a env var.

Issue #8248, opened 2025-01-01
Exception Unpacking Error in `process_chat_payload`


albinvar:
When an exception occurred in `process_chat_payload`, it was being returned instead of raised.

![image](https://github.com/user-attachments/assets/8a56737f-5729-4077-9531-ed8aabea2205)

Instead of

![image (1)](https://github.com/user-attachments/assets/49ce9784-85a9-434d-957d-05ef62965ce8)

 This caused the caller (chat_completion) to attempt unpacking the Exception object as if it were a tuple, leading to the error:
```json
{"detail": "cannot unpack non-iterable Exception object"}
```

The issue can be reproduced by intentionally raising an exception within the custom function ([example](https://gist.github.com/albinvar/902a0e3734232e742528147761269b76)). 

It seems to have been introduced in version `0.5.x`, whereas earlier versions handled this correctly.

PR #8249, opened 2025-01-01
fix: raise exceptions properly in process_chat_payload

albinvar:
Changed return to raise to ensure the exception propagates correctly to the caller and triggers proper error handling.

https://github.com/open-webui/open-webui/blob/4bc9904b3cd0726d3f9c3cbaeade972cf167b6c4/backend/open_webui/utils/middleware.py#L647

Using raise ensures the exception is thrown upwards, allowing the API handler to catch it and respond with a proper HTTP error code.

Issue Reference #8248


tjbck:
Good catch, Thanks!

PR #8281, opened 2025-01-02
feat: VECTOR_LENGTH configurable through environment variable

binaryYuki:
PR for https://github.com/open-webui/open-webui/discussions/8271

### Description

This pull request introduces several updates and improvements to the backend/open_webui/retrieval/vector.py file. These changes focus on making the system more flexible and configurable by replacing hardcoded values with environment-driven configurations, improving maintainability and adaptability to different deployment environments.

The following changes were made to enhance the functionality and configuration capabilities of the retrieval/vector.py module:

### Added
Import of os and int libraries from os and builtins respectively to enable environment-based configurations.

docs update: https://github.com/open-webui/docs/pull/348

### Changed
* [`backend/open_webui/retrieval/vector/dbs/pgvector.py`](diffhunk://#diff-28a6dc0ae8b8aecb38ddec790ac95fca054271e96eb3ecf2b2cd4cf7a7c1adb2L88-R94): Updated the `adjust_vector_length` method to use `self.vector_length` instead of a hardcoded value.
* [`backend/open_webui/retrieval/vector/dbs/pgvector.py`](diffhunk://#diff-28a6dc0ae8b8aecb38ddec790ac95fca054271e96eb3ecf2b2cd4cf7a7c1adb2L167-R172): Updated the `search` method to use `self.vector_length` for casting vectors.

### Removed
Removed static references to the hardcoded `VECTOR_LENGTH` to increase adaptability and eliminate redundancy.

### Additional Information
- Motivation: The hardcoded approach previously used for VECTOR_LENGTH restricted the flexibility of the system. Making VECTOR_LENGTH configurable allows the retrieval module to adapt to different deployments and environments without requiring code changes.

- Impact: These changes improve configurability and simplify deployment across environments by leveraging environment variables.

Relevant Sections Modified:

* [`backend/open_webui/retrieval/vector/dbs/pgvector.py`](diffhunk://#diff-28a6dc0ae8b8aecb38ddec790ac95fca054271e96eb3ecf2b2cd4cf7a7c1adb2R1): Imported the `os` module to access environment variables.
* [`backend/open_webui/retrieval/vector/dbs/pgvector.py`](diffhunk://#diff-28a6dc0ae8b8aecb38ddec790ac95fca054271e96eb3ecf2b2cd4cf7a7c1adb2L24-R40): Modified the `vector` column in the `DocumentChunk` class to use a configurable vector length from the `VECTOR_LENGTH` environment variable, with a default of 1536.
* [`backend/open_webui/retrieval/vector/dbs/pgvector.py`](diffhunk://#diff-28a6dc0ae8b8aecb38ddec790ac95fca054271e96eb3ecf2b2cd4cf7a7c1adb2L88-R94): Added `self.vector_length` to the `PgvectorClient` class to store the configurable vector length.

### DIFF
--- a/backend/open_webui/retrieval/vector/dbs/pgvector.py
+++ b/backend/open_webui/retrieval/vector/dbs/pgvector.py
@@ -1,3 +1,4 @@
+import os
 from typing import Optional, List, Dict, Any
 from sqlalchemy import (
     cast,
@@ -21,7 +22,6 @@ from sqlalchemy.ext.mutable import MutableDict
 from open_webui.retrieval.vector.main import VectorItem, SearchResult, GetResult
 from open_webui.config import PGVECTOR_DB_URL
 
-VECTOR_LENGTH = 1536
 Base = declarative_base()
 
 
@@ -29,7 +29,7 @@ class DocumentChunk(Base):
     __tablename__ = "document_chunk"
 
     id = Column(Text, primary_key=True)
-    vector = Column(Vector(dim=VECTOR_LENGTH), nullable=True)
+    vector = Column(Vector(dim=int(os.getenv("VECTOR_LENGTH", 1536))), nullable=True)
     collection_name = Column(Text, nullable=False)
     text = Column(Text, nullable=True)
     vmetadata = Column(MutableDict.as_mutable(JSONB), nullable=True)
@@ -37,6 +37,7 @@ class DocumentChunk(Base):
 
 class PgvectorClient:
     def __init__(self) -> None:
+        self.vector_length = int(os.getenv("VECTOR_LENGTH", 1536))
 
         # if no pgvector uri, use the existing database connection
         if not PGVECTOR_DB_URL:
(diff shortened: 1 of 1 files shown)


jk-f5:
A couple of things:
- This needs to follow the same pattern as all the other configuration environment variables. See `backend/open_webui/config.py` and consider naming the variable something like `PGVECTOR_INITIALIZE_MAX_VECTOR_SIZE` for consistency.
- Also, once set, this value cannot change without altering the vector column size, updating all existing vector entries, and then dropping and recreating the index with the new size. Therefore, we either need to prevent this value from being changed after the initial creation of the `document_chunk` table, or we need to write the code that alters the column size, updates all of the existing vectors to the new size, and then drops and recreates the index on the vector column. I would suggest the former option, as the latter is an expensive operation depending on the number of entries in the `document_chunk` table.

binaryYuki:
> * Also, once set, this value cannot change without altering the vector column size, updating all existing vector entries, and then dropping and recreating the index with the new size.

Hi,

Thank you for your review!

I have a few ideas for implementing this feature. For example, can we determine whether a migration has been performed before by checking whether the file “alembic.ini” exists? Or can we add a filter to migrations/env.py to prevent changes to the
DocumentChunk table?

If you have any comments or preferences, please let me know!

tjbck:
#8298

PR #8309, opened 2025-01-03
ci: add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

reincarnated 
- https://github.com/open-webui/open-webui/pull/8308

on top of `dev` - same/no new typos found


tjbck:
LGTM, Thanks!

Issue #7143, opened 2024-11-21
Degraded RAG results in v0.4.2 (compared to v0.3.35)


firoagni:
# Bug Report

## Installation Method

Using Docker to deploy Open WebUI

## Environment

- Open WebUI Version: v0.4.2
- Ollama: 0.4.2

- Operating System: macOS 15.0.1

**Confirmation:**

- [X ] I have read and followed all the instructions provided in the README.md.
- [X ] I am on the latest version of both Open WebUI and Ollama.
- [   ] I have included the browser console logs.
- [   ] I have included the Docker container logs.
- [X ] I have provided the exact steps to reproduce the bug in the "Steps to Reproduce" section below.

## Expected Behavior:

I am running two instances of Open WebUI:
	•	Port 2000: Open WebUI v0.4.2
	•	Port 3000: Open WebUI v0.3.35

Both are fresh installations. I created the same knowledge object and added an identical markdown file to it on both instances. Subsequently, I created a custom model based on `Llama:3.2` using the knowledge object in both instances, setting the TEMPERATURE value to 0 (accuracy over creativity).

Expectation: Both Open WebUI v0.4.2 and Open WebUI v0.3.35 should produce similar results since they are running the same custom RAG model.

## Actual Behavior:

For the same questions, my custom model in Open WebUI v0.4.2 consistently provides worse responses compared to Open WebUI v0.3.35. I tested with different files, including various file formats, and multiple questions. The answers from v0.4.2 are consistently inferior to those from v0.3.35. Strangely, when I click on citations, they are mostly the same between the two versions. However, v0.4.2 tends not to follow the citations as closely.

Even tried changing the TEMPERATURE values. No luck.

What changes might account for this discrepancy?

## Reproduction Details
1.	Setup Instances:
- Deploy two instances of Open WebUI on different ports:
	- Port 2000: Install Open WebUI v0.4.2
	- Port 3000: Install Open WebUI v0.3.35
  Ensure both installations are fresh and have no prior configuration.

2. Create Knowledge Object:
 - On both instances, create the same knowledge object.
 - Add an identical markdown file to the knowledge object in each instance.

3.	Configure Custom Model:
- In both WebUI instances, create a custom model based on Llama:3.2 using the knowledge object.
- Set the TEMPERATURE value to 0 for both models to prioritize accuracy over creativity.

4.	Run Tests:
- Ask identical questions to both models in each WebUI instance.
- Observe and document the responses from each version.

[...]

PR #5989, opened 2024-10-07
enh: citations show relevance score - compact citation view 

jannikstdl:
### Description

This PR includes the following key changes:

- **Relevance Score for Citations:** Citations now display a relevance score in both percentage and float distance formats.
~~**Sorted Citations:** Citations under the message are now sorted by relevance (e.g., source 1, source 2...).~~ -> Already implemented in the backend
- **Compact Source Display:** For more than 3 sources, a cleaner source display is shown under the message.

**Relevance Score for Citations:**
image

**Compact Source Display:**
https://github.com/user-attachments/assets/848300f3-edef-4074-b2d5-d5ba9c680714

**Testing**
- [x] Milvus DB
- [x] If error / distances are not available 
- [x] Other Embedding Engines than Ollama: OpenAI Embedding, Dafault integrated ones
- [x] If only one distance is in cosine similarity range -1 to 1

PR #8379, opened 2025-01-07
Fix the issue of inaccurate answers after enabling RAG query generation

qiaozhi199:
I upgraded the Open-WebUI version from v0.3.35 to v0.4.7. I found that the RAG response quality is not as good as in version v0.3.35. 
I also noticed that others have encountered similar issues: https://github.com/open-webui/open-webui/issues/7143

I read the source code of Open-WebUI and found that the `RAG query generation` feature was added starting from version v0.4.0.
I saw that after enabling the `RAG query generation` feature, the Ollama service is called to generate 1 to 3 queries, which are then used to query vectors.

For example, if 3 queries are generated and the top k is set to 6, there will be a total of 18 vector query results. However, the default sorting method of the `merge_and_sort_query_results function` is ascending order, so it will only return the 6 query results with the smallest distances, meaning the 6 results with the lowest vector similarity. This leads to inaccurate RAG responses.

Therefore, I believe the `merge_and_sort_query_results` function should use descending order sorting, so that it can return the top k vector query results with the highest similarity.


juananpe:
I think there's a misunderstanding about how vector similarity works. AFAIK, the current ascending sort in `merge_and_sort_query_results` is actually correct because:

* A distance score of 0 means vectors are identical (perfect match)
* Smaller distances = Higher similarity
* Larger distances = Lower similarity

So when the function sorts in ascending order, it's already returning the most similar results first (those with smallest distances).

It would be really helpful if you could share some example documents and queries where you're seeing worse results in v0.4.7 compared to v0.3.35. That way we could better investigate what might be causing the quality decrease.

longfei-zhang:
https://milvus.io/docs/metric.md
Vector similarity it's based on metric types:
image

qiaozhi199:
> I think there's a misunderstanding about how vector similarity works. AFAIK, the current ascending sort in `merge_and_sort_query_results` is actually correct because:
> 
> * A distance score of 0 means vectors are identical (perfect match)
> * Smaller distances = Higher similarity
> * Larger distances = Lower similarity
> 
> So when the function sorts in ascending order, it's already returning the most similar results first (those with smallest distances).
> 
> It would be really helpful if you could share some example documents and queries where you're seeing worse results in v0.4.7 compared to v0.3.35. That way we could better investigate what might be causing the quality decrease.

I configured and used the Milvus vector database in Open-WebUI，
I saw that in the Open-WebUI source code, a collection is created using the pymilvus client, with the index type set to HNSW and the metric_type set to COSINE.
https://github.com/open-webui/open-webui/blob/1dfb479d367e5f5902f051c823f9aef836e04791/backend/open_webui/retrieval/vector/dbs/milvus.py#L99-L104

The Milvus official documentation mentions that for the COSINE metric type, the higher the value, the greater the similarity.
https://milvus.io/docs/metric.md

image

tjbck:
@juananpe fyi, cosine similarity of 1 means perfect match not 0. This definitely seems to be a required fix, thanks everyone!

juananpe:
> @juananpe fyi, cosine similarity of 1 means perfect match not 0. This definitely seems to be a required fix, thanks everyone!

Please, can you double check that it is also the case in ChromaDB?

What I understand from Chroma's documentation:
https://docs.trychroma.com/docs/collections/configure#configuring-chroma-collections
is that in Chroma the cosine function is (1 - the cosine) to preserve the semantic of smaller distance means closer.

See:
image
and:
https://github.com/chroma-core/chroma/issues/1187#issuecomment-1773806170

tjbck:
@juananpe 😅 You're absolutely right—seems like chroma has a rather unconventional way to represent cosine similarity. It’s definitely non-standard so we should add better edge-case handling in our codebase and clear comments in the code to explain this behavior. Thanks for catching that! 🙌

tjbck:
c79b975ad05c3ec982a1e78516be0c4c18ef391d

rgaricano:
from https://cookbook.chromadb.dev/core/collections/#collection-properties
...
 (Note: cosine value returns cosine distance rather then cosine similarity. Ie. values close to 0 means the embeddings are more similar.)
 ...
 
 by the way, update to chromadb 0.6?
 (it need change has_collection function in chroma.py to:
   def has_collection(self, collection_name: str) -> bool:
        # Check if the collection exists based on the collection name.
        collections = self.client.list_collections()
        return collection_name in [collections]
)

jannikstdl:
> > @juananpe fyi, cosine similarity of 1 means perfect match not 0. This definitely seems to be a required fix, thanks everyone!
> 
> Please, can you double check that it is also the case in ChromaDB?
> 
> What I understand from Chroma's documentation: https://docs.trychroma.com/docs/collections/configure#configuring-chroma-collections is that in Chroma the cosine function is (1 - the cosine) to preserve the semantic of smaller distance means closer.
> 
> See: image and: [chroma-core/chroma#1187 (comment)](https://github.com/chroma-core/chroma/issues/1187#issuecomment-1773806170)

@juananpe @tjbck 
can confirm.
Had a conversation with a chroma dev in October because of PR #5989

They indeed have a different cosine range than other vector DBs.. They say its because of the wikipedia article about that.
image

tjbck:
@jannikstdl yeah it should really just be called `Angular distance` not `Cosine similarity` 😅

PR #8385, opened 2025-01-07
fix: Pgvector vector column size check was failing on initialization …

jk-f5:
# Changelog Entry

### Description
- `metadata.reflect` throws an exception rather than returning an empty list. Refactor vector column check to account for this.

### Fixed
- Correctly check for the pgvector document_check table on first load.


tjbck:
Thanks!

GanizaniSitara:
TYVM

PR #8397, opened 2025-01-08
fix: Exception raised when new user tries to login through OAuth for the first time

kagesenshi:
# Changelog Entry

### Description

When logging in the first time through OAuth, following exception is raised:

```
  File "/app/backend/open_webui/utils/oauth.py", line 298, in <module>
    auth_manager_config.WEBHOOK_MESSAGES.USER_SIGNUP(user.name),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/open_webui/config.py", line 258, in __setattr__
    self._state[key].value = value
    ~~~~~~~~~~~^^^^^
KeyError: 'WEBHOOK_MESSAGES'
```

This is caused by `WEBHOOK_MESSAGES` supposed to be coming from `constants.py` while this code tries to take it from `auth_manager_config` . 

### Fixed

- Fixed exception raised when first login through OAuth

### Additional Information

This has been reported by another user a month ago, but seems to not be addressed. I'm skipping creating a discussion post due to it.

https://github.com/open-webui/open-webui/discussions/7672

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## []
+
+### FIXED
+
+- **OAuth user registration fixed**: Resolved a bug causing exception raised during first user login through OAuth
+
 ## [0.5.4] - 2025-01-05
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## []
-
-### FIXED
-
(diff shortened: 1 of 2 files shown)


tjbck:
Good catch, fixed with b3c7ecaea17fe6b9896bd2effb7efb5a8cb04464


Thanks!

kagesenshi:
Erm, before I send this patch, I did try that same approach in https://github.com/open-webui/open-webui/commit/b3c7ecaea17fe6b9896bd2effb7efb5a8cb04464 get the same exception during backend startup because `WEBHOOK_MESSAGES` is a constant and not a `PersistentConfig`, thus could not be attached to `auth_manager_config` as that expects `dict[str, PersistentConfig]` in its state dictionary. 

following exception is raised during backend startup with this change https://github.com/open-webui/open-webui/commit/b3c7ecaea17fe6b9896bd2effb7efb5a8cb04464

```
  File "/app/backend/open_webui/main.py", line 301, in <module>
    from open_webui.utils.oauth import oauth_manager
  File "/app/backend/open_webui/utils/oauth.py", line 60, in <module>
    auth_manager_config.WEBHOOK_MESSAGES = WEBHOOK_MESSAGES
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/open_webui/config.py", line 258, in __setattr__
    self._state[key].value = value
    ~~~~~~~~~~~^^^^^
KeyError: 'WEBHOOK_MESSAGES'
```

PR #7752, opened 2024-12-10
Add option to add database schema as env var

MadsLang:
# Changelog Entry

### Description

- When using Postgresql as database, there can be conflicts between reserved keywords like "user" and the table names, e.g. "user". I have described this in [this issue](https://github.com/open-webui/open-webui/issues/7750). To fix this problem, I added the option to set DATABASE_SCHEMA as an environment variable, that sets a default schema to use for all tables for the internal database. 

### Added

- The options to use the environment variable DATABASE_SCHEMA. This sets a default schema in the sqlalchemy ORM definition using the MetaData class described [here](https://docs.sqlalchemy.org/en/14/orm/declarative_tables.html#explicit-schema-name-with-declarative-table). 

### Additional information

I have also made [a PR](https://github.com/open-webui/docs/pull/314) updating the docs.

Issue #7750, opened 2024-12-10
Schema clashes when using Postgresql as database


MadsLang:
## TLDR
No explicit schema definition clashes with reserved keywords, when migrating database to postgresql. Would it be a good solution to add schema definition together with the DATABASE_URL environment variable? 

## Description of the problem

I am trying to host my database for OpenWebUI as a postgresql database in Cloud SQL in Google Cloud Platform. I am deploying the Docker container for OpenWebUI and specifying the database connection string in the DATABASE_URL environment variable (as specified at [the documentation](https://docs.openwebui.com/getting-started/advanced-topics/env-configuration/#database_url)). 

The connection seems to be working fine, but when trying to log into the UI, I am getting the error: *"The email or password provided is incorrect. Please check for typos and try logging in again."*. I am also using trusted header to authenticate (which have been working fine before when I was using native sqlite as database). 

The problem seems to be that when authenticating the user, the backend is querying the table "user" in the database (using the ORM defined by Users.get_user_by_email - see https://github.com/open-webui/open-webui/blob/29a271959556743e6deb4d55a5a982983335d7ab/backend/open_webui/apps/webui/routers/auths.py#L316). 

In Postgresql the name "user" is [a reserved keyword](https://www.postgresql.org/docs/current/sql-keywords-appendix.html), which means that this query is getting the sql-users in my deployment of postgres - rather than the user table defined by OpenWebUI. I have checked this by just querying my postgresql-database directly: `select * from user` whereas `select * from "user"` and `select * from public.user` gets the correct table of users. 

## Possible solution

I suspect that simply being able to specify a schema for postgresql would suffice. I think that optimally, it would just be another environment variable that you could add. The value could then be added when creating the sqlalchemy connection. For example adding it to `declarative_base` in (https://github.com/open-webui/open-webui/blob/main/backend/open_webui/apps/webui/internal/db.py#L102)

See also [defining explicit schema with the MetaData class](https://docs.sqlalchemy.org/en/14/orm/declarative_tables.html#explicit-schema-name-with-declarative-table)

## PR? 

I am willing to make a PR, if this would be a good solution to fixing this issue?

PR #8510, opened 2025-01-13
Add DATABASE_SCHEMA as env var

MadsLang:
# Changelog Entry

### Description

When using Postgresql as database, there can be conflicts between reserved keywords like "user" and the table names, e.g. "user". I have described this in https://github.com/open-webui/open-webui/issues/7750. To fix this problem, I added the option to set DATABASE_SCHEMA as an environment variable, that sets a default schema to use for all tables for the internal database.

### Added
The options to use the environment variable DATABASE_SCHEMA. This sets a default schema in the sqlalchemy ORM definition using the MetaData class described [here](https://docs.sqlalchemy.org/en/14/orm/declarative_tables.html#explicit-schema-name-with-declarative-table).

### Additional information

- I will also make a PR to update the docs.
- This change was previously submitted as [a PR](https://github.com/open-webui/open-webui/pull/7752), but has been rebased and tested.


tjbck:
Thanks, more testing wanted in dev!

sbe-arg:
where is the documentation for this?

MadsLang:
> where is the documentation for this?

Forgot to renses and get it into the docs-repo, as you can see here: https://github.com/open-webui/docs/pull/373

PR #8547, opened 2025-01-14
fix: File deletion doesn't properly clean up database entries

juananpe:
# Pull Request Checklist

- [x] **Changelog:** Ensure a changelog entry following the format 
- [x] **Code review:** Have you performed a self-review of your code

# Changelog Entry

Fix file deletion to properly clean up resources

### Description

When deleting files from either knowledge bases or chat, the system wasn't properly cleaning up all associated resources, particularly in the sqlite DB. This caused issues with re-uploading previously deleted files and left orphaned entries in the database and the uploads/ folder.

### Known issues

- There still are orphaned collections (file-UID) in Chroma when removing a file from the chat input. I don't like to add more coupling in routers/files.py with something like this in the `delete_file_by_id` function:
```python
        file_collection = f"file-{id}"
        if VECTOR_DB_CLIENT.has_collection(collection_name=file_collection):
            VECTOR_DB_CLIENT.delete_collection(collection_name=file_collection)
```
So I added a comment there ( # We should add Chroma cleanup here) as a placeholder to remember that we should act here.

### DIFF
--- a/backend/open_webui/routers/files.py
+++ b/backend/open_webui/routers/files.py
@@ -345,6 +345,8 @@ async def get_file_content_by_id(id: str, user=Depends(get_verified_user)):
 async def delete_file_by_id(id: str, user=Depends(get_verified_user)):
     file = Files.get_file_by_id(id)
     if file and (file.user_id == user.id or user.role == "admin"):
+        # We should add Chroma cleanup here
+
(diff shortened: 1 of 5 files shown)


tjbck:
LGTM, Thanks!

Issue #8425, opened 2025-01-09
Use OAuth Profile Picture (e.g., Microsoft Entra ID) as User Avatar in OpenWebUI


jeannotdamoiseaux:
# Feature Request

**Is your feature request related to a problem? Please describe.**
When users log in via OAuth providers (e.g., Microsoft Entra ID), OpenWebUI does not currently fetch and display their profile picture from the OAuth service. This limits personalization and requires users to manually upload a profile picture or stick with the default avatar.

**Describe the solution you'd like**
OpenWebUI should fetch and use the profile picture from the user's OAuth account after login. For example:

Request the profile picture during the OAuth flow (e.g., using Microsoft Graph's GET /me/photo/$value for Microsoft Entra ID) after obtaining appropriate permissions like User.Read.
Save the profile picture in the user’s OpenWebUI profile and display it in relevant places (navigation bar, comments, etc.).

**Describe alternatives you've considered**
- Manual profile picture uploads (less seamless).
- Fixed default avatars (non-personalized).
- Gravatar integration (requires external accounts and additional configuration by users).

**Additional context**
I’m new to the repo but willing to help implement this feature. A pointer on where OAuth and user profile logic resides in the codebase would be greatly appreciated. Thank you!

rgaricano:
https://github.com/open-webui/open-webui/blob/1dfb479d367e5f5902f051c823f9aef836e04791/backend/open_webui/utils/oauth.py#L7

jeannotdamoiseaux:
> https://github.com/open-webui/open-webui/blob/1dfb479d367e5f5902f051c823f9aef836e04791/backend/open_webui/utils/oauth.py#L257

That’s strange—I’m not sure why it’s not working on my end. Could there be any specific configuration or setup difference that might be causing this issue?

robert-mcdermott:
@jeannotdamoiseaux, I'm currently using LDAP against a local Active Directory for authentication, but need to move to SAML/OAuth/SSO against our Entra ID tenant. Based on on you feature request, it appears that you have already gotten this working. The Open WebUI docs don't explain who I can accomplish this, can you point me in the right direction? Thanks.

jeannotdamoiseaux:
[...]

PR #8566, opened 2025-01-15
Add functionality to retrive picture for microsoft oauth

rragundez:
Microsoft oauth does not send the `picture` claim like other providers do. This PR enables the retrieval of the profile picture in those cases. 
Is quite simple TBH, there is a standard given by the Microsfot Graph API to retrieve it:
- https://learn.microsoft.com/en-us/graph/api/profilephoto-get?view=graph-rest-1.0&tabs=http#request

It is not a breaking change as the other providers still behave exactly in the same way.

I have tested  with my company credentials and it works. 

Solves: #8425

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -487,6 +487,7 @@ def load_oauth_providers():
             "server_metadata_url": f"https://login.microsoftonline.com/{MICROSOFT_CLIENT_TENANT_ID.value}/v2.0/.well-known/openid-configuration",
             "scope": MICROSOFT_OAUTH_SCOPE.value,
             "redirect_uri": MICROSOFT_REDIRECT_URI.value,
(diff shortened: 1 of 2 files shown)


nikmaksimenk0 (approved):

tjbck:
LGTM, Appreciate it!

rragundez:
Hi @tjbck, how often do you release or when do you think I will see this change incorporated into a release?

jeannotdamoiseaux:
I've implemented this, but it doesn't seem to be working as expected on my end. Could you please clarify if there are any additional changes or configurations needed to get this working?

rragundez:
@jeannotdamoiseaux can you be more specific on what is the error you see in the backend?

rragundez:
@jeannotdamoiseaux apologies, I somehow forgot to commit the part where I add the acess token in the headers. I made this PR to solve that. 
#8608

PR #8576, opened 2025-01-15
feat: User Feedback UI for Admins

Vojtech-Siler:
# Changelog Entry

### Description

- This feature was added because the Feedback functionality does not allow administrators to see user feedbacks. They could see that a model somehow won or lost, but could not collect the specific information.
- This feature allows admins to go to the specific chats from the Feedbacks section in the Admin Panel using a new column with the chat title in it, and see the highlited messages that have user feedback on them. These messages are highlited by a border (green for positive, red for negative feedback), a badge, and have the necessary feedback information under them, including the rating, reason and comment. This allows admins to receive straight forward feedback they can work on and improve the particular models if needed.
- This feedback is added into the /s/ page used for sharing read-only chats, it is only fetched if the current user is an administrator, and has a URL property to make sure the feedback doesn't automatically appear when sharing chats outside of the Feedbacks section.
- The PR includes 3 new migrations - one to add the new column chat_id to feedback (I know it's already in the meta property, however meta being a JSON, I could not find a way to filter feedbacks from the DB using that), one to migrate the chat_id information from meta to the new column for already existing feedbacks, and one for making the chat_id column required. This has been split into 3 migrations due to safety, because of a previous experience of a migration not being able to add a column and populate it all in one migration.
- The flow is as follows: In the Evaluations/Feedbacks section in the Admin Panel, the chat title where the feedback occured is shown. When clicked, the chat is shown using the /s/ sharing page, all feedbacks for that particular chat id are fetched if the user is admin, the URL property showFeedback is added, and it is determined for each message if it has feedback or not. If it does, style and additional info are added to it.

### Added

- Required chat_id column to the feedback table in DB
- Methods in DB, routing and evaluation index.ts to fetch feedbacks for a particulat chat id
- New column in the Feedbacks section with clickable chat titles
- Feedback visualization for admins when clicked through the Feedbacks section, highlighting messages with feedback and showing necessary information about the feedback

### Changed

- /s/ chat sharing page to include feedbacks, if clicked from the Feedbacks section (using a URL parameter) and only if user is admin
- feedback db model, added chat_id column populated from the meta property

### Breaking Changes

---

### Additional Information

- The screenshots and a video bellow show how the feature looks and how it works.

### Screenshots or Videos
image
image

https://github.com/user-attachments/assets/7c86a40c-f707-48e0-ac04-fd0a82637a7a

### DIFF
--- a/backend/open_webui/migrations/versions/790ecce592df_add_chat_id_to_feedback.py
+++ b/backend/open_webui/migrations/versions/790ecce592df_add_chat_id_to_feedback.py
@@ -1,0 +1,30 @@
+"""add_chat_id_to_feedback
+
+Revision ID: 790ecce592df
+Revises: 3781e22d8b01
+Create Date: 2025-01-08 15:13:16.063379
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+import open_webui.internal.db
+
+revision: str = '790ecce592df'
+down_revision: Union[str, None] = '3781e22d8b01'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade():
+
+    # Add chat_id column to feedback
+    op.add_column(
+        "feedback",
+        sa.Column("chat_id", sa.Text(), nullable=True),
+    )
+
+
+def downgrade():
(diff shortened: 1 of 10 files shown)


tjbck:
Definitely a great start, this won't be directly merged but will be kept opened as a reference. Thanks for the hard work!

Issue #5763, opened 2024-09-27
enh: allow to use S3 for uploaded files


hongbo-miao:
**Is your feature request related to a problem? Please describe.**

@tjbck clarified Postgres can be used for metadata (Open WebUI config and user chat history) at https://github.com/open-webui/helm-charts/issues/83#issuecomment-2379241585 except for user uploaded files.

**Describe the solution you'd like**
It would be great to support to use S3 for uploaded files.

**Describe alternatives you've considered**
Use EBS or EFS for uploaded files.

**Additional context**
None

tjbck:
PR Welcome!

ZhangChaoWN:
I'm willing to contribute to this feature.

DucNgn:
@ZhangChaoWN 
Thanks for taking this on! I'm interested in contributing to this feature as well.
Lmk if you need help finishing it!

ZhangChaoWN:
@DucNgn, I'm super excited that you're interested in helping out on this feature!

Here’s a quick overview of what's been done so far and what still needs attention.

Completed:
- In process_doc: Download file from S3
- In upload_file: Upload file to S3
- /{id}/content API: Download file to the client

TODO:
- Modify the configuration to enable switching between saving files to S3 or the local file system (currently hardcoded). And Retrieve the S3 bucket name from the configuration.
- Consider adding documentation about this feature.
- In process_doc, would it be better to delete the temporary file downloaded from S3 after storing it in the vector database?

I have forked this project and pushed my code changes. If you're interested in collaborating on the coding work, feel free to merge my code or ask me to merge yours. Feel free to point out any mistakes or suggest ways to improve the feature. If you have any ideas for additional tasks, Please feel free to share them as well.

ZhangChaoWN:
Squashed commits and rebased onto the latest main branch in the forked repo

LeoLiuYan:
The content of the uploaded file will be indexed in the vector database; is it still necessary to upload it to  S3?   @tjbck @ZhangChaoWN

tjbck:
@gmemstr Good catch, should be addressed in dev! More testing wanted here!

nickfixit:
What about mounting a JuiceFS filesystem?

davizucon:
Hey ! well done, I'm looking forward this feature :)

So, taking a look at the code, I started fixing it, but what do you think if we follow the same approach as for vector/dbs? 
[...]

PR #8580, opened 2025-01-15
Split the storage providers into separate classes in preparation for adding more cloud providers

rragundez:
Split the storage providers into separate classes in preparation for adding more cloud providers.

Suggested in #5763 

Once merged after any modifications I will add the GCS storage provider which is actually the reason for me to contribute as I need it, to not hack my way into it.

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -9,22 +9,22 @@ from urllib.parse import urlparse
 
 import chromadb
 import requests
-import yaml
-from open_webui.internal.db import Base, get_db
+from pydantic import BaseModel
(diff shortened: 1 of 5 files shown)


tjbck:
Amazing addition to our codebase! The PR generally looks good to me, but I have to ask: is this PR tested?

tjbck:
Also, please target our `:dev` branch going forward!

rragundez:
Noted. @tjbck I'll add some tests, I think I have most of them already laying around somewhere in other projects so no problem.

tjbck:
Looks amazing, Thanks for all the hard work!

PR #8610, opened 2025-01-16
🌟 Chore: Add Svelte Inspector for Easier Development 🌟

excho0:
So I thought it might be a great idea to add Svelte Inspector to the project. It's a developer tool that makes working on Svelte apps much easier by enabling us to inspect components, jump directly to their source code, and debug more efficiently—all right from the browser.

### Why does it matter?

- 🕵️ **Easier Debugging**: It helps you see the structure of your components and how they interact in real-time.
- 🚀 **Faster Development**: Jumping to the source code of a component is just a click away—no need to dig through folders manually.
- 🤝 **Better Collaboration**: It’s helpful for both new and experienced team members to navigate and understand the app's structure quickly.
- 🎯 **Focus on Code**: Spend more time building features and less time hunting for where things are defined.

Adding it is lightweight and only affects the development environment, so it won't impact the production build.

Cheers! 🚀

### DIFF
--- a/svelte.config.js
+++ b/svelte.config.js
@@ -16,8 +16,16 @@ const config = {
 			fallback: 'index.html'
 		})
 	},
+	vitePlugin: {
+		inspector: {
+			toggleKeyCombo: 'meta-shift', // Key combination to open the inspector
+			holdMode: false,             // Enable or disable hold mode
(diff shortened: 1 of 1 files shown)


tjbck:
LGTM, Thanks!

PR #7964, opened 2024-12-19
Fix Redis Lock Expiration in periodic_usage_pool_cleanup 

jk-f5:
# Changelog

### Fixed
- The `periodic_usage_pool_cleanup` coroutine was encountering an error due to the Redis lock expiring before it could be renewed in some environments.To resolve this timing issue, the lock acquisition was moved from module initialization into the `periodic_usage_pool_cleanup coroutine` itself.

PR #8660, opened 2025-01-17
Remove unnecessary decode statement from release_lock method

jk-f5:
# Changelog Entry

### Description
- Remove unnecessary decode statement from release_lock method. Since our Redis instance is created with `decode_responses=True` this isn't required. `decode` is also not a supported string method in python3.

Fixes:
- Address stack trace on Open Webui shutdown with Redis support enabled.

See #7964 for more details


tjbck:
Thanks!

PR #8712, opened 2025-01-21
Improve s3 provider tests

rragundez:
Move client and other S3 objects instantiation to the __init__ of the test class to avoid potential bugs of dealing with Class attributes


tjbck:
Thanks!

PR #8793, opened 2025-01-23
fix: model ids not being applied

tjbck:


NeuroWhAI:
I had an issue related to the ollama, and I think this might fix my issue.
Looking forward to the next release!

PR #9016, opened 2025-01-28
WIP: rectify bad title generation issue in reasoning models

kagesenshi:
# Changelog Entry

### Description

- When using reasoning model, user gets the reasoning block as the chat title. This patch

### Fixed

-  bad title generation issue in reasoning models

### Screenshots or Videos

image

### DIFF
--- a/backend/open_webui/routers/tasks.py
+++ b/backend/open_webui/routers/tasks.py
@@ -17,6 +17,7 @@ from open_webui.utils.task import (
     moa_response_generation_template,
 )
 from open_webui.utils.auth import get_admin_user, get_verified_user
+from open_webui.utils.misc import strip_reasoning_tag
 from open_webui.constants import TASKS
 
 from open_webui.routers.pipelines import process_pipeline_inlet_filter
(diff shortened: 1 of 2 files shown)


kagesenshi:
changing this fix to WIP .. the fix only works when conversation is short. if conversation is long, reasoning tokens still overtake the 200 char limit.

ZTGeng:
Met the same issue. Just my 2 cents: is it possible to modify the DEFAULT_TITLE_GENERATION_PROMPT_TEMPLATE in the config file by adding an instruction like "skip the reasoning and directly generate the title"?

Zambito1:
For the moment I just set my title generation model to something small enough to keep in memory along with the main model, which is probably better to do anyways. I'm using llama3.2 1b for title generation.

tjbck:
Closing in favour of 5420c165c60edb23a841f1b469d2fec381775cbf, Appreciate the effort!

kagesenshi:
@tjbck cool

btw .. you may want to strip the reasoning `detail` html from input given to the title generation prompt .. because that might confuse title generation to something that have nothing to do with the actual question-answer ..

PR #9088, opened 2025-01-29
**feat** Support CoT models for title generation.

joshuawu:
# Pull Request Checklist

### Note to first-time contributors: Please open a discussion post in [Discussions](https://github.com/open-webui/open-webui/discussions) and describe your changes before submitting a pull request.

**Before submitting, make sure you've checked the following:**

- [x] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [x] **Description:** Provide a concise description of the changes made in this pull request.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ ] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [x] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [ ] **Testing:** Have you written and run sufficient tests for validating the changes?
- [x] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [x] **Prefix:** To clearly categorize this pull request, prefix the pull request title, using one of the following:
  - **BREAKING CHANGE**: Significant changes that may affect compatibility
  - **build**: Changes that affect the build system or external dependencies
  - **ci**: Changes to our continuous integration processes or workflows
  - **chore**: Refactor, cleanup, or other non-functional code changes
  - **docs**: Documentation update or addition
  - **feat**: Introduces a new feature or enhancement to the codebase
  - **fix**: Bug fix or error correction
  - **i18n**: Internationalization or localization changes
  - **perf**: Performance improvement
  - **refactor**: Code restructuring for better maintainability, readability, or scalability
  - **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc.)
  - **test**: Adding missing tests or correcting existing tests
  - **WIP**: Work in progress, a temporary label for incomplete or ongoing work

# Changelog Entry

### Description

- Added a new setting to enable title generation by CoT models.
- Added fallback logic for title generation by CoT models (when the setting is off).

### Added

- Title generation fallback behavior when using a CoT model
- Interface setting "Title generation model uses chain of thought"

### Changed

- Refactored `max_tokens`/`max_completion_tokens` while adding the new behavior
- Truncate titles to 50 characters immediately before title update
- Refactored title fallback logic
- Moved reasoning tags names to a constant

### Deprecated

- None

### Removed

[...]

PR #9095, opened 2025-01-30
**fix** Fix title generation JSON enforcement changes

joshuawu:
# Pull Request Checklist

- [ ] **Testing:** Have you written and run sufficient tests for validating the changes?
- [x] **Prefix:** To clearly categorize this pull request, prefix the pull request title, using one of the following:

# Changelog Entry

### Description

Fix bug where task would crash if it generated a summary without strict JSON format.
However, this only mitigates the root issue: see **Additional Information**.

### Changed

- Refactored regex for stripping thoughts before title generation (to be more readable)

### Fixed

- Fix bug where task would crash if it generated a summary without strict JSON format

### Security

- Fix bug where task would crash if it generated a summary without strict JSON format

### Breaking Changes

---

### Additional Information

The cause of the new issue is due to the new change to enforce a strict JSON response when generating titles.
This fix only prevents the generation task from crashing. It simply ignores the bad response and falls back to the chat history.

## Example of query failure
This is an easy way to get title generation to fail:
Screenshot 2025-01-30 085816
Which produces the fallback title:
Screenshot 2025-01-30 085822
(Note: The issue is with the title generation response, not directly the initial response above.)

Also, I've seen it fail even with more unrelated prompts like topics about medicine, science, or history. With many models, sometimes the model disobeys and uses Markdown instead of JSON. An example was `**Title:** 💊 Health and Medicine`

For further context, earlier I suggested having an explicit setting for chain of thought models (see the closed PR https://github.com/open-webui/open-webui/pull/9088), but it was not preferred.

### DIFF
--- a/backend/open_webui/routers/tasks.py
+++ b/backend/open_webui/routers/tasks.py
@@ -167,13 +167,11 @@ async def generate_title(
     # Remove reasoning details from the messages
     for message in messages:
         message["content"] = re.sub(
-            r"<details\s+type=\"reasoning\"[^>]*>.*?<\/details>",
+            r'<details\s+type="reasoning".*?>.*?</details>',
             "",
             message["content"],
-            flags=re.S,
+            flags=re.DOTALL,
         ).strip()
-
-    print(messages)
 
     content = title_generation_template(
         template,
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -909,7 +909,7 @@ async def process_chat_response(
                                     "title", "New Chat"
                                 )
                             except Exception as e:
-                                pass
+                                title = None
 
                             if not title:
                                 title = messages[0].get("content", "New Chat")


joshuawu:
Minor edit to description: moved regex change from fix to refactor due to a misinterpretation.

joshuawu:
Another example of failing to use JSON (which this PR will ignore instead of crash):

```
<think>
Alright, I need to come up with a concise title and an emoji for this chat history. The user provided examples that are 3-5 words long with emojis. Let me look at what was discussed.

First, there's a mention of "this sentence," so it's about writing or composing something. Then the assistant talks about points: one, two, three. That sounds like a list or structured information. So the main themes are writing and lists.

I need an emoji that represents both writing and lists. The pencil emoji 📝 fits writing, and adding a list usually involves checks or numbers, but since we can't use symbols, maybe another approach. Alternatively, thinking about composition or structure, like building something, so the hammer and nails emoji might not fit. Wait, perhaps the pencil is sufficient because it's directly related to writing.

So combining that with the idea of points, I can phrase it as "Writing Points Clearly." That captures both the writing aspect and the list points. The title should be concise, so "Writing Points Clearly" works well within 3-5 words. The emoji 📝 is clear and relevant.
</think>

The main themes discussed were about writing and structuring information in a clear manner.

**Title:** 📝 Writing Points Clearly
```

It's not a huge issue but the current approach will hit this sporadically.

tjbck:
Updated, thanks!

PR #9163, opened 2025-01-31
Bug fix, configurable default if using oauth via the `DEFAULT_USER_ROLE` env variable

rragundez:
TBh I think this is a straight forward one @tjbck 

For some reason currently the default role behaviour is different if using the OAuth roles. This PR addresses that, it changes the hardcoded default role for the one available for the user to set via environment variables. If the user does not set the default role via the environment variables then it will default to the same behaviour as before.


tjbck:
LGTM, Thanks!

PR #9195, opened 2025-01-31
[EASY] Set default picture to user.png

rragundez:
If the user's picture cannot be retrieve via the URL provided by OAUTH then it is set to en empty string instead to the default `user.png` picture. This PR addresses that.


tjbck:
Thanks!

PR #9398, opened 2025-02-05
Optional user metadata forwarding to LiteLLM for cost tracking

elabbarw:
# feat: Optional user metadata forwarding to LiteLLM for cost tracking

## Description
This PR introduces an opt-in feature to forward user metadata to LiteLLM for enhanced cost tracking. A new environment variable enables the inclusion of additional user information in the payload sent to LiteLLM Proxy.

## Changes
- Added new environment variable `ENABLE_FORWARD_USER_INFO_LITELLM_TAGS` in env.py
- Implemented logic to include user metadata in payloads when enabled
- Modified payload structure in `routers/openai.py`, `routers/audio.py`, `routers/image.py`, and `retrieval/utils.py`

## Metadata Details
When enabled, the following tags are included:
- Source: openwebui
- Model name
- User email
- User name

## Implementation
- The feature is opt-in, requiring explicit activation via the environment variable
- Payload structure now includes a `metadata` object with a list of 'tags' when enabled

## Testing
- Conducted comprehensive testing with various user scenarios to ensure accurate metadata generation and transmission

## Privacy Considerations
- Feature is disabled by default to maintain user privacy
- Only activates when `ENABLE_FORWARD_USER_INFO_LITELLM_TAGS` is set

## Screenshots
User Metadata Forwarding

## Impact
This feature enhances cost tracking capabilities through LiteLLM while maintaining user privacy as an opt-in feature.


tjbck:
Should be implemented as a Function!

elabbarw:
> Should be implemented as a Function!

That would account for chat, but how would you track the costs of background task runs for titles and tags? What about embedding calls for adding files and memories, as well as querying them? Image generation might be captured in chat, but clicking the small image generation icon under a chat reply would be missed.

elabbarw:
Hey @tjbck , will I be able to retrieve metadata from all background tasks, including tags, titles, and audio and image runs, using functions or pipelines?

PR #9401, opened 2025-02-05
remove <think/> for title generation

WeiLai5432:
# Changelog Entry

### Description
remove <think> thinking </think> content for title generation of CoT models


tjbck:
This endpoint is not being used.

PR #9416, opened 2025-02-05
feat(ui): implement domain filter list for web search settings

abdalrohman:
Implement Domain Filter Lists in Web Search Settings

This allows for dynamic management of the lists without requiring code changes.


tjbck:
Thanks!

PR #9456, opened 2025-02-06
fix: check for email claim before skipping userinfo endpoint

### LINKED ISSUE #9452: OIDC login fails when email claim only available via userinfo endpoint
# OIDC authentication issue: userinfo endpoint not called when token contains partial claims

## Description
When an OIDC provider returns partial userinfo claims in the token but provides complete claims via the userinfo endpoint, authentication fails with "email is missing" error.

## Environment
- Open WebUI version: v0.5.10 (latest)
- OIDC Provider: Duke University OIDC
- OAuth Flow: Authorization Code Grant
- Affected File: `/app/backend/open_webui/utils/oauth.py`

## Current Behavior
The code checks for userinfo in the token and only calls the userinfo endpoint if no data is present. However, some OIDC providers (like Duke) include partial claims in the token but require the userinfo endpoint call for complete profile information.

## Expected Behavior
The code should call the userinfo endpoint if required claims (like email) are missing, even if some userinfo data exists in the token.

## Debugging Evidence
Debug logs show the issue clearly:

1. Token contains partial userinfo:
```python
'userinfo': {
    'sub': 'user123@duke.edu',
    'aud': 'client-id-123',
    'kid': 'rsa1',
    # ... but no email claim
}
```

2. Userinfo endpoint provides complete data:
```python
{
    'sub': 'user123@duke.edu',
    'netID': 'user123',
    'name': 'Example User, Ph.D.',
    'email': 'user@duke.edu',
    'email_verified': False
    # ... full profile data
}
```

## Current Code
In `/app/backend/open_webui/utils/oauth.py`:
```python
user_data: UserInfo = token.get("userinfo")
if not user_data:
    user_data: UserInfo = await client.userinfo(token=token)
```

## Proposed Fix
```python
user_data: UserInfo = token.get("userinfo")
if not user_data or "email" not in user_data:
    user_data: UserInfo = await client.userinfo(token=token)
```

## Impact
This affects any OIDC provider that:
1. Includes partial claims in the token response
2. Requires userinfo endpoint call for complete profile data
3. Follows the OIDC spec pattern of separating token and userinfo responses

The current implementation prevents users from these providers from logging in, even when all required scopes are properly configured.

mitomac:
# Fixes #9452

## Description
Fixes an issue where OIDC authentication fails when the email claim is only available via the userinfo endpoint but not in the token's userinfo data. This affects OIDC providers (like Duke) that follow the pattern of providing complete user information through the userinfo endpoint rather than in the token response.

### Fixed
- OIDC authentication now properly fetches email claim from userinfo endpoint when not present in token data
- Prevents authentication failures with OIDC providers that separate token and userinfo responses

### Changed
- Modified the userinfo data fetch logic to check specifically for required email claim

## Implementation Details
The fix modifies the OIDC authentication flow to check not just for the presence of userinfo data in the token, but specifically for the required email claim. If the email claim is missing, it will make a call to the userinfo endpoint to fetch complete user data.

### Code Changes
In `/app/backend/open_webui/utils/oauth.py`:
```python
# Before
user_data: UserInfo = token.get("userinfo")
if not user_data:
    user_data: UserInfo = await client.userinfo(token=token)

# After
user_data: UserInfo = token.get("userinfo")
if not user_data or "email" not in user_data:
    user_data: UserInfo = await client.userinfo(token=token)

### DIFF
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -193,7 +193,7 @@ class OAuthManager:
             log.warning(f"OAuth callback error: {e}")
             raise HTTPException(400, detail=ERROR_MESSAGES.INVALID_CRED)
         user_data: UserInfo = token.get("userinfo")
-        if not user_data:
+        if not user_data or "email" not in user_data:
             user_data: UserInfo = await client.userinfo(token=token)
         if not user_data:
             log.warning(f"OAuth callback failed, user data is missing: {token}")


tjbck:
Thanks!

PR #9486, opened 2025-02-06
Use DB for generated images

rragundez:
At the moment images generated have several issues:
- they are not managed by any external source
- their behaviour is different from other uploaded files
- there is no auth checks for users and images are just passed to the src if the img tag from the cache dir so they are basically public
- If multiple replicas since the logic depends on having them on disk it will fail to load since the image might have been written in a different replica when created.

This change makes the images to be treated exactly in the same way as upload files and backed up by the same mechanism. 
All the logic is pushed to the already proven 'files.py' logic. With this change all the problems mentioned above should be solved.

I tested using dall-e-3 openai model.

related to:  #5109 and #8186

cc @jk-f5

### DIFF
--- a/backend/open_webui/routers/files.py
+++ b/backend/open_webui/routers/files.py
@@ -56,11 +56,11 @@ def upload_file(
                     "id": id,
                     "filename": name,
                     "path": file_path,
-                    "data": file_metadata,
                     "meta": {
                         "name": name,
                         "content_type": file.content_type,
                         "size": len(contents),
+                        "data": file_metadata,
                     },
                 }
             ),
--- a/backend/open_webui/routers/files.py
+++ b/backend/open_webui/routers/files.py
@@ -3,30 +3,22 @@ import os
 import uuid
 from pathlib import Path
 from typing import Optional
-from pydantic import BaseModel
-import mimetypes
 from urllib.parse import quote
 
-from open_webui.storage.provider import Storage
-
+from fastapi import APIRouter, Depends, File, HTTPException, Request, UploadFile, status
+from fastapi.responses import FileResponse, StreamingResponse
+from open_webui.constants import ERROR_MESSAGES
+from open_webui.env import SRC_LOG_LEVELS
 from open_webui.models.files import (
     FileForm,
     FileModel,
     FileModelResponse,
     Files,
 )
-from open_webui.routers.retrieval import process_file, ProcessFileForm
-
-from open_webui.config import UPLOAD_DIR
-from open_webui.env import SRC_LOG_LEVELS
-from open_webui.constants import ERROR_MESSAGES
-
-
-from fastapi import APIRouter, Depends, File, HTTPException, UploadFile, status, Request
-from fastapi.responses import FileResponse, StreamingResponse
-
-
+from open_webui.routers.retrieval import ProcessFileForm, process_file
+from open_webui.storage.provider import Storage
 from open_webui.utils.auth import get_admin_user, get_verified_user
+from pydantic import BaseModel
 
 log = logging.getLogger(__name__)
 log.setLevel(SRC_LOG_LEVELS["MODELS"])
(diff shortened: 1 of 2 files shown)


jk-f5:
I'm not a maintainer (just an enthusiast), so I can't approve this PR but I do have a couple of comments.

backend/open_webui/routers/images.py
```diff
@@ -39,7 +34,7 @@
 
 
 @router.get("/config")
-async def get_config(request: Request, user=Depends(get_admin_user)):
+async def get_def(request: Request, user=Depends(get_admin_user)):
```
jk-f5, line 37: This rename seems less clear. Why not leave it as `get_config`?

backend/open_webui/routers/images.py
```diff
@@ -500,13 +465,23 @@ async def image_generations(
...
+                    headers={
+                        "content-type": content_type,
+                    },
+                )
+                file_item = upload_file(request, file, user)
+                url = request.app.url_path_for(
+                    "get_file_content_by_id", id=file_item.id
+                )
+                images.append({"url": url})
+                file_body_path = IMAGE_CACHE_DIR.joinpath(f"{file_item.id}.json")
```
jk-f5, line 481: So we're still storing the image json metadata to disk?

backend/open_webui/routers/images.py
rragundez: TBH I have no clue why that change. I put it back.

backend/open_webui/routers/images.py
rragundez: Yes, I have no idea why that file is needed. I will adapt to also store it.

rragundez:
> I'm not a maintainer (just an enthusiast), so I can't approve this PR but I do have a couple of comments.

I understand but help with a review would be great before pinging Tim for a final review. If you can also help me actually test that would be great. With and without a bucket as storage provider. 

I tested with and without bucket using the `openai` image generation with `dall-e-3` but there are other 3 types I would like to test as well. If you could help me with any of that would be great. Thanks.

backend/open_webui/routers/images.py
rragundez: I have changed the code such that there is no need to store a separate file but it will store this information together with the file object in the DB, since it is already supported by the File type in the DB

rragundez:
Hi @tjbck  can you also have a look? I think it is almost there.

backend/open_webui/routers/images.py
tjbck: Saving these metadata allows for image model training, which we most definitely should support out of the box.

rragundez: Hi @tjbck the metadata it is stored in the DB together with the other file properties like filename and url, I think this approach is better than managing 2 separate files. What do you think?

rragundez:
Hi @jk-f5  can you review again please? thanks

backend/open_webui/routers/files.py
```diff
@@ -61,6 +56,7 @@ def upload_file(
                     "id": id,
                     "filename": name,
                     "path": file_path,
+                    "data": file_metadata,
```
tjbck, line 59: `metadata` would be better to be placed inside of "meta" field (e.g. ["meta"]["data"])

BhavanaSBhat:
@rragundez , Previously, image generation failed in a multi-instance deployment of OpenWebUI on GCP because images were stored in a local cache. Since implementing sticky sessions was not ideal due to scaling concerns, we opted to store generated images in a GCS bucket for centralized storage. This ensures consistency across instances without affecting scalability. 

I have tested this solution and approve this change.

rragundez:
@BhavanaSBhat 
Using a bucket and implementing a custom way of backing up images breaks the pattern already established to store files, without mentioning the custom auth mechanism that has to be implemented for users.
This PR follows the already established patterns of using the `files.py` which make use of the Database patterns, the Storage provider patterns and the auth patterns without having to maintain extra logic for this for images.
Please review the PR.

backend/open_webui/routers/files.py
rragundez: Done, @tjbck  please have a another look.

tjbck:
More testing wanted after it's been merged to dev, thanks!

PR #9505, opened 2025-02-06
Update jina_search.py

binxn:
### Description

Updated Jina's search function in order to use POST and make use of the result count passed by the user, as well as adding a header to not return the images possibly found on the search reducing token usage.

### Added

- `"Content-Type": "application/json"` header
- `"X-Retain-Images": "none"` header
- Added a `payload `variable containing the query and result count (capped at 10)

### Changed

- Replaced GET request by a POST request

### Removed

- `[:count]` cap on the `data["data"]` loop because we are already asking Jina to return the result count specified by the user

---

### Additional Information

- Jina supports a max of 10 result count


tjbck:
Thanks!

PR #9542, opened 2025-02-07
Implement realtime filter

XingjianXie:
#7223

### DIFF
--- a/backend/open_webui/utils/chat.py
+++ b/backend/open_webui/utils/chat.py
@@ -44,6 +44,10 @@ from open_webui.utils.response import (
     convert_response_ollama_to_openai,
     convert_streaming_response_ollama_to_openai,
 )
+from open_webui.utils.filter import (
+    get_sorted_filter_ids,
+    process_filter_functions,
+)
 
 from open_webui.env import SRC_LOG_LEVELS, GLOBAL_LOG_LEVEL, BYPASS_MODEL_ACCESS_CONTROL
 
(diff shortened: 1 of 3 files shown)


XingjianXie:
This new function hook provides huge extensibility for open-webui while maintaining full compliance with OpenAI's API.

Q: Why not use `Pipe` instead?

- Existing `Filter` functions naturally extend to support real-time streaming. This ensures consistent content handling throughout the generation, preventing abrupt content disappearance due to `outlet` function.

- For developers consuming standard OpenAI APIs, this solution eliminates the need for redundant Pipe configurations solely for response customisation, such as customising rendering effect.

- This eliminates further issues regarding to API deviations, because we offer this new option for customisation.

XingjianXie:
Fixed a minor issue regarding to `tag_content_handler` behaviour. 
- `AA<tag>` -> `content_blocks = ["AA", ""]` (correct)
- `AA<tag>BB` -> `content_blocks = ["AA", "BB"]` (correct)
- `<tag>BB` -> `content_blocks = ["BB", "BB"]` (incorrect)
This fix it to:
- `AA<tag>` -> `content_blocks = ["AA", ""]` (correct)
- `AA<tag>BB` -> `content_blocks = ["AA", "BB"]` (correct)
- `<tag>BB` -> `content_blocks = ["", "BB"]` (correct)

tjbck:
Could we split this to two PRs? Atomic PRs would be appreciated!

XingjianXie:
Split into #9617 and #9619

PR #9241, opened 2025-02-02
feat: Better support for reasoning content

zeyugao:
# Pull Request Checklist

### Note to first-time contributors: Please open a discussion post in [Discussions](https://github.com/open-webui/open-webui/discussions) and describe your changes before submitting a pull request.

**Before submitting, make sure you've checked the following:**

- [x] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [x] **Description:** Provide a concise description of the changes made in this pull request.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ ] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [ ] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [x] **Testing:** Have you written and run sufficient tests for validating the changes?
- [ ] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [ ] **Prefix:** To cleary categorize this pull request, prefix the pull request title, using one of the following:
  - **BREAKING CHANGE**: Significant changes that may affect compatibility
  - **build**: Changes that affect the build system or external dependencies
  - **ci**: Changes to our continuous integration processes or workflows
  - **chore**: Refactor, cleanup, or other non-functional code changes
  - **docs**: Documentation update or addition
  - **feat**: Introduces a new feature or enhancement to the codebase
  - **fix**: Bug fix or error correction
  - **i18n**: Internationalization or localization changes
  - **perf**: Performance improvement
  - **refactor**: Code restructuring for better maintainability, readability, or scalability
  - **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc.)
  - **test**: Adding missing tests or correcting existing tests
  - **WIP**: Work in progress, a temporary label for incomplete or ongoing work

# Changelog Entry

### Description

- Sometimes the model does not output `\n` after `<think>` tag, which makes open webui fail to detect the reasoning.
image

### Changed

- Remove `\n` the enforcement when checking `start_tag`

### Additional Information

Follow up of https://github.com/open-webui/open-webui/pull/9225

I have made it more reasonable and support for the reasoning token from the api

https://github.com/open-webui/open-webui/discussions/9193#discussioncomment-12028093

### Screenshots or Videos

- Self hosted, `<think>\n </think>`

image

[...]

PR #9617, opened 2025-02-07
Implement realtime filter

XingjianXie:
#9542 (midlet)

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -1444,6 +1444,28 @@ async def process_chat_response(
                         try:
                             data = json.loads(data)
 
+                            extra_params = {
+                                "__event_emitter__": get_event_emitter(metadata),
+                                "__event_call__": get_event_call(metadata),
+                                "__user__": {
+                                    "id": user.id,
+                                    "email": user.email,
+                                    "name": user.name,
+                                    "role": user.role,
+                                },
+                                "__metadata__": metadata,
+                                "__request__": request,
+                                "__model__": form_data.get("model"),
+                            }
+
+                            data, _ = await process_filter_functions(
+                                request=request,
+                                filter_ids=get_sorted_filter_ids(form_data.get("model")),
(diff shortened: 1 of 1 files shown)


backend/open_webui/utils/middleware.py
```diff
@@ -1356,6 +1356,27 @@ async def stream_body_handler(response):
...
+                                "__event_emitter__": get_event_emitter(metadata),
+                                "__event_call__": get_event_call(metadata),
+                                "__user__": {
+                                    "id": user.id,
+                                    "email": user.email,
+                                    "name": user.name,
+                                    "role": user.role,
+                                },
+                                "__metadata__": metadata,
+                                "__request__": request,
```
XingjianXie, line 1369: User might also wish to have more fine grain control over content rendering. We can further add `__content__` and `__ content_blocks__` to `extra_params`. Then user Filter can read from data, parse data and modify `content` and `content_block`, then simply remove delta from data.

tjbck:
Will be taking sometime to reflect on this feature.

XingjianXie:
Thanks. A separate refactoring pull request is ready to merge if you wish for just removing duplication code between existing `inlet` and `outlet`.

Spectrelai:
@tjbck Hey, how's the consideration of this PR going? Since more and more sites are using attributes like reasoning_content (DeepSeek and other third parties) and reasoning ([OpenRouter Announcements](https://openrouter.ai/announcements/reasoning-tokens-for-thinking-models)) to store thinking processes, implementing this through pipes would be troublesome (requiring separate configuration for each). I saw that direct support was not possible in another PR discussion(#9241), but once the realtime filter is implemented in this PR, it could provide a convenient unified solution. Could you confirm when this might be considered?

tjbck:
Introduced as `stream` hook with 46c4da48642ccb670c5a696c0d8c62572d8b2b4c !

Example usage:
```py
def stream(self, event: dict) -> dict:
        print(event)
        return event
```

PR #9629, opened 2025-02-08
Handle no picture when retrieving from Microsoft EntraID

rragundez:
If the user has no picture in his Microsfot profile in EntraID then set the default picture.

Related #9327


rragundez:
Hi @tjbck , easy one to review.

cc @jeannotdamoiseaux

tjbck:
Thanks!

PR #9655, opened 2025-02-08
Fix - Usage statistics returned by the OpenAI compatible endpoint

Seniorsimo:
# Changelog Entry

### Description

- Fixed the usage statistics returned by the OpenAI compatible endpoint to reflect the acutal OpenAI standard as from their actual documentation:
https://platform.openai.com/docs/api-reference/chat/create

### Fixed

- Usage statistics returned by OpenAI endpoint

### DIFF
--- a/backend/open_webui/utils/response.py
+++ b/backend/open_webui/utils/response.py
@@ -5,51 +5,26 @@ from open_webui.utils.misc import (
     openai_chat_completion_message_template,
 )
 
+def convert_response_ollama_usage_to_openai(data: dict) -> dict:
+    return {
+        "prompt_tokens": int(data.get("prompt_eval_count", 0)),
+        "completion_tokens": int(data.get("eval_count", 0)),
+        "total_tokens": int(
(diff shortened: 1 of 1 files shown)


Seniorsimo:
Hi,

I saw that the PR was rejected.

If I may ask, was it because I didn't open a discussion about it? I thought the previous bug thread was enough.

In the meantime, today I rebuilt the Docker image with the fix from the PR, and it seems to work! Great!
But unfortunately, another issue came up: a small serialization problem with the arguments in tool_calls.
You can see it in the fork.

Buuut my misfortune didn't end there: after fixing this and replaying the LangChain pipeline, another problem came up. And this time, at least for now, I don’t know how to handle it.

I will collect more information and clarify everything somewhere. But how should I proceed?
Should I open a discussion and report everything I've done and found there?

Or should I open one or more bug reports for each fix?

Let me know, please.
Great work!

JTHesse:
What is the status here?
For me the usage statistics are still missing.

Seniorsimo:
hi, look here:
https://github.com/open-webui/open-webui/discussions/10089

PR #9733, opened 2025-02-10
Prevent SSRF and HTML injection

rragundez:
There is a vulnerability on the /api/v1/util/pdf endpoint via SSRF and HTML injection. This PR applies a patch such that contents in the request are escaped to html characters before rendeing the HTML service side.

@tjbck we have flagged this as a vulnerability which is of medium risk.


BhavanaSBhat:
👍

PR #9739, opened 2025-02-10
i18n: fix Spanish (es_ES) translations

yukiteruamano:
# Changelog Entry

### Description

- Changes for i18n support in Spanish (only es_ES distribution)

### Changed

- Only new strings for translations. Nothing more.


tjbck:
Thanks!

PR #9917, opened 2025-02-13
feat: Optional title generation

nachogmd:
# Changelog Entry

### Description

Allow to configure optional title generation

### Added

- The boolean configuration var ENABLE_TITLE_GENERATION makes title generation optionaL

### Changed

- Changed tasks, main, config, interface settings and corresponding i18n files


tjbck:
Thanks!

PR #10006, opened 2025-02-14
Avoid calling generate_queries when ENABLE_SEARCH_QUERY_GENERATION is false

nachogmd:
# Changelog Entry

### Description

- Currently, even if ENABLE_SEARCH_QUERY_GENERATION  generate_queries is called and returns a 400 error. This pull request avoids generating the query from the start and returns the user input

### Added

- Honor ENABLE_SEARCH_QUERY_GENERATION parameter

### Changed

- Changed middleware.py so that it doesn't call generate_queries and fixed a typo when logging web search and added a message stating the number of returned results


nachogmd:
Creating new branch and format files

PR #10036, opened 2025-02-14
Avoid calling generate_queries when ENABLE_SEARCH_QUERY_GENERATION is false instead of raising a 400 status

nachogmd:
# Changelog Entry

### Description

- Avoid calling generate_queries when ENABLE_SEARCH_QUERY_GENERATION is
false instead of raising a 400 status

### Changed

- Changed middleware.py in order to honour ENABLE_SEARCH_QUERY_GENERATION when set to false and fixed some typo error in retreival.py using logging.info instead of log.info

### Fixed

- Honour ENABLE_SEARCH_QUERY_GENERATION and if set to false return the prompt as a search query

### Additional Information

It would probably nice to have two additional properties: 

- CONCATENATE_SEARCH_QUERY_GENERATION which would join all user input in a single query
- USE_FIRST_MESSAGE_AS_SEARCH_QUERY so that the search query sticks to original prompt and not subsequent prompts that might have lost context

Let me know if these make sense in order to implement them!

### DIFF
--- a/backend/open_webui/routers/retrieval.py
+++ b/backend/open_webui/routers/retrieval.py
@@ -1312,7 +1312,7 @@ def process_web_search(
     request: Request, form_data: SearchForm, user=Depends(get_verified_user)
 ):
     try:
-        logging.info(
+        log.info(
             f"trying to web search with {request.app.state.config.RAG_WEB_SEARCH_ENGINE, form_data.query}"
         )
         web_results = search_web(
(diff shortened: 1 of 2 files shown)


nachogmd:
Hi there, I don't understand the Format Backend errors. Please advise! Is it related to formatting issues? Is it acceptable to use eclipse py as a development environment and use its formatters?

Thank you very much in advance. Looking forward to contribute as much as I can

nachogmd:
May I please get some feedback?

Thank you very much in advance

nachogmd:
Is this considered a bad idea?

Thank you very much for your help

PR #10041, opened 2025-02-14
fix: return local time from `getFormattedDate` and `getCurrentDateTime`

crizCraig:
Currently `getFormattedDate` returns UTC and `getCurrentDateTime` returns a mix of local and UTC. This currently only affects prompt templates, but I have another change to send this to openai backends to better serve users by knowing their time.

### DIFF
--- a/src/lib/utils/index.ts
+++ b/src/lib/utils/index.ts
@@ -1001,6 +1001,10 @@ export const bestMatchingLanguage = (supportedLanguages, preferredLanguages, def
 export const getFormattedDate = () => {
 	const date = new Date();
 	return date.toISOString().split('T')[0];
+	const year = date.getFullYear();
+	const month = String(date.getMonth() + 1).padStart(2, '0');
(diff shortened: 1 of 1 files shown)


tjbck:
`getCurrentDateTime` definitely returns local current date time, I'd suggest you try the latest version.

crizCraig:
Thanks Tim. Maybe I can use prompt variables for what I need. `getCurrentDateTime` in the latest dev actually _mixes_ local with UTC since `toISOString` returns UTC.

```
// Get the date in the format YYYY-MM-DD
export const getFormattedDate = () => {
	const date = new Date();
	return date.toISOString().split('T')[0];
};

// Get the time in the format HH:MM:SS
export const getFormattedTime = () => {
	const date = new Date();
	return date.toTimeString().split(' ')[0];
};

// Get the current date and time in the format YYYY-MM-DD HH:MM:SS
export const getCurrentDateTime = () => {
	return `${getFormattedDate()} ${getFormattedTime()}`;
};
```

PR #10042, opened 2025-02-15
feat: send user's local date to openai compatible APIs

crizCraig:
Building on #10041 - this sends the user's local date for better servicing time sensitive prompts like "what's the weather?" or "what's going on in the world today?" or "when are the suns playing next?" etc...  I didn't put it in a header like the other user data since it comes from the frontend. Since all frontend derived data is sent in the payload, I included it there instead.


tjbck:
We already have prompt variables for this exact use cases.

PR #10045, opened 2025-02-15
feat: allow disabling password signups

crizCraig:
Deploying public instances can lead to abuse of the email signup since emails don't get verified. Manually approving signups is not ideal as it prevents legitimate users from being able to easily signup. This PR allows disabling email/password signups in the UI and on the server. Oauth can then be use to help prevent people creating tons of accounts to get free usage. This as I charge for usage after a certain amount of free use.


crizCraig:
@tjbck, I fixed the frontend format here. The backend format is fixed in #10076 - please let me know if there are other changes you'd like to see here.

crizCraig:
Looks like some unrelated i8n failures starting with this [build](https://github.com/open-webui/open-webui/actions/runs/13350792740/job/37287015092) - but other than that this is ready!

tjbck:
I'll think a bit more about this feature, appreciate your patience!

crizCraig:
Thanks Tim. Appreciate you! ❤️

Issue #9912, opened 2025-02-13
Domain Filter List in WebSearch


liucoj:
Hi, On the latest version 0.5.11, I noticed that the new option 'Domain Filter List' in WebSearch doesn't work as expected. When I tried to insert a couple of domains separated by commas (e.g., example.com, example2.com), the filtered domains were not considered when retrieving results. I'm using Tavily as an API.
I couldn't find any instructions about this new option on official OWUI channels, such as Twitter, Discord, documentation, or here. Although I noticed that when saving preferences after inserting domains into the filter list, only the Tavily API update, concurrent requests, and connections are updated in the logs - not the domain filters themselves.
Any ideas?

liucoj:
I think Tavily module is not ready for this option, I didn't find anything related to "filter_list: Optional[list[str]]" within the code (tavily.py) as with the other search modules.

tjbck:
It's only supported by a select few providers, PR welcome here!

liucoj:
@tjbck I made a simple implementation in tavily.py (in **bold** ) and it works! I took as an example the searching module (searxng.py) working module. Considering that, in UI, as an admin, the search module permits only 3 options (max results , max concurrent requests , domain filter list ), and there aren't any other options to consider, I'll make a PR with these implementations. Do I miss something?

#Refactor function

def search_tavily(
    api_key: str,
    query: str,
    count: int,
    **filter_list: Optional[list[str]] = None**,
) -> list[SearchResult]:

#Add include_domain as requested by Tavily API
url = "https://api.tavily.com/search"
data = {"query": query, "api_key": api_key}
**include_domain = filter_list**
response = requests.post(url, include_domain, json=data)
response.raise_for_status()

nachogmd:
I will create a pull request which handles both "included" and "excluded" if this makes sense. 

I struggled too with this feature until I noticed what it does is to exclude those domains

It should involve passing all search engines a white list and a black list. I would need advise on how to name configuration parameters as if this approach is taken ideally we should have RAG_WEB_SEARCH_DOMAIN_FILTER_LIST as 

- RAG_WEB_SEARCH_WHITE_FILTER_LIST
- RAG_WEB_SEARCH_BLACK_FILTER_LIST

Best regards,
n

tjbck:
@nachogmd env var naming should follow the existing convention 

[...]

PR #10130, opened 2025-02-16
feat: Allow blocking and allowing domains in web search as per #9912

nachogmd:
# Changelog Entry

### Description

- Allow blocking and allowing domains in web search.

### Added

- Added two new env vars RAG_WEB_SEARCH_ALLOWED_DOMAIN_LIST and RAG_WEB_SEARCH_BLOCKED_DOMAIN_LIST

### Changed

- Created both environment variables in config.py
- Changed UI so that they are persisted
- Don't try to generate search query if ENABLE_SEARCH_QUERY_GENERATION = False
- Changed retrieval/web/main.py to enforce these variables
- Created a new class SearchParameters to facilitate integration of search engines and modified search engines implementations accordingly
- Detect if no search results are returned and in such case don't try to store them
- When informing the UI about searching include the search engine being searched and the query being made
- Fix typo when logging in process_web_search

### Deprecated

- RAG_WEB_SEARCH_DOMAIN_FILTER_LIST is no longer used

### Removed

- RAG_WEB_SEARCH_DOMAIN_FILTER_LIST is no longer used

### Fixed

- When no search results are found it doesn't raise an error

### Breaking Changes

--- stored value for RAG_WEB_SEARCH_DOMAIN_FILTER_LIST may require migration

### Additional Information

- Refer to issue #9912 where some additional features might come in


tjbck:
Please target `:dev` branch.

PR #10206, opened 2025-02-17
i18n: Improve RTL text handling in chat and sidebar with auto-direction

ttv20:
Enhances the handling of right-to-left (RTL) text in chat messages and sidebar titles by implementing automatic text direction detection.

Key Changes:
1. Added `dir="auto"` attribute to:
   - Chat message text elements
   - Sidebar chat titles

2. Fixed RTL-specific styling:
   - Changed blockquote borders from `border-left` to `border-inline-start`
   - Updated table header alignment from `text-align: right` to `text-align: start`

Technical Details:
There are two approaches to handle bi-directional text:
- Using `unicode-bidi: plain-text` on the text container element (doen't work on parent element)
- Adding `dir="auto"` attribute to the parent element (chosen approach)

Testing:
To verify these changes, use the following prompt:
```
write please a sample headers (level 1,2,3), text (bold/italic), list, numbered list, quotes, table, code
write each example in Hebrew
dont output markdown in code block
```

For a live example of the behavior, see: [https://codepen.io/ssfrzglx-the-lessful/pen/KwKdRXP](https://codepen.io/ssfrzglx-the-lessful/pen/KwKdRXP)
Before:
Screenshot_20250217_222818
After:
Screenshot_20250217_223201


ttv20:
I see now that the menu button on the sidebar and the download button on the table are overriding the text, for now disable the text-align fix in this element, the direction is the importent thing

tjbck:
Thanks!

PR #10304, opened 2025-02-18
i18n: Improve RTL text handling in chat and sidebar with auto-direction #2

ttv20:
Add some places that I missed on https://github.com/open-webui/open-webui/pull/10206 (thanks for the fast merge)

e.g. the feature of chat direction (https://github.com/open-webui/open-webui/pull/2318)  is kind of redundend now because it will affect only the model profile&name and action under the message position
@idomamia


tjbck:
Thanks!

PR #10311, opened 2025-02-19
Fix "Cannot read properties of undefined (reading 'startsWith')"

alexgleason:
I don't know the root cause of why `title` isn't available here, but it's crashing the UI because the types are wrong.

Looks like this was introduced in bbda717b69cd26f7a3cb1c87a383e0d5d522a30c

image

image


tjbck:
Thanks!

PR #10459, opened 2025-02-20
0.5.16

tjbck:


chrisportela:
It looks like there is a GHA billing/usage issue with publishing the latest docker image https://github.com/open-webui/open-webui/actions/runs/13443131263

PR #10617, opened 2025-02-23
fix: replace "system" with "developer" for o1 and o3 models

grandtele:
### Description

This implementation corrects the temporary fix in the codebase. Initially, the "o" series models only supported the "user" role. However, they recently introduced support for the "developer" role, which is equivalent to the "system" role.

### Fixed
- Fixed role mapping: changed `"system"` to `"developer"` for o1 and o3 models.

### Changelog Entry
- Fixed: Updated OpenAI o1 and o3 role handling by replacing "system" with "developer".

### Additional Information
- No new dependencies were introduced.
- No breaking changes expected.


tjbck:
Thanks!

PR #10634, opened 2025-02-23
fix: Handle no location permission

sebdanielsson:
# Changelog Entry

### Description

This PR contains a fix for when the user has not given Open-WebUI access to the users location. Instead of erroring out and refusing to send the message the variable {{USER_LOCATION}} is now replaced with LOCATION_UNKNOWN. Additionally when the user is using {{USER_LOCATION}} in a slash prompt, an error toast is displayed notifying the user that location data is not available.

### Fixed

- Fixes an issue where a prompt couldn't be sent if a location variable was present but location access was not allowed.


tjbck:
Thanks!

PR #10809, opened 2025-02-26
feat: update get_current_user to refresh last active timestamp asynchronously

TobiasGoerke:
Updating the users' last activity asynchronously has significant performance implications. 
It allows users to be authenticated with two less db queries to be waited for. 
For a page reload, this means there's 40-60 fewer queries waited for.

In my setup, page load time have decreased about 45%. 

Please see https://github.com/open-webui/open-webui/discussions/10808 for more info.


tjbck:
Please target our `dev` branch next time, Thanks!

PR #10885, opened 2025-02-27
fix env.var based on documentation

weisser-dev:
currently the documentation says, that the env var for LDAP_SEARCH_FILTERS is like this, but the implementation says LDAP_SEARCH_FILTER to be compatible with current implementation and doc we need to add this here.


gaby (approved):
👍 LGTM

gaby:
Fixes https://github.com/open-webui/open-webui/discussions/10882

tjbck:
70667d5cc8dd755f7e439bd6eb1f822fc76e38f0

PR #10962, opened 2025-02-28
i18n(html): sync html language with i18n language

Noxiven:
# Changelog Entry

### Added

- Improved Language Consistency: The HTML lang attribute now automatically synchronizes with the application's selected i18n language

---

### Screenshots or Videos

image

image

image

### DIFF
--- a/src/lib/components/chat/Settings/General.svelte
+++ b/src/lib/components/chat/Settings/General.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
 	import { toast } from 'svelte-sonner';
 	import { createEventDispatcher, onMount, getContext } from 'svelte';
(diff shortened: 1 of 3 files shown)


OrenZhang:
When the default HTML language attribute is set to 'en' (English), if the user's device language is different, some browsers will automatically display a "translate" option, even though the interface may already be in the user's current language.

It is recommended to set the HTML lang attribute to match the actual language being used by the user, to avoid unnecessary translation prompts. For example, for Chinese users, setting <html lang="zh"> or <html lang="zh-CN"> would be appropriate.

OrenZhang:
For example, Edge will display this on every page

IMG_0820

tjbck:
fe44e4d3448b934c0f316390d5141dec3348d906, Thanks!

PR #10983, opened 2025-02-28
fix: add redis lock timeout config

yvedeng:
# Changelog Entry

### Description

- The TTL of the redis lock is too short in our situation, and it caused the pool cleanup to exit unexpectedly. Lock TTL should be longer than the expected duration of the cleanup + the delay; however, the duration increases along with the increasing usage pool. Therefore, we added this Redis lock timeout environment variable to allow each application to adjust it based on the expected usage pool. 

### Added
- `WEBSOCKET_REDIS_LOCK_TIMEOUT_IN_SECONDS` environment variable. The default value is 1 minute.


tjbck:
Thanks!

PR #11009, opened 2025-03-01
fix: typing error in replace_messages_variable func args

amuwall:
# Changelog Entry

### Description

- The typing hint for `messgaes` args in `replace_messages_variable` is `str`. But it actually should be a `dict`.

### Fixed

- Modify `messages` typing to `dict`.


tjbck:
Thanks!

PR #11043, opened 2025-03-02
i18n: update zh-TW

TiancongLx:
Dear @tjbck Please help me revise a translation detail on line 300, which I think is better to keep as is here, rather than translate it. Thanks!

`token: 偏置值` => `token:bias_value`

```json
{
    "Enter comma-seperated \"token:bias_value\" pairs (example: 5432:100, 413:-100)": " 輸入逗號分隔的 \"token: 偏置值 \" 配對 (範例：5432:100, 413:-100)"
}
```


tjbck:
Thanks!

PR #11120, opened 2025-03-04
fix(jupyter): fix kernel_id not set and optimize code

Noxiven:
### Fixed

- Fixed an issue where kernel_id might not be set when using Jupyter as a code interpreter, causing exceptions in the finally block. Additionally, optimized the code structure for better performance and maintainability.
- Fixed an issue where incorrect URL concatenation was causing potential kernel closure failures. The improper URL formation was preventing proper communication with the Jupyter kernel, resulting in kernels remaining open after sessions ended.
---

#### Issue 

wrong format or url cause 403

iShot_2025-03-03_21 25 44

kernel_id missing when api is not 200

image

#### Testing

image

image

### DIFF
--- a/backend/open_webui/utils/code_interpreter.py
+++ b/backend/open_webui/utils/code_interpreter.py
@@ -1,14 +1,15 @@
 import asyncio
 import json
 import uuid
+from typing import Optional
+
+import httpx
 import websockets
-import requests
-from urllib.parse import urljoin
 
 
 async def execute_code_jupyter(
-    jupyter_url, code, token=None, password=None, timeout=10
-):
+    jupyter_url: str, code: str, token: str = None, password: str = None, timeout: int = 60
+) -> Optional[dict]:
     """
     Executes Python code in a Jupyter kernel.
     Supports authentication with a token or password.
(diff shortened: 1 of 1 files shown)


OrenZhang:
I've tested for both password and token authentication, and text or image output, it works well

backend/open_webui/utils/code_interpreter.py
```diff
@@ -1,148 +1,189 @@
...
+    async def init_kernel(self) -> None:
+        async with self.session.post(url="/api/kernels", params=self.params) as response:
+            response.raise_for_status()
+            kernel_data = await response.json()
+            self.kernel_id = kernel_data["id"]
 
+    def init_ws(self) -> (str, dict):
+        ws_base = self.base_url.replace("http", "ws")
+        ws_params = "?" + "&".join([f"{key}={val}" for key, val in self.params.items()])
+        websocket_url = f"{ws_base}/api/kernels/{self.kernel_id}/channels{ws_params if len(ws_params) > 1 else ''}"
```
OrenZhang, line 102: `ws_params` starts with "?" even if params is empty, so using `len(ws_params) > 1`

tjbck:
Thanks!

PR #11160, opened 2025-03-04
chore: Improve config.json migration

tidely:
# Description

Fixed a Time-of-Check Time-of-Use (TOCTOU) bug by handling exceptions instead of checking for a file existing beforehand. While this is extremely unlikely to occur naturally we should fix it anyway.

`load_json_config` was only being used by the migration, so we can remove it.

We previously saved the loaded config in the `data` variable. This has a global scope and would not get removed by the gc. We can fix this by moving the migration into its own function.

## Future

We should probably add a comment on how long of a migration period we want to have. This project moves very fast and the migration started 7 months ago already.

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 import shutil
+import base64
 
 from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


tidely:
There appear to be a lot of unused imports. If possible I'd recommend using [ruff](https://github.com/astral-sh/ruff) to check for them and automatically fix them. That would be a horrible pull request to review, so you would need run and merge it yourself.

Running these commands would fix a lot of warnings I'm getting using my editor

```sh
ruff check --fix
ruff format
```
image

PR #11208, opened 2025-03-05
fix(i18n): lang is always set to en-US

Noxiven:
### Fixed

- HTML lang is always set to en-US

---

### Additional Information

I use `console.log(i18next?.language)` and it show undefined, so `const lang` will always be `en-US`


tjbck:
Closing in favour of edf94d732b983f2d1661fdc033c3a6ed024abfc1, Thanks!

PR #11358, opened 2025-03-07
i18n(common): add i18n translation

Noxiven:
### Added

- Add i18n wrappers to previously untranslated text elements


tjbck:
LGTM, Thanks!

PR #11497, opened 2025-03-10
feat: Added new k_reranker parameter

mahenning:
# Pull Request Checklist

# Changelog Entry

### Description

- Adds a separate top k value for the reranker, making the RAG pipeline more flexible

### Added

- Adds a separate top k value for the reranker that can be set in the documents tab in admin settings

---

### Additional Information

- Example: Retrieve `top_k=100` results from the embedding space, then rerank them and give the `top_k_reranker=20` document chunks to the LLM. Right now you have to decide to set top k high and clutter the LLM context indow, or use a low top k i embedding space. Embedding 
- I know that the Minimum Score parameter already exists, but in my tests I didn't find a value that reliably works. Some queries hit 0.9 similarity score (especially if some document-related key words are in the query). Sometimes the similarity was only 0.01, but still with relevant information, which would be filtered out by a fixed threshold value.
- The top_k_reranker parameter is hidden in the UI if hybrid search is disabled, the same way the reranker model field works. This PR also don't break stuff for users who don't want to use two different top_k. Just set both to the same value to get the current behavior.


Phlogi:
Tested this when working on improving overall performance by running queries in parallel for hybrid search, see https://github.com/open-webui/open-webui/discussions/11729. 

Works fine for me, do we need to handle top_k < top_k_reranker?

mahenning:
> do we need to handle top_k < top_k_reranker?

Good point! Looking into it.

Phlogi:
I'd would suggest just a min() to cut back yes.
maybe add to docs that the k is a total of 2*k one for BM25 and vector if that's correct?

mahenning:
I've added the case, and added also sorting by distance for cutting off by k. I've read that the results are not guaranteed to be ordered by distance for all rerankers, although I personally would expect this from a so called "reranker". I added an explicit sorting before cutting in the case of k < k_reranker. For k >= k_reranker, no cutting off is necessary.

Review/feedback welcome.

Phlogi:
Works fine for me!

rgaricano:
I think that the sort have to be done, althrought k < k_reranker. In this way, the array is returned already sorted, although not all applications connected to the endpoint will verify the score and many accesses will probably understand the array as already sorted.and use it as is.
(And I still think a slider should be included to indicate the type of hybrid search, whether more lexical or more semantic... but that's another topic)

backend/open_webui/retrieval/utils.py
```diff
@@ -146,7 +146,10 @@ def query_doc_with_hybrid_search(
 
         # retrieve only min(k, k_reranker) items, sort and cut by distance if k < k_reranker
         if k < k_reranker:
-            sorted_items = sorted(zip(distances, metadatas, documents), key=lambda x: x[0], reverse=True)
+            if VECTOR_DB == "chroma":
+                sorted_items = sorted(zip(distances, metadatas, documents), key=lambda x: x[0], reverse=False)
```
ivanbaldo, line 150: Should maybe add this comment:
       # Chroma uses unconventional cosine similarity, so we don't need to reverse the results
        # https://docs.trychroma.com/docs/collections/configure#configuring-chroma-collections

mahenning:
It seems that for chromadb you have to sort in ascending order when cutting with `[:k]`, as used in 
https://github.com/open-webui/open-webui/blob/ba676b7ed6a4ce141474d7c31797ea2fd8aa513a/backend/open_webui/retrieval/utils.py#L313-L318
And in the function `merge_and_sort_query_results`:
https://github.com/open-webui/open-webui/blob/ba676b7ed6a4ce141474d7c31797ea2fd8aa513a/backend/open_webui/retrieval/utils.py#L213-L214

I've added that to the sorting and cutting in the k < k_reranker case. I've also added the k_reranker parameter for the cutting in `merge_and_sort_query_results` (only hybrid search of course), so that at max `k_reranker` results are forwarded to the chat LLM.

backend/open_webui/retrieval/utils.py
mahenning: Done, thanks.

ivanbaldo:
Option 2 seems to give a better mix of results from several queries...
In option 1 we can have all the results from a single query which gives less diversity...

rgaricano:
No, with option 1, you have 90 results before sorting; you're not limiting diversity. Option 2 does bias them, there may be more relevant results in the discarded results, or those not discarded may be repeated in other queries.
Option 2, however, saves you some time.

ivanbaldo:
@rgaricano in option 1 you are potentially limiting the diversity because if all the results for query 2 have higher scores from the reranker than for all the results in query 1 and 3, then you will get only the results from query 2.

In option 2 you can't cover the requested 30 results from a single query because at max you have 20 from a single query.

rgaricano:
But scores are a probability of the total, it is impossible for all the results of a query to be greater than all the results of the others.

ivanbaldo:
Ouch maybe I didn't understand the algorithm well.
My current understanding is that the scores are per query by the reranker, and the reranker scores each result against a single query, so they aren't a distribution of all the combined results from all queries.
If I am right, then maybe the reranker for a query may assign low scores for almost all results and then for another query assign high scores for almost all the results, resulting in that second query overriding most of the results from the first query in Option 1.

rgaricano:
yes, i understand the same, the score is a distribution of the all results of the query, sum of all query results scores=1
per query!,

ivanbaldo:
Ah, alright then, if it distributes all the results between 0 and 1 per each query, then it should bring a good mix from different queries already.

Phlogi:
Side note, I refactored the Chroma decision in my PR already (https://github.com/open-webui/open-webui/pull/11814, specifically https://github.com/open-webui/open-webui/pull/11814/commits/a969d6ee03eda54819eb4842ef05a5a1733b1d8b), here is a combined version of util.py that implements this PR and the parallel processing. Thanks for testing.

rgaricano:
For reference:

BM25
indirect used here: https://github.com/dorianbrown/rank_bm25
that implement this maths (functions, scoring,..): https://www.cs.otago.ac.nz/homepages/andrew/papers/2014-2.pdf )

A reranker sample from chromadb
https://blog.amikos.tech/chroma-daily-devlog-rerankers-a11eae856113 
https://github.com/amikos-tech/chromadbx/pull/78/files

mahenning:
I looked into the LangChain code again, especially the `EnsembleRetriever`, which uses both the `BM25Retriever` and the `VectorSearchRetriever`. The Algorithm is rather simle, it purely uses the **rank of the document** and is not using the score of the base retrievers. So a document which places "5th best" in one retriever gets the score `weight/(rank + c)` = `0.5/(5+60)~0.008` (c is a fixed offset). If the same chunk appears both in the BM25 and Vector retriever, the scores are summed up.

https://github.com/langchain-ai/langchain/blob/0ba03d8f3a40eafca6fe2f688046778653e117d2/libs/langchain/langchain/retrievers/ensemble.py#L310-L319

What does this mean? If we would have e.g. 20 chunks each for both base retrievers, and _assuming_ that both retrievers have different chunks, the score distributions of both base retrievers would be the same, as the e.g. 5th doc get's the same score (0.008, see above). Only when one chunks is in both retrievers, it can get a better score than just its rank. This discards the scores of the base retrievers (for VectorRetriever the cosine similarity), but it makes sense because the scores of BM25 and VectorRetriever are not comparable, even when we normalize the BM25 score.

If there are no hard opinions against it, I'd go with option 2. As I see it, it should be a very niche corner case either way because `k < k_reranker` makes little sense and most people would have `k > k_reranker`. So yes, it would make a difference, but `k > k_reranker` should just work better.

mahenning:
Now, when only the ranks of the base retrievers are used, I wonder if the results of the `ContextualCompressionRetriever` (so the reranker) should be reverse sorted if the vector database was a chroma db. The (reverse sorted) cosine similarity score of the chromadb is never used directly, only by the VectorSearchRetriever to get the best k documents. After that the Ensemble retriever uses only the ranks of the documents for scoring them, and after this point it should make no difference if the chromadb uses an inversely ranked cosine similarity??

mahenning:
I've reverted part of the code, as far as I see (and tested) it the reranker gives already results in descending order (1 is best, 0 is worst). The score of the chromadb is not used at this point. I also fixed some other chromadb-related ordering issues in this PR https://github.com/open-webui/open-webui/pull/11876.

tjbck:
Thanks!

PR #11737, opened 2025-03-16
Enhance OpenAI TTS Voice Parameter Handling

roeizavida:
This pull request introduces support for both string and JSON object formats for the voice parameter in the OpenAI TTS implementation.
Previously, the system only accepted a simple string for the voice setting, which limited compatibility with services requiring more complex configurations, such as Vertex AI TTS using the LiteLLM proxy.
With this update, the code now attempts to parse the TTS_VOICE configuration value. If it can be successfully parsed as a JSON object, it will be used as such; otherwise, the original string value will be utilized.

### DIFF
--- a/backend/open_webui/routers/audio.py
+++ b/backend/open_webui/routers/audio.py
@@ -264,7 +264,21 @@ async def speech(request: Request, user=Depends(get_verified_user)):
 
     if request.app.state.config.TTS_ENGINE == "openai":
         payload["model"] = request.app.state.config.TTS_MODEL
-
+        
+        # Handle voice parameter - support both string and JSON object formats
+        voice_setting = request.app.state.config.TTS_VOICE
+        if voice_setting:
+            try:
+                # Try to parse as JSON dict
+                voice_dict = json.loads(voice_setting)
+                if isinstance(voice_dict, dict):
+                    payload["voice"] = voice_dict
+                else:
+                    payload["voice"] = voice_setting
+            except (json.JSONDecodeError, TypeError):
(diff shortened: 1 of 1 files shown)


tjbck:
Could you provide us with more details? Screen recordings would be helpful here.

roeizavida:
> Could you provide us with more details? Screen recordings would be helpful here.

According to LiteLLM docs, you need to provide the voice as object in order to use Vertex AI TTS service: https://docs.litellm.ai/docs/providers/vertex#usage---basic
The request to the OpenAI server would look like this:
```python
import openai

client = openai.OpenAI(api_key="sk-1234", base_url="http://0.0.0.0:4000")

# see supported values for "voice" on vertex here: 
# https://console.cloud.google.com/vertex-ai/generative/speech/text-to-speech
response = client.audio.speech.create(
    model = "vertex-tts",
    input="the quick brown fox jumped over the lazy dogs",
    voice={'languageCode': 'en-US', 'name': 'en-US-Studio-O'}
)
print("response from proxy", response)
```

Open WebUI passes this as a string, even if I use `{'languageCode': 'en-US', 'name': 'en-US-Studio-O'}` for voice - it would not be passed as an object.
This PR allows providing the `voice` as dict, in addition to string which is used for OpenAI, therefore allowing the usage of Vertex AI TTS and other OpenAI compatible providers who may use the same structure for voices.

tjbck:
Your PR will always override the voice model set by the individual users. PLEASE test your PRs before opening one.

PR #11814, opened 2025-03-18
perf: Parallelize search_query in query_collection_with_hybrid_search and vector db data retrieval

Phlogi:
- Link to discussion opened before PR: https://github.com/open-webui/open-webui/discussions/11729

- [x]  **Target branch:** Please verify that the pull request targets the `dev` branch.
- [x]  **Description:** Provide a concise description of the changes made in this pull request.
- [x]  **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [x]  **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [x]  **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [x]  **Testing:** Have you written and run sufficient tests for validating the changes?
- [x]  **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [x]  **Prefix:** To cleary categorize this pull request, prefix the pull request title, using one of the following

# Changelog Entry

### Description

- I noticed that RAG with hybrid search is unusably slow. Currently, the hybrid search runs a nested 2 level for loop in serial to get all potential relevant parts of the document(s) and then reranks them sequentially too. 
- Both of these bottlenecks are fixed with the changes. Before the change only few cores were utilized on my system, with these changes all cores are full saturated. The estimated speed improvement is about 50 % on a CPU-only system. 

#### Function query_doc_with_hybrid_search updates:

- Replace nested for loop with a ThreadPoolExecutor to get all collection data in parallel and only once for all queries.
- The collection_data is then passed to the query_doc_with_hybrid_search function. 
- Execution of process_query over all potential documents in parallel

#### Refactoring of the Chroma sorting exception
- DRY-principle: move the reverse sorting flag based on the VECTOR_DB type, specifically for "chroma" into merge_and_sort_query_results

#### Function merge_and_sort_query_results Enhancements:

- Optimized memory allocation by estimating capacity and reserving space for combined results.​
- Implemented batch processing for document hash computations to improve performance.​
- Added an early return for cases with empty combined results
- Refined the sorting mechanism and truncated results to the top k entries.

### Removed

- Unneeded imports at top removed: uuid and asyncio

---

### Additional Information

- I measured the speed up based on log entries time stamps. On my CPU-only system with 32 cores, the improvement is about 50% less time to answer questions of a single complex document, length 360'000 chars.

### Testing
The following manual tests were performed:

Querying...
1. a collection of 5 book sized documents (about multiple megabytes)
2. a collection with a single document
3. a single document uploaded in chat
4. multiple documents uploaded in chat

Queries used, verification:
- Specific know-how queries that i knew the appropriate answer. Verified the returned relevant part of the document manually.


mahenning:
It was mentioned in https://github.com/open-webui/open-webui/pull/11876 that the reverse order is also necessary for pgvector. I'll expand the linked PR with an attribute for each database class do if reverse should be used or not, so that the `if "chromadb` check  is changed. Just as a heads up. Otherwise, great PR!

Phlogi:
This is now ready for testing again with further optimizations. 
@mahenning Maybe you can help testing it?

mahenning:
I'll take a look in the next days

tjbck:
This definitely needs to be broken down into several atomic PRs. I encourage you to reopen.

mahenning:
A quick google search later, and I'm not sure if all supported databases are actually thread safe. Chroma [is thread safe](https://cookbook.chromadb.dev/core/system_constraints/), but I'm not sure about pgvector or qdrant. Can you check that?
Can you also check if there is a speedup (or at least _not_ a slowdown) if open-webui only gets e.g. 6-8 (or even down to 4) cores to simulate a mid-range consumer pc? I think you can do that with `taskset -c 0-5`. I'm pretty sure with e.g. 6 cores, one call to a reranking model can already saturate the CPU. It could be the case that multiple concurrent threads slow "fight" for the cores and slow down the retrieval. If that's the case, the number of concurrent threads need to be handled.

@tjbck except for the heapq part (and some other old code like the ordering with/without chroma, which is not at this place anymore in dev), I'm actually not sure where this should be broken down tbh. Either the vector/embedding/reranking part is completely managed by threads, or it is not. Do you have an example in mind?

Phlogi:
@mahenning We're only fetching data, so i don't think thread safety is an issue.

I'll double-check the speed up on lower CPU cores. Do you know how langchain decides on how many CPU cores can be used? We'd need to consider that to find the optimal degree of parallelism on the higher level. 

I did start breaking down the PRs. 
1. https://github.com/open-webui/open-webui/pull/12132 (avoid fetching data multiple times from vector DB)
...

Phlogi:
@mahenning does the hybrid search from dev work for you? I don't get any results it seems. Not sure what's going on.

mahenning:
Thanks for noticing/checking. I also had problems in my fork. In my forked repo, I have a `dev` branch that should just mirror/copy the dev branch from this repo. I created a fetch-only "upstream" remote for that, that points to the original repo. 

For reasons I don't know exactly, `git fetch upstream && git merge upstream/dev` showed me "up to date", but it didn't incorporate all the changes from this dev branch. 
Steps to check I did:
git log upstream/dev --oneline -n 5 # checks for the latest comit hashes/messages the repo knows of the upstream, compare with the one from github, should be the same after fetching
git merge upstream/dev # said "Already up to date."
git diff dev upstream/dev # still showed differences, pointing to a divergence from my local dev branch and this repos dev branch
I fixed it with `git rebase upstream/dev`. 

After that, hybrid search worked for me again. If you have a similar setup, check the steps.

mahenning:
> Do you know how langchain decides on how many CPU cores can be used? 

In the normal setup, both embedding and reranking model are sentence-transformer models, which runs on pytorch I think. But as far as I read, when you start the webui with `taskset -c 0-5 [commands to start webui]`, the process and all processes/threads it spawns only get cpu core 0 to 5.

Phlogi:
@mahenning thanks for the tasket hint, result is here: https://github.com/open-webui/open-webui/pull/12239

PR #11850, opened 2025-03-19
Wrong pull, i write correction notes but another proposal had already been submitted

rgaricano:
# Pull Request

## FIX: Download PDF generator with MD

Replace markdown lib by fpdf2 recommended lib mistletoe (https://github.com/miyuchina/mistletoe) for convert MD to html and render to PDF improving view.

This commit replace markdown lib, does the same job and improve conversion to hml.
I tested I didn't find any incompatibility with previous uses of markdown here or in other places.
(mistletoe lib is MIT licence)

#11840

openwebui-pdf-mark

### DIFF
--- a/backend/open_webui/utils/pdf_generator.py
+++ b/backend/open_webui/utils/pdf_generator.py
@@ -4,7 +4,7 @@ from pathlib import Path
 from typing import Dict, Any, List
 from html import escape
 
-from markdown import markdown
+from mistletoe import markdown
 
 import site
 from fpdf import FPDF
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -72,7 +72,7 @@ dependencies = [
     "python-pptx==1.0.0",
     "unstructured==0.16.17",
     "nltk==3.9.1",
-    "Markdown==3.7",
+    "mistletoe==3.7",
```
rgaricano, line 75: 75 have to be `"mistletoe", `
without version

backend/open_webui/utils/pdf_generator.py
```diff
@@ -42,7 +42,7 @@ def format_timestamp(self, timestamp: float) -> str:
 
     def _build_html_message(self, message: Dict[str, Any]) -> str:
         """Build HTML for a single message."""
-        role = escape(message.get("role", "user"))
+        role = markdown(message.get("role", "user"))
```
rgaricano, line 45: I I replaced the wrong line, I'm asleep!!!
correct:
45 `role = escape(message.get("role", "user"))`
46 `content = markdown(message.get("content", ""))`

rgaricano (commented):
mistakes corrected

PR #11856, opened 2025-03-19
feat-11855 - Add new Admin Option: "enforce_temporary_chat_mode" (with locale de_DE & en)

weisser-dev:
### Description:
This PR introduces a new feature to enforce whether user chats are temporary or permanent by adding a new permission `enforce_temporary_chat_mode`:

- **Backend**: 
  - A new setting `USER_PERMISSIONS_ENFORCE_TEMPORARY_CHAT_MODE` was added in `config.py` to define whether the temporary chat mode is enforced globally.
  - The global permission is added to `DEFAULT_USER_PERMISSIONS` under the "chat" section.

- **Frontend**:
  - The `permissions.svelte`, `groups.svelte`, and `editGroupModal` components were updated to handle the new permission.
  - A UI switch was added to allow toggling the `enforce_temporary_chat_mode` setting for chat users.

### Purpose:
This feature allows administrators to enforce whether chats are temporary or permanent by setting the `enforce_temporary_chat_mode` permission. When enabled, user chats will automatically follow the temporary mode regardless of other settings.

### Changes:
- Backend: Added `enforce_temporary_chat_mode` permission in `config.py`.
- Frontend: Updated relevant components (`permissions.svelte`, `groups.svelte`, `editGroupModal`) to support the new permission.
- UI: Added a toggle switch for enforcing temporary chat mode in the frontend.

### Notes:
- The default value for `enforce_temporary_chat_mode` is `false`. It can be toggled via the UI for each user or group.
- This change ensures that admins can control whether user chats should remain temporary, overriding other configurations.

Feature: https://github.com/open-webui/open-webui/discussions/11855

Pull Request Checklist

✅ Target branch: This pull request targets the dev branch.
✅ Description: see above
✅ Changelog: Infos for the Changelog, see above
Documentation: I was not able to add this to the documentation yet, pls do this after merge, or i could open a PR for docs
✅ Dependencies: No new dependencies were introduced & also not a breaking change, current logic works without any problems
✅ Testing: The feature has been manually tested with different values & behaviours
✅ Code review: Self-review was conducted, ensuring adherence to coding standards & also reviewed by gpt-4o
✅ Prefix:
This pull request is categorized as: feat: Introduces a new feature to allow global default chat mode configuration.

image

⸻

This PR is now ready for review and merge! 🚀


weisser-dev:
@tjbck pls check this, it would be awesome if this could be merged and get a new feature in the next version

weisser-dev:
> I'm trying to understand why this is needed when you could just `/?temporary-chat=true`? Could you elaborate more?

We have very strict data protection regulations in Germany. Among other things, these regulations state that no user data may be stored by default unless users actively opt in. This means that by default, users must start with the "enforce temp chat" setting enabled and actively set it to false via the URL parameter.

In Germany, there are some rather peculiar regulations regarding data protection and the safeguarding of personal data... unfortunately. Due to data protection regulations, no chat history may be stored until users actively enable it for themselves—but even then, only until their session expires. If we cannot provide this functionality, we would have to look for alternatives to OpenWebUI.

However, since I am very excited about this tool and you are doing an excellent job, I decided to implement this feature. Fundamentally, nothing changes in the existing logic, as URL parameters and shortcuts still work as before. But as an admin, you now have the additional option to define a "default" chat mode for users, which in this case would be "temporary."

I’m not sure if this is a problem in the US, but in Europe—especially in Germany—data and privacy protection of users are extremely important for many companies. That’s also why AI is not widely used here. This would be an extremely important feature for this market!

tjbck:
@weisser-dev 

> We have very strict data protection regulations in Germany. Among other things, these regulations state that no user data may be stored by default unless users actively opt in. This means that by default, users must start with the "enforce temp chat" setting enabled and actively set it to false via the URL parameter.

Response from ChatGPT:

No, this statement is not true.

In Germany, data protection regulations, particularly under the GDPR (General Data Protection Regulation), require that users give informed and explicit consent before personal data is collected or stored. However, GDPR does not specifically mandate that "enforce temp chat" must be enabled by default.

What GDPR does require:

Privacy by Default & Design – By default, only the necessary personal data should be processed, but this does not necessarily mean temporary chat must be enforced unless personal data is automatically stored otherwise.
Opt-in Mechanism for Data Collection – If storing non-essential user data, users must actively consent (opt-in) before any data is retained.
Thus, whether "enforce temp chat" should be enabled by default depends on whether user data would otherwise be stored without explicit consent. If no personal data is stored without consent, enforcing temporary chat mode by default may not be legally required.

---

Correct me if I'm wrong but this also aligns with what I know, for instance, ChatGPT absolutely does not have temporary chat enabled by default.

weisser-dev:
> @weisser-dev
> 
> > We have very strict data protection regulations in Germany. Among other things, these regulations state that no user data may be stored by default unless users actively opt in. This means that by default, users must start with the "enforce temp chat" setting enabled and actively set it to false via the URL parameter.
> 
> Response from ChatGPT:
> 
> No, this statement is not true.
> 
> In Germany, data protection regulations, particularly under the GDPR (General Data Protection Regulation), require that users give informed and explicit consent before personal data is collected or stored. However, GDPR does not specifically mandate that "enforce temp chat" must be enabled by default.
> 
> What GDPR does require:
> 
> Privacy by Default & Design – By default, only the necessary personal data should be processed, but this does not necessarily mean temporary chat must be enforced unless personal data is automatically stored otherwise. Opt-in Mechanism for Data Collection – If storing non-essential user data, users must actively consent (opt-in) before any data is retained. Thus, whether "enforce temp chat" should be enabled by default depends on whether user data would otherwise be stored without explicit consent. If no personal data is stored without consent, enforcing temporary chat mode by default may not be legally required.
> 
> Correct me if I'm wrong but this also aligns with what I know, for instance, ChatGPT absolutely does not have temporary chat enabled by default.

Thats right, this is one reason why our company also is not allowed to use ChatGPT. And why we are looking for a OnPrem Tool with this option. In addition to explaining its purpose, the implementation simply extends the existing functionality without introducing a breaking change or requiring adjustments to existing installations. Maybe we can have a short chat about this in discord?

weisser-dev:
@tjbck you could ask these Question to ChatGPT: 
Is it possible that in a German insurance company with a works council, storing chats with tools like AI or OpenWebUI is prohibited to protect employees and ensure traceability?

Then you recieve the answer ^^

rgaricano:
yes weisser, in spain is similar, 
Perhaps not as restrictive as in Germany, I don't know, but here the issue of data protection is also legislated and is very protectionist and equally low compliance with [The AI Act Explorer](https://artificialintelligenceact.eu/ai-act-explorer/)

weisser-dev:
@tjbck i also fixed some translation issues for germany, cause i18n wasn´t use yet for temporary chat: 
image
but now it is working
image

va-nessi:
Hi @tjbck , @weisser-dev, I also work at a company with a strong works council, which prohibits us from using Microsoft Copilot & ChatGPT precisely because of the lack of such a feature. We have very strict regulations to protect employees (including technical newcomers), which is why the rule applies: Temporary chat should be the default, and a regular chat must be explicitly enabled (e.g. with url params like ?temporary-chat=false). Therefore, this feature is also relevant for us. An alternative to this would only be to encrypt all chats 100% so that they are no longer analyzable, and no one (not even a database admin) has access to unencrypted chats.

tjbck:
Agreed, this will be added but will require some refactoring/renaming on our end. Appreciate your input and patience!

gaby:
👍 For adding this, we have an instance with multiple users and we do not want to keep history of chats. This would solve this, since there's no way to auto-prune chats in Open-WebUI.

weisser-dev:
> 👍 For adding this, we have an instance with multiple users and we do not want to keep history of chats. This would solve this, since there's no way to auto-prune chats in Open-WebUI.

Currently we prune chats every 20 Seconds with a CronJob and delete the chats Table... We doing this with a Kubernetes Sidecar... but this is just a bad workaround

tjbck:
Closing in favour of 5f792d27717abc0747f9bfb2847658124fc04b5c, Thanks!

Issue #11822, opened 2025-03-18
issue: Bug by open citations / context pop up using files (with the usage of reranking models)


etlstrauss:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.5.19 and v0.5.20

### Ollama Version (if applicable)

0.6.1

### Operating System

Ubuntu 24.04.2 LTS and Debian GNU/Linux 12 (bookworm)

### Browser (if applicable)

Firefox 136.0.1 and Chrome 134.0.6998.35

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [ ] I have included the Docker container logs.
- [x] I have listed steps to reproduce the bug in detail.

### Expected Behavior

By using knowledge bases or direct file upload, the interface provides (at the bottom of the messages) a link to the used context and specific data. This button should open a pop up dialogue with this informations.

### Actual Behavior

By clicking on the button an error occurred an the whole interface freezes. No pop up dialogue will be opened.

### Steps to Reproduce

1. Go to Admin Panel -> Documents -> enable Hybrid Search + choose a rerank model
3. Open a new chat
4. Ask somethings and include a file upload or knowledge base
5. Click on the button under the actual chat message, where the information of the used context is provided.

### Logs & Screenshots

```
Uncaught (in promise) TypeError: n[9].distance.toFixed is not a function
    vi CitationsModal.svelte:135
    pi CitationsModal.svelte:123
    vt utils.js:165
    ai tippy.esm.js:2482
    Pt Component.js:148
    li Tooltip.svelte:39
    _i CitationsModal.svelte:142
    Tl CitationsModal.svelte:115
    hi CitationsModal.svelte:111
    Rl CitationsModal.svelte:84
    wi CitationsModal.svelte:78
    vt utils.js:165
    M Immutable
    p Modal.svelte:64
    dt scheduler.js:119
    _t scheduler.js:79
    promise callback*ut scheduler.js:20
    ht Component.js:81
    ctx Component.js:139
    I Citations.svelte:108
    g Citations.svelte:120
    D dom.js:361
    m Citations.svelte:121
    m Citations.svelte:124
    m Citations.svelte:205
    m Citations.svelte:99
    _t Component.js:44
    m ResponseMessage.svelte:806
[...]

Issue #9106, opened 2025-01-30
Error at RAG function chromadb sorting and RAG optimization


Schwenn2002:
### Discussed in https://github.com/open-webui/open-webui/discussions/9001

<div type='discussions-op-text'>

<sup>Originally posted by **Schwenn2002** January 27, 2025</sup>
There is a faulty code for sorting the document list in ChromaDB because the values ​​of the cosine vector are misinterpreted (see below).

Furthermore, the RAG function can significantly speed up the query speed due to much smaller context lengths if the sorted return list from the RAG is limited to the best X hits:

A suggestion for optimizing the RAG function.

Currently, you can improve the quality of the query using Top k and ReRanking.

It would be optimal to filter out the most important documents using Top k=50 or 70 and ReRanking (0,6 - 0.8). Then another parameter would be optimal if, after the reranking, maximum the best (Top Best=10 or 20) hits were passed on to the LLM as context.

This way, the context length can be kept shorter and the response time improved. This also ensures that the context cannot become larger than configured in the LLM.

The current behavior of the LLM is that a context that is too large cannot be processed (i.e. it is interpreted as empty). It would therefore make sense to truncate the context from the RAG query after reranking to the context length of the LLM, so that only the best hits are passed on</div>

---

### backend/open_webui/retrieval/utils.py
**Why is there a distinction made here for sorting for ChromaDB? Is this a bug?**

```
    if VECTOR_DB == "chroma":
        # Chroma uses unconventional cosine similarity, so we don't need to reverse the results
        # https://docs.trychroma.com/docs/collections/configure#configuring-chroma-collections
        return merge_and_sort_query_results(results, k=k, reverse=False)
    else:
        return merge_and_sort_query_results(results, k=k, reverse=True)
```

The result of cosine similarity always lies within the range of -1 to 1.

---

### **Interpretation of Values:**
- **1 (Maximum Similarity):**  
  The two vectors are identical or perfectly aligned.  
  The angle between the vectors is **0°**.

- **0 (No Similarity):**  
  The vectors are orthogonal (perpendicular) to each other.  
  They have no common direction or meaning.

- **-1 (Maximum Oppositeness):**  
  The two vectors are exactly opposite in direction.  
  The angle between the vectors is **180°**.

#### **Cosine Similarity in the Context of Embeddings:**
- A high value (close to 1) indicates that two embeddings share similar meanings or characteristics.
- A low value (close to -1) indicates that the embeddings are opposite in meaning.
[...]

PR #11876, opened 2025-03-19
Fix: wrong citation order for chromadb, wrong order for hybrid search

mahenning:
# Pull Request Checklist

# Changelog Entry

### Description

This PR fixes multiple embedding distance related issues with chromadb.

### Fixed

- Fix wrong order of citations in non-hybrid search when using chromadb
- Fix wrong ordering of documents when merging and picking the k best docs in hybrid search with chromadb
- Now picks best score for a document when it has multiple scores from different queries

---

### Additional Information

- Fix 1: Chromadb gives the actual distance (instead of a score) in [0, 2] with 0 distance as the best. While this is handled in merging the queries and picking the best k, the citations are ordered by "Relevance", which assumes 0 is worst and 1 is best. Chroma distance is 0 (best), 1 (unrelated) and 2 (worst). This resulted in the worst distance being at the top of citations. After picking the k best results, now transforms the [0, 2] distance into [-1, 1], where 1 is best.
- Fix 2: In hybrid search, it still switched the `reverse` parameter in `merge_and_sort_query_results` when chromadb is used. But the results of the hybrid search uses scores of the reranker, which are already correctly ordered (0 is worst, 1 is best). This led to picking the k worst results instead of the best ones.
- Fix 3: When merging documents, a set of hashes was used to check if a document was already added. This led to a "first comes first serves" scenario, where a document with distance x was inserted, but when the same doc with distance y from a different query comes in later in the loop, it isn't inserted even if score y > score x. It now checks if the new distance is better and overwrites the entry in that case, leading to the best score for this document chunk remaining.

### DIFF
--- a/backend/open_webui/retrieval/utils.py
+++ b/backend/open_webui/retrieval/utils.py
@@ -178,8 +178,7 @@ def merge_and_sort_query_results(
     query_results: list[dict], k: int, reverse: bool = False
 ) -> dict:
     # Initialize lists to store combined data
-    combined = []
-    seen_hashes = set()  # To store unique document hashes
+    combined = dict()  # To store documents with unique document hashes
 
     for data in query_results:
         distances = data["distances"][0]
@@ -192,10 +191,19 @@ def merge_and_sort_query_results(
                     document.encode()
                 ).hexdigest()  # Compute a hash for uniqueness
 
-                if doc_hash not in seen_hashes:
-                    seen_hashes.add(doc_hash)
-                    combined.append((distance, document, metadata))
+                if doc_hash not in combined.keys():
+                    combined[doc_hash] = (distance, document, metadata)
+                    continue  # if doc is new, no further comparison is needed
 
+                # if doc is alredy in, but new distance is better, update
+                if not reverse and distance < combined[doc_hash][0]:
+                    # Chroma uses unconventional cosine similarity, so we don't need to reverse the results
+                    # https://docs.trychroma.com/docs/collections/configure#configuring-chroma-collections
+                    combined[doc_hash] = (distance, document, metadata)
+                if reverse and distance > combined[doc_hash][0]:
+                    combined[doc_hash] = (distance, document, metadata)
+
+    combined = list(combined.values())
     # Sort the list based on distances
     combined.sort(key=lambda x: x[0], reverse=reverse)
 
@@ -203,6 +211,12 @@ def merge_and_sort_query_results(
     sorted_distances, sorted_documents, sorted_metadatas = (
         zip(*combined[:k]) if combined else ([], [], [])
     )
+
+    # if chromaDB, the distance is 0 (best) to 2 (worse)
+    # re-order to -1 (worst) to 1 (best) for relevance score
+    if not reverse:
+        sorted_distances = tuple(-dist for dist in sorted_distances)
+        sorted_distances = tuple(dist + 1 for dist in sorted_distances)
 
     # Create and return the output dictionary
     return {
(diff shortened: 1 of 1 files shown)


rgaricano:
fix https://github.com/open-webui/open-webui/issues/11822  here?

mahenning:
Okay this seems to be more of a general case instead of a chromadb-only case. For each supported db, there seems to be a wrapper class. A solution I would advocate for is to add a variable "reverse_score" (or something) to each db class constructor, which indicates how the db scores should be handled. Then this can be used with 
```
if VECTOR_DB_CLIENT.reserse_score:
    # code
```
This make the code cleaner instead of bloating a `if "chromadb":` to `if ["chromadb", "pgvector", "otherdb"...]`

mahenning:
I could implement this. The question is, should I do it in this PR too? Would make sense to me, because the wrong ordering is addressed here. On the other hand, the PR gets quite big.

rgaricano:
Something that i didn't test and that i haven't clear is if this a database issue or a emmbeded data,
I mind, e.g. if I made a first default openwebui install with chromadb and late i migrate the openwebui.db to posgresql &pgvector, how this affect this particulatity? is done in embedded vectors? or is independet for the case?

mahenning:
I hope I understand your question right.
It seems that different databases give different scores for cosine similarity. Some databases (like chromadb, pgvector) return the score as "distance", so 0 is best and 1 is worse. Some other databases (like qdrant) return the score as "similarity", so 1 is best and 0 is worse. If you migrate your vectors to another db, it does not affect the database and the vectors directly, but the score of query results can change from "distance" to "similarity" '(or the other way around). Open-webui has to handle "distance" and "similarity" databases and their results different.

rgaricano:
I remember that in a conversation this was discussed with some basis and references, but I can't find it.

I found thats, both of [Schwenn2002](https://github.com/Schwenn2002)
https://github.com/open-webui/open-webui/discussions/9001
https://github.com/open-webui/open-webui/issues/9106

rgaricano:
I found it, was a PR: https://github.com/open-webui/open-webui/pull/8379

mahenning:
> I remember that in a conversation this was discussed with some basis and references, but I can't find it.
> 
> I found thats, both of [Schwenn2002](https://github.com/Schwenn2002) #9001 #9106

I've linked them the PRs. Looks like they want more or less the features I implemented.

> I found it, was a PR: https://github.com/open-webui/open-webui/pull/8379

This PR added the `reverse` parameter, it is just used wrong in hybrid search.

> and a sugestion for speed it the proccess caching BM25 retriever: https://github.com/open-webui/open-webui/discussions/8582#discussioncomment-11912522

Looks interesting! I'm not that deep into this topic though.

naliwai:
Since the wrapper classes for each database are mostly implemented maybe it makes sense to compute the right metric directly there at query time? And not postpone it to later stages? Relation between cosine_similarity and cosine_distance is 
cosine_similarity = 1 - cosine_distance.

tjbck:
Thanks!

rgaricano:
I update a branch with this commit + k_rerank https://github.com/open-webui/open-webui/pull/11497 +parallelize https://github.com/open-webui/open-webui/pull/11814
and two pdf fix, pdf generator (markdown code) & pdf/ocr (unstructuredPDF) (there are a couple of PR closed)
with unrestricted python dependencies to work without downgrade anyone installed (pyproject.toml & requirements changes & i think that tsconfig.json too for work with never node)
and with main icons changed to be distingible on browser tabs (I was very confused with so many OIs)

If anyone want test this changes, i think that i mixed its correctly, this branch:
https://github.com/rgaricano/open-webui/tree/main-updated

PR #11911, opened 2025-03-20
fix: error message propagation from filter-pipeline

diwakar-s-maurya:
# Changelog Entry
Fix error message propagation from filter-pipeline

### Description

Error message returned from filter-pipeline was not being shown on UI. It always showed "Connection closed".
With this fix it will show the error message on the UI from the filter-pipeline properly.

### Screenshots or Videos

- Before
image

- After
For filter https://github.com/open-webui/pipelines/blob/main/examples/filters/rate_limit_filter_pipeline.py
image


tjbck:
Thanks!

PR #12167, opened 2025-03-29
Fix alignament buttons text

rgaricano:
Fix text buttons alignament

Fixed align tag on user & admin settings buttons

- correction of some align tag assigned to buttons that are right and
have to be left, this means that although it is not noticeable in
English, in other languages, especially if several lines appear, it is
obvious.

openwebui_aligns_config

openwebui_aligns_config1


tjbck:
Unwanted changes, please review your PR before creating one!

rgaricano:
fck hell,
sorry, i have a mixed things.... WIP

PR #12236, opened 2025-03-31
Fix middleware.py missing prefix on embedding function

Phlogi:
Missing prefix after commit https://github.com/open-webui/open-webui/commit/433b5bddc159f50d8e2183f18050481ac49e8cce


tjbck:
Thanks!

PR #12259, opened 2025-04-01
[FEAT] adjust Translations for temporary chat

weisser-dev:
i fixed, that the temporary chat translations will work, currently in all languages it is "Temporary Chat" cause the i18n is not used within the Placeholder.svelte and ChatPlaceholder.svelte i also add translation for enforce temporary chat in german


src/lib/components/chat/ChatPlaceholder.svelte
```diff
@@ -67,15 +69,15 @@
...
-				content="This chat won't appear in history and your messages will not be saved."
-				className="w-fit"
-				placement="top-start"
-			>
-				<div class="flex items-center gap-2 text-gray-500 font-medium text-lg my-2 w-fit">
-					<EyeSlash strokeWidth="2.5" className="size-5" /> Temporary Chat
-				</div>
-			</Tooltip>
+		<Tooltip
+			content={$i18n.t(tempChatTooltipText)}
```
tjbck, line 73: `$i18n.t('This chat won’t appear in history and your messages will not be saved.')`

weisser-dev: ok i will resolve this

weisser-dev: @tjbck fixed

tjbck:
Thanks!

PR #12260, opened 2025-04-01
[improvement] default permission for new groups is false for enforce …

weisser-dev:
currently in the Groups.svelte and Permissions.svelte the value is "true"; so if a admin creates a new group, the temporary chat is set to true by default, i guess most of the users need this permission set to false


tjbck:
Thanks!

PR #12266, opened 2025-04-01
[improvement] insert temporary-chat=false logic & remove console.log()

weisser-dev:
Improved Temporary Chat Behavior
**Current Behavior:**
- If the URL parameter temporary-chat is set to "true", temporary chat is enabled.
- If no URL parameter is set, the system checks for the user's permission.
- If the permission exists, the user has no option to disable temporary chat.

**Improvement:**
- Now, the user can explicitly set the URL parameter to "false" to disable temporary chat.
- This gives users more control, as the logic now follows:
- URL parameter > Permission

This change ensures that users can override enforced permissions via the URL parameter if needed. 🚀

Maybe this is needed? If not, just close it. When I thought about implementing this feature, I thought this logic would be nice for the user.

### DIFF
--- a/src/routes/(app)/+layout.svelte
+++ b/src/routes/(app)/+layout.svelte
@@ -195,13 +195,16 @@
 				showChangelog.set($settings?.version !== $config.version);
 			}
 
-			if ($page.url.searchParams.get('temporary-chat') === 'true') {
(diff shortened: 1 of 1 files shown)


tjbck:
Enforced temporary chat should disable toggling off temporary chat.

PR #12289, opened 2025-04-01
Fix: num_keep param explanation/exemplified (en-US)

rgaricano:
Def. n_keep / num_keep:
- It's used when the n_ctx limit is reached. A new prompt will be constructed with the first n_keep characters of the original prompt plus the second half of the output to free up space for more conversation.
- It's the amount of tokens after your system prompt that are always kept, when the start of the history is thrown away.

First N, no Last N !!!

In Params Configs it's wrong exemplified, as:
"This option controls how many tokens are preserved when refreshing the context. For example, if set to 2, the **last** 2 tokens of the conversation context will be retained. Preserving context can help maintain the continuity of a conversation, but it may reduce the ability to respond to new topics."

When have to be, as:
"This option controls how many tokens are preserved when refreshing the context. For example, if set to 2, the **first** 2 tokens of the conversation context will be retained. Preserving context can help maintain the continuity of a conversation, but it may reduce the ability to respond to new topics."

Ollama's num_keep func.: https://github.com/ollama/ollama/blob/c001b98087e45b7b60509127d4d2e9d9ba809444/runner/ollamarunner/runner.go#L124-L131


rgaricano:
Frontend build error by other global causes.

rgaricano:
I made other changes to spanish translation in other commit, but i don't know why it was added to this pull request.

UPD: Spanish Translation (es-ES) v0.6.0
- minors corrections

agree if it's merged too, i will be more cautious

PR #12376, opened 2025-04-03
feat: Add option to disable OAuth profile pictures

### LINKED ISSUE #12325: issue: Microsoft SSO Profile Pictures Causing Severe Performance Degradation in Admin Panels
### Check Existing Issues

### Installation Method

Git Clone

### Open WebUI Version

v0.6.0

NA

### Operating System

Amazon Linux (EC2, Docker container)

### Browser (if applicable)

NA

### Confirmation

- [x] I have listed steps to reproduce the bug in detail.

### Expected Behavior

Microsoft SSO authentication should work efficiently without causing significant performance degradation.
There should be an option to use default profile images instead of loading images from the OAuth provider.
When setting a non-existent claim in OAUTH_PICTURE_CLAIM, it should fall back to the default user image rather than the provider's picture URL

### Actual Behavior

Microsoft SSO authentication works, but it causes extreme performance issues when loading user data.
The Knowledgebase section and Users section in the admin panel for us took approximately 30 seconds to load just 5 users, pulling large base64 strings. (there are some large images in our tenancy)

Setting OAUTH_PICTURE_CLAIM to a non-existent claim (e.g., "no-claim") doesn't work as expected because the code falls back to the provider's picture URL rather than using the default image.
The application retrieves full-size profile pictures (1-10MB) rather than thumbnails, causing significant load times.

### Steps to Reproduce

1. Set up Open WebUI v0.6.0 with Docker on EC2 (Amazon Linux)
2. Use external postgres database
3. Configure Microsoft Entra SSO integration with the proper environment variables
4. Have users authenticate with Microsoft SSO
5. Attempt to access the Knowledgebase section or Users section in the admin panel
6. Observe longer than expected load times for the knowledgebase, and users section in admin panel
7. Attempt to resolve by setting OAUTH_PICTURE_CLAIM="no-claim" in your environment variables
8. Notice that this doesn't resolve the issue as the code falls back to the OAuth provider's hardcoded picture URL (OAUTH_PROVIDERS)

### Logs & Screenshots

NA

### Additional Information

Root Cause:
The issue is in backend/open_webui/utils/oauth.py at line 330, where the code is designed to fall back to the OAuth provider's picture URL rather than an empty string when the specified claim doesn't exist:

`picture_url = user_data.get(
    picture_claim, OAUTH_PROVIDERS[provider].get("picture_url", "")`

This means that when OAUTH_PICTURE_CLAIM is set to a non-existent claim, instead of using an empty string (which would trigger the default user image), it falls back to whatever is defined in 
`OAUTH_PROVIDERS[provider].get("picture_url", "").`

Solution:
Change line 330 in backend/open_webui/utils/oauth.py from:

`picture_url = user_data.get(
    picture_claim, OAUTH_PROVIDERS[provider].get("picture_url", "")`

To:

`picture_url = user_data.get(picture_claim, "")`

This change ensures that when a claim doesn't exist, it falls back to an empty string, which causes the application to use the default user image (user.png) instead of attempting to load the images from Microsoft Entra.

I think that a better solution would be to create an environment variable for OAUTH_USE_PICTURE_CLAIM (bool) to toggle between using the picture claim, or the user.png

MushroomLamp-COB:
### Pull Request Description
This PR adds a new configuration option OAUTH_USE_PICTURE_CLAIM that allows administrators to toggle whether OAuth profile pictures are used or if the default user.png avatar is used instead. This addresses issue #12325, providing a simple way to disable the use of profile pictures from OAuth providers.

### Changelog Entry
**Description**
Added a new environment variable and configuration option to control whether profile pictures from OAuth providers are used or if the default avatar is used instead. This provides administrators with greater control over user avatars in their Open WebUI deployment.
Added

Added OAUTH_USE_PICTURE_CLAIM environment variable and corresponding PersistentConfig entry
Added conditional logic in oauth.py to respect this setting when processing user profile pictures

**Changed**

Modified the OAuth authentication flow to check the new configuration option before attempting to fetch and process profile pictures

**Fixed**

Fixes #12325: Added option to turn off the use of profile pictures from OAuth claims

### Additional Information
This change is minimal and focused on providing administrators with greater control over avatar usage. The default setting is True, which maintains backward compatibility with existing behavior.

Tested on a docker, expected functionality for True, False, and not specified values.
(this is my first time doing a PR, hopefully this is ok)

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -331,6 +331,12 @@ JWT_EXPIRES_IN = PersistentConfig(
 # OAuth config
 ####################################
 
+OAUTH_USE_PICTURE_CLAIM = PersistentConfig(
+    "OAUTH_USE_PICTURE_CLAIM",
+    "oauth.oidc.use_picture_claim",
+    os.environ.get("OAUTH_USE_PICTURE_CLAIM", "True").lower() == "true",
+)
+
 ENABLE_OAUTH_SIGNUP = PersistentConfig(
     "ENABLE_OAUTH_SIGNUP",
     "oauth.enable_signup",
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -23,6 +23,7 @@ from open_webui.config import (
     OAUTH_PROVIDERS,
     ENABLE_OAUTH_ROLE_MANAGEMENT,
     ENABLE_OAUTH_GROUP_MANAGEMENT,
+    OAUTH_USE_PICTURE_CLAIM,
     OAUTH_ROLES_CLAIM,
     OAUTH_GROUPS_CLAIM,
     OAUTH_EMAIL_CLAIM,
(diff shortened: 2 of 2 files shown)


tjbck:
LGTM!

PR #12409, opened 2025-04-03
Feat: optional update toast as environment variable

victorstevansuse:
# Changelog Entry

### Description

These changes add a feature flag `ENABLE_UPDATE_NOTIFICATION` to disable the update toast notification. Deployments may choose not to upgrade if their infrastructure is not ready to handle it, with that in mind it is not always possible for the OI admin to update, given these external factors. The update sometimes might not be up to the OI admin.

### Added

- `ENABLE_UPDATE_NOTIFICATION` environment variable to disable the update notification through the global configs. Default `True`.
- Changes in the UI to only show the 'Show "What's New" modal on login' setting interface option when the mentioned environment variable is truthy. 

### Changed

- Conditional rule to render the toast in the frontend.

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -460,3 +460,8 @@ OTEL_TRACES_SAMPLER = os.environ.get(
 
 PIP_OPTIONS = os.getenv("PIP_OPTIONS", "").split()
 PIP_PACKAGE_INDEX_OPTIONS = os.getenv("PIP_PACKAGE_INDEX_OPTIONS", "").split()
+
+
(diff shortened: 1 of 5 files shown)


tjbck:
While we understand the intent to respect infrastructure readiness, the update notification is meant to encourage timely upgrades, particularly because we occasionally ship important security patches. It’s essential that all users remain aware of new versions and known vulnerabilities.

If an individual OI admin prefers not to see the update toast, this can already be configured on a per-user basis via the existing "Show 'What’s New' modal on login" setting — which should address your immediate concern without requiring an environment-wide toggle.

Issue #12441, opened 2025-04-04
issue: Bad Request (400 no Body) from OpenAI Chat Completion API Endpoint in n8n-Tool-Integration


floriankick:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

0.6.0

### Ollama Version (if applicable)

0.6.2

### Operating System

Linux

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have listed steps to reproduce the bug in detail.

### Expected Behavior

The API endpoint /ollama/v1/chat/completions should be compatible with the corresponding endpoint of the Ollama instance. 

### Actual Behavior

The WebUI API endpoint /ollama/v1/chat/completions does not accept a message content=null whereas the corresponding Ollama API endpoint /v1/chat/completions can handle the same request.
This issue causes n8n OpenAI Chat model node to fail on the second request after calling a tool, because the tool call message from the LLM (that is sent as part of the conversation history within the context window) has an empty content field.

### Steps to Reproduce

Send a request to http://webui.instance/ollama/v1/chat/completions with the following body:
```
{
    "model": "<insert any model name>",
    "messages": [{"role": "system", "content": "You are a helpful assistant"}, {"role": "user", "content": "calculate the square root of 55"}, {"role": "assistant", "content": null, "tool_calls": [{"id": "call_0o4z52ag", "type": "function", "function": {"name": "calculator", "arguments": "{\"input\":\"sqrt(55)\"}"}}]}, {"role": "tool", "content": "7.416198487095663", "tool_call_id": "call_0o4z52ag"}],
    "stream": false
}
```
Use any model available, it doesn't matter because the request won't reach the model anyways.

For Reference send the same Request to http://ollama.instance/v1/chat/completions which works as expected.

### Logs & Screenshots

Docker logs:
```
[...]

PR #12447, opened 2025-04-04
fix: make content parameter optional in OpenAI chat completion API endpoint

floriankick:
# Changelog Entry

### Description

- Solves  #12441
- Improve compatibility of OpenAI chat completion endpoint with the corresponding endpoint in Ollama

### Changed

- `content` parameter in OpenAI chat completion endpoint is now optional.


tjbck:
Thanks!

PR #12461, opened 2025-04-04
fix: Don't cache index.html

phlmn:
# Changelog Entry

### Description

This PR adds `Cache-Control` headers to force the browser to always check `index.html` for changes. Without this, after updating open-webui users will receive the old index.html from their browser cache and sometimes run into errors.

### Fixed

- 🔄 **Caching Bug:** `index.html` is now revalidated on every request to prevent errors after updates

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -377,16 +377,26 @@ log.setLevel(SRC_LOG_LEVELS["MAIN"])
 
 
 class SPAStaticFiles(StaticFiles):
+    def __init__(self, *args, no_cache_paths, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.no_cache_paths = no_cache_paths
+
     async def get_response(self, path: str, scope):
         try:
-            return await super().get_response(path, scope)
+            res = await super().get_response(path, scope)
+            if path in self.no_cache_paths:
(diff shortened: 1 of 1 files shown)


rgaricano:
maybe is easier add header in https://github.com/open-webui/open-webui/blob/main/src/app.html ?

bet0x:
> # Changelog Entry
> ### Description
> 
> This PR adds `Cache-Control` headers to force the browser to always check `index.html` for changes. Without this, after updating open-webui users will receive the old index.html from their browser cache and sometimes run into errors.
> ### Fixed
> 
>     * 🔄 **Caching Bug:** `index.html` is now revalidated on every request to prevent errors after updates

This should not be forced on everyone.

tjbck:
We definitely don't want the cache to be removed everytime, would be nice if we could somehow intelligently remove cache on every update. Feel free to reopen!

Classic298:
@phlmn  what if there is a little function that gets the version number of openwebui and sends the version number to the webbrowser/client? This way the client can keep track of the versioning of the files and if the version changes, it's time to refresh the cache. This is a common way to do it intelligently :)

phlmn:
@tjbck bump
As mentioned in #13582 I think this would fix some issues.

PR #12467, opened 2025-04-04
feat: LDAP Group-Based Authorization in OpenWebUI #12456 via Auth headers

codespg:
Added Variable - WEBUI_AUTH_TRUSTED_GROUP_HEADER
Used the same variable to Identify the Request Header with Groups to sync in OpenWeb UI.

Group Configuration: Allow administrators to define groups in OpenWebUI and map them to existing LDAP groups.

Setup Apache/Reverse Proxy to send the Request header with Group details in Semicolon delimited.

Automatic Group Assignment: Upon login (via SSO Auth Headers) automatically assign users to their corresponding OpenWebUI groups based on their LDAP group memberships.

Note: The group parsing assumes a semicolon delimited string in the header. If different delimters needed we can configure variable accordingly and add another update.


codespg:
Hello Maintainer,

Can you verify and merge this pull request?

yu-xiaoli:
Thank you for this helpful task.

About the removal update process, I want to confirm whether Openwebui can still synchronize normally if when user is deleted from all groups, or should we not consider this scenario?

Assuming the original userA is member of Group 1 and Group 2, 
when the userA is removed from all existing Groups (Group 1, and Group 2), 
the Group information in the Header is empty， leading to the user_trusted_groups being []. 
Condition below would  cause the removal update processing  to be skipped incorrectly?

```
    # Remove groups that user is no longer a part of
    for group_model in user_current_groups:
        if user_trusted_groups and group_model.name not in user_trusted_groups:
            log.debug(
                f"Removing user from group {group_model.name} as it is no longer in their trusted header groups"
            )
```

Is it necessary to distinguish between the two scenarios of not obtaining group information in Headers and obtaining group information but the group is empty?

codespg:
Hello,
I have reused same group remal code used in oAuth group management. 

Since its better to add this condition to remove all group when empty list is received, i will add it and update here.

codespg:
Hello @yu-xiaoli,
Thanks for your suggestion, I have removed the extra validation to not remove groups if user_trusted_groups is empty.

tjbck:
Closing in favour of cce5f024bd63ba3a83630c3d3a69d48251b71d1d

PR #12515, opened 2025-04-06
chore: Fix dependabot configuration

gaby:
### Description

- Migrate package ecosystem to `uv` for python dependencies. See https://github.blog/changelog/2025-03-13-dependabot-version-updates-now-support-uv-in-general-availability/
- Fix directory for python, it should be the directory where the uv.lock/pyproject.toml are, not the code.
- Add missing entry for `npm` dependencies.
- Add target-branch to `github-actions` tracker.

With these fixes you should see +10 new PR's get created with dependency updates since the current dependabot configuration was not working at all.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,12 +1,20 @@
 version: 2
 updates:
-  - package-ecosystem: pip
-    directory: '/backend'
+  - package-ecosystem: uv
+    directory: '/'
     schedule:
       interval: monthly
     target-branch: 'dev'
+
+  - package-ecosystem: npm
+    directory: '/'
+    schedule:
+      interval: monthly
+    target-branch: 'dev'
+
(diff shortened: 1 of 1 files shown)


tjbck:
Thanks!

gaby:
@tjbck They will show up, once the change is merged to `main`

gaby:
@tjbck I just noticed that there's 3 files tracking dependencies

- pyproject.toml
- uv.lock
- backend/requirements.txt

Which one is the one that's supposed to be used?

tjbck:
`pyproject.toml` is used for pypi and `backend/requirements.txt` is used for our docker images.

gaby:
> `pyproject.toml` is used for pypi and `backend/requirements.txt` is used for our docker images.

In that case the previous entry needs to be re-added.

uv entry for top level dir
pip entry for `backend/` dir

PR #12521, opened 2025-04-06
chore: Only run GitHub workflows when needed

gaby:
### Description

- Currently workflows are run all the time, they should only run when needed.
- The backend workflow will only run when changes to `backend/`, `pyproject.toml` and `uv.lock` files are made.
- The frontend workflow will run when changes are *not* related to `Python`.
- Added `python3.12` to workflow


tjbck:
LGTM, Thanks!

PR #12526, opened 2025-04-07
chore: Dependabot update

gaby:
### Description

- Re-added `pip` tracker for `backend/requirements.txt`  since that file is separate from the main `pyproject.toml`.

Related to #12515


tjbck:
👍

PR #12594, opened 2025-04-08
feat: Add support for searching files by name

gaby:
### Description

- Add support for searching files by name, or wildcard.


gaby:
@tjbck Before I mark it as ready, wanted to get your opinion on this. We had a user point out that searching for files by `file_id` wasn't intuitive since it required multiple calls.

tjbck:
Merging this for now but we may want to turn this into db query later down the line, Thanks!

Related: https://github.com/open-webui/open-webui/blob/63533c9e3ab41edd7bd4124ef94f6b6dc09aa175/backend/open_webui/models/chats.py#L531

PR #12615, opened 2025-04-08
chore: Add debug logging to RAG utils

robert-norberg:
# Changelog Entry

### Description

Adds debug logging to RAG utils

### Added

Additional details will be logged if `RAG_LOG_LEVEL` is set to `DEBUG`.

---


backend/open_webui/retrieval/utils.py
```diff
@@ -257,6 +261,7 @@ def query_collection(
 ) -> dict:
     results = []
     for query in queries:
+        log.debug(f"query_collection:query {query}")
```
bbartels, line 264: Might be best not to log the raw prompt, at least in production. Could feature flag it to `dev`?

robert-norberg: I'm assuming you're not running in prod with `RAG_LOG_LEVEL` set to `DEBUG`. If you are, it's already logging all of the extracted text from every document you upload.

tjbck:
Thanks!

PR #12664, opened 2025-04-09
Updated Hungarian language to full. translation.json

Beni34:
I completely translated the program into Hungarian.


tjbck:
Thanks!

Issue #12690, opened 2025-04-10
security: Private Prompts in Workspace are visible to all users regardless


Classic298:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Pip Install

### Open WebUI Version

0.6.2

### Operating System

Debian 12

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have listed steps to reproduce the bug in detail.

### Expected Behavior

When a prompt is saved in the workspace and the prompt is saved as a **private prompt**, it remains private and is not available to all users in the chat interface by accessing the stored prompts with the `/` command.

### Actual Behavior

All prompts, even when stored as a private prompt, are available to all users in the chat interface. By writing `/` all the available prompts get listed and can be accessed by anyone

### Steps to Reproduce

1. As a normal user, go to workspace
2. Create a new prompt - make sure to select 'private' for the permissions OR have the admin disable the public sharing permission altogether
3. Save the prompt
4. Ask another user to enter `/` into their chat interface and observe the full list of all prompts being shown, even those of other users.

### Logs & Screenshots

Browser Logs not applicable

Issue #8586, opened 2025-01-15
Prompts created by regular users should not be public by default


DmitriyAlergant:
# Feature Request

**Is your feature request related to a problem? Please describe.**

Using Open WebUI in an enterprise setting, we would like to grant end-users (non admins) an ability to create custom prompts.  This has become possible with the recent introduction of Groups feature, and Default Permission setting.

However, currently, prompts created by end-users by default have "public" visibility.  We are afraid of creating a mess and therefore cannot enable the feature.

**Describe the solution you'd like**

Prompts created by non-admin users should only have private visibility (by default), or, optionally, visible to Group that this user is a member of.

Ideally, the same logic should apply to Models and Knowledge Documents - but the current issue in pacticular is about Prompts.

**Describe alternatives you've considered**

There is no meaningful alternatives currently.

**Additional context**

tarmst:
Plugging my PR that was closed: https://github.com/open-webui/open-webui/pull/8364

flefevre:
We have also the same difficulty with model assistants. A model assistant should be made public only by a privileged user to ensure, we do not have a mess in assistant listing.
Typically the admin/privileged user should ensure that a logo has been defined, the description is fulfilled, the tags have been pitted etc... To ensure quality of model assistant. Ideally we could have a star,like feature on each model assistant given by the user of given by the analysis through giskard.

In the admin panel, it could be good to have a toogle button to autorise or not the switch by enduser or privileged user(ideally)

davizucon:
Hey Folks, 
Is there any intention to continue discussing the feature? Or has anyone found a workaround?

Thanks !!!

Classic298:
It is implemented indeed but it is faulty!

Even though private prompts are hidden in the workspace for other users, other users can still access privately saved prompts in the chat interface by accessing the prompts with the / command. Literally all prompts even of other users will be listed there, even if set to private!

Related

https://github.com/open-webui/open-webui/issues/12690

Classic298:
Created PR: https://github.com/open-webui/open-webui/pull/12757

PR #12757, opened 2025-04-11
fix: Security - Default prompt permissions to respect user access rights

Classic298:
# fix: Default prompt permissions to respect user access rights

**Pull Request Checklist**

## Pull Request Checklist
- [x] **Description:** A detailed description of the security vulnerability and fix is provided
- [x] **Changelog:** Added below
- [X] **Documentation:** No documentation updates needed as this maintains existing functionality but fixes security
- [X] **Dependencies:** No new dependencies added
- [ ] **Testing:** I've verified the logic, but request maintainer testing with various permission configurations
- [x] **Code review:** Self-reviewed for security implications and code correctness
- [x] **Prefix:** Used "fix" as this addresses a security vulnerability

## Description
This PR addresses a security vulnerability where prompts could be unintentionally shared with all users, bypassing access control restrictions. When users create prompts without explicitly setting access control permissions (by not clicking the "Access" button), the system defaults to public access (`null`) _even when users don't have permission to create public prompts or the admin has disabled public sharing._

**Vulnerability details:**
What is the vulnerability / how to execute the vulnerability:

- By default, all prompts are created with `accessControl = null` (public access), unless the user manually, explicitly opens the access-settings when creating a new prompt - in that case, his actual permissions would be loaded and correctly set. If the user never opens the access settings during prompt creation, the prompt is **always saved as a public prompt, regardless of public sharing permissions.**
- When a user doesn't open the access control modal during prompt creation, they inadvertently, always create public prompts
- This results in private/sensitive prompts being exposed to all users when they type `/` in the chat interface

**Important note to maintainer (Tim):**
While I've carefully reviewed the changes, I kindly request your additional verification and testing. The scope was intentionally kept narrow to address just the vulnerability without introducing regressions.

**Impact:**
Based on CVSS, this would likely score as a Medium severity vulnerability due to the unauthorized information disclosure across user boundaries with no special privileges required to exploit.

## **If you are affected: How to fix?**
If you are affected by this here is a fix for affected prompts:
**As a user:**
Edit all your prompts, open the access settings and ensure it is set to private, then save the prompt.

**As an admin:**
Edit all the prompts that your users created and check the access settings. If the access settings show "Public", then this prompt is affected by the vulnerability (user did not open access settings during prompt creation). Change it to private and save the edited prompt. Do this for all prompts.

## Solution
The fix modifies `PromptEditor.svelte` to default prompt access based on user permissions:
- Users with public sharing permission still default to public prompts (backward compatible)
- Users without public sharing permission now default to private prompts
- Existing prompts maintain their current access control settings
- \*\* **Does NOT change the default behaviour of the access settings when creating a prompt. If you have the ability to create public prompts, the default behaviour will remain to create public prompts. If you do not have the ability to create public prompts, your prompts will only be created privately** \*\*

This ensures the principle of least privilege - users can only share prompts publicly when they explicitly choose to AND have the permission to do so.

### Changed
- Changed default access control value to be based on user permissions rather than always public, regardless of the user's permissions set by admins

### Fixed
- Fixed vulnerability where prompts could be unintentionally shared with all users
- Fixed access control defaults to respect user permissions set by admins

### Security
- Addressed information disclosure vulnerability with prompt sharing

---

### Additional Information
- Fixed missing comma for coding-style-guideline adherence in unrelated file
- The fix should maintain full backward compatibility with existing prompts
- This implements proper default security based on user permissions rather than requiring explicit action to make prompts private
- This fix was co-created with AI (Claude 3.7 Sonnet)

### Related Issue / Discussion
- Originally identified in issue: https://github.com/open-webui/open-webui/issues/12690
- Later discussed here: https://github.com/open-webui/open-webui/discussions/12718
- Related: https://github.com/open-webui/open-webui/issues/8586


Classic298:
Alternative fix was deployed by Tim it seems: https://github.com/open-webui/open-webui/commit/59dd79815651625e84097365f492223d077413fa

PR #12890, opened 2025-04-15
Fix: Remove own vector collection from knowledges

mahenning:
# Pull Request Checklist

# Changelog Entry

### Description

- Removed the file processing step from "adding file to knowledge" as it led to a second entry in the vector db 
- Removed knowledges own vector collection

### Changed

- Knowledges now have no own vector "collection"

### Fixed

- Removed doubled file processing when added file to knowledge

---

### Additional Information

- **Before the fix** the process of "adding a file to a knowledge through the UI" was as followed:
file upload -> add to knowledge
#### file upload:
- file is uploaded into the "uploads" directory
- file is added to the "file" table in the `webui.db` database
- **file content is chunked and added to the vector_db with the collection name "file-{id}"**

#### add to knowledge
- file is checked if it is in `file` table of `webui.db`
- **file content is chunked and added to the vector db with the collection name "{knowledge-id}" (added to all other vectors in this collection)**
- file id is added to the knowledge entry

The problem is marked in bold. The file is added two times to the vector db, and the first vector entry "file-{id}" is never used again because each knowledge manages its own vector data on top of each file and the vector data of "{knowledge-id}" is used. This leads to bloat in the vector db and orphaned collections. This PR fixes that as the knowledge now uses the correct "file-{id}" vector entries that are created at file upload.

Two potentially negative points arise from that:
- Befor the PR, there was one (huge) call to the vector db, as the vector knowledge entry held every document in one collection. This is now split into a "retrieve-and-rerank" step for each file, which could be a bit slower
- If you now try to add an unprocessed file to the knowledge, the adding fails. Before that, the knowledge processed the file itself anyway. I see this as an absolute corner case, as one has to explicitly call the file upload API endpoint with "process=False" and I don't see the point in adding unprocessed files. The case that the file is not found in the `file` table is already covered.

Edit: I've so far reworked the knowledges handling of vector collections. The logic is that only files should have an entry in the vector db, and knowledges should only be a collection of files (not vector "collection" here). If a file is added to multiple knowledges with the API, a knowledge checks if that file is used in any other knowledge on deletion  (in the UI, each upload is a separate file with its own UUID, so this is already a corner case).

### DIFF
--- a/backend/open_webui/retrieval/utils.py
+++ b/backend/open_webui/retrieval/utils.py
@@ -490,7 +490,7 @@ def get_sources_from_files(
                 if file.get("legacy"):
                     collection_names = file.get("collection_names", [])
                 else:
-                    file_ids = file.get("data", []).get("file_ids", [])
+                    file_ids = (file.get("data") or {}).get("file_ids", [])
                     file_ids = [f"file-{file_id}" for file_id in file_ids]
                     collection_names.extend(file_ids)
             elif file.get("collection_name"):
--- a/backend/open_webui/retrieval/utils.py
+++ b/backend/open_webui/retrieval/utils.py
@@ -490,7 +490,9 @@ def get_sources_from_files(
                 if file.get("legacy"):
                     collection_names = file.get("collection_names", [])
                 else:
-                    collection_names.append(file["id"])
+                    file_ids = file.get("data", []).get("file_ids", [])
+                    file_ids = [f"file-{file_id}" for file_id in file_ids]
+                    collection_names.extend(file_ids)
             elif file.get("collection_name"):
                 collection_names.append(file["collection_name"])
             elif file.get("id"):
--- a/backend/open_webui/routers/knowledge.py
+++ b/backend/open_webui/routers/knowledge.py
@@ -182,8 +182,11 @@ async def reindex_knowledge_files(request: Request, user=Depends(get_verified_us
         try:
             files = Files.get_files_by_ids((knowledge_base.data or {}).get("file_ids", []))
 
+            reindexed_files = []
             failed_files = []
             for file in files:
+                if file.id in reindexed_files:
+                    continue  # file already reindexed in another knowledge
                 try:
                     if VECTOR_DB_CLIENT.has_collection(
                         collection_name=f"file-{file.id}"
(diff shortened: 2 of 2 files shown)


mahenning:
TODOs:
- resolve: knowledge update/deletions
- resolve: upload of the same file is now possible

tjbck:
"file-{id}" IS used. You can reference individual files from the chat input. PLEASE check for previous PRs and discussions.

mahenning:
Yeah, and file-{id} IS used here too! It just removes the vector space of {knowledge id}. Please reconsider your decision or give me an example why this here is not working. Again, I *did not* remove "file-{id}", but make the knowledge also use it.
To make my point hopefully clearer, an example from the current code:

Case: A file gets uploaded to a knowledge base in the UI. 
`upload_file` @ routers/files.py is called, which calls `process_file` to upload the file into the vector db. A collection_name parameter is not proveded, so "file-{file-id}" is used as collection name (second code block):
https://github.com/open-webui/open-webui/blob/07d8460126a686de9a99e2662d06106e22c3f6b6/backend/open_webui/routers/files.py#L132-L133
https://github.com/open-webui/open-webui/blob/07d8460126a686de9a99e2662d06106e22c3f6b6/backend/open_webui/routers/retrieval.py#L964-L967


After upload, `add_file_to_knowledge_by_id` @ /routers/knowledge.py is called,  which *also* calls `process_file` with the parameter `collection_name=id` (which is the knowledge id):
https://github.com/open-webui/open-webui/blob/07d8460126a686de9a99e2662d06106e22c3f6b6/backend/open_webui/routers/knowledge.py#L357-L363


And in `process_file`, this lines are called in both cases:
https://github.com/open-webui/open-webui/blob/07d8460126a686de9a99e2662d06106e22c3f6b6/backend/open_webui/routers/retrieval.py#L1081-L1094

Only the second time, `collection_name` is not "file-{file-id}", but "{knowledge-id}", and "add=True", adding the file content to the collection "{knowledge-id}". This is redundant, as every (properly uploaded) file has already a collection with "file-{file-id}".

tjbck:
Won't this cause a significant performance overhead when querying against 1,000+ documentst?

mahenning:
Hm, fair point. I'll test it with a stack of dummy data and report back. I admit, I didn't fully consider that now for each doc, the vector->embed->rerank pipeline is executed and if/how this will affect performance with the new parallel thread pool.

mahenning:
You can leave this on closed. The performance hit is measurable (1s vs 30s for 1500 pdfs with 1-5 pages of text) and I can't efficiently manage the multi-collection reads with chroma in the asyncio/threadpool environment. I'm still not happy about the doubled vector space, especially for the 1000+ docs case but I do see the point in exchanging disk space for ease and performance. Thanks for checking with the performance question.

PR #12947, opened 2025-04-17
feat: Add gRPC support to Qdrant

gaby:
### Description

- Adds support for gRPC as an alternative method for Qdrant. Using gRPC provides higher performance.
- Current behavior is kept of default to HTTP.

Support for gRPC is documented here: https://github.com/qdrant/qdrant-client#grpc

Fixes https://github.com/open-webui/open-webui/issues/12677
Related https://github.com/open-webui/open-webui/discussions/12722


tjbck:
LGTM, Thanks!

Issue #12811, opened 2025-04-13
issue: Web search results *once again* all get source id=1


Ithanil:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

0.6.4

### Operating System

Debian 12

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have listed steps to reproduce the bug in detail.

### Expected Behavior

One source id per result page.

### Actual Behavior

Same source id for all results.

### Steps to Reproduce

Execute web search with embedding&retrieval or without, look at the source ids in the content sent to LLM.

### Logs & Screenshots

![Image](https://github.com/user-attachments/assets/7ab99adb-aff1-4257-926e-d0515f1e1c79)

### Additional Information

This was fixed by https://github.com/open-webui/open-webui/pull/12506 and https://github.com/open-webui/open-webui/pull/12507 in 0.6.2 and is now broken again in 0.6.4 due to https://github.com/open-webui/open-webui/pull/12562 , despite my comments and request to check for it.

Reverting https://github.com/open-webui/open-webui/pull/12562 fixes the issue for web search, but of course then source ids will be broken again for knowledge bases.

almajo:
Too bad fix #12562 for RAG broke Web Search citations again. As listed in the PR, I haven't tested this route tbh because I never use it and didn't think it would impact web search.

Before going back and forth on this, I would really recommend using identifiers for llm citations instead of numbers. Something like the first 6 letters of the file_id for example. This would make citations way more obvious to handle, be it tools, web search or rag search. Additionally, we would be able to implement something for https://github.com/open-webui/open-webui/discussions/12247 . This is basically impossible to nicely fix without clear identifiers that don't rely on an order or the full source list.

Elmolesto:
@almajo I Just noticed it's happening with the collection citations, too. The PR didn't solve the issue 😅

https://github.com/open-webui/open-webui/issues/12655  is still happening 🫤

CC @tjbck

Ithanil:
fixed in v0.6.6

PR #12562, opened 2025-04-07
fix: Correct Citations for knowledge base models

almajo:
# Pull Request Checklist

### Note to first-time contributors: Please open a discussion post in [Discussions](https://github.com/open-webui/open-webui/discussions) and describe your changes before submitting a pull request.

**Before submitting, make sure you've checked the following:**

- [X] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [ ] **Description:** Provide a concise description of the changes made in this pull request.
- [ ] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ ] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [ ] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [ ] **Testing:** Have you written and run sufficient tests to validate the changes?
- [X] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [ ] **Prefix:** To clearly categorize this pull request, prefix the pull request title using one of the following:
  - **BREAKING CHANGE**: Significant changes that may affect compatibility
  - **build**: Changes that affect the build system or external dependencies
  - **ci**: Changes to our continuous integration processes or workflows
  - **chore**: Refactor, cleanup, or other non-functional code changes
  - **docs**: Documentation update or addition
  - **feat**: Introduces a new feature or enhancement to the codebase
  - **fix**: Bug fix or error correction
  - **i18n**: Internationalization or localization changes
  - **perf**: Performance improvement
  - **refactor**: Code restructuring for better maintainability, readability, or scalability
  - **style**: Changes that do not affect the meaning of the code (white space, formatting, missing semi-colons, etc.)
  - **test**: Adding missing tests or correcting existing tests
  - **WIP**: Work in progress, a temporary label for incomplete or ongoing work

# Changelog Entry

### Description

The RAG-pipeline was not citing correctly when using a model equipped with one or more knowledge bases. This small change corrects the citations by giving the model the actual document id of the chunk-context-source (instead of the knowledge base id). This works all the way through frontend citation parsing due to the consistent order of the source objects.

Cases citation might break: When the metadata.name is not unique within the sources (e.g. multiple files with the same ' name'). Then the deduplication step in citation rendering takes some sources out and the numbers are not aligned any more.

### Fixed

[...]

PR #12971, opened 2025-04-17
fix: Align backend <source> tag indexing with frontend citation grouping

tth37:
### Related Issue and PR:

#12811 #12562

### Problem

The current backend logic for generating citation indexes within `<source>` tags in the context string, leads to inconsistencies with how the frontend groups and display citations:

1. **Web Search Results:** Web search results lack a `file_id` in their metadata. The original code used `file_id` as the key for indexing, causing all web results (where `file_id` is `None`) to map to the same index, even though they represent distinct sources. This directly causes issue #12811
2. **Frontend Grouping Mismatch:** Sources that the frontend treats as a single citation group (e.g., potentially multiple uploads of files with the same name, or multiple chunks from the same logical source) could be assigned different sequential indices by the backend based on their distinct `file_id`. This results in generated text like `[1][2][3]` where the frontend might only display clickable citations for `[1][2]`, leaving subsequent numbers as plain text because they refer to indices not uniquely represented in the frontend's grouped view.

### Root Cause

The backend used an internal identifier (`file_id`) to create unique sequential indices for *each document chunk*. However, the frontend aggregates citation sources based on a different logic: `metadata?.source ?? source?.source?.id ?? 'N/A'`. 

https://github.com/open-webui/open-webui/blob/c5636ff68c4e9ca095cd6458cc740f4a3aa53831/src/lib/components/chat/Messages/Citations.svelte#L47-L57

This mismatch causes the mapping between the backend-generated `<source id="...">` tags and the frontend's citation display to break.

### Solution

This PR modifies the backend context generation logic to use the **exact same identifier logic as the frontend** for creating the citation indices within the `<source>` tags.

Specifically, the identifier used to group sources and assign the `id` attribute in `<source id="{index}">` is now calculated using:
`doc_meta.get("source", None) or source.get("source", {}).get("id", None) or "N/A"`

This ensures that:
- Each unique logical citation source (as identified by the frontend logic) gets a single, consistent index number.
- Documents belonging to the same logical citation group (e.g., multiple web results grouped under one citation, or multiple chunks from the same file) will now correctly share the same `<source id="...">` index in the backend-generated context.
- The indices (`[1]`, `[2]`, etc.) referenced in the model's generation will directly correspond to the citation blocks rendered by the frontend.

This alignment will hopefully ensure citation references function correctly.

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -888,16 +888,20 @@ async def process_chat_payload(request, form_data, user, metadata, model):
     # If context is not empty, insert it into the messages
     if len(sources) > 0:
         context_string = ""
-        citated_file_idx = {}
-        for _, source in enumerate(sources, 1):
+        citation_idx = {}
+        for source in sources:
             if "document" in source:
                 for doc_context, doc_meta in zip(
                     source["document"], source["metadata"]
                 ):
-                    file_id = doc_meta.get("file_id")
-                    if file_id not in citated_file_idx:
-                        citated_file_idx[file_id] = len(citated_file_idx) + 1
-                    context_string += f'<source id="{citated_file_idx[file_id]}">{doc_context}</source>\n'
+                    citation_id = (
+                        doc_meta.get("source", None)
+                        or source.get("source", {}).get("id", None)
+                        or "N/A"
+                    )
+                    if citation_id not in citation_idx:
+                        citation_idx[citation_id] = len(citation_idx) + 1
+                    context_string += f'<source id="{citation_idx[citation_id]}">{doc_context}</source>\n'
 
         context_string = context_string.strip()
         prompt = get_last_user_message(form_data["messages"])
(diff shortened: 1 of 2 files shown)


tth37:
### Integration Tests

**Test on web search results (results are mocked data)**

Before:
image


After:
image

**Test on same uploaded document name**

image

Before:
image

After:
image

almajo:
Very cool! thanks for the fix and also testing with uploaded docs. Can you also test with a model using knowledge bases? Just want to make sure we don't play another round of knowledge vs web search. :D

tth37:
> Very cool! thanks for the fix and also testing with uploaded docs. Can you also test with a model using knowledge bases? Just want to make sure we don't play another round of knowledge vs web search. :D

Of course! Could you please help me design a test case that can best verify its correct functionality? (like creating knowledge base `foo`, uploading `foo.md`; and then creating knowledge base `bar`, uploading `bar.md`)?

tjbck:
Great one, LGTM!

PR #12958, opened 2025-04-17
fix: use run_in_threadpool for search_web to prevent blocking

Youggls:
Used fastapi's run_in_threadpool function to execute the search_web function, preventing the synchronous function from blocking the entire web search process.

# Pull Request Checklist

### Note to first-time contributors: Please open a discussion post in [Discussions](https://github.com/open-webui/open-webui/discussions) and describe your changes before submitting a pull request.

**Before submitting, make sure you've checked the following:**

- [x] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [x] **Description:** Provide a concise description of the changes made in this pull request.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ ] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [ ] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [x] **Testing:** Have you written and run sufficient tests to validate the changes?
- [x] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [x] **Prefix:** To clearly categorize this pull request, prefix the pull request title using one of the following:
  - **BREAKING CHANGE**: Significant changes that may affect compatibility
  - **build**: Changes that affect the build system or external dependencies
  - **ci**: Changes to our continuous integration processes or workflows
  - **chore**: Refactor, cleanup, or other non-functional code changes
  - **docs**: Documentation update or addition
  - **feat**: Introduces a new feature or enhancement to the codebase
  - **fix**: Bug fix or error correction
  - **i18n**: Internationalization or localization changes
  - **perf**: Performance improvement
  - **refactor**: Code restructuring for better maintainability, readability, or scalability
  - **style**: Changes that do not affect the meaning of the code (white space, formatting, missing semi-colons, etc.)
  - **test**: Adding missing tests or correcting existing tests
  - **WIP**: Work in progress, a temporary label for incomplete or ongoing work

# Changelog Entry

### Description
- Enhanced web search functionality by implementing thread-based execution, preventing synchronous operations from blocking the application flow and improving overall responsiveness.

### Added
- None

### Changed
- Modified web search implementation to use FastAPI's `run_in_threadpool` function for non-blocking execution

### Deprecated
- None

### Removed
- None

### Fixed
[...]

PR #13031, opened 2025-04-18
feat: Add support for configuring FastAPI/AnyIO Thread Pool

gaby:
### Description

- FastAPI uses AnyIO which by default has a limit of `40` threads for the whole application. These threads are used by any non-async or threaded function. This PR adds support for configuring this value via `ENV`, ~it also bumps the default value to 64 threads.~
- This should help the server from getting blocked when handling web-searches or any other `non-async` calls.

Implementation Source: https://github.com/Kludex/fastapi-tips#2-be-careful-with-non-async-functions
AnyIO Docs: https://www.starlette.io/threadpool/#concurrency-limitations

Related: https://github.com/open-webui/open-webui/pull/12958

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -1204,7 +1204,7 @@ ENABLE_USER_WEBHOOKS = PersistentConfig(
 )
 
 # FastAPI / AnyIO settings
-THREAD_POOL_SIZE = int(os.getenv("THREAD_POOL_SIZE", "64"))
+THREAD_POOL_SIZE = int(os.getenv("THREAD_POOL_SIZE", "0"))
 
 
 def validate_cors_origins(origins):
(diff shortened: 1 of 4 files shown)


tth37:
For the ThreadPoolExecutor, maybe we should set an upper bound for `max_workers` to prevent system thrashing? (such as [number of CPUs the current process can use](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.cpu_count))

gaby:
@tth37 This is for AnyIO internal Thread Pool which is used by FastAPI for non-async calls. It defaults to 40 threads.

Classic298:
Would it be sensible to add docs for this env var? Or would we expect that 64 is a good default value for this? @gaby

gaby:
@Classic298 Will be added once merged. Those are in a different repo. I can change it back to 40, but 64 is a multiple of 2/4/8.

@tth37 Not really, just that its a multiple of 8. I can change it back to 40.

gaby:
@Classic298 @tth37 Changed default to "0", so current behavior isn't changed.

tjbck:
Thanks!

Issue #13027, opened 2025-04-18
feat: Asynchronous `process_chat_payload` in chat completion


tth37:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.

### Problem Description

The `/api/chat/completion` endpoint is designed to support streaming output and return immediately with a `task_id`. This allows the frontend to use the existing WebSocket connection to receive status updates and the final response, and crucially, enables early stopping of the generation process via `task_id`.

However, when features like **Web Search** or **Tools** are enabled, the endpoint currently **does not return immediately**. Instead, it waits for the `process_chat_payload` phase (which includes potentially long-running operations like web searches or tool executions) to complete before returning the `task_id`.

This synchronous behavior during the payload processing phase leads to two significant issues:

1. **Delayed Early Stopping**: The frontend does not receive the `task_id` until after web search/tool execution finishes. This prevents users from stopping the request during this initial, potentially lengthy (30-60s+), phase.
2. **Network Timeouts**: The extended wait time for the endpoint to respond increases the risk of network errors, such as gateway timeouts or client-side request timeouts, degrading the user experience.

### Cause Analysis

The chat completion process can be broadly divided into two phases:

1. `process_chat_payload`: Handles request preprocessing, including web searches, tool calls, and injecting results into the context for the language model.
2. `process_chat_response`: Handles the actual generation of the AI response by LLM and streams results back via WebSocket.

Currently, `process_chat_response` is correctly handled asynchronously using `create_task`, as seen:

https://github.com/open-webui/open-webui/blob/b8fb4e528dc2629acf68b9a555a59fd0173aaa51/backend/open_webui/utils/middleware.py#L1209-L1210

However the `process_chat_payload` remains a **synchronous** function, user have to wait until `process_chat_payload` finishes and then they can receive the background `task_id`. Things get worse when web search feature is enabled as it might take up to 30s-60s, in this period user cannot early stop the request, and facing the risk of connection timeout.

### Desired Solution you'd like

Refactor the `chat_completion` handler in `main.py` to make the entire processing pipeline (both payload processing and response generation) **asynchronous** from the start. This can be achieved by wrapping all time-consuming logic within a single background task created immediately upon receiving the request.

```python3
async def all_time_consuming_jobs(request, form_data, user, metadata, model):
    form_data, metadata, events = await process_chat_payload(
        request, form_data, user, metadata, model
    )
    response = await chat_completion_handler(request, form_data, user)
[...]

PR #13061, opened 2025-04-19
perf: Asynchronous `process_chat_payload` in chat completion

tth37:
### Problem Description

The complete problem description is detailed in #13027.

For short, when using the `/api/chat/completions` endpoint asynchronously (typically from browser Web UI), the server should return a `task_id` immediately so the UI can track process and allow early cancellation.

However, if features like Web Search or Tools are enabled, the server waits for this preprocessing (`process_chat_payload`) to finish **before** returning `task_id`, leading two main issues:

1. **Delayed Cancellation**: Users cannot stop the request during the long preprocessing phase because UI doesn't have the `task_id` yet.
2. **Network Timeouts**: Long wait of this endpoint increase the chance of requests failing due to network timeouts, _commonly_ when using reverse proxies or intranet penetrations.

Image

### Solution

This PR changes the behavior for asynchronous requests:

1. Check if the request is async (`event_emitter` and `event_caller` exists) at the very beginning of the chat completion handler.
2. If it *is* an async request:
    - Immediately create a background task that handle *both* `process_chat_payload` and `process_chat_response`
    - Return the `task_id` to the client right away.
3. If it *is not* a streaming request:
    - Keep the original behavior: process everything synchronously and return the StreamingResponse

image

### Error handling

I extended the previously unhandled `task-cancelled` event. This event can be triggered by either `asyncio.CancelledError`, or when the `process_chat_payload_and_response` is running as a background job.

Formerly, when errors are encountered during `process_chat_payload` or `chat_completion_handler`, the `/api/chat/completions` endpoint would directly return the error message. Now that all the jobs are processed as a backend task, we must use WebSocket events to tell the browser about error details.

When the browser receives the `task-cancelled` event, it will mark `currentMessage.done = true` and re-fetch current taskIds, to ensure that the messages and the stop button behaves properly. This ensures the exact same user experience as the previous error handling logic, and robust error handling when chatting with multiple models.

### Additional Messages

I've tried my best to keep the original structure of codebase, make minimal changes, and ensure backward-compatibility.

### Tests

For sync requests, I used `curl` to test the chat completion endpoints, and they are behaving the same as original version:

```bash
curl -X POST "http://localhost:8080/api/chat/completions" \
    -H "Content-Type: application/json" \
    -H "Authorization: Bearer <my_token>" \
    -d '{
             "model": "deepseek-v3",
             "messages": [{
                   "role": "user",
                   "content": "hi"
              }],
              "stream": true
        }'
```

```bash
curl -X POST "http://localhost:8080/api/chat/completions" \
    -H "Content-Type: application/json" \
    -H "Authorization: Bearer <my_token>" \
    -d '{
             "model": "deepseek-v3",
             "messages": [{
                   "role": "user",
                   "content": "hi"
              }],
              "stream": false
        }'
```

For async requests, I tested the endpoint via browser Web UI. The endpoint returns with `task_id` very quickly and *before* the web search process ended. Upon receiving `task_id`, I am able to cancel the generation task at any time.

Error handling:
- [x] Error in `process_chat_payload` phase
- [x] Error in `process_chat_response` phase
- [x] `task-cancelled` error when user early stop a request
- [x] Stop button behavior when chatting with multiple models

image

More test cases / discussions are welcomed!

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -52,6 +52,7 @@ from open_webui.utils.logger import start_logger
 from open_webui.socket.main import (
     app as socket_app,
     periodic_usage_pool_cleanup,
+    get_event_emitter,
 )
 from open_webui.routers import (
     audio,
@@ -1166,7 +1167,7 @@ async def chat_completion(
         )
 
     async def process_chat_payload_and_response(
-        request, form_data, user, metadata, model
+        request, form_data, user, metadata, model, event_emitter=None
     ):
         try:
             form_data, metadata, events = await process_chat_payload(
@@ -1183,6 +1184,15 @@ async def chat_completion(
                         "error": {"content": str(e)},
                     },
                 )
+            if event_emitter is not None:
+                await event_emitter(
+                    {
+                        "type": "task-cancelled",
+                        "data": {
+                            "error": str(e),
+                        },
+                    }
+                )
             raise HTTPException(
                 status_code=status.HTTP_400_BAD_REQUEST,
                 detail=str(e),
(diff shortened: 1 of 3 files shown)


tjbck:
Appreciate your effort, I'll personally refactor this part.

Issue #12832, opened 2025-04-14
issue: Web Search Feature Fails When Encountering Empty Content


manwallet:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

v0.6.4

### Ollama Version (if applicable)

Not applicable

### Operating System

macos sequoia 

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have listed steps to reproduce the bug in detail.

### Expected Behavior

When encountering empty content during a web search:
1. The system should log the issue but skip the problematic resource
2. The search should continue processing other valid resources
3. The user should receive partial results from the successfully processed pages
4. At minimum, provide a more helpful error message that doesn't terminate the entire request

### Actual Behavior

When performing a web search and one of the retrieved pages returns empty content:
1. The system raises a `ValueError: The content provided is empty. Please ensure that there is text or data present before proceeding.` exception
2. The entire web search functionality fails with HTTP 400 error
3. The user receives no results, even from successfully retrieved and processed pages
4. The API request fails completely instead of returning partial results

### Steps to Reproduce

1. Send a chat message that triggers web search (with a query like "托福词汇记忆方法")
2. Observe that the search attempts to fetch multiple resources
3. When one of the resources returns empty content, the entire search fails
4. No search results are returned despite some resources being successfully retrieved

### Logs & Screenshots

2025-04-14 04:33:24.278 | INFO | open_webui.routers.retrieval:save_docs_to_vector_db:821 - save_docs_to_vector_db: document https://m.ease100.com/toefl/vocabulary/631.html web-search-260deb2eb1113a0545a9112e9d642238a8fb233195726e7d6624 - {}
2025-04-14 04:33:24.278 | ERROR | open_webui.routers.retrieval:process_web_search:1545 - The content provided is empty. Please ensure that there is text or data present before proceeding. - {}
Traceback (most recent call last):
...
ValueError: The content provided is empty. Please ensure that there is text or data present before proceeding.
...
File "/app/backend/open_webui/routers/retrieval.py", line 1546, in process_web_search
raise HTTPException(
      └ <class 'fastapi.exceptions.HTTPException'>
[...]

PR #13079, opened 2025-04-20
fix: Web RAG empty content

tth37:
### Related Issue and Commit
#12832 70718dda90af07370414dea28f9a93058623a33b

### Problem Description
After web pages are loaded, the backend proceeds to `save_doc_to_vector_db`. However, if one of the documents has an empty `page_content`, the `save_docs_to_vector_db` operation fails, subsequently causing the entire web search to fail. Commit 70718dda90af07370414dea28f9a93058623a33b partially addresses this issue by adding a check for empty `doc.page_content`.

However, there are cases where `doc.page_content` contains **only whitespace**, which also causes the `save_docs_to_vector_db` process to fail. This happens because `save_docs_to_vector_db` performs preprocessing on the `page_content`, including cleaning and splitting, which cannot handle whitespace-only content.

https://github.com/open-webui/open-webui/blob/70718dda90af07370414dea28f9a93058623a33b/backend/open_webui/routers/retrieval.py#L862-L863

### Proposed Solution
To ensure the web search process remains robust, we can wrap the `save_docs_to_vector_db` process in a try-catch block. This will handle potential errors gracefully, preventing them from failing the entire web search.

### DIFF
--- a/backend/open_webui/routers/retrieval.py
+++ b/backend/open_webui/routers/retrieval.py
@@ -1557,19 +1557,24 @@ async def process_web_search(
             collection_names = []
             for doc_idx, doc in enumerate(docs):
                 if doc and doc.page_content:
-                    collection_name = f"web-search-{calculate_sha256_string(form_data.query + '-' + urls[doc_idx])}"[
-                        :63
-                    ]
+                    try:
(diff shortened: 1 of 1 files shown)


tth37:
Resolved in 43efff0fe62a0ac8e41342a9158dcc387a1ec6d5, I'll close this PR :-)

PR #13107, opened 2025-04-21
fix: `max_results` in Tavily search handler

tth37:
# Changelog Entry

### Description

Related: #13075 
Fixed some small bugs when calling Tavily search API.

1. API Key Handling
    - Previously, the API key was passed in the request body, which is not documented in Tavily's [official API reference](https://docs.tavily.com/documentation/api-reference/endpoint/search).
    - Updated to pass the `api_key` in the `Authorization` header, following the recommended practice.
2. `max_results` Parameter
    - The Tavily API natively supports `max_results` (default: 5) parameter, facing issues when `count` is greater than 5.
    - Pass the `count` parameter as `max_results` to get the correct number of search results.
3. Domain Filter Fix
    - The Tavily search handler previously ignored the domain filter list.

---

### DIFF
--- a/backend/open_webui/retrieval/web/tavily.py
+++ b/backend/open_webui/retrieval/web/tavily.py
@@ -2,7 +2,7 @@ import logging
 from typing import Optional
 
 import requests
-from open_webui.retrieval.web.main import SearchResult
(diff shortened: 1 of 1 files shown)


tjbck:
Thanks!

PR #13045, opened 2025-04-18
**perf** | Multi-thread web searching

TomBayne:
# Pull Request Checklist

**Before submitting, make sure you've checked the following:**

- [x] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [x] **Description:** Provide a concise description of the changes made in this pull request.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [x] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [x] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [x] **Testing:** Have you written and run sufficient tests to validate the changes?
- [x] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [x] **Prefix:** To clearly categorize this pull request, prefix the pull request title using one of the following:
  - **BREAKING CHANGE**: Significant changes that may affect compatibility
  - **build**: Changes that affect the build system or external dependencies
  - **ci**: Changes to our continuous integration processes or workflows
  - **chore**: Refactor, cleanup, or other non-functional code changes
  - **docs**: Documentation update or addition
  - **feat**: Introduces a new feature or enhancement to the codebase
  - **fix**: Bug fix or error correction
  - **i18n**: Internationalization or localization changes
  - **perf**: Performance improvement
  - **refactor**: Code restructuring for better maintainability, readability, or scalability
  - **style**: Changes that do not affect the meaning of the code (white space, formatting, missing semi-colons, etc.)
  - **test**: Adding missing tests or correcting existing tests
  - **WIP**: Work in progress, a temporary label for incomplete or ongoing work

# Changelog Entry

### Description

- Multi-threaded web search to improve performance. Probably could squeeze more performance out of this.
- During my tests, significantly reduced search time on a Raspberry Pi 5 server.

### Changed

- Web search runs across multiple threads.

---

### Additional Information

Settings: 10 Web Searches, 10 Concurrent
Test Query: 'Who is the current US president?'
Test A (Multi-threaded) : 1m 22s
Test B (Old Code) : 4m 23s
Test C (Multi-threaded) : 1m 40s
Test D (Old Code): 2m 23s

PR #13111, opened 2025-04-21
perf: Concurrent processing for web search queries

tth37:
### Problem Description

This PR addresses the same issue as #13045 -- improving the performance of multiple web search queries -- but with a simpler and cleaner approach.

Currently, the `async process_web_search` handler processes queries **sequentially**, performing searches and loading pages one at a time. This creates a bottleneck when multiple queries are generated, significantly slowing down overall performance.

### Solution

Since `process_web_search` is an **I/O-intensive** operation, we can easily optimize it by parallelizing the execution using `asyncio` tasks and `asyncio.gather`. This allows multiple search queries to run concurrently, effectively reducing latency.

### Breaking Change

Previously, the backend emitted individual events (e.g., `"Searching {{searchQuery}}"`) for each query. Due to parallel processing, this is no longer feasible. Instead, we now emit a single event (`"Searching the web"`), aligning with common practices in other LLM web UIs.

### Testing

| Test Case | Sequential | Parallel | Speedup |
|-----------------------|------------|----------|---------|
| “what is open webui”, “how to install open webui”, “how to install open webui with docker”(max_results=5) | 20.01 (5.40+9.25+5.36) | 7.99 | 2.5x |
| “what is lorem”, “example website”, “vue vs svelte”(max_results=5) | 20.96 (7.80+5.34+7.81) | 7.58 | 2.8x |
| "svelte", "vue", "react", "angular", "nextjs", "nuxtjs", "jquery", "qwik", "astro"(max_results=1) | 18.36 | 3.11 | 5.9x |
| “melbourne”, “sydney”, “canberra”(max_results=5) | 35.01 (12.26+11.71+11.03) | 13.96 | 2.5x |

In typical cases, parallel execution introduces minimal overhead, with total latency approaching the lower-bound duration of the slowest single query.

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -366,21 +366,17 @@ async def chat_web_search_handler(
         }
     )
 
-    web_search_tasks = [
-        asyncio.create_task(
+    gathered_results = await asyncio.gather(
+        *(
             process_web_search(
                 request,
-                SearchForm(
-                    **{
-                        "query": searchQuery,
-                    }
-                ),
+                SearchForm(**{"query": searchQuery}),
                 user=user,
             )
-        )
-        for searchQuery in queries
-    ]
-    gathered_results = await asyncio.gather(*web_search_tasks, return_exceptions=True)
+            for searchQuery in queries
+        ),
+        return_exceptions=True,
+    )
 
     for searchQuery, results in zip(queries, gathered_results):
         try:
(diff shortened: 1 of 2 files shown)


tth37:
When comparing to #13045, this PR specifically focuses on parallelizing **multiple search queries** (i.e., running `process_web_search` concurrently). While simple, this solution demonstrates its effectiveness in accelerating web searching. (As the performance bottleneck primarily occurs at the external search/load API calls rather than in our backend processing)

tjbck:
Looks reasonable, Thanks!

PR #13176, opened 2025-04-23
fix: FireCrawlLoader default mode to scrape

tth37:
Hi, after 09874ab83dcf7c39babde9e5142b4caf9b2c9193, the FireCrawlLoader's source URLs are now correctly handled!

However, there’s still a small point to address. I strongly recommend changing FireCrawlLoader's default mode to `'scrape'`. The `'scrape'` mode focuses on fetching a single URL, which is consistent with the behavior of other web loaders. In contrast, selecting `'crawl'` mode—particularly when using a **self-hosted** Firecrawl instance—can result in Firecrawl recursively fetching all linked pages from the given URL, leading to unacceptably long processing times for just one webpage.


tjbck:
PR template please!

PR #12571, opened 2025-04-07
feat: reindex knowledge files

hurxxxx:
# Pull Request Checklist

### Note to first-time contributors: Please open a discussion post in [Discussions](https://github.com/open-webui/open-webui/discussions) and describe your changes before submitting a pull request.

**Before submitting, make sure you've checked the following:**

- [x] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [x] **Description:** Provide a concise description of the changes made in this pull request.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ ] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [ ] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [x] **Testing:** Have you written and run sufficient tests to validate the changes?
- [x] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [x] **Prefix:** To clearly categorize this pull request, prefix the pull request title using one of the following:
  - **BREAKING CHANGE**: Significant changes that may affect compatibility
  - **build**: Changes that affect the build system or external dependencies
  - **ci**: Changes to our continuous integration processes or workflows
  - **chore**: Refactor, cleanup, or other non-functional code changes
  - **docs**: Documentation update or addition
  - **feat**: Introduces a new feature or enhancement to the codebase
  - **fix**: Bug fix or error correction
  - **i18n**: Internationalization or localization changes
  - **perf**: Performance improvement
  - **refactor**: Code restructuring for better maintainability, readability, or scalability
  - **style**: Changes that do not affect the meaning of the code (white space, formatting, missing semi-colons, etc.)
  - **test**: Adding missing tests or correcting existing tests
  - **WIP**: Work in progress, a temporary label for incomplete or ongoing work

# Changelog Entry

### Description
- There is an inconvenience of having to reupload all files when the embedding model is changed.
- Therefore, a feature has been added in the Document settings of the admin panel to reindex all files in the entire Knowledge base.

### Screenshots or Videos
- Video capture is linked due to attachment size issues.
[https://1drv.ms/v/c/3dbe8525a1b21172/IQR4qX7mvNnQSKdXoRfoWzEqAXLG76icmrU75LJ9ujYmkAo](url)

PR #13225, opened 2025-04-25
feat: reindex all memories, files and knowledges

mahenning:
# Pull Request Checklist

# Changelog Entry

### Description

- This PR adds a button into the Documents tab in admin view to re-index all memories, single files (chats, files in knowledges..) and knowledge bases, and indicates its progress

### Added

Frontend:
- Adds a button to the "Admin->Documents" tab to re-index all memories, files, knowledges that are currently in the database
- Adds the information on how many files/knowledges have to be re-indexed to the confirmation button, also a stop button
- Adds a progress bar to indicate how long the user should wait for it to finish
Backend:
- Adds /reindex endpoints in utils, memory, files and knowledge. The endpoint in /utils starts the reindexing of all 3 other parts and is the preferred starting point
- Adds a /reindex/stop endpoin to halt reindexing manually
- Adds a "/count" endpoint in memories, files, knowledges to get the total number of items without straining the db to retrieve the actual item contents -> fast & memory efficent
- Adds paginated db access for memories, files, knowledges -> prevents that e.g. all files are read simultaneously into RAM
- Adding redis support and non-blocking threading to support multi-instance deployments (**currently not tested -> Feedback is highly appreciated!**)
---

### Additional Information

- This PR builds on https://github.com/open-webui/open-webui/pull/12571, which added a button to re-index all knowledges. But this re-indexes only the "knowledge-as-a-whole", not the single files in the kowledge (or any other files). Here, the functionality is extended to single files, which may be in the chat but also in knowledges, and also to memories.

### Screenshots or Videos

The confirmation window shows how many files and knowledges will be worked on to get an estimation of the time it will take. It will also enable you to set the number of documents that are loaded into RAM at the same time, and if documents shouldbe processed completely from scratch (takes longer)
Screenshot 2025-09-05 142054

A loading bar indicates the progress of the re-indexing
Screenshot 2025-09-08 144039
Screenshot 2025-09-08 144119


mahenning:
I seek feedback for the following:

- "Reindex all Files Vectors" is not a very good description I think, it is very technical. Anyone has another wording? Maybe "Reindexing all single files"? The important part is that it also re-indexes files from knowledges, so if you include a file from a knowledge with "#" in the chat, the "single" file is used there.
- For full re-indexing, memory indexing is missing. The implementation is easy, but where to put the button in the UI? In the documents tab, where the embedding model is set and the other buttons for re-indexing are, or in another part of the UI?

mahenning:
Yes. Also when the chunk size/overlap is updated or the text splitter is changed.

jayteaftw:
@mahenning Hey nice work!
Quick question:could this be used for transitioning to a different vector database without losing data?
For example, you start off using Chroma and realize that it is too small to handle the user base so you switch to qdrant. 
Once they set env qdrant to be the DB, you could simply press the reindex all button in the Admin page and it would create the collections for you. 

I think it would make sense to do memory within this PR asweel.
However it also looks like there are some other things being indexed too scuh as web_searches and hashes (?)

```python
# backend/open_webui/retrieval/vector/dbs/qdrant_multitenancy.py
# lines 61-65 
self.MEMORY_COLLECTION = f"{self.collection_prefix}_memories"
self.KNOWLEDGE_COLLECTION = f"{self.collection_prefix}_knowledge"
self.FILE_COLLECTION = f"{self.collection_prefix}_files"
self.WEB_SEARCH_COLLECTION = f"{self.collection_prefix}_web-search"
self.HASH_BASED_COLLECTION = f"{self.collection_prefix}_hash-based"
```

Or are the web search and hash based collections just faster lookup and processing time? And if they were deleted, it wouldnt crash open webui and they would simply be regenerated upon request.

mahenning:
I did not test database migration, but in theory this should work, yes. It just iterates over every file, deletes it from vector storage (if it's there) and then re-processes it completely. If there is a new vector store, it just re-processes the files into it. I'd be great to get feedback on this, if possible.

I agree on the memory re-indexing but I got no real feedback on this (or if this PR gets merged at all), so it's on hold right now. Personally, I would not add a third button to re-index memory separately, but merge knowledge/files/memory re-indexing into a single button. While this means potentially heavy lifting on the server side, I don't see a use-case where you only want to re-index just a part of your vector db. Either all or nothing. Everything else is puzzling for the user.

I looked up where the get/insert/upsert methods of the vector db are used and it seems not in web search right now, so I think the lines you posted are just a precaution to future uses.

jayteaftw:
> I did not test database migration, but in theory this should work, yes. It just iterates over every file, deletes it from vector storage (if it's there) and then re-processes it completely. If there is a new vector store, it just re-processes the files into it. I'd be great to get feedback on this, if possible.
> 
> I agree on the memory re-indexing but I got no real feedback on this (or if this PR gets merged at all), so it's on hold right now. Personally, I would not add a third button to re-index memory separately, but merge knowledge/files/memory re-indexing into a single button. While this means potentially heavy lifting on the server side, I don't see a use-case where you only want to re-index just a part of your vector db. Either all or nothing. Everything else is puzzling for the user.
> 
> I looked up where the get/insert/upsert methods of the vector db are used and it seems not in web search right now, so I think the lines you posted are just a precaution to future uses.

Unless I am missing something, I totally agree! Especially since the embedding engine is global to every category that is we don't have a separate embedding engine for knowledge/files/memory then it doesn't make sense to just reindex knowledge and files when memory will have the previous embedding engine even if you reindex, effectively defeating the point of reindexing. 

Honestly I would add the changes as sometimes it can take a while to get PRs incorporated as there are a good of amount of PRs to be reviewed.

backend/open_webui/routers/files.py
```diff
@@ -221,6 +223,75 @@ async def search_files(
...
+            log.error(
+                    f"Error processing file {file.filename} (ID: {file.id}): {str(e)}"
+                )
+    log.info("Reindexing files completed sucessfully.")
+    return True
+
+
+# This function is just used to get a progress bar about the re-indexing to the UI
+@router.get("/reindex/stream")
+async def stream_progress():
```
tjbck: This won't work in multi-replica setup, should be marked as experiemental

backend/open_webui/routers/files.py
```diff
@@ -221,6 +223,75 @@ async def search_files(
...
+
+    log.info(f"Starting reindexing for {len(files)} files")
+    for i, file in enumerate(files, start=1):
+        try:
+            if VECTOR_DB_CLIENT.has_collection(collection_name=f"file-{file.id}"):
+                VECTOR_DB_CLIENT.delete_collection(collection_name=f"file-{file.id}")
+        except Exception as e:
+            log.error(f"Error deleting file 'file-{file.id}' from vector store: {str(e)}")
+        try:
+            if file.meta['content_type'] in [
```
tjbck: We might want to use a helper function here to ensure this matches with file upload endpoint

mahenning: Yeah this would be a good idea. I could write one, but I'm not sure where. Is there any place for helper functions?

backend/open_webui/routers/files.py
mahenning: Is it because of the global variable that is used to store the status of the indexing? Not sure how to solve this, I'll think about it. 
I could also include a toggle/button to turn off the progress bar. The default would then be "off" for compatibility with kubernetes, but I suppose some users wouldn't notice the progress bar option then. Is there a way to automatically detect multi-replica settings on start, and then set "off" as default, otherwise "on"?

mahenning: I did dig a bit through the code and saw that `AppConfig` uses redis (if configured). But it also enforces the values to be saved to be `PersistentConfig`, which is too much for a temporal runtime value I think. It would be nice if I can use redis for this. Is that an option or do you have another idea? I think that UI feedback on this functionality would be important, as it could run for a long time.

guenhter: I guess Redis would be nice, but it not running everywhere...

It is not the prettiest thing, but what about a file `/var/run/lock/foo/owui-file-reindex.lock` where the progress is written to? The reindex endpoint is writing it, and this `stream_progress` function is only reading.

Furthermore, it should then not be possible to trigger a second reindex-operation, if there is already a reindex task running. 

--

It still would not work when the instance is spun up on several Servers.

What do you think?

backend/open_webui/routers/files.py
guenhter: Maybe the whole block from line 248-269 can be reused from `files.py>upload_file`. I guess you anyway have this code from there. Maybe in the files.py, the interesting lines can be put to an own function (e.g. `process_file_by_content_type`) and you then just reuse that function.

backend/open_webui/routers/files.py
mahenning: Personally I only have experience using a single replica. I see 3 options:

A) Add a warning to the confirmation pop-up that this should only be used in single instances. Easiest to implement, but then it would be a feature that cant be used by everyone -> puzzling, when to use or not? What happens if a user in multi-replica setup presses "go" either way?

B) Use a local .lock file. I'm not sure if this is worth pursuing, this only works for very local multi-replica setups.

C) Use redis (somehow) or another method to enable this feature for multi-replica instances. I thought that redis is a dependency for multi-replica setups, because of [this blue Info box](https://docs.openwebui.com/getting-started/env-configuration/#redis) and also the [requirements for the helm charts](https://github.com/open-webui/helm-charts/tree/main/charts/open-webui#requirements). 

When going C), I can add the logic to use redis if configured, and fall back to the local logic that is implemented now, which works for single replicas.

guenhter: This sounds awesome.

mahenning:
- Added Redis support for reindexing status, fallback to local variable if redis is not configured
- Starting the reindexing and getting the loading bar blocks the reindexing button, so you can't start it twice
- Implemented that the loading bar re-appears if the page is re-loaded or one comes back to the page
- Added a "Counting files.. " dialogue to the start when one clicks the reindex button for responsiveness while actually counting files in the background, which can take a while

In my one-replica setup, this is working now, with full responsiveness of the UI while the reindexing is working. You can now even leave/reload the page and get back, and the loading bar is still there and updated.

!!! I could not test this with redis/multi-replica settings, because I dont have the setup for that. If someone has, and can test that the reindexing status is shared across instances, it would be great. I *think* that it should work in theory.

mahenning:
Potential TODOs (feeback wanted here):
- include memories into re-indexing
- include knowledge base (already implemented with another button) here
- create a helper funcion for the file type handling (where should it go? router/files.py?)

mahenning:
- The button now reindexes all memories, files and knowledges in this order. The confirmation dialogue shows the number of files and knowledges (screenshot below).
- The loading bar now stays after the progress is finished to better indicate that the progress is done. When reloading the page after reindexing, it will be hidden again.
- I've removed the other button to just reindex the knowledge bases, as it is included and would just be confusing to have an additional button for 1/3 functionality

One additional point: the files are completely re-read from the disk when reindexing, as I found no other way to retain the different pages for the storing as vectors. When reading from the `Files` table, the file content there is one long string, where the different pages are already lost. If one changes the way a file is read (e.g. a new content extraction engine) and then reindexes an old file, it will be extracted in the new way .

The helper function TODO is still open but maybe thats work for another PR.
In my opinion this PR is ready to merge. The multi-replica setting is still untested, but I cannot test it.

grafik

grafik

guenhter:
Awesome work. I can't wait to get this functionality.

mahenning:
Quick info: With chromadb, the size of the vector store almost doubles when reindexing. The problem is that chromadb does not delete the vector- and metadata when a collection is deleted (contrary to their [own documentation](https://docs.trychroma.com/docs/collections/delete-data)) 
> Chroma supports deleting items from a collection by id using .delete. The embeddings, documents, and metadata associated with each item will be deleted.

https://github.com/open-webui/open-webui/pull/16520 fixes the deletion of vector folders, but the internal chroma.sqlite3 database still holds all the metadata of the files, including full text blocks of the (deleted) documents. A workaround is to backup and delete the `vector_db` folder in `backend/data/` and start the re-indexing. This will build a fresh, clean vector store.

Classic298:
>  but the internal chroma.sqlite3 database still holds all the metadata of the files,

absolutely, i noticed that - despite running the VACUUM command on chroma db, which my PR does do.

I will try to identify a way to make it delete even the remaining data.

elabbarw:
Hey, 

As we discussed in another thread, you will **need** to add batching for KBs, files, and memories or else this will end up crashing OWUI when reindexing a large amount of KBs (https://github.com/mahenning/open-webui/blob/83a0639cb6d7796514a7d8b4ecfb0a815d21cccf/backend/open_webui/routers/knowledge.py#L182).  Also make sure that none of these jobs block the app (test calling /health when running the job - k8s liveness probes use this url and it would be assumed that the pod is unhealthy if not responsive). This should make it scalable and enterprise friendly 😄 

Apart from that, great job! 🚀

mahenning:
@elabbarw I've updated the code with your snippets (thanks again for them, it made it so much easier for me). If you have time to review, I'd be very glad!
I put the compute-heavy parts into async calls, and tested with both CPU/GPU the responsiveness of the app with a ping to /health every 1s.

Currently, the batch_size is hard-coded to a batch of 10. I did no yet test the difference in performance for e.g. 100/1000, but it should be more or less negligible as it only leads to more paginated db calls, which should be fast in comparison to the workload of embedding. But more user control would be nice, maybe with an env var?
Also, for knowledges it reads all files for the current knowledge. This _could_ lead to RAM problems on constrained systems with a very, very large knowledge base. To solve this, it could be also paginated/split:
- read out all file ids for current knowledge
- get `batch_size` files from the list of ids, process them, then get the next batch

Let me know if the env var or pagination for knowledge files should be implemented as well.

mahenning:
- Added a new `/reindex` API endpoint in backend/router/utils, which unifies the starting of the other 3 endpoints (memory/files/knowledge) in the backend. This shifts the work of starting the memory/files/knowledge reindexing from the frontend to the backend. Advantages:
  - the frontend only has to call one endpoint, not three
  - users can start a reindexing process with an API call to the backend
- The REINDEX_STATE now returns "progress" (0-100) as well as "status" ("idle", "running", "done") for each of memory/files/knowledge for the streaming endpoint. This makes it a lot easier to see the current status of the reindexing process
example output of stream endpoint:
`data: {"memories": {"progress": 100, "status": "done"}, "files": {"progress": 82, "status": "running"}, "knowledge": {"progress": 0, "status": "idle"}}`
- The progress bar is now divided into 3 parts for the 3 different reindexing states, when a part is finished, that part of the bar becomes green to visulalize progress better (see updated screenshot in initial post)

elabbarw:
> @elabbarw I've updated the code with your snippets (thanks again for them, it made it so much easier for me). If you have time to review, I'd be very glad! I put the compute-heavy parts into async calls, and tested with both CPU/GPU the responsiveness of the app with a ping to /health every 1s.
> 
> Currently, the batch_size is hard-coded to a batch of 10. I did no yet test the difference in performance for e.g. 100/1000, but it should be more or less negligible as it only leads to more paginated db calls, which should be fast in comparison to the workload of embedding. But more user control would be nice, maybe with an env var? Also, for knowledges it reads all files for the current knowledge. This _could_ lead to RAM problems on constrained systems with a very, very large knowledge base. To solve this, it could be also paginated/split:
> 
> * read out all file ids for current knowledge
> * get `batch_size` files from the list of ids, process them, then get the next batch
> 
> Let me know if the env var or pagination for knowledge files should be implemented as well.

- I think it would good to provide the option in admin settings with the default set at 10. Env var is the easy option, hehe 😉 
- I totally agree with batching the files as well! Good call 😄

mahenning:
> * Added a new `/reindex` API endpoint in backend/router/utils, which unifies the starting of the other 3 endpoints (memory/files/knowledge) in the backend. This shifts the work of starting the memory/files/knowledge reindexing from the frontend to the backend. Advantages:
>       
>       * the frontend only has to call one endpoint, not three
>       * users can start a reindexing process with an API call to the backend
> 
>     * The REINDEX_STATE now returns "progress" (0-100) as well as "status" ("idle", "running", "done") for each of memory/files/knowledge for the streaming endpoint. This makes it a lot easier to see the current status of the reindexing process
>       example output of stream endpoint:
>       `data: {"memories": {"progress": 100, "status": "done"}, "files": {"progress": 82, "status": "running"}, "knowledge": {"progress": 0, "status": "idle"}}`
> 
>     * The progress bar is now divided into 3 parts for the 3 different reindexing states, when a part is finished, that part of the bar becomes green to visulalize progress better (see updated screenshot in initial post)

- Added a /utils/reindex/stop endpoint, to halt reindexing. Note that the db is then in a half-reindexed state, but maybe the option is good to controlled end a process that takes too long. You can always just restart the reindexing to fix the embeddings. 
- Added a "Stop" Button to the UI that utilized the endpoint
- Added batching option in UI (see screenshot in main post) and API endpoint
- Added the option to not read file content from disk but from the internal db in the UI (seescreenshot) and API endpoint. This speeds up the reindexing process by 30-50%, but it also then ignores page limits for splitting, as the file content is saved as one big string in the db. More info on this in the next comment
- Added tooltips in the UI for the options

mahenning:
To give a few insights into the implementation:
I have decided not to use environment variables for batch size and the "read from disk" feature. It will use default vars that are set in the frontend and backend separately. A bit ugly, but it is much work to let the frontend load that from the backend. The impact of a (reasonable) value of batch_size (3, 5, 20..) is negligible. I got the same speed for batch size 3 as with batch size 20. As this only controls how many files are read from the db, and db access is usually fast, 10 (defult) should be fine for 99% of cases. The files are processed one after the other either way.

As for the "Process files from disk" feature.. I'm not completly happy with the solution, but thats based on how OWUI handles file reads right now. I saw that reading the file from disk took ~25-40% of the total reindexing time per file. And with the same context extraction parameters, the extracted content is (almost) the same that is already saved in the db.
There are 2 points that could be improved:
- Currently, the file content saved in the db is one long concatenated string. The file content read from disk usually is a list of strings that are split by page, which means that a chunk of text cannot be over multiple pages. Reading the content from the db changes that, as there are no "pages" anymore. To change that, one have to change the way the content is saved in the db, which exceeds this PR
- When reading from disk, there are two steps: 
a) Actually reading the doc and extracting the content (I/O bound)
b) Embedding the content (usually CPU/GPU bound)
My idea was a queue, so that document A can be embedded while document B can already be read from disk at the same time. The problem here is that the `process_file` function does both steps, so I can not split them up into their own processes.

i-iooi-i:
That's fantastic! I eagerly anticipate the early availability of this functionality.

mahenning:
After some consideration, I'd say this PR is ready at least from its features. Feedback/questions/comments still welcome. Unfortunately, I still found no one to test this in multi replica environments..

blacktoby:
When will it be merged? Also eagerly awaiting this feature!

i-iooi-i:
https://github.com/open-webui/open-webui/pull/18565

Can the quantity of retrieved memories be controlled or adjusted? Integer (1-100), controls max memories retrieved

mahenning:
@i-iooi-i maybe I don't understand the question, but this PR is exclusively for re-indexing the vector data for memory (files, knowledges..). It does not touch how memories works for the user.

tomasloksa:
In the meantime, I created a simple script that is able to reindex all files (and folders) into the new vector DB.

Using the script i successfully migrated 2 instances of OWUI from ChromaDB to PGVector.
[OWUI File Reindex](https://github.com/tomasloksa/owui-file-reindex)

i-iooi-i:
@mahenning My apologies, it seems I was mistaken. The current functional experience with OpenWebUI is quite unsatisfactory. Despite storing numerous memories, it seems capable of recalling only a handful, or sometimes none at all. I hope the official team will soon consider bolstering this particular feature. In the interim, I've resorted to using a plugin as a replacement for the memory function; it performs admirably on the new OpenWebUI v0.6.41 and boasts excellent memory retrieval capabilities.

https://www.reddit.com/r/OpenWebUI/comments/1o24giv/comment/nqirhjk/
image

mahenning:
> In the meantime, I created a simple script that is able to reindex all files (and folders) into the new vector DB.
> 
> Using the script i successfully migrated 2 instances of OWUI from ChromaDB to PGVector. [OWUI File Reindex](https://github.com/tomasloksa/owui-file-reindex)

Thanks, I'll take a look at your implementation and if I can maybe transfer something of that in my PR! 
I alo have a script for cleaning up the files and vector DBs, but it would be nicer to not handle this with external, self written scripts, but with an "official", integrated UI solution.

mahenning:
@tomasloksa When I read this right, your script only indexes/embeds files which have no embeddings, right? My solution re-indexes everything. This can also be used for db migration/switching, but also changing an embedding model or e.g. the `chunks_size` parameter, and then needing new embeddings with this changes in the same database.

tomasloksa:
@mahenning yes, it's just a one-time script I used when migrating the vector DB I thought could be useful to someone, while waiting for this PR to be merged. Your solution would definitely be useful even in my case.

jayteaftw:
Any update on this?

guenhter:
@mahenning Would you mind reaching out to me on discord? My username is `guenther_44723`

stronk7:
It would so nice to get ALL contents (memories, knowledges, files, ...) reindexed each time ANY change in embeddings (model, vectors db, ... ) happens. Right now I'm blocked using an old embedding when, right now, there are greatly improved ones to be used. And, also, I want to move away from built-in chroma to a remote one... 

And this PR will solve, if I'm not wrong, all those situations.

Thanks for the effort and contributions, @mahenning, much appreciated. Hope this finally lands and we get all those (embedding-changes) use cases covered. They are real.

Ciao :-)

TomTheWise:
Does this PR also recreate the vector database (for example external pgvector) so that changing to another embedding model that has a different size of dimensions will work?

Classic298:
@TomTheWise

No or yes depending on your Vector Database.

The PR deletes each collection via VECTOR_DB_CLIENT.delete_collection() before re-embedding, so it depends on what your vector DB backend does with that call. If delete_collection fully drops the collection/table (like Milvus Multitenancy does), then yes, the new embeddings with different dimensions will work because the schema gets recreated from scratch on insert.

If your pgvector implementation only truncates rows but keeps the table structure (and its dimension constraint), then no, you'd still hit a dimension mismatch. You'd want to verify that your specific vector DB adapter actually drops the underlying table/index, not just the data.

nodox:
Have we stopped working on this?

I have 30gb  of files to reindex and I'm not sure how it'll respond or corrupt the database state

stevenmoto:
Trying to change the embedding model in my OWUI instance.

For existing files, the reindex button, causes an error:
chromadb.errors.InvalidArgumentError: Collection expecting embedding with dimension of 2560, got 2048

I believe this work would help to process them correctly?

madsgraaew:
> Trying to change the embedding model in my OWUI instance.
> 
> For existing files, the reindex button, causes an error: chromadb.errors.InvalidArgumentError: Collection expecting embedding with dimension of 2560, got 2048
> 
> I believe this work would help to process them correctly?

No, apparently they dont delete or overwrite the existing collection, so unless your new embedding model has the same dimensions, the reindex button does nothing.

PR #13253, opened 2025-04-27
fix: The value of the 'Access-Control-Allow-Origin' header in the res…

noexcs:
fix: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'.

# Changelog Entry

### Description

When the user does not configure anything, the default CORS configuration on the backend is "`*`", and the frontend's credentials mode is set to 'include'. 

This leads to a CORS issue: `The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'`. 

As a result, first-time users encounter difficulties in normal usage on their localhost.

To resolve this, I have added http://localhost:5173;http://localhost:8080 to the default `CORS_ALLOW_ORIGIN`.

### Fixed

- The `The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'` issue, when the user does not configure anything.

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -1255,7 +1255,7 @@ def validate_cors_origin(origin):
 # To test CORS_ALLOW_ORIGIN locally, you can set something like
 # CORS_ALLOW_ORIGIN=http://localhost:5173;http://localhost:8080
 # in your .env file depending on your frontend port, 5173 in this case.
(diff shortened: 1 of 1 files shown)


tjbck:
Agreed, Thanks!

Issue #12888, opened 2025-04-15
issue: requirements for tools and filters aren't installed after update


JTHesse:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.6.5

### Ollama Version (if applicable)

v0.6.5

### Operating System

Ubuntu 22.04

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have listed steps to reproduce the bug in detail.

### Expected Behavior

The Function or Tool should be active and usable after upgrading the docker image.

### Actual Behavior

The Function or Tool is deactivated and need to be saved again, then the requirements will be installed.

### Steps to Reproduce

1. Start OpenWebui container
2. Add Function or Tool with requirements in frontmatter
3. Update or recreate container

### Logs & Screenshots

```
Loading WEBUI_SECRET_KEY from file, not provided as an environment variable.
Generating WEBUI_SECRET_KEY
Loading WEBUI_SECRET_KEY from .webui_secret_key
/app/backend/open_webui
/app/backend
/app
INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
INFO  [open_webui.env] 'ENABLE_API_KEY' loaded from the latest database entry
INFO  [open_webui.env] 'ENABLE_API_KEY_ENDPOINT_RESTRICTIONS' loaded from the latest database entry
INFO  [open_webui.env] 'API_KEY_ALLOWED_ENDPOINTS' loaded from the latest database entry
INFO  [open_webui.env] 'JWT_EXPIRES_IN' loaded from the latest database entry
INFO  [open_webui.env] 'ENABLE_OLLAMA_API' loaded from the latest database entry
INFO  [open_webui.env] 'OLLAMA_BASE_URLS' loaded from the latest database entry
INFO  [open_webui.env] 'OLLAMA_API_CONFIGS' loaded from the latest database entry
INFO  [open_webui.env] 'ENABLE_OPENAI_API' loaded from the latest database entry
INFO  [open_webui.env] 'OPENAI_API_KEYS' loaded from the latest database entry
[...]

Issue #12314, opened 2025-04-01
issue: when pod with openwebui restarts, dependencies installed on runtime are lost


Lutherwaves:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Other

### Open WebUI Version

0.6.0

### Operating System

Ubuntu 22.04

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have listed steps to reproduce the bug in detail.

### Expected Behavior

When openwebui pod restarts, dependencies installed during runtime (when creating tools) are persisted via pvc and are available upon pod restart. These include

- previously setup embedding/re-ranking models when using non-transformers models
- previously setup tool additional dependencies are already present

### Actual Behavior

When a pod restarts (after upgrade and/or manual restart)

- playwright if enabled is scheduled to be installed and causes delays in restarting openwebuo
- tools that have additional dependencies do not work and need to be edited and saved
- if using re-ranking model with ollama/openai, it tries to re-install the model, but since it is not a known dependency available for transformers library, it fails and required embedding -> re-ranking to be resetup

### Steps to Reproduce

1. Install openwebui in kubernetes using helm
2. Install a tool with external library dependency https://openwebui.com/t/lutherwaves/todoist
3. Enable RAG hybrid search and select mx-embed-large as the model with ollama configured as the host
4. Delete open-webui-0 pod and watch the restart
5. Upon successful restart, test chatting with a model and a tool that has external dependencies, you get "library not found" since its' installation is lost
6. In the logs of the pod, you can see it fails to pull mx-embed-large into transformers; if you go to Admin Settings -> Documents -> Reset Hybrid search and Save, it will again point to ollama and be fixed

### Logs & Screenshots

Restarting pod and showing how playwright reinstall is scheduled (I had overwritten the open-webui name to prism)

`➜  ~ kubectl get pods
NAME                                     READY   STATUS    RESTARTS   AGE
open-webui-pipelines-869fd55cc7-rqlbs    1/1     Running   0          2d21h
prism-0                                  1/1     Running   0          15h
prism-ollama-684b78fb7f-cpqvx            1/1     Running   0          15h
prism-open-webui-tika-7995fb7dcb-v5tmj   1/1     Running   0          29h
[...]

Issue #12121, opened 2025-03-27
issue: Function deactivated after container restart due to missing extra dependencies


almajo:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.5.20

### Operating System

Linux

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have listed steps to reproduce the bug in detail.

### Expected Behavior

After I restart the docker container I want to have the same state as before. Every function that was activated before, should be activated again.

### Actual Behavior

When a function has extra dependencies, the function crashes on startup and thus is being deactivated by openwebui.

### Steps to Reproduce

1. Spin up a container
2. create a custom function with an extra dependency (add `requirements: ...` to the docstring on top)
3. Activate the function
4. restart the container
5. See that the function is deactivated and when you want to toggle it to activate, see the log of the container that it crashes because dependencies are missing

### Logs & Screenshots

...

### Additional Information

The problem can be manually resolved by opening the function and press "Save" - which triggers the install of relevant requirements. 
It would be great if we could trigger this process automatically on startup.

MuhammadMia:
+1 here, I lose dependencies. Planning to fix this with a tool server soon when its released but would be nice to have a temp solution until then

Lutherwaves:
+1, not only tool dependencies are lost, but also unless env variables are set to define RAG completely, if a re-ranking model setting exists, it causes conflicts as it tries to install it on transformers, which obviously fails. More details I posted in https://github.com/open-webui/open-webui/issues/12314

rndmcnlly:
+1 from me too.

[...]

PR #13358, opened 2025-04-30
fix: install external requirements of tools and functions to prevent deactivation

almajo:
# Pull Request Checklist

# Changelog Entry

### Description

Bug: #12121 

This PR installs requirements of **active** functions and **admin** tools on startup. This prevents automatically deactivating due to missing requirements them after the container has been restarted.

I'm happy to discuss the decisions I made here.

##### Blocking call

I decided to go for a blocking operation, i.e. the API is only functional after the requirements are installed or it failed with an error log. The reason is to prevent conurruency issues with users. When the get_models is called and the function is not working because of missing requirements it will get automatically deactivated. All while the requirements are actually installed in the background.

This should not be noticeable at all in practice, maybe add an extra startup time of 1-10 seconds to the applications (if many requirements are present.

##### Only active functions and admin tools

I decided to only go for active functions in order to be able to deactivate a (maybe not-working) function at any time and not have any impact of active functions. This reduces the risk of dependency problems between the tools and functions.

Only installing requirements of admin functions (`user.role == admin`), we prevent possible dependency resolution issues or dependency injections from user tools, directly on startup. 

### Changed

- On startup external dependencies are installed before the API accepts requests.

### Fixed

- #12121 
- #12314 
- #12888 

---

### Tests

Following the reproduction steps in #12121 I was able to solve this issue and active functions with requirements remain active and working while others remain untouched.

If anything fails in this method, it is only logged and there is no change to before this PR.

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -383,6 +383,7 @@ from open_webui.utils.auth import (
     get_admin_user,
     get_verified_user,
 )
+from open_webui.utils.plugin import install_admin_tool_and_function_dependencies
 from open_webui.utils.oauth import OAuthManager
 from open_webui.utils.security_headers import SecurityHeadersMiddleware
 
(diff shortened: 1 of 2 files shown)


tjbck:
LGTM, Thanks!

PR #13471, opened 2025-05-04
fix: prevent showing disabled setting-tabs to user

diwakar-s-maurya:
### Fixed
If some settings are disabled, eg direct connections, direct tool servers, then they should not be visible to the user role, either directly or through searching the setting-tabs.

### Screenshots
"Manage Direct Connections" is not visible to user role
image

But still they can search for it and see it. Same goes for "Manage Tool Servers".
Screenshot_20250504_110122

### DIFF
--- a/src/lib/components/chat/SettingsModal.svelte
+++ b/src/lib/components/chat/SettingsModal.svelte
@@ -295,16 +295,43 @@
 	];
 
 	let search = '';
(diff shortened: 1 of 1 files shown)


tjbck:
Closing in favour of ed5de96d1d445f4946e8219743b865cb18bb906b

PR #13576, opened 2025-05-06
Fix: Fix Image parameter for Automatic1111 `/generations` endpoint

torisetxd:
### Description

- When sending requests to the `/generations` endpoint for images with a `model` body parameter and the image generation engine is `Automatic1111` an error occurs since the `set_image_model` method is used wrong and missing the request parameter.

### Fixed

- `model` param errors when using `Automatic1111` in image generation API.

### DIFF
--- a/backend/open_webui/routers/images.py
+++ b/backend/open_webui/routers/images.py
@@ -623,7 +623,7 @@ async def image_generations(
             or request.app.state.config.IMAGE_GENERATION_ENGINE == ""
         ):
             if form_data.model:
-                set_image_model(form_data.model)
+                set_image_model(request, form_data.model)
 
             data = {
                 "prompt": form_data.prompt,


tjbck:
PR template needed here.

torisetxd:
Thanks for the feedback. I was trying to provide a quick fix for the Automatic1111 image parameter issue that I encountered with image generation in my tool. Since this is just a one-word change to add the missing 'request' parameter, I kept the PR description concise. Happy to update with the full template if needed, though I didn't think a discussion thread was necessary for such a minimal fix.

Classic298:
@torisetxd i think the main issue is the missing contributor agreement. Discussion probably not needed for a 1 line change

torisetxd:
Oh maybe, I didn't notice one if there was inside of the PR format I simply put only the description and fixes areas since I assumed those were the only ones relevant. But regardless I agree to whichever terms apply.

Classic298:
Try opening a new PR and leave the contributor agreement in there, i think it'll be fine

torisetxd:
Alright, thanks for the quick response.

PR #14005, opened 2025-05-18
fix: prevent hidden models from being selected as default model

JimHeo:
# Pull Request Checklist

I opened [Discussions #14004](https://github.com/open-webui/open-webui/discussions/14004) to describe the issue and proposed solution.

- [x] **Target branch:** verify 'dev' branch
- [x] **Description:** see below
- [x] **Changelog:** see below
- [x] **Documentation:** n/a
- [x] **Dependencies:** none
- [x] **Testing:** yes
- [x] **Code review:** yes
- [x] **Prefix:** 

# Changelog Entry

### Description

1. Problem
When a model is marked Hide Model in Admin Panel, it disappears from the picker but is still auto-selected as the default when starting a new chat.

2. Steps to reproduce

- Go to Admin Panel → Settings → Models.

- Select Model A and Click "more" button and the "Hide Model" button to toggle.

- Click “New Chat” → Model A appears as the selected model.

3. Expected behaviour
A hidden model should not be chosen as the default. The first visible model (or a user-defined default) should be used instead.

Note: The model should still be available for title-generation and tag-generation tasks under user permissions, but modifications are required to prevent users from initiating direct chat with it.

### Changed
- `src/lib/components/chat/ModelSelector/Selector.svelte`: replaced the simple reactive assignment of `selectedModel` with logic that skips hidden models and, if the current value points to a hidden or missing model, updates `value` (and `selectedModel`) to the first visible model or clears it when none are available.

### Fixed
- Prevented hidden models from remaining selected as the default when opening a new chat, so users will now fall back to the first non-hidden model.

---

### Screenshots or Videos

Public Model: gpt-4.1, gpt-4.1-mini, gpt-4.1-nano
Hide Model: gpt-4.1-nano

Admin Panel

User Accessed View

### DIFF
--- a/src/lib/components/chat/ModelSelector/Selector.svelte
+++ b/src/lib/components/chat/ModelSelector/Selector.svelte
@@ -58,7 +58,21 @@
 	let tags = [];
 
 	let selectedModel = '';
-	$: selectedModel = items.find((item) => item.value === value) ?? '';
+	$: {
+			const foundModel = items.find((item) => item.value === value);
+			if (foundModel && !(foundModel.model?.info?.meta?.hidden ?? false)) {
+				selectedModel = foundModel;
+			} else {
+				// hidden model or cannot find model
+				const firstVisibleModel = items.find(item => !(item.model?.info?.meta?.hidden ?? false));
+				if (firstVisibleModel) {
+					value = firstVisibleModel.value; // update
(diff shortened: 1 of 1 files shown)


tjbck:
Hidden models should still be selectable via other methods.

PR #14095, opened 2025-05-20
**fix** if typo in external document loader

runyournode:
ExternalDocumentLoader was never selected as it was outside of the main if/elif/else loop.


tjbck:
Already fixed on dev.

PR #14152, opened 2025-05-22
fix(auth): correctly use password hash when duplicate email records exist

Noxiven:
# Changelog Entry

### Description

The Auth table may contain multiple records with the same email address, particularly if a user was deleted and then signed up again. This PR will use the user ID to select the specific Auth record, ensuring that we use the correct password hash.

### Fixed

- Fixed an invalid password hash use in auth


tjbck:
Not entirely sure when exactly this would be needed as if you follow the typical user deletion process it'll clean up both auth and user table, but wouldn't hurt to be more precise so merging this anyway, Thanks!

PR #14498, opened 2025-05-29
Revision of Spanish Translation (es-ES) v.0.6.13

rgaricano:
Adapted to Open-WebUI v.0.6.13
---
Has been done:

- Translation Review.

- Reverted traduction of "Prompt/indicador" (commit https://github.com/open-webui/open-webui/commit/16a9210321bd9f59abd6df3c6086dbeec0fbafdf   done by phpia).

The reason for reverting it is that while the term "prompt" is used in the more technical AI environment, more colloquial terms prevail in the user environment (e.g., "instructiones" is used in GPT). For the non-specialist user, adopting the meaning of one term or the other is similarly "strange," with the meaning in the native language term being much more descriptive.
It was not implemented at the time due to the speed with which this commit was integrated (it would be appreciated if commits of this type were maintained longer and with prior discussion).

On the other hand, we also have the possibility of a localization that integrates these types of "bilingual" ("Spanglish") terms natively without detriment to a pure localization, such as:
es-US


rgaricano:
Yes Tim, I will do, I dedicated several hours,
but how I have to do? is enought as was in previous commit referenced?

PR #14624, opened 2025-06-03
Patch security risk with public /cache endpoint

rragundez:
The `/cache` endpoint is unprotected and any file that is created in that directory will automatically be available without any auth. 
This PR adds protection over this endpoint to be available only to users of the app but also to only the user specific that created the image. 

In addition it is recommended to remove the extension of the file being written as load balancer and similar systems could activate cache  if the extension is kept  as it is identified as a file, and if cache is activated then it defeats the purpose of authentication as it will serve the file directly without triggering the logic of the endpoint. 
For now the only place I see it being used is for the images, which I already patched in this PR.

I tested but Please test from your side.

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -37,7 +37,7 @@ from fastapi import (
 from fastapi.openapi.docs import get_swagger_ui_html
 
 from fastapi.middleware.cors import CORSMiddleware
-from fastapi.responses import JSONResponse, RedirectResponse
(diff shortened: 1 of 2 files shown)


tjbck:
PR template wanted here.

rragundez:
Why? This is clearly needed and not implemented and it is an easy PR to review

Ithanil:
@rragundez Please just reopen with proper title&template, this seems to be important. 🙏

rragundez:
Done

PR #14629, opened 2025-06-03
feat(i18n): add locale support for Uzbek in Latin (uz-Latn-UZ) and Cyrillic (uz-Cyrl-UZ) scripts

Bek85:


tjbck:
PR template needed.

PR #14649, opened 2025-06-04
Auth cache

rragundez:
# Pull Request Checklist

The /cache endpoint is unprotected and any file that is created in that directory will automatically be available without any auth.
This PR adds protection over this endpoint to be available only to users of the app but also to only the user specific that created the image.

In addition it is recommended to remove the extension of the file being written as load balancer and similar systems could activate cache if the extension is kept as it is identified as a file, and if cache is activated then it defeats the purpose of authentication as it will serve the file directly without triggering the logic of the endpoint.
For now the only place I see it being used is for the images, which I already patched in this PR.

I tested but Please test from your side.

- [ ] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [ ] **Description:** Provide a concise description of the changes made in this pull request.
- [ ] **Testing:** Have you written and run sufficient tests to validate the changes?
- [ ] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [ ] **Prefix:** To clearly categorize this pull request, prefix the pull request title using one of the following:

# Changelog Entry

### Description

- [Concisely describe the changes made in this pull request, including any relevant motivation and impact (e.g., fixing a bug, adding a feature, or improving performance)]

### Added

- [List any new features, functionalities, or additions]

### Changed

- [List any changes, updates, refactorings, or optimizations]

### Deprecated

- [List any deprecated functionality or features that have been removed]

### Removed

- [List any removed features, files, or functionalities]

### Fixed

- [List any fixes, corrections, or bug fixes]

### Security

- [List any new or updated security-related changes, including vulnerability fixes]

### Breaking Changes

---

### Additional Information

- [Insert any additional context, notes, or explanations for the changes]
  - [Reference any related issues, commits, or other relevant information]

### Screenshots or Videos

- [Attach any relevant screenshots or videos demonstrating the changes]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -2219,13 +2219,14 @@ async def process_chat_response(
 
                                                 # ensure the path exists
                                                 os.makedirs(
-                                                    os.path.join(CACHE_DIR, "images"),
+                                                    os.path.join(CACHE_DIR, user.id, "images"),
                                                     exist_ok=True,
                                                 )
 
                                                 image_path = os.path.join(
                                                     CACHE_DIR,
-                                                    f"images/{id}.png",
+                                                    user.id,
+                                                    f"images/{id}",
                                                 )
 
                                                 with open(image_path, "wb") as f:
@@ -2236,7 +2237,7 @@ async def process_chat_response(
                                                     )
 
                                                 stdoutLines[idx] = (
-                                                    f"![Output Image {idx}](/cache/images/{id}.png)"
+                                                    f"![Output Image {idx}](/cache/images/{id})"
                                                 )
 
                                         output["stdout"] = "\n".join(stdoutLines)
@@ -2251,12 +2252,13 @@ async def process_chat_response(
 
                                                 # ensure the path exists
                                                 os.makedirs(
-                                                    os.path.join(CACHE_DIR, "images"),
+                                                    os.path.join(CACHE_DIR, user.id, "images"),
                                                     exist_ok=True,
                                                 )
 
                                                 image_path = os.path.join(
                                                     CACHE_DIR,
+                                                    user.id,
                                                     f"images/{id}.png",
                                                 )
 
@@ -2268,7 +2270,7 @@ async def process_chat_response(
                                                     )
 
                                                 resultLines[idx] = (
-                                                    f"![Output Image {idx}](/cache/images/{id}.png)"
+                                                    f"![Output Image {idx}](/cache/images/{id})"
                                                 )
 
                                         output["result"] = "\n".join(resultLines)
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -37,7 +37,7 @@ from fastapi import (
 from fastapi.openapi.docs import get_swagger_ui_html
 
 from fastapi.middleware.cors import CORSMiddleware
-from fastapi.responses import JSONResponse, RedirectResponse
+from fastapi.responses import FileResponse, JSONResponse, RedirectResponse
 from fastapi.staticfiles import StaticFiles
 
 from starlette_compress import CompressMiddleware
@@ -1628,7 +1628,19 @@ async def healthcheck_with_db():
 
 
 app.mount("/static", StaticFiles(directory=STATIC_DIR), name="static")
-app.mount("/cache", StaticFiles(directory=CACHE_DIR), name="cache")
+
+@app.get("/cache/{path:path}")
+async def serve_cache_file(
+    path: str,
+    user=Depends(get_verified_user),
+):
+    file_path = os.path.abspath(os.path.join(CACHE_DIR, user.id, path))
+    # prevent path traversal
+    if not file_path.startswith(os.path.abspath(CACHE_DIR)):
+        raise HTTPException(status_code=404, detail="File not found")
+    if not os.path.isfile(file_path):
+        raise HTTPException(status_code=404, detail="File not found")
+    return FileResponse(file_path)
 
 
 def swagger_ui_html(*args, **kwargs):


backend/open_webui/utils/middleware.py
tjbck, line 1: Not wanted.

tjbck:
Closing in favour of 9964ad0a5b4243d752709f6efb9da7703e43206c, Thanks!

backend/open_webui/utils/middleware.py
rragundez: What part exactly? Because a penetration test will flag this as at least a medium security risk for sure.

rragundez:
@tjbck  I saw the commit. As I mentioned in the description this will not work in a caching mechanism because of the extension (normally they trigger if it is understood that the endpoint is actually a direct file, similar to .js files that's why Svelte regenerates the files hash on every built). The browser, load balancer or security tool like Akamai will cache the image and bypass the auth of the endpoint

rragundez:
Also the changes you made do not address the cross user access. So if sensitive information like graphs of sales or other financials are generated they are not secure per user

Ithanil:
@tjbck Users being able to freely access _other_ users cached images just by knowing the path can't be acceptable.

tjbck:
@Ithanil how would you handle chat sharing in this case?

Ithanil:
> @Ithanil how would you handle chat sharing in this case?

OK, fair point. Maybe make an exception for shared chats? That would still be better than having it open for all chats.

Any other idea @rragundez ?

tjbck:
@Ithanil the term `freely` is also not quite accurate here, I understand the concern but if you were to apply the same token (attack vector) you could also argue users can "freely" log in to other users account once they know their password.

Classic298:
> @tjbck Users being able to freely access _other_ users cached images just by knowing the path can't be acceptable.

Honestly I do not understand the issue.
So...

the user shares a chat.
And then another user guesses the (very long) share link and now has access to the shared chat.
Where is the issue?
The user shared the chat, another user found it.
This is like uploading a video to youtube in "unlisted" mode or like publicly sharing a google docs "to anyone who has the link"?

rragundez:
I just flagged it. This app has a bunch of security risks that we have been closing. I'm committing back the main ones, but if you don't want that's fine I just flagged them. Like a few months ago I closed one where a user could take full control via code injection of the internal network the app is deployed on.

My advice just pay for a penetration test worth its name and you will see the flags. This one at an enterprise level will be flagged as medium in my experience. Specially if the sell is that anyone can use it in ab enterprise setting. 

Just to add for example there is also no TTL for information/docs being uploaded. 

We can have a chat but I also understand if you don't want it. I can just flag it and help out if you want.

Ithanil:
> > @tjbck Users being able to freely access _other_ users cached images just by knowing the path can't be acceptable.
> 
> Honestly I do not understand the issue. So...
> 
> the user shares a chat. And then another user guesses the (very long) share link and now has access to the shared chat. Where is the issue? The user shared the chat, another user found it. This is like uploading a video to youtube in "unlisted" mode or like publicly sharing a google docs "to anyone who has the link"?

Yeah, you can argue like that for *shared* chats, but as I understand it currently the images are accessible regardless of shared status. That's more like you being able to brute force Google docs links until you eventually hit a *private* document and still have access. And I think a password-based login that is attackable like that would also be considered unsafe.

Not saying this is super critical, at least after https://github.com/open-webui/open-webui/commit/9964ad0a5b4243d752709f6efb9da7703e43206c, but still I'm sure it would be flagged in a test. Also, I know other projects that we use, with pretty much exactly the same problem, that had to put in extensive work to work around it.

rragundez:
Let me clarify it a bit better @tjbck  @Ithanil, the commit that Tim made does not address 2 points:

1. Any file under `/cache` are accessible by any user. This is the discussion you are having above. Not sure why this comparison with the password. Anyone that has developed an enterprise ready application immediately will tell you this is not a best practice. So that is just a fact, the point is more if based on functionality you want to keep it like that and accept the risk, instead of fighting the problem statement. Is normal to find a compromise based on functionality.

2. Since the changes in `middleware.py` were not accepted, the images created are still served with a `.png` extension in the endpoint. Any "formal" best practice deployment will have some sort of caching layer somewhere based on a CDN service, either at a load balancer level, or using more enterprise tooling like Akamai, but there are others. These CDN services by default will cache things that "look" like files. Right now if you leave the extension of the image endpoint, the CDN will see the endpoint ending as a file and the response mimetype is a file so most probably it will cache it. **Next time it will receive the same request from ANYWHERE (not only logged users but also public internet) it will return the file directly without actually calling the endpoint (as it does with `favicon.ico` or images in the `/static` dir but that is a known convention and they are needed to be public for other reasons), rendering the changes in the commit useless.**  Removing the extension will solve this problem for most cases.

So I would suggest you at least accept Nr. 2 as it doesn't affect your discussion. I can submit another PR for this if you want.

BTW for Nr 2, there is a related problem when the application gets updated and the svelte JS files hash changes. There is a logic in the app that when the CDN service tries to load the JS files from cache of the previous version it actually returns a 302 which does not break cache it will just fail to load the files and the app won't load at all, to break the cache it needs to return a 404 so that the CDN does not load from cache but actually serve the new JS files

Classic298:
you can try to create a new (minimal) PR. Let's see what tim thinks

but thanks now i understand. Generated Images are accessible even if you are not a user and even if the chat isn't shared.

Yes I can see this perhaps as a security issue-

tjbck:
I believe the solution here is leave the current naive implementation as it is currently, and introduce a factory structure like how we handle storage to enable local storage as well as a remote storage. PR welcome!

rragundez:
I disagree but is better than the current implementation. Will create the PR.

rragundez:
Please take a look #14774

I also added another security related PR in #14703

PR #14786, opened 2025-06-08
feat: Add OAuth support for Tool Server authentication

jescalada:
This PR adds OAuth support for Tool Server authentication.

**Note:** This PR enables logging in through a provider configured in the backend, and getting an "access token" which can be used to validate subsequent requests. Available providers are Google and Atlassian. Feel free to extend this for the providers that are most relevant!

This PR **does not** implement the entire OpenWebUI -> OAuth + MCP flow discussed in #14121. I hit a few roadblocks along the way and I believe some extra refactoring work is needed to get it to work. I'll write some suggestions about how we could approach that and some clues I found.

# Changelog Entry

### Description

- This PR enables logging in through a provider configured in the backend, and getting an "access token" which can be used to validate subsequent requests. Available providers are Google and Atlassian. This can be easily extended with any provider needed.

### Added

- Added OAuth authentication option for Tool Servers
- Added Tool Server provider configuration variables and provider loading
- Added relevant endpoints for handling OAuth login for tool servers
- Extended `AddServerModal` to show providers and handle entire OAuth flow, including **manual** token refresh

---

### Additional Information

- Relevant issues
  - #14121
  - #13457

#### Demo Video
[Demo](https://github.com/user-attachments/assets/b992d959-a7d4-445e-896a-b872c896fb9f)

Note that my cursor isn't showing in the video, sorry about that!

### Testing

If you'd like to test this PR, you can create your own Google auth client here: https://console.cloud.google.com/apis/credentials

Otherwise, you can do the same with an Atlassian OAuth 2.0 app: https://developer.atlassian.com/console/myapps

Otherwise, I can post my credentials and then manually add a reviewer's email (otherwise it won't work, since it's a test auth app).


backend/open_webui/utils/oauth.py
```diff
@@ -543,3 +544,44 @@ async def handle_callback(self, request, provider, response):
...
+
+        redirect_uri = request.url_for("toolserver_oauth_callback", provider=provider)
+        return await client.authorize_redirect(request, redirect_uri, state=state)
+
+    async def handle_toolserver_callback(self, request, provider):
+        client = self.oauth.create_client(provider)
+        token = await client.authorize_access_token(request)
+
+        try:
+            state = json.loads(base64.urlsafe_b64decode(request.query_params["state"]))
```
akaihola: `state` variable is unused.
```suggestion
            _ = json.loads(base64.urlsafe_b64decode(request.query_params["state"]))
```

limbik-martinj:
Just a question:
How would this PR effect the following usecase?

I have a keycloak OAuth Server deployed and the open-webui is connected with OpenID Oauth.
My OpenAPI server tools also look for a Bearer access token, same one that the frontend webui just acquired.
I have a  customized version that allows me to pass the access token from the webui to the tool server.

Would this PR require me to login separately to aquire a separate token, that's not the same from the main login screen?

jescalada:
@limbik-martinj I believe that my OAuth implementation will generate a token based on the `TOOL_<PROVIDER>_<VARIABLE-NAME>` envs configured in the backend. If this matches the configuration for your OAuth provider, we would expect the login flow to **release a new token** (no login needed since already authenticated). However, there might be issues since the old token would no longer be valid. I would need to test this first to make sure.

Does your use case require sharing/forwarding the user's login token to authenticate to the Tool Server? And if so, is there a reason why the existing Session auth option doesn't enable it?

backend/open_webui/utils/oauth.py
jescalada: The state *may* contain useful information such as the Server URL if we want a complete flow (login -> callback -> verification). Right now, it's only used to verify that the request wasn't forged. I'll commit the suggestion though since it's the cleaner version!

limbik-martinj:
@jescalada 
Yes, I am playing around with OpenAPI/Swagger and Remote MCP servers that require the same `access_token` passed around for SSO issued by my keyclaok Oauth server.

You can see how I customized my fork for that [here in this link](https://github.com/limbik/open-webui/commit/d11c457a7f900ae9ed5e325a75c929e070288ecc#diff-26c3a187bc639161f3919b3be06019a4f03359031559e516d24e546db8cca478R102-R105)

The existing Session auth token by open-webui is good for some useful info provided by the app itself. 
However, not good enough for my APIs that are guarded by my  SSO `access_token` that might have come in from 3rd Party user login.

I understand that my use-case might not be what you are solving here.  Nevertheless, I am very interested changes around SSO and passing that down to remote tooling that is exposed and secured under OpenID SSO.

Screenshot From 2025-06-18 13-14-41

jescalada:
### Update

I just updated this PR to complete the OAuth flow. I created this [Tool Server Demo repo](https://github.com/jescalada/tool-server-demo) to test that everything works.

image

### Important Note

Now, the OAuth flow extracts **two** important auth variables: `id_token`, which is **required** by OAuth flows for identifying the user, and `access_token` which is **optional** and provides access to a specific service or API.

image

limbik-martinj:
I like how you are showing both tokens `id_tokens` for Identifying a user after authentication. and `access_token` which represents authorization to a Resource Server (API).

I am not sure however if you have the headers correct.  Do you think you can revisit that?

Here is my research using Gemini 2.5 Pro:

Standard Practice (RFC 6750 - OAuth 2.0 Bearer Token Usage): Access tokens (especially bearer tokens) are almost universally sent in the Authorization header using the Bearer scheme.

Example: Authorization: Bearer <access_token>

X-Access-Token header: This is a custom header. While technically possible for an API to implement this, it deviates from the widely adopted and recommended Authorization: Bearer header, which is understood by virtually all OAuth 2.0 and OIDC compliant systems and libraries

The id_token is for the client to verify the user's identity. Sending it as a Bearer token to a resource server for authorization is a misuse of the id_token's intended purpose.

Hope this is useful, If I misunderstood your implementation feel free to disregard.

Classic298:
### **WARNING**
**You are trying to merge into `main`!**
Dangerous! This might get your PR closed without a comment or worst case: gets your changes accidentally merged into production!

jescalada:
@limbik-martinj Sorry for the slow update - thanks for the insights! I'll be revisiting this soon.

I misunderstood what the typical use case would look like, and the Python library I used as an example required **both** the `id_token` and the `access_token` in order to authenticate. There are some use cases where we want to figure out who the user is (`id_token`) and then validate that they can access a certain resource (`access_token`) in one go.

As you mentioned, the typical use case only needs an `access_token` to reach an API endpoint (in our case a tool server API).

I'll be fixing this sometime in the next week - mostly to revert the extra `id_token` stuff (the bug fixes and main flow should still work).

jescalada:
I just fixed the implementation to only use the `access_token` - the `id_token` stuff was a misunderstanding on my part. As you mentioned, the `access_token` is the [best practice for accessing an API resource](https://auth0.com/blog/id-token-access-token-what-is-the-difference/).

I also fixed up the [sample Python tool server](https://github.com/jescalada/tool-server-demo) so that it doesn't need the `id_token` anymore (in fact, it removes the JWT library entirely). Meanwhile, the user's identity is fetched through the `/userinfo` endpoint instead.

The fixed flow should be working as expected now:

image

Would love a quick look at this @tjbck, let me know if anything in the sample flow isn't clear. Note that this does not set up the MCP support, but it's a step in the right direction since many users require simpler auth flows for their Tool Servers.

Thank you!

jescalada:
Note: I'm still working on fixing an issue for the `access_token` not refreshing correctly.

jescalada:
Should be good to take a look @tjbck. Sorry about the extra ping! 😃

tjbck:
Closing in favour of OAuth2.1, OAuth2.1 support is implemented in such a way it could theoretically be extended to Tool servers as well.

PR #14841, opened 2025-06-10
Fix return type of update_user_api_key_by_id

mluckydream:
[Fix return type of update_user_api_key_by_id](https://github.com/open-webui/open-webui/commit/f5869f9579c9c3f104ab122c0409d67d4b571e58)


tjbck:
PR template needed

PR #14957, opened 2025-06-13
Fix: postgresql null value search

Classic298:
# Pull Request Checklist

- [X] **Testing:** Have you written and run sufficient tests to validate the changes?
- [X] **Prefix:** To clearly categorize this pull request, prefix the pull request title using one of the following:

# Changelog Entry

### Description

Fixes: https://github.com/open-webui/open-webui/discussions/9318

### Changed

- Replaces null values \u0000 so that no errors happen during chat search with PostgreSQL databases

### Fixed

- Fixed the search

---


Classic298:
I tested these changes on Open WebUI with PostgreSQL: works.

I will test them later on SQLITE too, if my changes break SQLITE.

If it works on SQLITE too, i will set the PR ready for review. For now it is a draft.

Classic298:
Tested the search in an SQLITE db setup. Still works. That being said, my PR should finally fix this old issue for the Search in PostgreSQL environments without breaking anything on SQLITE :)

Ithanil:
I think we really need to get this merged, if it works as intended. I just hit this issue hard when trying to run my cleanup script (https://github.com/open-webui/scripts/blob/main/scripts/cleanup_pg/cleanup_pg.py) in production. Need to replace all \u0000 on the fly in the select statements, which sucks and eats tons of memory on the DB.

Classic298:
I see the connection now. Your cleanup script is hitting the same underlying PostgreSQL issue with \u0000 characters.
My PR actually makes two changes:

Sanitizes message content before storing (prevents new null chars)
Sanitizes search input in get_chats_by_user_id_and_search_text (helps with search)

However, your cleanup script uses direct SQL queries that bypass the Open WebUI API layer entirely, so it would still hit the existing corrupted JSON data in the database.

**The issue is that PostgreSQL fails when parsing ANY JSON containing \u0000, regardless of the query.**

But you're absolutely right that we need this merged to prevent the issue from continuing to spread!

Yet, you still need to adjust your script as it doesn't seem like my PR will address _your_ issue

Classic298:
And yes, @tjbck it would be great to see this soon, as every single person with a PostgreSQL instance has had a broken search for a long time, this would fix it (at least this specific issue of course :D )

Ithanil:
> I see the connection now. Your cleanup script is hitting the same underlying PostgreSQL issue with \u0000 characters. My PR actually makes two changes:
> 
> Sanitizes message content before storing (prevents new null chars) Sanitizes search input in get_chats_by_user_id_and_search_text (helps with search)
> 
> However, your cleanup script uses direct SQL queries that bypass the Open WebUI API layer entirely, so it would still hit the existing corrupted JSON data in the database.
> 
> **The issue is that PostgreSQL fails when parsing ANY JSON containing \u0000, regardless of the query.**
> 
> But you're absolutely right that we need this merged to prevent the issue from continuing to spread!
> 
> Yet, you still need to adjust your script as it doesn't seem like my PR will address _your_ issue

Yes, what I meant is that your PR helps to prevent creation of new entries with \u0000 and thus, when all the old ones are eventually deleted, the problem is finally solved.

backend/open_webui/models/chats.py
```diff
@@ -580,7 +585,7 @@ def get_chats_by_user_id_and_search_text(
         """
         Filters chats based on a search query using Python, allowing pagination using skip and limit.
         """
-        search_text = search_text.lower().strip()
+        search_text = search_text.replace("\u0000", "").lower().strip()
```
tjbck, line 588: qq: why is this part needed?

backend/open_webui/models/chats.py
```diff
@@ -232,6 +233,10 @@ def upsert_message_to_chat_by_id_and_message_id(
         if chat is None:
             return None
 
+        # Sanitize message content for null characters before upserting
+        if isinstance(message.get("content"), str):
+            message["content"] = message["content"].replace("\x00", "")
```
tjbck, line 238: qq: when do we get `"\x00"`?

backend/open_webui/models/chats.py
Classic298: This sanitization is needed because when users search their chat history, the search query itself might contain null characters (either typed directly or pasted from somewhere). If these null characters reach the PostgreSQL JSON queries, they'll cause the same ERROR: unsupported Unicode escape sequence that's breaking the search functionality.

backend/open_webui/models/chats.py
Classic298: The "\x00" (null character) gets introduced when LLM responses contain null characters in their output, or when users paste content from external sources that include embedded null characters. PostgreSQL's JSON functions cannot handle null characters within JSON string values, causing the \u0000 cannot be converted to text error during chat searches.

\x00 is hexadecimal notation and \u0000 is Unicode notation, but they're identical - both represent the ASCII/Unicode null character with byte value 0.

tjbck:
I wonder if there's a more elegant way to address those edge cases, but regardless, I'll merge this for now. Thanks!

PR #14976, opened 2025-06-14
Update follow ups handling to be more robust

skuovo:
Follow ups seem to be received in one of the 3 formats, depending on how the LLM responds.

1. JSON dict as requested, with a "follow_ups" key and list value.
2. JSON list value, without the dict.
3. A numeric bullet-point delimited string, using 1. XXX\n2. XXX\n3. XXX

This change allows all 3 to be handled gracefully, increasing the chances the follow ups are displayed.

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -1069,15 +1069,31 @@ async def process_chat_response(
                         else:
                             follow_ups_string = ""
 
-                        follow_ups_string = follow_ups_string[
-                            follow_ups_string.find("{") : follow_ups_string.rfind("}")
-                            + 1
-                        ]
+                        if '{' in follow_ups_string and '}' in follow_ups_string:
+                            follow_ups_string = follow_ups_string[
+                                follow_ups_string.find("{") : follow_ups_string.rfind("}")
+                                + 1
+                            ]
+                        elif '[' in follow_ups_string and ']' in follow_ups_string:
+                            follow_ups_string = follow_ups_string[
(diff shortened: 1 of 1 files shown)


backend/open_webui/utils/middleware.py
```diff
@@ -1069,15 +1069,31 @@ async def background_tasks_handler():
...
+                            ]
+                        elif '[' in follow_ups_string and ']' in follow_ups_string:
+                            follow_ups_string = follow_ups_string[
+                                follow_ups_string.find("[") : follow_ups_string.rfind("]")
+                                + 1
+                            ]
+                        elif '1.' in follow_ups_string:
+                            lines = follow_ups_string.split('\n')
+                            follow_ups_list = [
+                                line[2:].strip()
```
skuovo, line 1085: This will not handle any numeric prefixes that are multiple digits, but I haven't run into that with the default prompt asking for 3-5 responses.  This could be made more robust if a custom prompt is used requesting more.

backend/open_webui/utils/middleware.py
```diff
@@ -1069,15 +1069,31 @@ async def background_tasks_handler():
...
+                                for line in lines
+                                if len(line) > 2 and line[0].isdigit() and line[1] == '.'
```
skuovo, line 1087: This will not handle multiple digits (eg. more than 10 results in this format), but will fail gracefully and return the first 9.  I haven't seen this behaviour as the default prompt only requests 3-5.

tjbck:
Model dependent behaviour the prompt template should be customized instead.

Issue #15411, opened 2025-07-01
issue: upstream headers are dropped for non-streaming chat completions request


Simon-Stone:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

v0.6.15

### Operating System

Any

### Browser (if applicable)

Any

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

When making requests against `/api/chat/completions`, headers from the upstream APIs (OpenAI, Anthropic, LiteLLM) should be forwarded. 

### Actual Behavior

Headers are currently only passed through from the upstream for streaming requests, but not for non-streamed requests because of this [bit of code](https://github.com/open-webui/open-webui/blob/de018f091260ae757a61e9e4d8691be6442e3ea6/backend/open_webui/routers/openai.py#L849):

The `else` branch only returns the actual payload of the response, but not the headers.

### Steps to Reproduce

 This can be confirmed with the following two curl commands:

```
curl -v --location '<your_owui_url>/api/chat/completions' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer YOUR_API_KEY' \
--data '{
  "model": <model_id>,
  "stream": true,
  "messages": [
    {
      "role": "user",
      "content": "what llm are you"
    }
  ]
}'
```
versus
```
curl -v --location '<your_owui_url>/api/chat/completions' \
--header 'Content-Type: application/json' \
[...]

PR #15412, opened 2025-07-01
fix: Forward response headers

Simon-Stone:
# Changelog Entry

### Description

- Instead of only returning the payload of the response, we make sure that a proper `Response` object is constructed that includes the upstream API's response headers. 

### Added

- Response headers are preserved for non-streaming chat completion requests.

### Changed

- Changed the response construction for non-streaming requests

### Fixed

- Fixes issue #15411

---

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -1309,7 +1309,7 @@ async def process_chat_response(
                         media_type=response.media_type,
                     )
                 else:
-                    return json.dumps(merged_data)
+                    return merged_data
 
             return response
         else:
@@ -1333,7 +1333,7 @@ async def process_chat_response(
                         media_type=response.media_type,
                     )
                 else:
-                    return json.dumps(merged_data)
+                    return merged_data
 
             return response
 
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -1216,9 +1216,11 @@ async def process_chat_response(
 
     # Non-streaming response
     if not isinstance(response, StreamingResponse):
+        response_data = json.loads(response.body.decode())
+
         if event_emitter:
-            if "error" in response:
-                error = response["error"].get("detail", response["error"])
+            if "error" in response_data:
+                error = response_data["error"].get("detail", response_data["error"])
                 Chats.upsert_message_to_chat_by_id_and_message_id(
                     metadata["chat_id"],
                     metadata["message_id"],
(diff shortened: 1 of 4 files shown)


backend/open_webui/utils/middleware.py
```diff
@@ -1216,41 +1216,37 @@ async def background_tasks_handler():
 
     # Non-streaming response
     if not isinstance(response, StreamingResponse):
+        response_data = json.loads(response.body.decode())
```
tjbck: Can we safely assume this will always be the case?

Simon-Stone: The previous code expects `response` (now `response_data`) to be a dict-like. It is possible that `response` (in the updated version) is a dict-like already and does not need to be parsed. I will add another check around the JSON parsing to make sure it only happens if `response` is not already a dict.

backend/open_webui/utils/middleware.py
```diff
@@ -1216,41 +1235,40 @@ async def background_tasks_handler():
 
     # Non-streaming response
     if not isinstance(response, StreamingResponse):
+        if isinstance(response, Response):
+            response_data = json.loads(response.body.decode())
```
rgaricano, line 1239: why new var?
just:
        if isinstance(response, Response):
            response= json.loads(response.body.decode())
        ~~else~~
        and isn't necessary replace response by response_data

rgaricano:
I see a lot of
```
if res and isinstance(res, Response):
            res = res.body.decode()
```
if the idea is have a complete response with  both non/stream, in what cases `res` is not going to be instance of `Response`?
in this case, there is any case when `res`is going to be `[dict]`? or other kind of instance?

I don't see any refac in `backend/open_webui/utils/response.py`, shouldn't it adapt?

(

Simon-Stone:
> I see a lot of
> 
> ```
> if res and isinstance(res, Response):
>             res = res.body.decode()
> ```
> 
> if the idea is have a complete response with both non/stream, in what cases `res` is not going to be instance of `Response`? in this case, there is any case when `res`is going to be `[dict]`? or other kind of instance?
> 
> I don't see any refac in `backend/open_webui/utils/response.py`, shouldn't it adapt?
> 
> (

`StreamingResponse` [is a subclass](https://github.com/encode/starlette/blob/fa5355442753f794965ae1af0f87f9fec1b9a3de/starlette/responses.py#L220) of `Response`, so `isinstance(response, Response)` would also be `True` if `response` is a `StreamingResponse`.

The various checks are there to preserve previous behavior, since I figured the middleware should be decoupled from the functions generating responses. Otherwise it could be a breaking change.

What do you think needs refactoring in `response.py`?

PR #15414, opened 2025-07-01
Feat: add color picker as an input

Simon-Stone:
# Changelog Entry

### Description

I am working on [a tool ](https://github.com/dartmouth/dartmouth-chat-tools/blob/main/src/dartmouth_chat_tools/create_document.py) that has some `UserValves` that let the user define some RGB color values (in this case for some spreadsheet styling). I thought that it would be nice to have a proper color picker when choosing values for these valves in the Chat Controls so I went ahead and created one.

The element shows up if the default value for a valve is a valid RGB hex code. Seemed reasonably unlikely that a valve would fit that format but _not_ need a color picker, so I think this is a pretty solid heuristic.

### Added

- Adds a color picker as the input element if the default value of a Valve is a valid RGB hex code.

### Breaking Changes

- **BREAKING CHANGE**: Valves that contain a string as the default value that *looks* like an RGB hex code (i.e., `"#123ABC"`) but is not representing an RGB color will no longer be able to receive a string that is not a valid RBG hex code through the input element. I would argue that's a very unlikely scenario, but it is technically a breaking change.

### Screenshots or Videos

image

### DIFF
--- a/src/lib/components/common/Valves.svelte
+++ b/src/lib/components/common/Valves.svelte
@@ -7,6 +7,45 @@
 
 	export let valvesSpec = null;
 	export let valves = {};
+
+	function isHexColor(value) {
+		return typeof value === 'string' && /^#([0-9A-F]{3}){1,2}$/i.test(value);
+	}
+
+	function validateHexColor(value) {
+		let cleaned = value.replace(/[^0-9A-Fa-f]/gi, '');
+		cleaned = cleaned.slice(0, 6);
+		// If we have 3 or 6 valid hex chars, it's a valid color
+		if (cleaned.length === 3 || cleaned.length === 6) {
+			return '#' + cleaned.toUpperCase();
+		}
+		// For incomplete values, still prefix with # but don't validate fully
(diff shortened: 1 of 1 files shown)


tjbck:
Closing in favour of f55e93cabea08bc2de917cf25a2b560e6f4fd6f8, Thanks!


```py
class Pipe:
    class Valves(BaseModel):
        """System administrator-level configuration."""

        a_debug: bool = Field(default=True, description="Enable or disable debug mode.")
        pass

    class UserValves(BaseModel):
        """User-level configuration (overrides Valves if set)."""

        color: str = Field(
            default="none",
            description="A color selector.",
            json_schema_extra={"input": {"type": "color"}},
            max_length=100,
        )

        map: str = Field(
            default="none",
            description="A map.",
            json_schema_extra={"input": {"type": "map"}},
            max_length=100,
        )

        pass

    def pipe():
        print("hi")
```

Simon-Stone:
Very nice solution, thanks!

PR #15509, opened 2025-07-03
feat: Allow configuring the PDF backend for Docling

gkachru:
### Description

Implement a setting to capture the PDF backend to pass as a configuration to Docling in case PDFs are not getting processed by the default.

### Added

- A new configurable setting for Docling to set the PDF Backend


tjbck:
CLA required and i18n:parse is not needed.

PR #15575, opened 2025-07-07
Fix: Correct redirect after OAuth login

anslwy:
# **Changelog Entry**

### **Description**

**Current behaviour:** When a user is on a specific page (e.g., a chat or settings) and is prompted to log in, the redirect functionality works correctly for the standard email/password login. But when using an OAuth provider (like Google, GitHub, etc.), the user is always redirected to the root URL (homepage) after a successful login, instead of the last page they were on.

### **Fixed**

This fix ensures the redirect works correctly for the OAuth flow by storing the redirect url inside the OAuth `state` parameter.

### **Contributor License Agreement**

By submitting this pull request, I confirm that I have read and fully agree to the Contributor License Agreement (CLA), and I am providing my contributions under its terms.

### DIFF
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -6,6 +6,7 @@ import uuid
 import json
 
 import aiohttp
+import urllib.parse
 from authlib.integrations.starlette_client import OAuth
 from authlib.oidc.core import UserInfo
 from fastapi import (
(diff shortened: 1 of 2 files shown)


tjbck:
Closing in favour of 4f2b821088367da18374027919594365c7a3f459, Thanks!

PR #15586, opened 2025-07-08
FEAT: Add User Info Headers to External Rerank Model Retrieval

rgaricano:
## Add User Info Headers to External Rerank Model Retrieval

Feature Petition Discussion : https://github.com/open-webui/open-webui/issues/15572#issuecomment-3048584786
-------------

----

### Description

- Added User Info Headers to External Rerank Model Retrieval.
-Added user parameter to ExternalReranker call.

### DIFF
--- a/backend/open_webui/retrieval/models/external.py
+++ b/backend/open_webui/retrieval/models/external.py
@@ -4,7 +4,6 @@ from typing import Optional, List, Tuple
 
 from open_webui.env import ENABLE_FORWARD_USER_INFO_HEADERS, SRC_LOG_LEVELS
 from open_webui.retrieval.models.base_reranker import BaseReranker
-from open_webui.utils.auth import get_admin_user, get_verified_user
 
 log = logging.getLogger(__name__)
 log.setLevel(SRC_LOG_LEVELS["RAG"])
@@ -16,11 +15,12 @@ class ExternalReranker(BaseReranker):
         api_key: str,
         url: str = "http://localhost:8080/v1/rerank",
         model: str = "reranker",
-        user=Depends(get_verified_user),
+        user=user,
     ):
         self.api_key = api_key
         self.url = url
         self.model = model
+        self.user = user
 
     def predict(self, sentences: List[Tuple[str, str]]) -> Optional[List[float]]:
         query = sentences[0][0]
@@ -42,10 +42,10 @@ class ExternalReranker(BaseReranker):
                 "Authorization": f"Bearer {self.api_key}",
             }
             if ENABLE_FORWARD_USER_INFO_HEADERS:
-                headers["X-OpenWebUI-User-Name"] = user.name
-                headers["X-OpenWebUI-User-Id"] = user.id
-                headers["X-OpenWebUI-User-Email"] = user.email
-                headers["X-OpenWebUI-User-Role"] = user.role
+                headers["X-OpenWebUI-User-Name"] = self.user.name
+                headers["X-OpenWebUI-User-Id"] = self.user.id
+                headers["X-OpenWebUI-User-Email"] = self.user.email
+                headers["X-OpenWebUI-User-Role"] = self.user.role
 
             r = requests.post(
                 f"{self.url}",
--- a/backend/open_webui/retrieval/models/external.py
+++ b/backend/open_webui/retrieval/models/external.py
@@ -2,7 +2,7 @@ import logging
 import requests
 from typing import Optional, List, Tuple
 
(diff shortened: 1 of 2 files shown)


tjbck:
is this tested?

backend/open_webui/routers/retrieval.py
```diff
@@ -167,6 +167,7 @@ def get_rf(
                         url=external_reranker_url,
                         api_key=external_reranker_api_key,
                         model=reranking_model,
+                        user=user,
```
athoik: Is user a global variable? I don't see where is defined in get_rf.

tjbck: yeah same concern here

rgaricano: added  user=Depends(get_verified_user)
to ensure that user exists and is verified

backend/open_webui/retrieval/models/external.py
```diff
@@ -36,12 +37,19 @@ def predict(self, sentences: List[Tuple[str, str]]) -> Optional[List[float]]:
             log.info(f"ExternalReranker:predict:model {self.model}")
             log.info(f"ExternalReranker:predict:query {query}")
 
+            headers={
+                "Content-Type": "application/json",
+                "Authorization": f"Bearer {self.api_key}",
+            }
+            if ENABLE_FORWARD_USER_INFO_HEADERS:
```
athoik: Maybe we can use the same logic we have in other places?

```
 r = requests.post( 
     f"{self.url}", 
     headers={ 
         "Content-Type": "application/json", 
         "Authorization": f"Bearer {self.api_key}",
          **(
              {
                  "X-OpenWebUI-User-Name": user.name,
                  "X-OpenWebUI-User-Id": user.id,
                  "X-OpenWebUI-User-Email": user.email,
                  "X-OpenWebUI-User-Role": user.role,
              }
              if ENABLE_FORWARD_USER_INFO_HEADERS and user
              else {}
     }, 
     json=payload, 
 ) 
```

Above logic will work if both ENABLE_FORWARD_USER_INFO_HEADERS is enable and user is not None.

backend/open_webui/routers/retrieval.py
athoik: Great! Please check also my other comment about headers, using same logic as in other places is preffered.
Also the other logic checks if user is not None.

backend/open_webui/retrieval/models/external.py
rgaricano: I saw it, I thought about doing it this way, but I'm not very comfortable with that notation and I was in a bit of a hurry.
I'll take note.
Also I prefer have the call clear and do the asignation before, but I keep as is used in other places.

backend/open_webui/retrieval/models/external.py
```diff
@@ -41,6 +42,15 @@ def predict(self, sentences: List[Tuple[str, str]]) -> Optional[List[float]]:
...
                     "Content-Type": "application/json",
                     "Authorization": f"Bearer {self.api_key}",
+                    **(
+                        {
+                            "X-OpenWebUI-User-Name": self.user.name,
+                            "X-OpenWebUI-User-Id": self.user.id,
+                            "X-OpenWebUI-User-Email": self.user.email,
+                            "X-OpenWebUI-User-Role": self.user.role,
+                        }
+                        if ENABLE_FORWARD_USER_INFO_HEADERS and user
```
tjbck, line 52: `self.user`

tjbck (changes requested):

backend/open_webui/retrieval/models/external.py
athoik: Just a trivial change, but still required, i think it's `if ENABLE_FORWARD_USER_INFO_HEADERS and self.user`

I had difficulty understanding how i should get user variable too, it's tricky ...

backend/open_webui/retrieval/models/external.py
rgaricano: yes, and a closing parenthesis,
but I have problems when testing, no matter how you do it, it always seems that user is not defined.

rgaricano:
with some change in /backend/open_webui/retrieval/utils.py
I have working the user, but I have a method no permited

openwebui_rerank_error_method

I figure that it's because it's trying to reseach /v1/rerank that doesn't exist in openwebui or ollama (my mistake)
I can't do more test because I haven't configurated any reranker engine, I'm going to revise all changes and just check that user is forwared properly.

athoik:
Supersede by #15615

PR #15729, opened 2025-07-15
Fix: some FileItemModal.svelte strings to i18n

rgaricano:
### Added
 - i18n function for strings
 - strings in en-US locales
_____

### DIFF
--- a/src/lib/components/common/FileItemModal.svelte
+++ b/src/lib/components/common/FileItemModal.svelte
@@ -133,9 +133,9 @@
 							>
 								<div class="flex items-center gap-1.5 text-xs">
 									{#if enableFullContent}
-										Using Entire Document
+										$i18n.t('Using Entire Document')
 									{:else}
-										Using Focused Retrieval
+										$i18n.t(`Using Focused Retrieval')
 									{/if}
 									<Switch
 										bind:state={enableFullContent}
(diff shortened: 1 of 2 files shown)


EntropyYue:
You can run `npm run i18n:parse` to add this entry to the localization files

rgaricano:
So, there is no need to push the locale?
just the changed code?

Classic298:
@rgaricano Tim will run the npm run i18n:parse command every time before a new version, so yes, only code pushes, no need to modify the language files unless you are adding translations

EntropyYue:
My point is, this will add the i18n entries referenced in your code to all localization files, and it will be an empty dictionary
There's no need to manually add new keys to the localization files, as this command will do it automatically
You can see that the frontend formatting has failed because this check determines whether `npm run i18n:parse` has been run

EntropyYue:
This will add the i18n entries referenced in your code to all localization files, and it will be an empty dictionary key
There's no need to manually add new keys to the localization files, as this command will do it automatically
You can see that the frontend formatting has failed because this check determines whether `npm run i18n:parse` has been run

rgaricano:
It's what I imagined. Thanks for confirming it. I know for another occasion.

tjbck:
Thanks!

PR #14288, opened 2025-05-24
feat: Optionally hide private workspace items from admins

Classic298:
# Pull Request Checklist

### Related issue:

https://github.com/open-webui/open-webui/discussions/14083

**Before submitting, make sure you've checked the following:**

- [X] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [X] **Description:** Provide a concise description of the changes made in this pull request.
- [X] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [X] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources? https://github.com/open-webui/docs/pull/560
- [ ] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [X] **Testing:** Have you written and run sufficient tests to validate the changes?
- [X] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [X] **Prefix:** To clearly categorize this pull request, prefix the pull request title using one of the following:
  - **feat**: Introduces a new feature or enhancement to the codebase

# Changelog Entry

### Description

- Prev.: If a user created a private item in a workspace section (private prompt, private tool, private knowledgebase, private model), then any admin would always see all items, even if set to private and not shared with any group.
  - This would lead to clutterered UI, obstruct the admin's ability to manage workspaces properly in larger instances
  - Admin does not need to see private user's items - the introduction of a feature to hide private objects would lead to more privacy for users
    - Users would potentially expect private items to be somewhat private

Fixes: https://github.com/open-webui/open-webui/discussions/14083

### Added

- Added a new env var to optionally hide all workspace items that are "private" from all admins.

### Changed

- If the env var is not set, current behaviour does not change.

### Security

- Optional privacy improvements for users.

---

### Additional Information

- Behaviour of prompts/knowledge/etc. that is shared WITH A GROUP does not change
  - Admins are, by default, member of all groups. So if a user creates a prompt and shares it with a group, the admins will see that workspace item.
  - Only if the workspace item was created with "private" permissions and the env var is set, will it be hidden from admins.

Related docs PR: https://github.com/open-webui/docs/pull/560

### Screenshots or Videos

- [Attach any relevant screenshots or videos demonstrating the changes]

### Contributor License Agreement

[...]

PR #15746, opened 2025-07-15
feat: Add admin workspace privacy control with ENABLE_ADMIN_USER_WORKSPACE_ACCESS environment variable

Classic298:
# Pull Request Description

This pull request introduces a new environment variable `ENABLE_ADMIN_USER_WORKSPACE_ACCESS` that allows administrators to control admin workspace access privileges. When disabled, administrators are treated exactly like regular users and only see workspace items (knowledge bases, models, prompts, and tools) that they have explicit access to through the existing access control system.

**Key Features:**
- New `ENABLE_ADMIN_USER_WORKSPACE_ACCESS` environment variable (defaults to `True` for backward compatibility)
- When disabled, admins use the same access control logic as regular users
- Affects all workspace sections: Knowledge, Models, Prompts, and Tools
- Leverages existing user permission functions instead of creating new logic

**Use Cases:**
- Multi-tenant environments where admins should not automatically access all user content
- Organizations requiring explicit permission grants, even for administrators
- Compliance scenarios where data access must be explicitly controlled

**How it works:**
Instead of admins seeing everything by default, they now follow the same permission rules as regular users. If an admin needs access to a user's private content, that user must explicitly share it with the admin or add the admin to a group with appropriate permissions - just like sharing between any other users.

## Checklist Status
- [x] **Target branch:** Targeting `dev` branch
- [x] **Description:** Concise description provided above
- [x] **Changelog:** Added below following Keep a Changelog format
- [X] **Documentation:** Docs PR: https://github.com/open-webui/docs/pull/625
- [x] **Dependencies:** No new dependencies added
- [x] **Testing:** Manual testing performed - admin privacy filtering works correctly in all workspace sections
- [x] **Code review:** Self-reviewed, follows existing code patterns and standards
- [x] **Prefix:** Using `feat:` prefix for new feature

---

# Changelog Entry

### Description
Added `ENABLE_ADMIN_USER_WORKSPACE_ACCESS` environment variable to enable admin workspace access control. When disabled, administrators follow the same access control rules as regular users for workspace items, requiring explicit permission grants to access user content.

### Addresses
Addresses these issues/discussions:

https://github.com/open-webui/open-webui/pull/14288

https://github.com/open-webui/open-webui/discussions/14968

https://github.com/open-webui/open-webui/discussions/14083

These can be closed if this is merged.

### Related Docs PR

https://github.com/open-webui/docs/pull/625

### Added
- New `ENABLE_ADMIN_USER_WORKSPACE_ACCESS` environment variable (defaults to `True`)
- Admin access control for workspace sections (Knowledge, Models, Prompts, Tools)
- Automatic fallback to existing user-based access control when admin access is disabled

### Changed
- Admin access logic in knowledge, models, prompts, and tools routers
- Conditional routing based on admin access setting - admins use regular user access paths when admin access disabled

### Deprecated
- None

### Removed
- None

### Fixed
- None

### Security
- Enhanced data separation capabilities for multi-tenant deployments
- Prevents administrators from accessing private user content unless explicitly granted access

### Breaking Changes
- None - feature maintains admin access by default and maintains full backward compatibility

---

### Additional Information

**Environment Variable Details:**
```bash
# Enable admin workspace access (default: true)
ENABLE_ADMIN_USER_WORKSPACE_ACCESS=true # or false
# When set to false, admins get the same viewing rights as normal users
```


Classic298:
Docs PR ready, PR description done, testing performed, ready for review

This PR aimed to make minimal changes to introduce this privacy preserving feature that is definitely needed.

If a different env var name is desired, changes can be made.

guenhter:
Awesome feature. This improves data security even more. Nice.

One word on the name of the variable: There is already one existing var which is dong this for chats: `ENABLE_ADMIN_CHAT_ACCESS`

Maybe your var name can also follow the `ENABLE_` ... pattern like `ENABLE_ADMIN_USER_WORKSPACE_ACCESS`

Classic298:
@guenhter Do you have another proposal? Enable admin user workspace access doesn't quite fit the description of making admin's workspace behave like normal user access control

guenhter:
I guess it is fine. I just thought keeping the `ENABLE_...` pattern for booleans. As always, the hardest part in software engineering is finding good names :)

mahenning:
You could use `ENABLE_ADMIN_USER_WORKSPACE_ACCESS` and default to `true` to get the current behavior, and set to `false` for the new/PR behavior?

Classic298:
@mahenning yes i could, but as I said I am not quite fond of this name, because it doesn't quite convey what this env var is about.

Classic298:
(editing the code to rename the env var - DRAFTED)

Classic298:
Updated all code sections, PR template and the Docs PR to change to the new environment variable naming.

Thanks guys for your feedback, i changed my mind

This new naming scheme is more in-line with ENABLE_ADMIN_CHAT_ACCESS and also behaves the same: -> if set to true, for both variables, admin has full access. If set to false for both vars -> admin loses full access.

It makes more sense this way.

Classic298:
Tested logic changes after renaming env var just now, works.

Just had to un-negate the if statements.

Classic298:
@tjbck ready for review

guenhter:
@Classic298 one thing came to me mind: What if a user gets deleted, then all his knowledge bases, ... remain but are not accessible (and not deletable) any more, are they?

If this setting is activated, maybe it would be good if admins see also the Knowlede Basisc, Prompts, ... of deleted users.

What do you think?

(this can also be addressed in later PRs and should not block this one. I just want to draw the attention to this so it is not forgotten)

All in all this is an awesome feature and improves data privacy a lot.

Classic298:
@guenhter unrelated to this PR

should be discussed in data retention discussions/issues about orphaned data, and yeah this is an existing problem also for orphaned chars and orphaned files if a user gets deleted

tjbck:
Closing in favour of 55ad48d1c3e8a608193f8d1cf2943f25852d4603, Thanks!

rgaricano:
to me that it is a nosense, an admin role that need and aditional permission? another env var for this?
Is not easier just add a new "pseudoadmin" role? that can work as admin but without admin permissions for workspace.... too much easy for manage future permissions, as it's now is needed to drag this new variable to use that new "permission".

Classic298:
@rgaricano the default behavior doesn't change. This is not a permission setting, this is for optionally restricting the admin

rgaricano:
> @rgaricano the default behavior doesn't change. This is not a permission setting, this is for optionally restricting the admin

If isn't a permission, why is added to permission logic?
you can call as you want, but a var that can restric a permission is a permission setting, external and optional, but a permission, & this can be managed by permission roles,
Why don't have a superadmin, admin or psadmin (with this new restriction)?,  if it's managed with roles just changing the "admin" user role is enought and it doesn't need change env var or restart open-webui.

I understand the funcionality, but I don't share the way to manage it.

What is the problem for not manage it as a new role?

Classic298:
@rgaricano because adding new role types would probably require major refactoring of huge parts of the backend and frontend too.

This PR's scope is limited to fixing the issue that is a privacy invasion into normal user's prompts and other private items in the workspace area by globally disabling admin's access to those items.

A similar variable for disabling the admin's access to the user's chats has also existed for a LONG time. So it is, for now, the only logical way to add a new env var to limit admin access to workspace area private items as well.

It follows the way it was done before.

And refactoring the entire permission and role system just for this one feature is not a good goal.

If you feel like we need new/custom roles quicker, feel free to discuss this on the many discussions that exist on this topic and then propose a PR.

This very PR was also based off of multiple user's questions and complaints that admins can see user's private items in the workspace.

Maxi-J:
Just tested it in the new version of Open-WebUI. Working great in the Workspace section. Is it on purpose, that the users' models are still visible in the model picker?

Classic298:
@Maxi-J Help me understand 

What is the scenario?

You are an admin
User creates model that is fully private or shared with a group that you are not part of (definitely not a public model) and you can see it in the model selector?

Maxi-J:
@Classic298 Yes, exactly.

The custom model (created in Workspace) is only shared with the user's team (not public). The admin is not part of that team and can therefore, with `ENABLE_ADMIN_WORKSPACE_CONTENT_ACCESS=false`, not see this model in the admin's Workspace. For the admin, it is still present in the model selector when chatting though.

Thanks for the quick reply!

Classic298:
Thanks for confirming

Yes this is unintended - I'll take a look at it

Classic298:
@Maxi-J this was a b**** to debug and find out where the root cause was located, but - i figured it out

https://github.com/open-webui/open-webui/pull/16498

check the new PR

Maxi-J:
@Classic298 Looks like it needed 3 additional checks for superficially one thing (filtering the model selector list)... I can only imagine how 'fun' that was to figure out.

Thanks a lot for delving into it and adding the necessary changes, much appreciated!

rgaricano:
this kind of problems are the ones that I was speaking ..."...as it's now is needed to drag this new variable to use that new `permission'..."
And probably some others are going to arise from the frontend logics.

Classic298:
@rgaricano This is not a "problem". I simply forgot about the fact that the model selector is also a place where custom items (in this case, models) from the workspace may appear and that you need to enforce visibility permissions there as well.

I focused entirely on the workspace section when creating THIS original PR, and forgot about other sections.

But the fix: PR i created yesterday should address this.

I think you are being a bit too negative here.
Again, i see your ideas with different roles and such - but you seem to forget that implementing such a massive role system is not trivial and would require MUCH MUCH more changes and variables and checks on all the same places to ensure it works properly.

rgaricano:
I`m not being negative, i'm trying to be propositive.
I understand that this is a band-aid for a feature that need a more deeper implementation, for me the way is through roles & permissions.
Yes you are focusing in workspaces, the most inmediate requested...and to extend to other places?, same bondage?

I don't think that it need a massive role system just some little changes, just use of has_access &| has_permission functions for all permission/access controls, and in those functions deny admin access/permission (others than for change roles, update_user) for _pseudoadmin_ role, then just it's needed change the role of the account to _admin_<->_pseudoadmin_ to access all resources as user or admin.
The only function that need bypass controls is update_user, in which _pseudoadmin_ have to be treated as _admin_.

If any role/permission system need MUCH MUCH more changes is because it's bad implemented, I think that isn't the case!

Classic298:
hmmmmmm

 i get where you're coming from kind of - let's see if this can be done in one of the next major versions

rgaricano:
I will try to do some tests too...if time permit.

PR #15759, opened 2025-07-16
bugfix/ avoid 404 response and handle the posibility that id is not i…

alvarellos:
Description: Avoid the return of 404 error which it is actually happening when id is missing.


tjbck:
PR template with CLA required here.

PR #15810, opened 2025-07-17
FIX: Add posthog lib forcing version to avoid opentelemetry issue

rgaricano:
### Description

Add "posthog==5.4.0" lib to pyproject.toml & backend/requirements.txt to avoid chromadb issue for lib versions incompatibility with that lib (chromadb  requiere posthog>=2.4.0,<6.0.0 ) which produce error in opentelemetry.

issue: https://github.com/open-webui/open-webui/discussions/15624

### Added

Add "posthog==5.4.0" lib to pyproject.toml & backend/requirements.txt 
______

### DIFF
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -51,6 +51,7 @@ langchain-community==0.3.26
 
 fake-useragent==2.1.0
 chromadb==0.6.3
+posthog==5.4.0
 pymilvus==2.5.0
 qdrant-client==1.14.3
 opensearch-py==2.8.0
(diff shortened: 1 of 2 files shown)


tjbck:
qq: do we have any other upstream libs that depend on posthog?

rgaricano:
No, it's used for chromadb & for opentelemetry task, but not directly for open-webui.
It's an issue of chromadb and probably is solved in new versions (i have to see) but as open-webui force to a specific chromadb pre version probably the error are going to persist.

This is a temporaly fix to prevent that the error arise until here & avoid repetitive comments about that error, it could be removed in future revisions/versions.

tjbck:
Appreciate the insight, Thanks!

PR #15935, opened 2025-07-22
FIX: STT default whisper trascription language

rgaricano:
### FIX: STT default whisper trascription language

Fix the transcripcion language used by default whisper, setting as WHISPER_LANGUAGE if it is setted in env var, even if a language is detected in the file's metadata.
It is understood that if a language is set as an environment variable for transcriptions, this should be the preferred one and the one that should be used for that purpose.

It would be advisable to add this variable as configurable in UI

### Changed

- Logic for select transcription language

_____


tjbck:
Agreed, Thanks!

PR #15985, opened 2025-07-24
feat: add support access_control field on folder, like knowledge

blitzh:
### Description

This pull request adds support for the `access_control` field when creating folders. Previously, folder permissions were hardcoded to default values (e.g. current user only), and any `access_control` provided by the frontend was ignored.

With this change:
- The frontend can now pass a full `access_control` structure (including `read` and `write` groups/users)
- The backend accepts and stores the value as-is if provided
- Falls back to default values only if `access_control` is not specified

This is useful for controlling folder-level visibility across users and groups, supporting upcoming multi-user scenarios and group-based sharing in the Open WebUI interface.

---

#### Added

- New field `access_control` (type: JSON) added to `folders` table
- Support for `access_control` in `FolderForm` Pydantic model
- Insertion logic in `insert_new_folder()` updated to store frontend-provided access controls

#### Changed

- Backend now uses `form_data.access_control` if available during folder creation

#### Fixed

- Bug where folder `access_control` was always default, even if provided by the frontend

No Alembic migration is included in this PR — please add the column manually depending on your DB engine:

#### Database Migration

Ensure the `folders` table has a new column added:

```sql
ALTER TABLE folder ADD COLUMN access_control JSON DEFAULT NULL;
```
Or if using SQLite:
```sqlite
ALTER TABLE folder ADD COLUMN access_control TEXT DEFAULT NULL;
```


gdzikowskitt:
Hello, is there a reason this PR was closed? I don't see this feature on main branch, unless I'm missing something

PR #15999, opened 2025-07-24
fix: oauth logout

guenhter:
# Pull Request Checklist

# Changelog Entry

### Description

The current logout handling requried the variable `OPENID_PROVIDER_URL` to be set. If not, the logout gets an error and the user stays logged in. But this variable is usually only necessary  for the generic OIDC provider but not for Google, Microsoft, or Github because theier variables are well  known an contained in the code. 
This PR reuses the correct logout variable depending on the used provider.

Details can be found: https://github.com/open-webui/open-webui/discussions/15007

### Fixed

- OAuth logout when Google, Microsoft or Github provider is used

---

### Additional Information

I probably have not considered all use cases for all providers, but this PR makes it definitely better as before (where no logout at all was possible).

### Screenshots or Videos

Logout when Microsoft is used:
microsoft_sign_out

### DIFF
--- a/backend/open_webui/routers/auths.py
+++ b/backend/open_webui/routers/auths.py
@@ -18,7 +18,7 @@ from open_webui.models.auths import (
     UpdateProfileForm,
     UserResponse,
 )
-from open_webui.models.users import Users
+from open_webui.models.users import UserModel, Users
 from open_webui.models.groups import Groups
 
 from open_webui.constants import ERROR_MESSAGES, WEBHOOK_MESSAGES
@@ -43,15 +43,17 @@ from open_webui.utils.auth import (
     create_api_key,
     create_token,
     get_admin_user,
+    get_current_user_optional,
     get_verified_user,
     get_current_user,
     get_password_hash,
     get_http_authorization_cred,
 )
+from open_webui.utils.oauth import get_oauth_provider_config_for_user
 from open_webui.utils.webhook import post_webhook
 from open_webui.utils.access_control import get_permissions
 
-from typing import Optional, List
+from typing import Optional
 
 from ssl import CERT_NONE, CERT_REQUIRED, PROTOCOL_TLS
 
@@ -667,22 +669,30 @@ async def signup(request: Request, response: Response, form_data: SignupForm):
 
 
 @router.get("/signout")
-async def signout(request: Request, response: Response):
+async def signout(
+    request: Request,
+    response: Response,
+    user: Optional[UserModel] = Depends(get_current_user_optional),
+):
     response.delete_cookie("token")
     response.delete_cookie("oui-session")
+    response.delete_cookie("oauth_id_token")
 
     if ENABLE_OAUTH_SIGNUP.value:
         oauth_id_token = request.cookies.get("oauth_id_token")
-        if oauth_id_token:
+        oauth_provider_config = (
+            get_oauth_provider_config_for_user(user) if user else None
+        )
+        discovery_endpoint = oauth_provider_config.get("discovery_endpoint")
+
+        if oauth_id_token and discovery_endpoint:
             try:
                 async with ClientSession(trust_env=True) as session:
-                    async with session.get(OPENID_PROVIDER_URL.value) as resp:
+                    async with session.get(discovery_endpoint) as resp:
                         if resp.status == 200:
                             openid_data = await resp.json()
                             logout_url = openid_data.get("end_session_endpoint")
                             if logout_url:
-                                response.delete_cookie("oauth_id_token")
-
                                 return JSONResponse(
                                     status_code=200,
                                     content={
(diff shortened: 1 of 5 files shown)


tjbck:
@Classic298 thoughts?

backend/open_webui/routers/auths.py
```diff
@@ -667,22 +669,30 @@ async def signup(request: Request, response: Response, form_data: SignupForm):
...
     response.delete_cookie("oui-session")
+    response.delete_cookie("oauth_id_token")
 
     if ENABLE_OAUTH_SIGNUP.value:
         oauth_id_token = request.cookies.get("oauth_id_token")
-        if oauth_id_token:
+        oauth_provider_config = (
+            get_oauth_provider_config_for_user(user) if user else None
+        )
+        discovery_endpoint = oauth_provider_config.get("discovery_endpoint")
```
guenhter, line 686: `oauth_provider_config` can be `None`

Classic298:
Why would it be beneficial to hardcode it for some providers but not for others? This makes the env var inconsistent in behaviour.

In recent weeks and months there have been many discussions around correct OAUTH configuration, and it was well established that OIDC provider env var needs to be configured.

Why would it be good to change this behaviour?
Rather it would be good to update the docs a bit.

Also, a lot of your changes seem to be style related changes to the code, not quite focused on just changing this behaviour.

guenhter:
What do you mean by "hardcoding"? I have not added any URL but just reused the existing ones.
If two providers are used at the same time, it is not enough to only support one logout url...

Classic298:
@guenhter  the google endpoint is now hardcoded from what I see

Furthermore, this PR introduces a requirement to set the `MICROSOFT_CLIENT_LOGIN_BASE_URL` var for Microsoft, which was NOT needed before.

Furthermore, it introduces an inconsistent behaviour for the OPENID_PROVIDER_URL, because now it is no longer needed for some providers, but it is needed for other providers.

For Microsoft OAUTH for example, OPENID_PROVIDER_URL was necessary to be set for it to work properly - now this would no longer be needed but instead you would need to set MICROSOFT_CLIENT_LOGIN_BASE_URL. We are just trading one env var with another. This doesn't get us anywhere.

Furthermore, the URL you are dynamically generating in the code does not look correct to me, at least I use a different one and it works for me - don't change a running system.

Also this PR introduces a breaking change for most of the OAUTH setups out there, as OPENID_PROVIDER_URL is no longer necessary, but MICROSOFT_CLIENT_LOGIN_BASE_URL becomes necessary, which it never was in the past.

> I probably have not considered all use cases for all providers, but this PR makes it definitely better as before (where no logout at all was possible).

Logout was only not possible, if you did not set the OPENID_PROVIDER_URL env var, which is a required variable. If you didn't set it, of course your OAUTH setup would not work properly, as it is an incomplete setup. Yes the docs are lacking here (although i think they got recently updated on that page) - but this was said again and again in other issues, discussions and PRs that the OPENID_PROVIDER_URL is required. Without it, it's an incomplete setup.

guenhter:
Thx for the feedback. I will work on that.
What is your sughestion how to handle the logout when two providers are used?

Classic298:
@guenhter perhaps the behaviour of OPENID_PROVIDER_URL  can be changed to allow multiple URLs.

This _should_ require minimal changes.

I am thinking of something like

OPENID_PROVIDER_URL = "URL1;URL2;URL3"

This would retain functionality for all other setups with just one provider, introduce compatibility for setups with multiple providers, and prevent the introduction of yet another environment variable?

But the problem you are describing is a tricky one.
When using multiple providers, indeed you have to have different OPENID_PROVIDER_URL URLs for logout to work for all of them

Please update your PR description as well, as this is the motivation behind the PR. I now understand why you would want to change this. At least this makes more sense now. Originally I was so confused why anyone would want to make such changes and make the OPENID_PROVIDER_URL behave differently depending on the provider you use.

But I see now, you attempted to mold the oauth to work so that you can have two or three providers at the same time without the issue of the OPENID_PROVIDER_URL env var.

The linked discussion also seemingly did not discuss issues when having multiple providers at the same time.
So for me it was not clear what the goal of this PR is.
And "fix: oauth logout" was also ambiguous, as oauth logout is not broken.
It should rather be a feature to introduce a way to have multiple providers at the same time with proper configuration options such as a modified OIDC var that also has backwards compatibility.

guenhter:
I tend to close the PR and disable the configuration in our installation to only a single provider. 
Documenting that only a single provider is supported maybe is the easier approach for the monemt. What do you think?

Classic298:
I agree. I will edit the docs to cover this.

Once you have a good `feat:` PR ready to `add multiple OAUTH provider support` we can change it again :)

Classic298:
Related PRs:
https://github.com/open-webui/docs/pull/634
https://github.com/open-webui/docs/pull/635

Issue #16037, opened 2025-07-26
issue: Saving tools as a group user fails due to automatically appending "Loading click Loaded click" at the beginning


vincent0408:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

0.6.18

### Operating System

Windows 11/Debian

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

When giving tools permission to group users, users should be able to simply save their tools.

### Actual Behavior

It formats the code correctly, then adds "Loading click Loaded click" at the beginning.

### Steps to Reproduce

1. Create a user and assign them to a group
2. Give workspace permissions to said group
3. Login as the user and navigate to Tools
4. Create a tool(default template works) and try saving
5. The code successfully formats but "Loading click Loaded click" appears.

### Logs & Screenshots

```
2025-07-26 13:05:31.425 | 2025-07-26 05:05:31.425 | INFO     | open_webui.utils.plugin:install_frontmatter_requirements:241 - No requirements found in frontmatter. - {}
2025-07-26 13:05:31.425 | 2025-07-26 05:05:31.425 | ERROR    | open_webui.utils.plugin:load_tool_module_by_id:111 - Error loading module: f: invalid syntax (<string>, line 1) - {}
[...]

PR #16044, opened 2025-07-26
FIX: User WorkSpace Create Tool Error

rgaricano:
### FIX: User WorkSpace Create Tool Error

- When Creating new tool in user workspace arise an error due to lack of permission when code is formatted.
- It solve https://github.com/open-webui/open-webui/issues/16037

### Changed

- _backend/open_webui/routers/utils.py_
Permission changed to allow verified users to format code; there are more specific permission checks later, in formatPythonCodeHandler.

- _src/lib/components/common/CodeEditor.svelte_
Added permission for the user with code_interpreter permission.
The tools permission could also be used, but I assume the code_interpreter permission is more appropriate for that role; if the user doesn't have code_interpreter permission, they shouldn't have to format the code either.
However, I'm not sure; both permissions could also be used.

### Security

- **REQUIRES REVIEW ON APPLIED PERMISSION.**
__________

### DIFF
--- a/backend/open_webui/routers/utils.py
+++ b/backend/open_webui/routers/utils.py
@@ -33,7 +33,7 @@ class CodeForm(BaseModel):
 
 
 @router.post("/code/format")
-async def format_code(form_data: CodeForm, user=Depends(get_admin_user)):
+async def format_code(form_data: CodeForm, user=Depends(get_verified_user)):
     try:
         formatted_code = black.format_str(form_data.code, mode=black.Mode())
         return {"code": formatted_code}
--- a/src/lib/components/common/CodeEditor.svelte
+++ b/src/lib/components/common/CodeEditor.svelte
@@ -178,7 +178,7 @@ print(black.format_str("""${code.replace(/\\/g, '\\\\').replace(/`/g, '\\`').rep
 	export const formatPythonCodeHandler = async () => {
 		if (codeEditor) {
 			const res = await (
-				$user?.role === 'admin'
+				($user?.role === 'admin' || $user?.permissions?.features?.code_interpreter)
 					? formatPythonCode(localStorage.token, _value)
 					: formatPythonCodePyodide(_value)
 			).catch((error) => {


tjbck:
This is an intended behaviour, /code/format endpoint can be exploited to DDoS the server.

rgaricano:
@tjbck 
Ok, but then if not permission + tool should be hidden, if not is so confusing for users.

coleb95:
> This is an intended behaviour, /code/format endpoint can be exploited to DDoS the server.

Ok, how will this issue be fixed then? I am also having this problem.

vincent0408:
Super confusing bug but the PR was rejected due to unrelated reason, don't know why this is the case. How is manually appending "Loading click Loaded click" in front of  **any tools template** considered as a prevention for DDoS?
If you think the endpoint should be hidden to users, don't give them tool creation permissions and show a warning message instead.

rgaricano:
> Super confusing bug but the PR was rejected due to unrelated reason, don't know why this is the case. How is manually appending "Loading click Loaded click" in front of **any tools template** considered as a prevention for DDoS? If you think the endpoint should be hidden to users, don't give them tool creation permissions and show a warning message instead.

Yes, was reasoned: **native code editor need admin permissions**, **for non admin it use pyodide**, maybe the problem is there. I have to recheck because I did it so fast and I bypass this requirement (I adviced in PR about doubts about security)

tjbck:
@vincent0408 again "Loading click" is NOT a valid python syntax.

@rgaricano Non admin users should also be able to format their code using pyodide, so everything *should* work as intended here.

rgaricano:
Yes, _should_, but it doesn't...I'll continue trying to find the issue cause ;)

vincent0408:
> @vincent0408 again "Loading click" is NOT a valid python syntax.
> 
> @rgaricano Non admin users should also be able to format their code using pyodide, so everything _should_ work as intended here.

No. Read the issue, it shouldn't take multiple comments by different users to let you know what the problem, let alone implying that I can't identify a _valid python syntax_ multiple times

tjbck:
@vincent0408 https://github.com/open-webui/open-webui/issues/16037#issuecomment-3132088731

PR #16098, opened 2025-07-28
feat: Add confirm password field to sign up

silentoplayz:
# Pull Request Checklist
- [X] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [X] **Testing:** Have you written and run sufficient tests to validate the changes?
- [X] **Prefix:** To clearly categorize this pull request, prefix the pull request title using one of the following:

# Changelog Entry

### Description
- This pull request introduces a "Confirm Password" field to the user sign-up page. This enhancement improves user experience by verifying that the user has entered their desired password correctly, reducing the chance of password-related errors upon account creation.

### Added
- **Confirm Password Field**: A new input field has been added to the sign-up form for password confirmation.
- **Password Match Validation**: Both frontend and backend validation have been implemented to ensure that the entered passwords match.
- **Error Messaging**: A user-friendly error message is now displayed if the passwords do not match.

### Changed
- **Sign-Up Form**: The sign-up form in `src/routes/auth/+page.svelte` has been updated to include the new "Confirm Password" field.
- **Backend API**: The `/auths/signup` endpoint and its corresponding `SignupForm` model have been updated to handle the `confirm_password` field.
- **API Function**: The `userSignUp` function in `src/lib/apis/auths/index.ts` has been modified to send the `confirm_password` to the backend.

### Security
- Spammers may need to re-write their automatic sign-up attack scripts to confirm their password upon sign-up now? Idk. Sounds like a win to me though!

### Screenshots or Videos
image

image

### DIFF
--- a/backend/open_webui/constants.py
+++ b/backend/open_webui/constants.py
@@ -47,6 +47,7 @@ class ERROR_MESSAGES(str, Enum):
     INVALID_PASSWORD = (
         "The password provided is incorrect. Please check for typos and try again."
     )
+    PASSWORD_MISMATCH = "The passwords do not match. Please re-enter your passwords and try again."
     INVALID_TRUSTED_HEADER = "Your provider has not provided a trusted header. Please contact your administrator for assistance."
 
     EXISTING_USERS = "You can't turn off authentication because there are existing users. If you want to disable WEBUI_AUTH, make sure your web interface doesn't have any existing users and is a fresh installation."
--- a/backend/open_webui/models/auths.py
+++ b/backend/open_webui/models/auths.py
@@ -87,6 +87,7 @@ class SignupForm(BaseModel):
     name: str
     email: str
     password: str
+    confirm_password: str
     profile_image_url: Optional[str] = "/user.png"
 
 
--- a/backend/open_webui/routers/auths.py
+++ b/backend/open_webui/routers/auths.py
@@ -571,6 +571,9 @@ async def signup(request: Request, response: Response, form_data: SignupForm):
                 status.HTTP_403_FORBIDDEN, detail=ERROR_MESSAGES.ACCESS_PROHIBITED
             )
 
+    if form_data.password != form_data.confirm_password:
+        raise HTTPException(400, detail=ERROR_MESSAGES.PASSWORD_MISMATCH)
+
     user_count = Users.get_num_users()
     if not validate_email_format(form_data.email.lower()):
         raise HTTPException(
--- a/src/lib/apis/auths/index.ts
+++ b/src/lib/apis/auths/index.ts
@@ -290,6 +290,7 @@ export const userSignUp = async (
 	name: string,
 	email: string,
 	password: string,
+	confirm_password: string,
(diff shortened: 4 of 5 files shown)


guenhter:
@silentoplayz Very useful extension. Wish we had this already earlier, because mistyping happens more then one would think.

I just have one remark: wouldn't it be sufficient if this feature stays completely in the UI? The UI can easily check for equality. In the backend I guess it has not so much value, because when the endpoint is used via API, there is no gain having the password twice, has it?

Like here `src/lib/components/chat/Settings/Account/UpdatePassword.svelte`

silentoplayz:
> @silentoplayz Very useful extension. Wish we had this already earlier, because mistyping happens more then one would think.
> 
> I just have one remark: wouldn't it be sufficient if this feature stays completely in the UI? The UI can easily check for equality. In the backend I guess it has not so much value, because when the endpoint is used via API, there is no gain having the password twice, has it?
> 
> Like here `src/lib/components/chat/Settings/Account/UpdatePassword.svelte`

That's a very insightful question. You're right that for an API-first approach, requiring the password to be sent twice might seem redundant. The primary benefit of a "confirm password" field is indeed for the user interface, to prevent users from making typos in their password during sign-up.

However, there are a few reasons why having a backend validation is still a good practice, even if it seems redundant:

    Security Best Practices: It's a common security principle to not trust the client. While the frontend can and should perform validation, a malicious actor could bypass the frontend validation and send a request directly to the API. If the backend doesn't validate, it could lead to unexpected behavior.

    Consistency: The UpdatePassword.svelte component you referenced is a good example of client-side validation. However, the sign-up process is arguably more critical. Enforcing the check on the backend ensures that all new accounts are created with a password that the user has confirmed, regardless of how the request is made.

    API Clarity: By having the confirm_password field in the API, it makes the API's intent clear: the user is expected to confirm their password. This can be helpful for developers who are integrating with the API in the future.

While it's true that the value is diminished when using the API directly, the benefits of backend validation in terms of security and consistency generally outweigh the minor inconvenience of sending the password twice. It's a defense-in-depth approach.

I can revert the backend changes if you'd prefer, but I'd recommend keeping them for the reasons mentioned above. What are your thoughts?

guenhter:
:) I guess my preference is not overruling yours. I'd go with whatever you are fine because you are way more involved. I just thought to align it with the code here `src/lib/components/chat/Settings/Account/UpdatePassword.svelte` because there the confirmation is also handled in UI only.

tjbck:
Closing in favour of b510f21a5b4b4f2039ac5c157dd27ac12ed185db, Thanks!

silentoplayz:
> Closing in favour of [b510f21](https://github.com/open-webui/open-webui/commit/b510f21a5b4b4f2039ac5c157dd27ac12ed185db), Thanks!

I've tested this and can confirm it is working with the environment variable `ENABLE_SIGNUP_PASSWORD_CONFIRMATION="true"` set for my Open WebUI instance. Thanks!

PR #16139, opened 2025-07-30
[WIP] revert: "i18n: improve chinese"

ShirasawaSama:
Reverts open-webui/open-webui#16123

The meanings of the two translations may conflict, and there is no distinction between where they are actually used. After I correctly distinguish them later, I will submit a better pull request (PR)

The translation of the "变量" in Chinese last modified was not done well. Thanks for @qingchunnh's reminder.

Perhaps displaying "管道参数 (Pipelines Valves)" (which meaning "Pipelines Parameters") in Chinese would be a better solution.

image


ShirasawaSama:
The word "阀门" in Chinese only means image

However, the appearance of this term in the current Chinese interface context is quite confusing, as it bears no relevance to pipelines.

image

image

And we can see in the code that currently, Valves are more like a way to store variables. If they are "variables" in the context of the user interface, it would be easier to understand.

https://github.com/open-webui/pipelines/blob/main/examples/pipelines/providers/azure_openai_manifold_pipeline.py#L8

image

https://github.com/open-webui/open-webui/blob/b8da4a8cd8257d4846f3608e299618a0b4f185ed/src/lib/components/admin/Functions/FunctionEditor.svelte#L77

image

>  Initialize 'valves' with specific **configurations**. Using 'Valves' instance helps encapsulate settings,

> Although the term "valve" indeed has a function of "adjusting pipelines" in Chinese, the context here does not contain any content related to "pipelines", making it extremely difficult for users to understand. Users may only comprehend how a product that "must be purchased for home decoration" suddenly appears.

However, there is currently no direct distinction between these two terms, so there is no good solution for the time being.

Perhaps displaying "管道参数 (Pipelines Valves)" (which meaning "Pipelines Parameters") in Chinese would be a better solution.

If it refers to a "valve" in the true sense, users may only understand it if this valve can actually process data, rather than just storing some values.

Take Valve in Tomcat as an example, it can really process data.

image

We have received some user feedback, for the end user, a lone "Valves" word on a fully Chinese interface is very confusing.

ShirasawaSama:
image
image

tjbck:
@ShirasawaSama is this ready to be merged or does this require a discussion amongst the translators?

ShirasawaSama:
> @ShirasawaSama is this ready to be merged or does this require a discussion amongst the translators?

I think we can keep this modification (not revert it). I have also consulted some native English speakers and native Chinese speakers, and the current translation of "变量" is not unacceptable.

However, this translation may still be controversial. We can turn it into a discussion, and if someone has a better translation, they can participate in the discussion. The core issue is that the function of "Valve" is not particularly well-explained in the Chinese interface.

On the other hand, it seems that the main maintainer of the Chinese language, @qingchunnh, has disappeared from his account. If there is a new language that needs to be translated, I may be able to temporarily participate.

PR #16336, opened 2025-08-06
feat: ollama think level

EntropyYue:
# Changelog Entry

### Description

Added support for ollama's think level, if "think" is set to false, it will not work

### Added

- Added support for ollama's think level


tjbck:
We already have think advanced param support, why do we need this logic?

EntropyYue:
ollama's think parameter supports "high", "medium", "low", true, or false. We need this logic to convert Reasoning Effort into ollama's format.
https://github.com/ollama/ollama/issues/11752

PR #16385, opened 2025-08-08
feat: Propagate upstream OpenAI router errors

gaby:
### Description

The main purpose of this PR is to change how errors are handled by the `openai` router. Currently when a bad request happens, Open-WebUI doesn't propagate the error back to the user, instead of just has a generic "Bad Request". This makes debugging for API users very hard, since they can't see which fields are wrong/missing.

- Propagate upstream OpenAI error payloads instead of generic messages.
- Handle both JSON and Plaintext responses from the router.
- Handle error statuses in verification, chat completion, embedding and proxy routes.

---


tjbck:
LGTM, Thanks!

PR #16734, opened 2025-08-19
UPD: Enable ExceptionGroup catching in tasks.py & handler in main.py

rgaricano:
Implementation of ExceptionGroup catching as commented in https://github.com/open-webui/open-webui/discussions/16732#discussioncomment-14155085

In some scenarios with concurrent tasks, for example when uploading files (e.g. issue https://github.com/open-webui/open-webui/discussions/16964), multiple tasks may fail simultaneously, resulting in "ExceptionGroup: Unhandled errors in a TaskGroup." errors.

To solve this it's necessary manage this faults as ExceptionGroup, and that is what this implementation does.

As also arised errors `ERROR | open_webui.tasks:redis_task_command_listener:55 - Error handling distributed task command: Expecting value: line 1 column 1 (char 0)` due to a Redis pub/sub message contains empty or invalid JSON data,  a check for this situation is implemented.

This errors are "dificult" to reproduce/verify, for this reason I also tested with a python script  that generate failed concurrent tasks: [test_exceptiongroup.py](https://github.com/user-attachments/files/21860833/test_exceptiongroup.py) (place&execute in backend dir)

I tested in diferent ways, uploading multiples files, running the script some times between other tasks,...
Now only appear a ` | WARNING  | open_webui.tasks:redis_task_command_listener:46 - Invalid JSON in Redis message: invalid_json... Error: Expecting value: line 1 column 1 (char 0)`

Changes:

in tasks.py
Added caching for multiple concurrent exceptions.
Modified:
- async def redis_task_command_listener
- async def cleanup_task

in main.py
Added caching for concurrent exceptions handler.
Modified:
- redis_task_command_listener

Regards
Ricardo

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -578,8 +578,19 @@ async def lifespan(app: FastAPI):
 
     yield
 
-    if hasattr(app.state, "redis_task_command_listener"):
-        app.state.redis_task_command_listener.cancel()
+# In the lifespan shutdown  
+    if hasattr(app.state, "redis_task_command_listener"):  
+        try:  
+            app.state.redis_task_command_listener.cancel()  
+            await app.state.redis_task_command_listener  
+        except ExceptionGroup as eg:  
+            log.error(f"Multiple errors during listener shutdown: {len(eg.exceptions)} exceptions")  
+            for exc in eg.exceptions:  
+                log.error(f"Shutdown error: {type(exc).__name__}: {exc}")  
+        except asyncio.CancelledError:  
+            pass  # Expected during shutdown  
+        except Exception as e:  
+            log.error(f"Error during listener shutdown: {e}")
 
 
 app = FastAPI(
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -578,17 +578,19 @@ async def lifespan(app: FastAPI):
 
     yield
 
-# In the lifespan shutdown  
-    if hasattr(app.state, "redis_task_command_listener"):  
-        try:  
-            app.state.redis_task_command_listener.cancel()  
-            await app.state.redis_task_command_listener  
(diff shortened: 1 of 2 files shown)


backend/open_webui/main.py
```diff
@@ -578,8 +578,21 @@ async def lifespan(app: FastAPI):
 
     yield
 
+    # In the lifespan shutdown
     if hasattr(app.state, "redis_task_command_listener"):
-        app.state.redis_task_command_listener.cancel()
+        try:
+            app.state.redis_task_command_listener.cancel()
+            await app.state.redis_task_command_listener
```
rgaricano, line 585: Added for inmediate cancelation. This line waits for the task to actually complete its cancellation process. Without this await, the application shutdown would continue immediately after requesting cancellation, potentially causing issues .

rgaricano:
Other error that this PR catch & address: 
```
| ERROR    | open_webui.tasks:cleanup_task:88 - Task exception was never retrieved
2025-08-21T00:15:08.753099711Z future: <Task finished name='Task-1774301' coro=<process_chat_response.<locals>.response_handler() done, defined at /app/backend/open_webui/utils/middleware.py:1422> exception=ClientPayloadError("Response payload is not completed: <TransferEncodingError: 400, message='Not enough data to satisfy transfer length header.'>")>
```
https://discord.com/channels/1170866489302188073/1210732510250541067/1407881916522168363

This error can occurs when:
- The AI model's response stream is interrupted or corrupted
- Network connectivity issues cause incomplete HTTP responses
- The upstream model server returns malformed chunked transfer encoding

Classic298:
This conflicts with dev and has had no activity since 21 August 2025. Closing it as stale. If you still want this in, open a discussion so it can be discussed and maybe merged in

Issue #16625, opened 2025-08-14
issue: ENABLE_ADMIN_CHAT_ACCESS does not remove "export Database" and "Export all chats" functionality from admin ui


dwt:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.6.18

### Ollama Version (if applicable)

0.11.2

### Operating System

MacOS 15.6 (24G84)

### Browser (if applicable)

Safari 18.6

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

When setting the einvironment variable `ENABLE_ADMIN_CHAT_ACCESS`, the admin settings should not provide any way to download individual or collective user chats.

### Actual Behavior

When setting the einvironment variable `ENABLE_ADMIN_CHAT_ACCESS` it is no longer possible to view individual users chat history in the admin -> users view. But it is still possible to switch to the database admin view and download all chats from all users, as well as the full database, which also gets those.

<img width="746" height="267" alt="Image" src="https://github.com/user-attachments/assets/ccb0822a-adce-40cf-9398-8427f8c9f713" />

### Steps to Reproduce

1. Install via Helm-Chart https://open-webui.github.io/helm-charts
2. Enable setting `ENABLE_ADMIN_CHAT_ACCESS=false, in the values.yaml like this: (excerpt)
    ```yaml
    extraEnvVars:
      - name: ENABLE_ADMIN_CHAT_ACCESS
        value: "false"
    ```
[...]

PR #16738, opened 2025-08-19
feat/fix: Use ENABLE_ADMIN_CHAT_ACCESS as a default for ENABLE_ADMIN_EXPORT

dwt:
This still retains the ability to manually configure ENABLE_ADMIN_EXPORT, but defaults its value to the value of ENABLE_ADMIN_CHAT_ACCESS if set, as the export contains the chats.

This is a followup to the discussion in #16625 where I found it highly unintuitive that disabling admin chat access, still retains his ability to regain that access by exporting all chats.

Looking at the code, there are examples like

```python
if PGVECTOR_PGCRYPTO and not PGVECTOR_PGCRYPTO_KEY:
    raise ValueError(
        "PGVECTOR_PGCRYPTO is enabled but PGVECTOR_PGCRYPTO_KEY is not set. Please provide a valid key."
    )
```

and

```python
if OLLAMA_BASE_URL == "" and OLLAMA_API_BASE_URL != "":
    OLLAMA_BASE_URL = (
        OLLAMA_API_BASE_URL[:-4]
        if OLLAMA_API_BASE_URL.endswith("/api")
        else OLLAMA_API_BASE_URL
    )
```

So dependant environment variables are not a new thing, which is why I hope you can accept this pull request.

# TODO

- [ ] [update Documentation](https://docs.openwebui.com/getting-started/env-configuration/#enable_admin_export) once we know this pull request has a chance to get merged

# Changelog Entry

### Description

- The environment variable `ENABLE_ADMIN_EXPORT`, now defaults to the value of `ENABLE_ADMIN_CHAT_ACCESS`, as the export contains the chats of all users.

### Changed

- If you have set `ENABLE_ADMIN_CHAT_ACCESS` to `false` but have not set `ENABLE_ADMIN_EXPORT`, it will now also be set to false and the respective controls in the admin interface will be hidden.

### Security

- This might fix a security violation where you wanted your admins to not be able to read user chats, but they still where able to via the export.

---

### Additional Information

- This was first reported by @dwt in #16625

### Screenshots or Videos

- 
image

- [x] By submitting this pull request, I confirm that I have read and fully agree to the [Contributor License Agreement (CLA)](/CONTRIBUTOR_LICENSE_AGREEMENT), and I am providing my contributions under its terms.

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -1349,14 +1349,18 @@ WEBHOOK_URL = PersistentConfig(
     "WEBHOOK_URL", "webhook_url", os.environ.get("WEBHOOK_URL", "")
 )
 
-ENABLE_ADMIN_EXPORT = os.environ.get("ENABLE_ADMIN_EXPORT", "True").lower() == "true"
-
 ENABLE_ADMIN_WORKSPACE_CONTENT_ACCESS = (
     os.environ.get("ENABLE_ADMIN_WORKSPACE_CONTENT_ACCESS", "True").lower() == "true"
 )
 
 ENABLE_ADMIN_CHAT_ACCESS = (
     os.environ.get("ENABLE_ADMIN_CHAT_ACCESS", "True").lower() == "true"
+)
+
+ENABLE_ADMIN_EXPORT = (
+    os.environ.get("ENABLE_ADMIN_EXPORT",
+        os.environ.get("ENABLE_ADMIN_CHAT_ACCESS", "True")
+    ).lower() == "true"
 )
 
 ENABLE_COMMUNITY_SHARING = PersistentConfig(


tjbck:
Those two are distinct env vars, `ENABLE_ADMIN_EXPORT` should be better documented on our docs instead.

Classic298:
docs changes are live on docs:dev

dwt:
@tjbck Sure these are distinct env vars, but why not use one of them as a default for the other? I know I was quite surprised that turning off `ENABLE_ADMIN_CHAT_ACCESS` did not disable `ENABLE_ADMIN_EXPORT`.

It is very likely that this kind of footgun is hurting other people too.

@Classic298 I appreciate moving the variables nearer together at https://github.com/open-webui/docs/blob/main/docs/getting-started/env-configuration.md#enable_admin_export, as that vastly increases visibility, but if we do not have any default here, we should at least document that you usually want to also disable `ENABLE_ADMIN_EXPORT` if you disable `ENABLE_ADMIN_CHAT_ACCESS`, as it is not obvious from the description yet, that admin export also grants access to chats. (I mean, if you know you know, but if you don't you don't)

Classic298:
@dwt PR welcome for docs, will happily merge that

and i can understand Tims point, both variables are independent of another, and there should be no reason for them that one sets the other - i can imagine many scenarios where one may want to enable one of them without enabling the other

dwt:
@Classic298 Will do.

Regarding the use case of enabling the one without enabling the other: This pull request perfectly supports that. The only thing it changes is the _default_. That is if you set _just_ `ENABLE_ADMIN_CHAT_ACCESS` to false, without also setting anything for `ENABLE_ADMIN_EXPORT` - then, and only then will `ENABLE_ADMIN_EXPORT` also be set to false. Both are of course still settable independently. (Of course this will also need docs, which are forthcoming).

Classic298:
Yes i understand that but it will be an unexpected change in behaviour for some instances that rely on both variables being independent of another and that only set one of them.

dwt:
@Classic298 But that's the point behind this pull request? If you set enable chat admin chat access to false, _you would expect_ that admins cannot access chats anymore. It is still possible though. If you actually wanted that, you should enable it.

What I'm saying is that there is an overwhelming likelihood that way more people actually would like admin export to be disabled but didn't see / notice this, than the amount of people that intentionally disabled admin access, but still wanted admins to get around that restriction by downloading a json file.

So the argument is that the current status quo breaks the setup of more people than the fix will.

Classic298:
I get where you're coming from but both of us don't have real number based arguments.

We are both speculating on what users are expecting and trying to achieve when setting either environment variable.

My point remains that if i disable admin access to user's chats, then i expect the direct chat access in the user overview to get hidden.

Database exports, which are meant for management/backup purposes should remain untouched from this.

Database exports only work if SQLite is used anyways.
And if you use SQLite, you can make use of database encryption and allow for encrypted database exports.

My point remains: the two settings should not be directly linked. Tim decided this as well by closing the PR and referencing the docs instead. It's not wanted to link these two env vars because there are scenarios where it's not wanted that they are linked.

The docs changes are live on the website now and I hope they'll clarify what admins want to set or not to set.

PR #16763, opened 2025-08-21
notes: prevent title caret jump by ignoring socket title updates while input focused

elijah-g:
Problem
When editing a note title, characters could be deleted and the cursor would jump unexpectedly. This happened because real-time “note-events” (socket) updates were overwriting the local input value while the user was actively typing, resetting the input and caret.

Fix
In src/lib/components/notes/NoteEditor.svelte, guard applying server-driven title updates while the title input is focused:
•  Only update note.title from socket if the input is not focused and the incoming title differs.

Why this helps
This avoids mid-typing overwrites that caused the caret to move and occasional character loss while preserving real-time updates when the user isn’t editing.


tjbck:
CLA required.

PR #16898, opened 2025-08-25
FEAT: Add Setting for Code Editor Indentation, tab/spaces

rgaricano:
### FEAT: Add Setting for Code Editor Indentation, tab/spaces

Implementation of setting for Code Editor Indentation as required in https://github.com/open-webui/open-webui/discussions/16892

Add:
- config var `CODE_EDITOR_USE_TABS`
- User Setting var `codeEditorUseTabs`
- swith of CodeMirror indentUnit to fit with that setting.
- i18n Strings: _Code Editor Use Tabs_, _Use tabs_ & _Use spaces_

Note: If necessary refresh the page with code blocks if the change of the User Setting Interface is done with the same page opened.
____


tjbck:
This does not change model code block responses, I believe this is not as trivial as just adding this in.

Issue #16896, opened 2025-08-25
issue: HybridSearch ZeroDivisionError: division by zero


Schwenn2002:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.6.25

### Ollama Version (if applicable)

ollama 0.11.6

### Operating System

Ubuntu 24.04

### Browser (if applicable)

chrome

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

The expected result is that documents are displayed from the RAG and the API returns corresponding chunks.``

### Actual Behavior

I get an error when querying RAG (http status 400) and in the Docker log there is an error for the hybrid search.

Test via the API:
/api/v1/retrieval/query/Collection

```
{
  "collection_names": [
    "3ce4b804-6c65-46c2-b93c-b6aec5345742"
  ],
  "query": "XXX some Text XXX",
  "k": 10,
  "k_reranker": 10,
  "r": 0.8,
  "hybrid": false,
  "hybrid_bm25_weight": 0.6
}
```

### Steps to Reproduce

Go to the URL /docs#/retrieval/query_collection_handler_api_v1_retrieval_query_collection_post and execute a test API query.

### Logs & Screenshots

 ██████╗ ██████╗ ███████╗███╗   ██╗    ██╗    ██╗███████╗██████╗ ██╗   ██╗██╗
[...]

PR #16901, opened 2025-08-25
FIX: hybrid param in query collection not handled

rgaricano:
### FIX: hybrid param in query collection not handled 

Fix  Issue: https://github.com/open-webui/open-webui/issues/16896

Problem:
- The QueryCollectionsForm model defines a hybrid field, but the handler logic only checks the global configuration ENABLE_RAG_HYBRID_SEARCH.
- division by zero error in the query_doc_with_hybrid_search function when BM25Retriever.from_texts() is called with an empty document list.

This PR fix this two issues adding this param to logic checks & logic for check emptry result sets in Hybrid Search (utils.py)

Also, same is added to _query_doc_handler_ that have same logic check.

Note: In the mentioned issue https://github.com/open-webui/open-webui/issues/16896 is showed another "issue" that arise in hybrid search & when there are empty result sets or malformed similarity scores in the specific collection data. This issue doesn't arise if hybrid=false.
Note1: Added Fix for solve both.

____

### DIFF
--- a/backend/open_webui/retrieval/utils.py
+++ b/backend/open_webui/retrieval/utils.py
@@ -124,20 +124,20 @@ def query_doc_with_hybrid_search(
     hybrid_bm25_weight: float,
 ) -> dict:
     try:
-        if not collection_result.documents[0]:  
+        if not collection_result.documents[0]:
             log.info(
                 f"Collection {collection_name} is empty, falling back to vector search only"
-            )  
-            # Use only vector search when no documents are available  
-            vector_search_retriever = VectorSearchRetriever(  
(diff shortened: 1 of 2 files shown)


tjbck:
Closing in favour of 07357afcf6a9f06b56cdfc85eec8b8b1512c66d7, Thanks!

rgaricano:
@tjbck 
& another commit for the division by zero error when no results? (I included both issues in same PR)
https://github.com/open-webui/open-webui/issues/16896#issuecomment-3223080188

or let the error appear and the fallback  as now?

PR #17049, opened 2025-08-29
FIX: Hybrid Search 

rgaricano:
### FIX Error in Hybrid Search

Fix 2 errors in hibryd Search:
- lexical-semantic terms are inverted
BM25 weight=1 --> lexical
BM25 weight=0 --> semantic

- Errors when Hybrid Search when bm25_weight=0.
   As noted in https://github.com/open-webui/open-webui/issues/17046 & https://github.com/open-webui/open-webui/discussions/16957 errors arise in this condition due to return empty list instead of fetching the actual collection data.
   
I opted for revert  the bypass when bm25_weight=0 done in PR https://github.com/open-webui/open-webui/commit/74b1c801 because this proposal sought to save calls if bm25 weight=0, which makes sense, since that setting practically implies the same thing as disabling hybrid search.
However, it was not considered that running hybrid search with bm25 weight=0 can also be used to perform a new reranking with different cutoff values ​​(both in terms of the number of results and the relevance threshold).

--------------------

### DIFF
--- a/backend/open_webui/retrieval/utils.py
+++ b/backend/open_webui/retrieval/utils.py
@@ -128,8 +128,6 @@ def query_doc_with_hybrid_search(
             log.warning(f"query_doc_with_hybrid_search:no_docs {collection_name}")
             return {"documents": [], "metadatas": [], "distances": []}
 
-        # BM_25 required only if weight is greater than 0
-        if hybrid_bm25_weight > 0:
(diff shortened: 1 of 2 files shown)


rgaricano:
@mahenning 
This PR revert bypass when bm25_weight=0 done in PR https://github.com/open-webui/open-webui/commit/74b1c801

tjbck:
Thanks!

PR #17061, opened 2025-08-31
fix: for forming TTS request URL when there is a slash at the end of the base URL

vasudan:
# Pull Request Checklist

- [ ] **Testing:** Have you written and run sufficient tests to validate the changes?

# Changelog Entry

### Description
Currently, the URL to the TTS API is made by concatenating strings together. If a user adds a slash to the end of their base URL in their settings, then requests will fail. To resolve this, `urljoin` is used instead so it doesn't matter how they format the end of their base URL.

### Changed
Used urllib.parse.join instead of string concatenation to create the request to the TTS server
Other thing: Removed unused `pathlib.Path` from this file.

---

### DIFF
--- a/backend/open_webui/routers/audio.py
+++ b/backend/open_webui/routers/audio.py
@@ -4,7 +4,6 @@ import logging
 import os
 import uuid
 from functools import lru_cache
-
 from pydub import AudioSegment
 from pydub.silence import split_on_silence
 from concurrent.futures import ThreadPoolExecutor
@@ -338,7 +337,6 @@ async def speech(request: Request, user=Depends(get_verified_user)):
             async with aiohttp.ClientSession(
                 timeout=timeout, trust_env=True
             ) as session:
-                
                 r = await session.post(
                     url=urljoin(request.app.state.config.TTS_OPENAI_API_BASE_URL, "/audio/speech"),
                     json=payload,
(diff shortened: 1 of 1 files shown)


vasudan:
Here is the discussion post
https://github.com/open-webui/open-webui/discussions/17060

tjbck:
Thanks!

rsxdalv:
This breaks API endpoints.


audio.py:

```python
urljoin(request.app.state.config.TTS_OPENAI_API_BASE_URL, "/audio/speech")
```

For example:
urljoin("http://localhost:7778/v1", "/audio/speech") -> http://localhost:7778/audio/speech

rsxdalv:
I will make a PR for reverting/fixing this issue.

rsxdalv:
There are multiple problems with this approach:
1. it's not consistent, not all places that rely on request.app.state.config.TTS_OPENAI_API_BASE_URL are converted to this.
2. Open-WebUI does not rely on urljoin except for search_firecrawl specifically.
3. Even if I fix it by using "audio/speech", such as `urljoin("http://localhost:7778/v1/", "audio/speech")`, now the trailing slash is mandatory.
4. The rest of the WebUI shows hints for what the final URL will be, this appears to be a more consistent and fundamentally stable approach.

PR #17231, opened 2025-09-05
dependency: Adding LibreOffice in Dockerfile

minashaygan:
adding LibreOffice to Dockerfile and change tag in docker image


tjbck:
https://docs.openwebui.com/license/

PR #17240, opened 2025-09-06
feat: model provider's web search

diwakar-s-maurya:
# Changelog Entry

### Description

Models like [gemini](https://ai.google.dev/gemini-api/docs/google-search) and [openai](https://platform.openai.com/docs/guides/tools) provide web search directly from their end. Currently there is no way to indicate to openwebui or pipelines to make use of model provider's web search tool instead of "web_search" implemented within openwebui.

To utilize model provider's web search tool:
- If `native_web_search` capability is on, pass "web_search" as a tool to model pipeline just like we do for "function calling" feature, and skip running the web_search implemented within openwebui
- in pipeline, detect the tool definition presence `body["tools"]` and add the web search tool of model provider in inference call.

### Added

- Flag to using model provider's web search

---

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -895,9 +895,30 @@ async def process_chat_payload(request, form_data, user, metadata, model):
             )
 
         if "web_search" in features and features["web_search"]:
-            form_data = await chat_web_search_handler(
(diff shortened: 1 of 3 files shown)


tjbck:
Can be done via custom parameters.

PR #16677, opened 2025-08-16
Fix- Missed deps in package.json

rgaricano:
Fix for errors when building due to a missed dependency and module.

Added missed dependency:  "@tiptap/suggestion": "^3.0.9",
and module: "y-protocols":"^1.0.6",

### Contributor License Agreement

By submitting this pull request, I confirm that I have read and fully agree to the [Contributor License Agreement (CLA)](/CONTRIBUTOR_LICENSE_AGREEMENT), and I am providing my contributions under its terms.

PR #17250, opened 2025-09-06
UPD: Unified_use_hightlight.js

rgaricano:
### UPD:   Unifying the duplicate highlight.js usage to reduce bundle size.

# Changelog Entry

### Description

-  To solve the bundle duplication problem where both CodeMirror's native language packages and highlight.js were being included separately.

### Changed

- Configure CodeMirror without language-specific packages.
- Created custom CodeMirror HighlightJSExtension.js, to use highlight.js as the highlighter for CodeMirror's syntax highlighting.
- Removed unused libs due to changes from package.json & package-lock.json:
    - "codemirror-lang-elixir": "^4.0.0"
    - "codemirror-lang-hcl": "^0.1.0"
_____


ShirasawaSama:
Thanks, I will recheck to confirm that the duplicate highlight.js usage have been correctly removed.

> _But I briefly looked at the commit, did you rerun `npm install --force`, it seems to me that the `package-lock.json` file was changed manually?_

rgaricano:
Yes, with package.json changed & after I had to fix some issues for that.

But I didn't test with different browsers & systems.

ShirasawaSama:
QQ_1757266426565

The current code appears to be failing compilation. I recommend converting it to draft to prevent accidental merging.

QQ_1757266504965

rgaricano:
Thanks, @ShirasawaSama ,
 Yes, I removed wrong lib when made the PR (I edited on my github repo directly).
the libraries to be removed are:
		"codemirror-lang-elixir": "^4.0.0",
		"codemirror-lang-hcl": "^0.1.0",

src/lib/components/common/CodeEditor.svelte
```diff
@@ -6,8 +6,8 @@
 	import { acceptCompletion } from '@codemirror/autocomplete';
 	import { indentWithTab } from '@codemirror/commands';
 
-	import { indentUnit, LanguageDescription } from '@codemirror/language';
-	import { languages } from '@codemirror/language-data';
+	import { highlightJSExtension } from './CodeEditor/HighlightJSExtension.js';
```
ShirasawaSama: should be `'./HighlightJSExtension.js'`

ShirasawaSama:
image

It looks like there are still two copies.

You can run `npm install rollup-plugin-visualizer -f` then edit `vite.config.ts`:

```ts
import { visualizer } from "rollup-plugin-visualizer";

import { viteStaticCopy } from 'vite-plugin-static-copy';

export default defineConfig({
	plugins: [
		// ......
		visualizer(),
	]
	// ......
})
```

And run `npm run build`.

You can see that the `stats.html` file has been created and can be viewed directly by opening it in the browser

ShirasawaSama:
On the other hand, I think it would be better if the language supports dynamic import?

To me, I think default to loading language support for JavaScript, TypeScript, Python, HTML, and CSS is necessary.

rgaricano:
OK,
& also Elixir, HCL and Mermaid

rgaricano:
it's duplicated because langs also are loaded by lowlight in RichtTextInput.
2 options for unify:
- Configure Vite to bundle all highlight.js languages into a single chunk (reasonable bundle size?)
- Lazy load languages for lowlight.

I'll implement a dynamic languages lazy for lowlight to use same as defined for codeblock, then we'll have all  syntax languages highligh definitions in hightlight.js

ShirasawaSama:
Hello, is there any recent progress on this matter? If you need any help, please feel free to let me know.

If you have any new ideas, feel free to share them with me anytime.

I plan to improve the initial page load speed for OpenWebUI. The next optimization phase will rely on highlight.js, so your pull request needs to be merged first.

In version 0.6.28, the JavaScript load size on the homepage has already been reduced by 4MB, but this is still far from sufficient.

rgaricano:
(just some hardware & other personal issues...)
Heard in the kitchen!. I'll get right on it.

ShirasawaSama:
> (just some hardware & other personal issues...) Heard in the kitchen!. I'll get right on it.

Thank you ever so much for your efforts!

rgaricano:
@ShirasawaSama,
I think that all is fixed.
i had to add HCL language definition to HightlightJSExtension.js to mantain actual languages supported.

ShirasawaSama:
Great job! I will have a code review.

image

package.json
```diff
@@ -128,6 +126,7 @@
 		"prosemirror-tables": "^1.7.1",
 		"prosemirror-view": "^1.34.3",
 		"pyodide": "^0.28.2",
+		"rollup-plugin-visualizer": "^6.0.3",
```
ShirasawaSama: I suggest removing this dependency before merging

package.json
```diff
@@ -138,6 +137,7 @@
 		"uuid": "^9.0.1",
 		"vite-plugin-static-copy": "^2.2.0",
 		"y-prosemirror": "^1.3.7",
+		"y-protocols":"^1.0.6",
```
ShirasawaSama: It seems that this dependency was not used in this code? Can it be removed?

src/lib/components/common/RichTextInput.svelte
```diff
@@ -152,15 +152,44 @@
...
+	lowlight.register('css', css);
+	lowlight.register('elixir', elixir);
+
+	const loadLanguageForLowlight = async (lang) => {
+	  if (lowlight.registered(lang)) {
+    	return true;
+	  }
+
+	  try {
+    	const languageModule = await import(`highlight.js/lib/languages/${lang}`);
```
ShirasawaSama: Dynamic import seems to be ineffective because all language dependencies have already been introduced in `highlight.js/lib/index.js`

image

From the results, it can also be seen that the parsers for each language were not individually packaged into small files:

image

I think it may not be possible to directly import the 'index.js' file or it can only be imported from the 'highlight.js/lib/core'.

ShirasawaSama:
@rgaricano After replacing `import hljs from ‘highlight.js’;` with `import hljs from ‘highlight.js/lib/core’;` in the following two files:

`src/lib/components/chat/Messages/CodeBlock.svelte` and `src/lib/utils/index.ts`

I was able to achieve dynamic language support.

Therefore, I recommend packaging the highlight.js files in these locations into a separate file, then replacing the `import hljs from ‘highlight.js’;` imports in these three places with the packaged file.

package.json
rgaricano: Yes, but when I build the code always have errors with it, I made a PR about:
https://github.com/open-webui/open-webui/pull/16677
It seem that it's a y-prosemirror dependent: https://github.com/open-webui/open-webui/pull/16677#issuecomment-3197929163

I don't know why the error on my end ??

I'll remove both (visualizer and y-protocols)

ShirasawaSama: I recommend trying to directly delete `node_modules` and then rerunning `npm install -f`. This should resolve 99% of the issues. 😂

src/lib/components/common/RichTextInput.svelte
rgaricano: I'm trying with a separate language loader.
openwebui_hightlight_stats0

ShirasawaSama: > I'm trying with a separate language loader.

When you can't see a large block of highlight.js dependency in stats.html, it means it's successful. 👍

rgaricano:
Yes, it what i did, I create a hightlightLanguageLoader.js for help.

now seem that it work dynamically.

src/lib/components/common/RichTextInput.svelte
```diff
@@ -983,7 +993,7 @@
 				StarterKit.configure({
 					link: link
 				}),
-				Placeholder.configure({ placeholder: () => _placeholder }),
+				Placeholder.configure({ placeholder }),
```
ShirasawaSama: This commit appears to contain other commits from the dev branch and may not have been rebased correctly.

We should not use `merge` to incorporate changes from the latest dev branch. Instead, use `git rebase dev` or `git rebase upstream/dev` (if your upstream source is the original repository).

rgaricano: I'll revise it, but I think that when I finish the tests I'm going to create a new clean PR.

ShirasawaSama:
Could you please squash all your commits into a single commit (`git rebase -i HEAD~20` then modify `pick` to `squash`) and then rebase onto the latest dev branch (`git rebase upstream/dev` and `git push -f`)? I'd like to make some code changes on your branch and then submit a pull request to your repository. Thanks!

rgaricano:
@ShirasawaSama,
I did the revase but i didn't correctly the squash, the commits isn't in a single one.

ShirasawaSama:
> @ShirasawaSama, I did the revase but i didn't correctly the squash, the commits isn't in a single one.

The simplest way to squash: https://stackoverflow.com/a/61171280

rgaricano:
https://github.com/open-webui/open-webui/pull/17250/commits/5be31f3a007932a6ecfc1f68f1f455ce4ba57149

ShirasawaSama:
@rgaricano I noticed that the file src/lib/components/chat/Messages/CodeBlock.svelte contains over 1,400 lines of new additions. Is this file required to be included in this PR?

It seems like the code block contains an entire editor, along with features for multi-user collaboration. I think this might be a bit problematic.

> Of course, I think the reason for my confusion is simply that the number of lines changed appears excessive. I'm unsure whether this section includes modifications from other branches or if it must necessarily include them.

In my personal opinion, the current changes may have exceeded the principle of minimal changes, and Tim might not merge this PR.

ShirasawaSama:
QQ_1758279657001

I suggest you can put this task on hold for now. I'll provide a brand-new PR within the next couple of days. After that, you can use my code to directly create a new PR.

rgaricano:
No, in that file must to be only one line changed, I revised and seem that I make something wrong when rebasing.

Now it seem that all is ok.
(I think that package.json & package-lock.json are ok, but as I have to edit it to add "y-protocols":"^1.0.6" for build and then remove for pull,...,  only this lib have to be removed:
- "@codemirror/lang-javascript": "^6.2.2",  
- "@codemirror/lang-python": "^6.1.6",  
- "@codemirror/language-data": "^6.5.1",  
- "codemirror-lang-elixir": "^4.0.0",  
- "codemirror-lang-hcl": "^0.1.0",
)

ShirasawaSama:
I've reworked it based on my approach. What do you think of this version? Could you share your thoughts or suggestions?

https://github.com/ShirasawaSama/open-webui/commit/121837b1c6dfeb420e1f3e777d827ab9bf055645

My modification enables dynamic language loading for code blocks in the tiptap editor, eliminating the limitation to a fixed set of languages.

Since CodeMirror's language support already supports dynamic loading, I left it unchanged.

 Additionally, I fixed an issue where Vite did not support importing files from node_modules using wildcards.

The only drawback is that the author of the lowlight library hard-coded the creation of the hljs instance. I think a better approach might be for me to submit a pull request for the lowlight library as well.

rgaricano:
I think that isn't a good idea hardcode the highlight languages to import, if its are changed it's necessary to change the code (& maybe some users use other lib versions in its devs).
I don't see any problem in the warning.

ShirasawaSama:
> I think that isn't a good idea hardcode the highlight languages to import, if its are changed it's necessary to change the code (& maybe some users use other lib versions in its devs). I don't see any problem in the warning.

When this warning appears, attempts to load language support will fail immediately. You can also inspect the bundled output and find that no language support was included at all—it was simply ignored.

In fact, CodeMirror's dynamic language loading is hard-coded as multiple dynamic import statements.

Alternatively, we could explore whether any existing Vite plugins are available.

ShirasawaSama:
https://github.com/wooorm/lowlight/blob/0f36148072cd096ca86753d6f1ff01589d30d78f/lib/index.js#L94

It appears to break other hljs instances. Therefore, we must also maintain a separate lowlight language list.

ShirasawaSama:
```js
import { defineConfig, type Plugin } from 'vite';
import path from 'path';
import fs from 'fs/promises';
const HighlightJSLanguagesPlugin = (): Plugin => {
	const virtualModuleId = 'virtual:highlightjs-languages';
	const resolvedVirtualModuleId = '\0' + virtualModuleId;

	return {
		name: 'highlightjs-languages-plugin',
		resolveId(id) {
			if (id === virtualModuleId) {
				return resolvedVirtualModuleId;
			}
		},
		async load(id) {
			if (id !== resolvedVirtualModuleId) {
				return;
			}
			const languagesPath = path.join(import.meta.resolve('highlight.js').replace('file:', ''), '../languages');
			const languages = (await fs.readdir(languagesPath)).filter(file => !file.endsWith('.js.js')).map(file => file.replace(/\.js$/, ''));

			const { default: hljs } = await import('highlight.js');
			const generateLanguageConfig = (language: string) => `() => import(${JSON.stringify(path.join(languagesPath, language))})`;
			const kvPairs = languages.map(language => [language, generateLanguageConfig(language)]);
			languages.forEach(language => {
				const config = hljs.getLanguage(language);
				if (config?.aliases?.length) {
					config.aliases.forEach(alias => kvPairs.push([alias, generateLanguageConfig(language)]));
				}
			});

			return `export default { ${kvPairs.map(([key, value]) => `'${key}': ${value}`).join(', ')} }`
		},
	}
}
```

QQ_1758350972656

ShirasawaSama:
> I think that isn't a good idea hardcode the highlight languages to import, if its are changed it's necessary to change the code (& maybe some users use other lib versions in its devs). I don't see any problem in the warning.

QQ_1758351163244

ShirasawaSama:
https://github.com/ShirasawaSama/open-webui/commit/069f58d40d64d1956b05d801cd2bf2e118bd9ed2

PR #17264, opened 2025-09-07
UPD: Add Chat Tittle to navbar

rgaricano:
### UPD: Add Chat Tittle to navbar

Sometimes, when we are in a chat is difficult to now in what we are, there isn't hightlight or focus, just the chat_id on address url. Easier to know were we are with chat tittle visible.

I set it with small font because is less disturbing on mobile.

With PR:
openwebui_chat_title0

mobile
openwebui_chat_title1
____


tjbck:
Webpage title should be used instead (current behaviour)

rgaricano:
Ohhh I don't see it on mobile, on PC with Chromium neither, it's too crowded!
Also when other instance of open-webui is embedded it's impossible to see it.

PR #17265, opened 2025-09-07
UPT:  iframe renderer - HTMLToken.svelte

rgaricano:
### UPD: Improvement of iframe renderer

Added 
- detection of src & srcdoc parameters and parsing atributes
- detection of multiples iframes in message
- autoresize of the iframe to adjust height.
- same security that for artifacts.

(first test failed is normal, is due to the way the code is structured and executed. When the script runs, it's trying to display the results of the security checks, but there's a timing issue with how the HTML is being parsed and displayed.)

openwebui_iframe_embedd_security

____

### DIFF
--- a/src/lib/components/chat/Messages/Markdown/HTMLToken.svelte
+++ b/src/lib/components/chat/Messages/Markdown/HTMLToken.svelte
@@ -18,6 +18,59 @@
 	} else {
 		html = null;
 	}
+
+    //iframes detection
+    let iframes = [];
+    let iframeEls = [];
+
+    $: {
+        iframes = [];
+        if (token?.text) {
+            const tempDiv = document.createElement('div');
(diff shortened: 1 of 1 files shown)


ShirasawaSama:
Can current iframes prevent XSS attacks? It seems that if the executed code contains fetch absolute paths (such as `/app/v1/models`), it will bring cookies along and request them.

rgaricano:
Is the same as for artifacts.
& Endpoints need authentification. 
Of course, as always, a bad configuration opens doors, e.g. WEBUI_AUTH=false or iframeSandboxAllowSameOrigin=true.

Note, 
I made the height dynamically to embedd open-webui instances, but I think the width and height should also be controlled/limited for the same attributes, if indicated in the tag (I made a version but lost it yesterday...sdd RIP)

PR #17410, opened 2025-09-12
feat: Private sharing permissions for Workspace entities

Simon-Stone:
##
# Description

This PR introduces permissions for "Private Sharing" of Workspace entities. If Private Sharing is allowed for a Group, a member of that Group can share their creations with this Group.  "Private Sharing" can be toggled independently for each Workspace component.

## Documentation
Documentation has not yet been updated.

## Dependencies
This PR introduces no new external dependencies. 

## Testing
I have not written any tests but have validated the intended behavior manually.

## Code review
I have self-reviewed my code, and tried to adhere to existing style and coding standards

# Changelog Entry

### Description

This PR addresses the points mentioned in discussion #15100 by introducing the concept of "Private Sharing" permissions: A Group now has an additional set of permissions that govern whether members of this Group can share Workspace creations with the Group privately. 

With this feature, the following scenarios can be avoided by deactivating private sharing permissions (from the discussion linked above):

> **Universities:** Students in a shared "Students" group can spam each other's interfaces by sharing personal models, uploading irrelevant knowledge bases, creating joke prompts, or publishing buggy tools
>
> **Large organizations:** Employees can flood departmental groups with personal workspace items, making it difficult for colleagues to find legitimate shared resources
>
> **Multi-tenant environments:** Users can clutter shared spaces with experimental or low-quality content that affects the entire group's productivity

### Added

- A set of permissions on Groups that govern privately sharing Workspace entities within the group

### Changed

- Added a prop `allowedPrivate` to `AccessControl`: This array holds the IDs of all groups that should appear in the group selector when sharing is set to Private.
- Added the same prop to `AccessControlModal` (a little bit of prop drilling, but follows existing patterns)
- Initialized `allowedPrivate` in all components that use an `AccessControl` or `AccessControlModal`:
   - If the component is part of one of the Workspace entities, use the new permissions to find the appropriate groups for private sharing (or all if admin)
   - If the component is not one of the Workspace entities (e.g. non-direct Tool Server connections, Notes, ...), allow all groups for private sharing. 

---
### Screenshots or Videos

The new options in the Permissions dialog:

permissions dialog

### DIFF
--- a/src/lib/components/AddServerModal.svelte
+++ b/src/lib/components/AddServerModal.svelte
@@ -14,6 +14,7 @@
 	import Tags from './common/Tags.svelte';
 	import { getToolServerData } from '$lib/apis';
 	import { verifyToolServerConnection } from '$lib/apis/configs';
+	import { getGroups } from '$lib/apis/groups';
 	import AccessControl from './workspace/common/AccessControl.svelte';
 	import Spinner from '$lib/components/common/Spinner.svelte';
 	import XMark from '$lib/components/icons/XMark.svelte';
@@ -34,6 +35,7 @@
 	let key = '';
 
 	let accessControl = {};
+	let allowedPrivate = [];
 
 	let id = '';
 	let name = '';
@@ -153,8 +155,13 @@
 		init();
 	}
 
-	onMount(() => {
+	onMount(async () => {
 		init();
+
+		let groups = await getGroups(localStorage.token);
+
+		// There is no setting governing private sharing this entity, so allow all groups
+		allowedPrivate = groups.map((group) => group.id);
 	});
 </script>
 
@@ -397,7 +404,7 @@
 
 							<div class="my-2 -mx-2">
 								<div class="px-3 py-2 bg-gray-50 dark:bg-gray-950 rounded-lg">
-									<AccessControl bind:accessControl />
+									<AccessControl bind:accessControl  {allowedPrivate}/>
 								</div>
 							</div>
 						{/if}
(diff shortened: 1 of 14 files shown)


ricdikulous:
This would be a great addition!

Classic298:
love it, but would perhaps prefer a different naming of the permission

Simon-Stone:
> love it, but would perhaps prefer a different naming of the permission

Thanks! What would be a better naming scheme?

Classic298:
Perhaps:

"Share Models with Groups"
"Share Prompts with Groups"
etc.

makes it very clear what this is about
also defines the scope more clearly.

If in the future (and it will happen someday) some form of direct 1-on-1 sharing gets added, then this permission name will be confusing because it reads as if you want to disable all forms of private sharing

Simon-Stone:
> Perhaps:
> 
> "Share Models with Groups" "Share Prompts with Groups" etc.
> 
> makes it very clear what this is about also defines the scope more clearly.
> 
> If in the future (and it will happen someday) some form of direct 1-on-1 sharing gets added, then this permission name will be confusing because it reads as if you want to disable all forms of private sharing

I see what you're saying. I like the dichotomy of Public vs Private, though, and I wonder if the more verbose version would be better placed in a tooltip. That could say a full sentence like "Allow members to share Models with the group" and so on.

ricdikulous:
Love your work here @Simon-Stone! When it comes to the naming I tend to agree with @Classic298 - specific is good even if it is a bit more wordy and less elegant.

Either way this is incredibly useful functionality I would love to see it in the product

Simon-Stone:
> Love your work here @Simon-Stone! When it comes to the naming I tend to agree with @Classic298 - specific is good even if it is a bit more wordy and less elegant.
> 
> Either way this is incredibly useful functionality I would love to see it in the product

Thanks! Valid point. I will let @tjbck decide whether this is worthy of inclusion before making any further changes, though.

tjbck:
#18888

PR #17457, opened 2025-09-15
feat: Study mode 

Simon-Stone:
# Study Mode

# Changelog Entry

### Description

This PR introduces Study Mode, which is inspired by [OpenAI's Study Mode](https://openai.com/index/chatgpt-study-mode/) and [Anthropic's Learning Mode](https://www.anthropic.com/news/introducing-claude-for-education).

Discussion in #17424 

When Study Mode is active, a system prompt is injected that instructs the LLM to act like a tutor/instructor, who walks the user through the thought process of discovering the solution to their problem instead of offering up a straight answer.

See this example:

screenshot of study mode

 
### Added

- Study Mode config parameters
- Env variable to set the default system prompt for Study Mode
- Settings tab to enable Study Mode globally and set a custom prompt
- Button in message input to activate/deactivate Study Mode
- TODO: User- and group-level permission for Study Mode (similar to Code Interpreter)

---

### Additional Information

- If Study Mode is active, the Study Mode system prompt is used as the chat-level system prompt and takes precedence over the user settings system prompt as normal.
    - If the chat-level system prompt is changed in the Chat Controls, Study Mode is deactivated.
- The implementation of Study Mode is modeled after the implementation of Code Interpreter (regarding the button behavior) and the task model prompts (regarding defining the default prompt and setting a admin-defined one)
 


rgaricano:
I still don't understand why you don't just create a studio mode model in the workspace.

Simon-Stone:
> I still don't understand why you don't just create a studio mode model in the workspace.

Because a Workspace Model will only use a single base model. If I want the user to be able to choose between different LLMs in Study Mode, it would require a separate Workspace Model for each base model.

Having a button to turn on this mode is also a nicer UX, in my opinion.

tjbck:
Should be implemented as a Function.

https://docs.openwebui.com/features/plugin/functions/filter#--toggle-filter-example-adding-interactivity-and-icons-new-in-open-webui-0610

Simon-Stone:
I had no idea this existed. Amazing!

dmanresa-saes:
Could this be the prompt for study mode? Would it be the same in Spanish? Anyone interested in creating a filter with this to see how it works? [https://chatgpt.com/c/68d0fe39-1b14-832c-8b22-e1dab82832ed](https://chatgpt.com/c/68d0fe39-1b14-832c-8b22-e1dab82832ed)

rgaricano:
@dmanresa-saes

Dario, I can't see the linked conversation (chatgt ask me for a superior plan), but
Yes, if the model used have enought knowledge about languages any prompt in english can be used for spanish.

I made a filter function for do that.
- Study Mode managed by admin that can enable it and set the prompt (valves), also can permit for user change mode or not.
- User can change mode, if admin not restrict it, throught Chat Control, valves, functions, Study_Mode,...


@Simon-Stone,
I changed the prompt, yours (from PR) entered in loops:
```
Yes, the injected prompt directly relates to the issue you were experiencing. The repeated lines you saw were a result of the system's attempt to follow these strict rules and guidelines, which included multiple references to the importance of following these rules. This caused a loop where the system kept processing the same information repeatedly, leading to the repetitive behavior you observed.

The prompt you shared includes detailed guidelines on how to interact with the user, including the specific rule not to give answers directly, but instead to guide the user through the problem-solving process one step at a time. This aligns with the issue you were facing where the system was stuck in a loop of repeating the same information.
```

XXX

Attached function:
[Filter_Function_StudyMode.py](https://github.com/user-attachments/files/22463893/Filter_Function_StudyMode.py)

i-iooi-i:
@rgaricano Could a toggle switch be integrated into the chat interface? I've activated the function, but I can't seem to find any corresponding switch options within the chat window.

Classic298:
Have you activated the function too? Globally?

rgaricano:
or activate the function in the config of the used model

i-iooi-i:
https://openwebui.com/t/magro07/learn_study
Found a great tool, tried it out, the experience was quite good.

PR #17516, opened 2025-09-17
fix: handle None usage data in chat response processing

NibbinNone:
Update usage data handling to use 'or {}' instead of default parameter to properly handle None values from API responses.

some api server return data like 

```
data: {"id": "chatcmpl-5jR1Fhd9KUk4P6N3zWo2y4coMscGu", "object": "chat.completion.chunk", "created": 1758113545, "model": "claude-neptune-v3", "choices": [{"index": 0, "delta": {"role": "assistant", "content": "3会更精确。"}, "logprobs": null, "finish_reason": null}], "usage": null, "system_fingerprint": "fp_d576307f90"}
```

The code `usage = data.get("usage", {})` gets None, which causes the open-webui to not display the message correctly.

Screenshot_20250917_211643_ B@95c440d

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -2033,7 +2033,7 @@ async def process_chat_response(
                                     choices = data.get("choices", [])
 
                                     # 17421
-                                    usage = data.get("usage", {})
+                                    usage = data.get("usage") or {}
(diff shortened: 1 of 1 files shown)


tjbck:
0cfc8f52561e474d0204cb3dcd3b017e821de277

PR #17528, opened 2025-09-17
0.6.30

tjbck:


anengineerdude:
Quick fix much appreciated @tjbck!  I was banging my head with v0.6.29 crashing earlier today in my dev env... started entering a issue, only to find out you sneaked out .30 tonight!

PR #17550, opened 2025-09-18
feat: support per-model RAG template override

athoik:
# Pull Request Checklist

# Changelog Entry

Add `rag_template` as a configurable parameter in model settings, allowing each model to define its own RAG template instead of always using the global default.
- Middleware now selects the model-specific RAG template if provided.
- Model editor UI updated to allow editing and saving `rag_template`.
- Fallback to global `RAG_TEMPLATE` remains when no override is set.

### Added

-  Support per-model RAG template override

### Breaking Changes

- **BREAKING CHANGE**: NONE

---

### Screenshots or Videos

By default the model screen will have an empty template and global rag template will be used.

image

We can override the default by setting our customized rag template.

image

Here is an example rag template, based on default, without citations.

```
### Task:
Respond to the user query using the provided context.

### Guidelines:
- If you don't know the answer, clearly state that.
- If uncertain, ask the user for clarification.
- Respond in the same language as the user's query.
- If the context is unreadable or of poor quality, inform the user and provide the best possible answer.
- If the answer isn't present in the context but you possess the knowledge, explain this to the user and provide the answer using your own understanding.
- Do not cite any <source> tag
- Do not use XML tags in your response.

### Output:
Provide a clear and direct response to the user's query.

<context>
{{CONTEXT}}
</context>

<user_query>
{{QUERY}}
</user_query>
```

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -1170,10 +1170,15 @@ async def process_chat_payload(request, form_data, user, metadata, model):
         if prompt is None:
             raise Exception("No user message found")
 
+        model_rag_template = (
+            model.get("info", {}).get("params", {}).get("rag_template", "")
+            or request.app.state.config.RAG_TEMPLATE
+        )
+
         if context_string != "":
             form_data["messages"] = add_or_update_user_message(
                 rag_template(
-                    request.app.state.config.RAG_TEMPLATE,
+                    model_rag_template,
                     context_string,
                     prompt,
                 ),
--- a/src/lib/components/workspace/Models/ModelEditor.svelte
+++ b/src/lib/components/workspace/Models/ModelEditor.svelte
@@ -63,6 +63,7 @@
 	}
 
 	let system = '';
+	let rag_template = '';
 	let info = {
 		id: '',
 		base_model_id: null,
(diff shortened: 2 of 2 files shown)


ghost:
hi, Is this pr still open

athoik:
Hi, 

PR is open, but most probably a Rebase / rework against current develop is required.

Classic298:
Closing this for now; please open a discussion outlining why this is needed. I have thought about this for a bit but I don't really see when/how this is needed or ever has been needed. When do you need a different rag template for some models versus other models; so as to, under what circumstances? And wouldn't the system prompt help you here with that?

PS: we actively recommend AGAINST using file_context today, which means, we actively recommend against using the RAG Template at all anymore because it is cache breaking.

And one more question/reason: the RAG template is basically a prompt injection into the user message that tells the model to not say something if the sources don't clearly state it. The same thing CAN be put in the system prompt and it would work. Why is it of need to inject it into the user message (and break cache while doing so) - and the other part, which is citations, can also be solved in the system prompt by having the model e.g. use Markdown to cite inline source links to the citations or reference the files directly (we also have new tools for that).

So generally a discussion would be wanted here with the use cases explained.
**IF the use case is related to legacy tool calling** it is extremely unlikely to ever get merged/added in, as the legacy tool calling path has been unsupported since 0.7.0 and we only keep it running (with low prio) and do not want to add new features to that path.

Closing it for the reasons above and also because this has merge conflicts. Thanks!

athoik:
It is related with "legacy" RAG path.

I do understand that the new "agentic" way is now only actively supported, as long as legacy stays I am fine with it.

I will try to keep legacy path just working, without introducing any new stuff, from now on.

PR #17617, opened 2025-09-20
FIX: hybrid_search-instances_checker

rgaricano:
### FIX: hybrid_search-fails  due to inconsistence in parameters type received in some circunstances.

In some circunstances hybrid search fail, as it's exposed in issue https://github.com/open-webui/open-webui/issues/17046

- query_doc_with_hybrid_search function expects collection_result to be an object with a documents attribute, but it's receiving a list or None instead (e.g. qdrant return None if collections don't exist, which could cause the issue)

- when query is as list (e.g. query_doc_with_hybrid_search function called for queries)

This PR add a couple of checkers for that.

____

### DIFF
--- a/backend/open_webui/retrieval/utils.py
+++ b/backend/open_webui/retrieval/utils.py
@@ -127,11 +127,19 @@ def query_doc_with_hybrid_search(
     hybrid_bm25_weight: float,
 ) -> dict:
     try:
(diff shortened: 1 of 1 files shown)


tjbck:
Closing in favour of 6e4a2f18e17f47c9d0f3b50822bfaae3a5638864, Thanks!

PR #17650, opened 2025-09-22
fix: pin datasets==3.6.0 in backend/requirements.txt

colpari:
Thix fixes the problem described in https://github.com/open-webui/open-webui/discussions/16301

### Description

- Adding datasets==3.6.0 in backend/requirements.txt because the newer versions have breaking changes which result into Text-to-Speech not working.  #16301 

### Added

- `datasets==3.6.0` in backend/requirements.txt

### Fixed

- Text-to-speech default configuration works again

### DIFF
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -49,6 +49,8 @@ tiktoken
 langchain==0.3.26
 langchain-community==0.3.27
 
+datasets==3.6.0
+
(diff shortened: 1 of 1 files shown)


tjbck:
This would cause more compatibility issues with other models, it would be prudent to raise an issue on their issue instead.

PR #17722, opened 2025-09-24
Fix: Hide hidden models in notes section

Classic298:
This commit fixes a bug where hidden models were still visible and selectable in the notes section. The following changes were made:

- The model dropdown in the chat section of the note editor is now filtered to exclude hidden models.
- The default model selection logic in the note editor is now filtered to exclude hidden models.


tjbck:
LGTM, Thanks!

PR #17723, opened 2025-09-24
Fix: Sanitize chat and other user content to prevent PostgreSQL errors

Classic298:
This commit introduces a comprehensive sanitization mechanism to prevent null bytes (\u0000) and other C0 control characters from being stored in the database. These characters were causing `UntranslatableCharacter` errors during search operations in PostgreSQL deployments.

The solution consists of a new centralized function, `sanitize_json_content`, which recursively traverses data structures and removes problematic characters from strings.

This sanitization is now applied at all relevant data entry points, including:
- Chat creation and updates
- Prompt creation and updates
- File upload metadata
- Content extracted from documents for retrieval

This proactive approach ensures data integrity and prevents future occurrences of the search error, providing a robust and long-term fix for the issue.


rgaricano:
I don't think that sanitazion the json is the best option, yes, it can solve the problem and maybe usefull for title and other IA generated text, but it can change the sense of the information of other data provided.
The de/encryption json strings & binary storage solution seem most robust for preserving data.

Classic298:
hm how it could change the sense of the information?

I am basically only substituting null strings and other ones with no real information value

rgaricano:
Ok, but then, what happend with other possible untraslatedCharacters? persists the issue?

Classic298:
what others?

Classic298:
will redo it due to major merge conflicts...

PR #17744, opened 2025-09-25
Fix: Prevent RAG queries when all files are in full context

Classic298:
This commit fixes an issue where Retrieval-Augmented Generation (RAG) queries were still being generated even when all attached files were set to 'full context' mode. This was inefficient as the full content of the files was already available to the model.

The `chat_completion_files_handler` in `backend/open_webui/utils/middleware.py` has been updated to:
- Check if all attached files have the `context: 'full'` property.
- Skip the `generate_queries` step if all files are in full context mode.
- Pass a `full_context=True` flag to the `get_sources_from_items` function to ensure it fetches the entire document content instead of performing a vector search.

This change ensures that RAG queries are only generated when necessary, improving the efficiency of the system.

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -631,48 +631,53 @@ async def chat_completion_files_handler(
     sources = []
 
     if files := body.get("metadata", {}).get("files", None):
+        # Check if all files are in full context mode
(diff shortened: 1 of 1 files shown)


tjbck:
LGTM, Thanks!

Issue #17088, opened 2025-09-01
issue: milvus error because the limit set to None by default


aaronchenlei:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

v0.6.26

### Operating System

ubuntu 22.04

### Browser (if applicable)

EDGE

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

I use milvus as the vector database, but I met the following exception, it seems because the limit is set to None in backend/open_webui/retrieval/vector/dbs/milvus.py", line 252:  

File "/home/jovyan/dev/open-webui-0.6.26/backend/open_webui/retrieval/utils.py", line 647, in get_sources_from_items
    query_result = query_collection_with_hybrid_search(
                   └ <function query_collection_with_hybrid_search at 0x7fcf926245e0>

  File "/home/jovyan/dev/open-webui-0.6.26/backend/open_webui/retrieval/utils.py", line 353, in query_collection_with_hybrid_search
    collection_results[collection_name] = VECTOR_DB_CLIENT.get(
    │                  │                  │                └ <function MilvusClient.get at 0x7fcfd7e62480>
    │                  │                  └ <open_webui.retrieval.vector.dbs.milvus.MilvusClient object at 0x7fcfe4d4d150>
    │                  └ '98d31974-93f5-4992-a3f6-4a41f6f77847'
    └ {}

[...]

PR #17769, opened 2025-09-26
Fix: milvus error because the limit set to None by default

Classic298:
The pymilvus library expects -1 for unlimited queries, but the code was passing None, which caused a TypeError. This commit changes the default value of the limit parameter in the query method from None to -1. It also updates the call site in the get method to pass -1 instead of None and updates the type hint and a comment to reflect this change.

Related: https://github.com/open-webui/open-webui/issues/17088


colabbear:
good

tjbck:
Thanks!

PR #17843, opened 2025-09-28
feat/chore: Clone default group permissions for new groups and remove redundant modal

Classic298:
FEAT: When creating a new user group, the permissions for the new group will now be pre-populated with the same permissions as the "default user group".

This is achieved by removing the separate `AddGroupModal` and instead using the `EditGroupModal` for both creating and editing groups.

CHORE: **Both modals were extremely similar anyways - this PR removes the redundant AddGroupModal and instead uses the EditGroupModal for all actions going forward**

When creating a new group, the `EditGroupModal` is now pre-populated with the default permissions, saving administrators from having to manually configure them each time.

This change **simplifies the codebase** by removing a redundant component and directly addresses the user's request to streamline the group creation process.


Classic298:
tested; works

Classic298:
Discussed here: https://github.com/open-webui/open-webui/discussions/16793

tjbck:
Thanks!

PR #17908, opened 2025-09-30
enh: add filters to embedding models

Simon-Stone:
# Filter function application for embedding models

Currently, embedding requests do not pass through Filter functions, even though they can be attached to embedding models in the Models settings.

This PR adds the application of filters to embedding models.

An example use case is the injection of additional parameters for upstream usage tracking (e.g. with LiteLLM).


tjbck:
1. CLA required. 2. We'll consider this suggestion, Thanks!

Simon-Stone:
Thanks for considering it!

PR #17976, opened 2025-10-01
FIX: N+1 database query in feedbacks

rgaricano:
### FIX: N+1 database query in feedbacks

Fix for solve the slowness when retrieving feedbacks, due to a N+1 database query problem in the implementation of evaluations endpoint.

Modify to use a single optimized query that joins feedback and user data.

____

### DIFF
--- a/backend/open_webui/models/feedbacks.py
+++ b/backend/open_webui/models/feedbacks.py
@@ -266,10 +266,7 @@ class FeedbackTable:
 
             result = []
             for feedback, user in query:
-                result.append({
-                    "feedback": feedback,
-                    "user": user
-                })
+                result.append({"feedback": feedback, "user": user})
 
             return result
         except Exception as e:
--- a/backend/open_webui/models/feedbacks.py
+++ b/backend/open_webui/models/feedbacks.py
@@ -5,6 +5,7 @@ from typing import Optional
 
 from open_webui.internal.db import Base, get_db
 from open_webui.models.chats import Chats
+from open_webui.models.users import User
 
 from open_webui.env import SRC_LOG_LEVELS
 from pydantic import BaseModel, ConfigDict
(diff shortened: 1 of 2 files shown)


YetheSamartaka:
Awesome work, but maybe I would also add another route to get all_feedbacks_ids which would return simplified list, where instead of returning the whole feedback, it would return just its id. Because when you have external system that would fetch these, right now, it fetches all the data and it is really slow. With simplified variant, it would just return those ids and then you could fetch individual feedbacks by their id. At top of that, providing a param "count" which would return just X amount of last feedbacks could further improve the performance and reduce the network strain for the use-cases.

rgaricano:
@YetheSamartaka 
A simple way, adding params,  would it work? :
```
@router.get("/feedbacks/all", response_model=list[FeedbackUserResponse] | list[str])
async def get_all_feedbacks(
    ids_only: bool = False,
    count: Optional[int] = None,
    user=Depends(get_admin_user)
):
    feedbacks = Feedbacks.get_all_feedbacks()

    if ids_only:
        # Return only IDs
        feedback_ids = [feedback.id for feedback in feedbacks]
        if count:
            return feedback_ids[:count]
        return feedback_ids

    # Existing full response logic
    if count:
        feedbacks = feedbacks[:count]

    feedback_list = []
    for feedback in feedbacks:
        user = Users.get_user_by_id(feedback.user_id)
        feedback_list.append(
            FeedbackUserResponse(
                **feedback.model_dump(),
                user=UserResponse(**user.model_dump()) if user else None,
            )
        )
    return feedback_list

YetheSamartaka:
@rgaricano Yes, absolutely. Thank you very much 👍🏻

rgaricano:
PR modified,
Added optional params to router /feedbacks/all:
- `ids_only`   # default False, if set to True only feedbacks ids are returned
- `count`        # default None, if set with a x int number only last x feedbacks are returned

Note:
Since the use of these optional parameters is geared towards non-Frontend requests, no changes are made to https://github.com/open-webui/open-webui/blob/53e98620bff38ab9280aee5165af0a704bdd99b9/src/lib/apis/evaluations/index.ts#L65

YetheSamartaka:
@rgaricano Perfect 👍🏻. Thank you very much.

tjbck:
Closing in favour of 60839606556922a4f565f405752e20ff400e24e3, Thanks!

YetheSamartaka:
Hi @rgaricano. Could you open a new PR where you would adjust the endpoint with /api/v1/evaluations/feedbacks/all so we can omitt it. The PR that was merged in favour does not have this and fetching via API is still awfully slow since it is getting all the feedbacks including all the chat history with it and we would need just feedback IDs..

rgaricano:
@YetheSamartaka,
sorry, i'm not on this & I didn't review/tested the refac finally committed... feel free to PR it!

PR #18008, opened 2025-10-02
feat: add dynamic enum support for UserValves/Valves with API-driven dropdowns

Simon-Stone:
# feat: Add dynamic enum support for UserValves/Valves with API-driven dropdowns

### Description

This pull request introduces dynamic enum population for UserValves and Valves in Functions and Tools. Instead of hardcoding dropdown options, developers can now specify API endpoints that are queried at runtime with user authentication, allowing for user-specific, permission-based dropdown options.

This enhancement solves a critical limitation where dropdown options couldn't be personalized based on the authenticated user's context. For example, a UserValve for "preferred model" can now show only the models the specific user has access to, rather than all models in the system.

**Key Benefits:**
- User-specific dropdown options based on permissions
- Real-time data from API endpoints
- Smart caching with configurable TTL
- Manual refresh capability for testing
- Zero breaking changes to existing code

### Added

- **Dynamic enum configuration via `json_schema_extra`**: Pydantic Field definitions now support a `dynamic_enum` parameter that tells the frontend which API endpoint to query
- **Frontend dynamic enum fetching in `Valves.svelte`**: 
  - Automatic detection and initialization of dynamic enum fields
  - HTTP requests with Bearer token authentication
  - Response path extraction for nested data structures
  - Per-endpoint caching with configurable TTL (default 5 minutes)
  - Request deduplication to prevent simultaneous duplicate calls
- **Refresh button UI**: Manual cache invalidation button next to each dynamic enum dropdown for testing and development

### Changed

- **`src/lib/components/common/Valves.svelte`**: Extended to detect and render dynamic enum fields with API-driven options
- Enhanced valve rendering logic to support three dropdown types: static enum, dynamic enum, and boolean switches

### Deprecated

- None

### Removed

- None

### Fixed

- None (this is a new feature)

### Security

- All API requests include user Bearer token authentication
- Endpoints validate user permissions server-side
- Frontend sanitizes API responses before rendering
- No CORS issues (uses existing WEBUI_API_BASE_URL)

### Breaking Changes

- None. This is a fully backward-compatible opt-in feature. Existing Valves/UserValves continue working unchanged.

---

### Additional Information

**Configuration Example:**

```python

class UserValves(BaseModel):
    preferred_model: str = Field(
        default="",
        description="Select your preferred chat model",
        json_schema_extra={
            "dynamic_enum": {
                "endpoint": "/models",
                "response_path": "data",  # Extract array from nested response
                "value_field": "id",
                "label_field": "name",
                "cache_ttl": 300,  # Cache for 5 minutes
                "fallback": []     # Fallback options if fetch fails
            }
        }
    )
```

**Technical Implementation:**
- Pydantic extracts `json_schema_extra` fields to the top level of the JSON schema, making `dynamic_enum` directly accessible on property specs
- Frontend reactive initialization ensures dynamic enums load when valve modal opens
- Smart caching strategy balances performance with data freshness
- Graceful degradation with fallback options and error messages

**Use Cases:**
- Model selection filtered by user permissions
- Knowledge base selection showing only user-accessible bases
- Tool/resource dropdowns based on user role
- Any scenario requiring user-context-dependent options

### Screenshots or Videos

image

### DIFF
--- a/src/lib/components/common/Valves.svelte
+++ b/src/lib/components/common/Valves.svelte
@@ -6,9 +6,167 @@
 	import Switch from './Switch.svelte';
 	import MapSelector from './Valves/MapSelector.svelte';
 	import { split } from 'postcss/lib/list';
+	import { WEBUI_API_BASE_URL } from '$lib/constants';
 
 	export let valvesSpec = null;
 	export let valves = {};
+
+	// Dynamic enum support
+	let dynamicEnumCache = {};
+	let dynamicEnumLoading = {};
+	let dynamicEnums = {};
+
+	/**
+	 * Fetch dynamic enum options from an API endpoint
+	 */
+	async function fetchDynamicEnum(property, config) {
+		// Normalize config to object format
+		const normalizedConfig = typeof config === 'string' ? { endpoint: config } : config;
+
+		const {
+			endpoint,
+			method = 'GET',
+			response_path = null,
+			value_field = 'id',
+			label_field = null,
+			cache_ttl = 300,
+			fallback = []
+		} = normalizedConfig;
+
+		const cacheKey = endpoint;
+		const now = Date.now();
+		const fullUrl = `${WEBUI_API_BASE_URL}${endpoint}`;
+
+		// Check cache
+		if (
+			dynamicEnumCache[cacheKey] &&
+			now - dynamicEnumCache[cacheKey].timestamp < cache_ttl * 1000
+		) {
+			return dynamicEnumCache[cacheKey].data;
+		}
+
+		// Prevent duplicate simultaneous requests
+		if (dynamicEnumLoading[cacheKey]) {
+			return dynamicEnumLoading[cacheKey];
+		}
+
+		// Start fetch
+		dynamicEnumLoading[cacheKey] = fetch(fullUrl, {
+			method,
+			headers: {
+				Authorization: `Bearer ${localStorage.token}`,
+				'Content-Type': 'application/json'
+			}
+		})
+			.then(async (res) => {
+				if (!res.ok) {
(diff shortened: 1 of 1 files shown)


tjbck:
Interesting feature! Could you provide more context on how this is/planned to be used?

Simon-Stone:
I have two use cases that I currently want to use this in:

- A deep research pipeline that uses different models for each of the tasks of planning the research, summarizing, and report writing
- An automatic Memory extraction Filter where the chat history is scanned by an LLM for things to extract

In both cases, I want to be able to let the user choose which model they want to use for these tasks, independently of the model they are currently chatting with.

Simon-Stone:
Any interest in taking a closer look at this, @tjbck?

tjbck:
Any specific reason why they have to be dynamic per-se?

Simon-Stone:
Can you achieve the intended behavior in any other way?

What I mean is: how can you use client-side parameters (like the user ID used to filter models by access privileges) in Valves otherwise?

Again, one example use case would be to allow the configuration of task models for specific pipelines (like automatic memory extraction, deep research, ...) based on what is available to the user.

tjbck:
You could just set them as a static list, which should be possible today without introducing this modification. FYI, We're leaning towards "No" to this one but might change our minds later down the line.

Simon-Stone:
I can see how that would be much simpler, but seems clunky. It also wouldn't work at all if you have different user groups with access to different models. Which list would you hardcode? Even if you were to keep a separate list for each group, how would you decide on the server side, which one to use for the client?

It's serving us well for our purposes, but I of course totally understand if it's not a good fit for the project.

Thanks for considering it!

tjbck:
This should be addressed!

Issue #3023, opened 2024-06-11
enh: add TTS playback options


bannert1337:
**Is your feature request related to a problem? Please describe.**
Currently TTS playback is not stopping when it should from a QoL standpoint.
I will list some areas where the playback should be stopped or continued.

**Describe the solution you'd like**

- stop TTS playback when starting a new voice recording
- stop TTS playback when regenerating
- stop TTS playback when leaving the chat window (creating new chat, accessing settings/admin panel, switching to other chats, deleting chat)

**Describe alternatives you've considered**
During TTS playback the user can switch chats and access settings/admin panel, without stopping the playback. A floating player should be displayed on top of the screen, allowing the user to pause, resume or stop the TTS playback. The player should also act as a progress bar with the ability to forward or rewind the playback.

**Additional context**

silentoplayz:
Related - #2152

tjbck:
PR Welcome!

silentoplayz:
Related - https://github.com/open-webui/open-webui/issues/16150

PR #18193, opened 2025-10-09
Fix: The sidebar no longer re-opens automatically

amithadityacp:
This pull request resolves issue #3023 by implementing a persistent, floating player for Text-to-Speech (TTS) playback, significantly improving the user experience.

### Description

Previously, TTS audio would continue to play even after a user navigated away from the chat or performed another action, which could be confusing. This PR introduces a floating player that provides users with full manual control over the audio playback from anywhere in the application.

### Key Changes

* **Centralized State Management:** Introduced a new Svelte store (`src/lib/stores/ttsStore.js`) to globally manage the active `Audio` object.
* **Floating Player Component:** Created a new component, `src/lib/components/FloatingPlayer.svelte`, which displays and controls the TTS playback. The player is persistent across all routes.
* **Component Integration:**
    * Modified `Message.svelte` to use the new store, putting the audio object into the store when TTS is activated.
    * Modified the main `+layout.svelte` to include the `<FloatingPlayer />`, making it globally available.

This approach provides a robust and user-friendly solution to the problem outlined in the original issue.

---

I have read and agree to the Contributor License Agreement.

### DIFF
--- a/src/lib/components/chat/FloatingPlayer.svelte
+++ b/src/lib/components/chat/FloatingPlayer.svelte
@@ -1,0 +1,60 @@
+<script>
+  import { activeTTSAudio } from '$lib/stores/ttsStore.js';
+
+  let isPlaying = false;
+
+  activeTTSAudio.subscribe(audioInstance => {
+    if (audioInstance) {
+      // Listen to the audio object's events to update the UI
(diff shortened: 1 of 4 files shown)


silentoplayz:
This PR's title does not match the description or diff provided. Please update the PR's title accordingly.

Issue #6154, opened 2024-10-13
Artifact renders parts of content twice


Robotino04:
# Bug Report

## Installation Method

Docker

## Environment

- **Open WebUI Version:** v0.3.32
- **Ollama (if applicable):** 0.3.9
- **Operating System:** NixOS
- **Browser (if applicable):** Firefox 130.0

**Confirmation:**

- [x] I have read and followed all the instructions provided in the README.md.
- [ ] I am on the latest version of both Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [ ] I have included the Docker container logs.
- [x] I have provided the exact steps to reproduce the bug in the "Steps to Reproduce" section below.

## Expected Behavior:
The content renders once.

## Actual Behavior:
Some of the content is rendered twice.

## Description

**Bug Summary:**
The artifacts view sometimes renders parts twice.

## Reproduction Details

**Steps to Reproduce:**
Prompt a model with "Show me a code snippet of a website's sticky header and footer please." and open the artifact view.
Alternatively replace some other artifact's code with the following and save:

```html
<!DOCTYPE html>
<html>
<head>
<style>
body {
  margin: 0;
}

.header {
  position: fixed;
  top: 0;
  left: 0;
  width: 100%;
  background-color: #f1f1f1;
  padding: 20px;
}

.footer {
  position: fixed;
  bottom: 0;
  left: 0;
  width: 100%;
  background-color: #333;
  color: white;
  padding: 20px;
}

.content {
  margin-top: 80px; /* Account for header height */
  margin-bottom: 60px; /* Account for footer height */
  padding: 20px;
}
</style>
</head>
<body>

<div class="header">
  <h1>My Website</h1>
</div>

<div class="content">
  <h2>Welcome to my website!</h2>
  <p>This is some example content. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</p>
  <h3>What We Offer</h3>
  <p>Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.</p>
  <ul>
    <li>Service 1</li>
    <li>Service 2</li>
    <li>Service 3</li>
  </ul>

</div>

<div class="footer">
  <p>&copy; 2023 My Website</p>
</div>

</body>
</html>
[...]

PR #18195, opened 2025-10-09
fix: Artifacts duplicate matching

EntropyYue:
# Changelog Entry

### Description

- This solves the problem of potential duplicate HTML in Artifacts. The original logic was to try to match both code blocks and non-code blocks simultaneously, but this could lead to duplication. The solution is to assume that the model will always use code blocks in this response when a code block is matched

### Fixed

- Fixed the issue of duplicate HTML in Artifacts


silentoplayz:
Does this solve https://github.com/open-webui/open-webui/issues/6154?

EntropyYue:
I tested #6154 and https://github.com/open-webui/open-webui/issues/6154#issuecomment-2704465405, and they are both solved.

tjbck:
Thanks!

PR #18253, opened 2025-10-11
UPD_FEAT-Refresh_Sidebar_ChatList_after_actions

rgaricano:
### UPD_FEAT-Refresh_Sidebar_ChatList_after_actions

Add function for refresh sidebar chat list.
Fix some realtime rendering issues.

Added:
- store for chatList refresh trigger.
- refresh actions in Folders & Sidebar.
- refresh trigger and calls after chat list actions in ChatItem.
- refresh trigger after unarchive chat/s in ArchiveChatsModal.

https://github.com/user-attachments/assets/abbf8c93-2b6e-4b5b-9bd6-fefe62e10556

____

### DIFF
--- a/src/lib/components/chat/Placeholder/FolderTitle.svelte
+++ b/src/lib/components/chat/Placeholder/FolderTitle.svelte
@@ -9,7 +9,7 @@
 
 	import { toast } from 'svelte-sonner';
 
-	import { selectedFolder } from '$lib/stores';
+	import { selectedFolder, chatListRefresh } from '$lib/stores';
 
 	import { deleteFolderById, getFolderById, updateFolderById } from '$lib/apis/folders';
 	import { getChatsByFolderId } from '$lib/apis/chats';
@@ -93,6 +93,10 @@
 			});
 
 			await selectedFolder.set(_folder);
+			chatListRefresh.update((v) => ({
+				timestamp: Date.now(),
+				folderId: folder.id
+			}));
 			onUpdate(_folder);
 		}
 	};
--- a/src/lib/components/layout/ArchivedChatsModal.svelte
+++ b/src/lib/components/layout/ArchivedChatsModal.svelte
@@ -127,7 +127,7 @@
 				// Trigger global refresh
 				timestamp: Date.now(),
 				folderId: null // Refresh all since we don't have folder context here
-			}));			
+			}));
 			onUpdate();
 			await init();
 		} catch (error) {
--- a/src/lib/components/layout/ArchivedChatsModal.svelte
+++ b/src/lib/components/layout/ArchivedChatsModal.svelte
@@ -11,7 +11,7 @@
 		unarchiveAllChats
 	} from '$lib/apis/chats';
 	import { chatListRefresh } from '$lib/stores';
-	
+
 	import ChatsModal from './ChatsModal.svelte';
 	import UnarchiveAllConfirmDialog from '$lib/components/common/ConfirmDialog.svelte';
 	import Spinner from '../common/Spinner.svelte';
@@ -108,10 +108,11 @@
 			toast.error(`${error}`);
 		});
 
-	    chatListRefresh.update(v => ({    // Trigger global refresh  
-	        timestamp: Date.now(),
-	        folderId: null  			// Refresh all since we don't have folder context here  
-	    }));
+		chatListRefresh.update((v) => ({
+			// Trigger global refresh
+			timestamp: Date.now(),
+			folderId: null // Refresh all since we don't have folder context here
+		}));
 		onUpdate();
 		init();
 	};
@@ -122,10 +123,11 @@
 			await unarchiveAllChats(localStorage.token);
 			toast.success($i18n.t('All chats have been unarchived.'));
 
-		    chatListRefresh.update(v => ({    // Trigger global refresh  
-		        timestamp: Date.now(),
-		        folderId: null  			// Refresh all since we don't have folder context here  
-		    }));			
+			chatListRefresh.update((v) => ({
+				// Trigger global refresh
+				timestamp: Date.now(),
+				folderId: null // Refresh all since we don't have folder context here
+			}));			
 			onUpdate();
 			await init();
 		} catch (error) {
--- a/src/lib/components/layout/ArchivedChatsModal.svelte
+++ b/src/lib/components/layout/ArchivedChatsModal.svelte
@@ -10,7 +10,8 @@
 		getArchivedChatList,
 		unarchiveAllChats
 	} from '$lib/apis/chats';
-
+	import { chatListRefresh } from '$lib/stores';
+	
 	import ChatsModal from './ChatsModal.svelte';
 	import UnarchiveAllConfirmDialog from '$lib/components/common/ConfirmDialog.svelte';
 	import Spinner from '../common/Spinner.svelte';
@@ -107,6 +108,10 @@
 			toast.error(`${error}`);
 		});
 
+	    chatListRefresh.update(v => ({    // Trigger global refresh  
+	        timestamp: Date.now(),
+	        folderId: null  			// Refresh all since we don't have folder context here  
+	    }));
 		onUpdate();
 		init();
 	};
@@ -116,6 +121,11 @@
 		try {
 			await unarchiveAllChats(localStorage.token);
 			toast.success($i18n.t('All chats have been unarchived.'));
+
+		    chatListRefresh.update(v => ({    // Trigger global refresh  
+		        timestamp: Date.now(),
+		        folderId: null  			// Refresh all since we don't have folder context here  
+		    }));			
 			onUpdate();
 			await init();
 		} catch (error) {
--- a/src/lib/components/layout/Sidebar.svelte
+++ b/src/lib/components/layout/Sidebar.svelte
@@ -209,7 +209,7 @@
 	};
 
 	$: if ($chatListRefresh) {
-    	initChatList();
+		initChatList();
 	}
 
 	const loadMoreChats = async () => {
--- a/src/lib/components/layout/Sidebar.svelte
+++ b/src/lib/components/layout/Sidebar.svelte
@@ -25,6 +25,7 @@
 		isApp,
 		models,
 		selectedFolder,
+		chatListRefresh,
 		WEBUI_NAME
 	} from '$lib/stores';
 	import { onMount, getContext, tick, onDestroy } from 'svelte';
@@ -62,6 +63,8 @@
 	import PinnedModelList from './Sidebar/PinnedModelList.svelte';
 	import Note from '../icons/Note.svelte';
 	import { slide } from 'svelte/transition';
+
+	let lastRefreshTimestamp = 0;
 
 	const BREAKPOINT = 768;
 
@@ -204,6 +207,10 @@
 		// Enable pagination
 		scrollPaginationEnabled.set(true);
 	};
+
+	$: if ($chatListRefresh) {
+    	initChatList();
+	}
 
 	const loadMoreChats = async () => {
 		chatListLoading = true;
--- a/src/lib/components/layout/Sidebar/ChatItem.svelte
+++ b/src/lib/components/layout/Sidebar/ChatItem.svelte
@@ -27,7 +27,8 @@
 		showSidebar,
 		currentChatPage,
 		tags,
-		selectedFolder
+		selectedFolder,
+		chatListRefresh
 	} from '$lib/stores';
 
 	import ChatMenu from './ChatMenu.svelte';
@@ -69,6 +70,15 @@
 		}
 	};
 
+	const refreshChatList = () => {
+	    if (chat?.folder_id) {
+	        chatListRefresh.update(v => ({
+	            timestamp: Date.now(),
+ 	           folderId: chat.folder_id
+	        }));  
+ 	   }  
+	};
+
 	let showShareChatModal = false;
 	let confirmEdit = false;
 
@@ -90,6 +100,7 @@
 			await chats.set(await getChatList(localStorage.token, $currentChatPage));
 			await pinnedChats.set(await getPinnedChatList(localStorage.token));
 
+			refreshChatList();
 			dispatch('change');
 		}
 	};
@@ -107,6 +118,7 @@
 		});
 
 		if (res) {
+			refreshChatList();
 			goto(`/c/${res.id}`);
 
 			currentChatPage.set(1);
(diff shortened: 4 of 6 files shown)


mauritzn:
Well I finally couldn't take the wait and manually applied the patch _(to `v0.6.34`)_, from my basic testing it does seem to fix the issues I was having, but I did stumble across one issue in my limited testing so far. When you archive the active chat it keep the chat open, I feel like it should close it since that chat is now archived, similar to what is done when the active chat is deleted.

Unless this is the intended behavior? I haven't really used the archive feature so I don't really know how it's intended to work, but from my perspective if you archive a chat it means you're done with it, but just in a less permanent way than a deletion. As such I'd expect a similar behavior as the deletion of the active chat would have.

I'll be testing this patch much more either later today or tomorrow. So I'll post an update comment after that, but the initial result does seem promising.

rgaricano:
archive only mean that isn't on main chat list, but there isn't such close/d chat, or remove it from history,... it have same id and can be called with same link.
It can be done, only it's needed open another chat after archive, but the "problem" is what chat open, the most recent, the nearest to the archived chat,....
I think this is not neccessary, you can click on new chat or in another chat and continue working easily.

mauritzn:
Alright, personally I just found it a bit confusing, since depending on your current sidebar scroll it really gives no visual feedback that the active chat has been archived. Also when testing it seems like the archive chat button from the active chat window doesn't work.

**This one seems to be broken:**
image

At least for me it feels like there should at least be some kind of feedback to the user if the active chat is archived, other than it just disappearing from the sidebar.

mauritzn:
Found another issue, pinning and unpinning does not trigger an update, I have to collapse and uncollapse the folder to see the folder update.

mauritzn:
Also when using the active chat's menu to move a chat to another folder it does move the chat correctly but an update is not triggered.

**This one:**
image

rgaricano:
I would recheck, certanly I don't use/tested this feats, I begin with this PR just to solve the folder icon refreshing, and then I added it in other places,... then in the folders,...,by the way also added notes drag & drop... same for collections  and finished with another PR that had this refreshes in, but it has been rejected ...al least  I have working it on my end & now i have a more clear view of the internals of folders and sidebar elements.

If this simple PR isn't merged yet, surely is because another way is on the table!

mauritzn:
mmh, well seems like a more robust solution would be ideal, something that would be less likely to break or be forgotten if another feature got added in the future. I personally haven't really checked the frontend code of this project, apart from the minimal checks I did when I manually applied this patch.

Normally when I've worked with Svelte I've never had to force updates like this, since that's one of the beauties of Svelte... its reactivity. But I haven't looked deeply into how the folder data is laid out, but I feel like it might be better to look at why the UI isn't automatically updating from the data changes, since it seems like a misuse of Svelte's core reactive design.

Otherwise I'd consider maybe using dispatches rather than watching for a store value change, e.g. dispatch an event whenever a chat is moved, deleted, cloned, renamed, etc., but again I really haven't checked this project's code enough to see what kind of structure they are going for. From my viewpoint trying to get Svelte's native reactivity to work would be the most ideal solution.

I'm personally a bit apprehensive about digging further into the source code, since the more and more this project develops makes me more and more inclined to just make my own smaller app, since I don't use any of the advanced features...

mauritzn:
Also out of curiosity have you tried looking into what broke folder updates with the newer versions? because at least for cloning, renaming, moving and deleting it worked perfectly fine in `v0.6.32`, so obviously something got updated that broke things. Might be useful to look at what broke it first, if you haven't already.

mauritzn:
Found another issue, when the chat title is updated using the "Title Auto-Generation" feature it doesn't update correctly

rgaricano:
> Found another issue, when the chat title is updated using the "Title Auto-Generation" feature it doesn't update correctly

It work fine with this refresh.


> Also when using the active chat's menu to move a chat to another folder it does move the chat correctly but an update is not triggered.
For this only it's neccessary implement the same logic that for delete:
Change:
https://github.com/open-webui/open-webui/blob/7a83e7dfa367d19f762ec17cac5e4a94ea2bd97d/src/lib/components/layout/Sidebar/ChatItem.svelte#L137-L140
by:
```
	const archiveChatHandler = async (id) => {
		if ($chatId === id) {
        	// Find the closest chat in the list
        	const currentChats = $chats || [];
        	const currentIndex = currentChats.findIndex(chat => chat.id === id);

	        // Navigate to the next chat, or previous if at the end, or home if it's the only chat
    	    if (currentChats.length > 1) {
        	    const nextChat = currentChats[currentIndex + 1] || currentChats[currentIndex - 1];
            	await goto(`/c/${nextChat.id}`);
	        } else {
    	        await goto('/');
        	}
	        await chatId.set('');
    	    await tick();
    	} 
		await archiveChatById(localStorage.token, id);

		refreshChatList();
		dispatch('change');
	};
```
I tested and work fine, when a chat is archived the closest one is loaded.
(the refreshChatList() function  is the refresh function of this PR)



About refresh with reactive elements,
yes It could be done but as the refresh is done with diferents actions and elementss I think that is easier this way, controlling when the refresh is done and also I avoid have to add all the reactive actions and elements to the reactive logic.

About previous versions,
If I'm not wrong, I think that is because now the load of those elements are dynamic, to improve the load and render time and the refresh hasn't been updated...for that we need this PR (or similar)

mauritzn:
> This message was written while I was looking through the source code, so it might be a bit odd, but I felt like documenting my research into the code was the best solution... Hope it makes some sense at least... _(top: oldest, bottom: latest)_

As I'm looking more over the code and where you've added calls to `refreshChatList()` pretty much all those cases included proper dispatch events, wouldn't it make more sense to just use those events to trigger the refresh in that case? Most of them seem to send out `dispatch('change');` whenever a folder refresh would be needed.

Actually when looking closer if the `change` event was used for `open-webui/src/lib/components/layout/Sidebar/ChatItem.svelte` it would actually fix the pinning/unpinning behavior as well, since it seems to dispatch a `change` event when it gets a `change` event from the `ChatMenu` component, which is only triggered by the `pinHandler` function.

The only unique `refreshChatList()` call I found in `ChatItem.svelte` was for the `cloneChatHandler` function, it would be worth checking to see if adding a `change` dispatch there would break anything, but if not, then everything could be handled by the `change` trigger rather than a Svelte store state. _(**UPDATE:** when checking the code closer it seems like a `change` event could be added to the `cloneChatHandler` function, thus removing the need for the `refreshChatList` function)_

But then looking at the other calls you do to update the `chatListRefresh` store directly it seems like in those places they decided to use `onUpdate` functions instead... so it seems like there are 2 separate practices for handling updates, rather than using Svelte's dispatch system in all places... So it might be best to avoid using your store updating system, since that would introduce a third "on update" event system... confusing things even more than they already are...

Wow... okay, looking through this project code is quite... interesting to say the least... it has become quite fragmented with no clear "everyone should do this like so", but rather it seems like there's just more and more confusing patches on top of old patches...

Well, now I found a place that uses both `dispatch('change');` and `dispatch('update');` on the same component... both triggering `initChatList()`... From what I've seen now, please do a proper checking of all the events and their event chains, because this project could desperately need a single clear standard and have all legacy calls converted to that new standard... because this is a mess and it will only cause more and more bugs in the future...

Okay, so from what I can tell the only component that uses `dispatch('update')` is `open-webui/src/lib/components/layout/Sidebar/RecursiveFolder.svelte` and `open-webui/src/lib/components/layout/Sidebar/Folders.svelte`. `Folders.svelte` only relays the event from `RecursiveFolder.svelte`, so it should be fine to convert `RecursiveFolder.svelte` to only use `dispatch('change');`, thus clearing that confusion and allowing the single event to be used for updating the folder instead. But I still recommend more investigation, just incase I missed something.

If you want, I could maybe create a fork that shows the changes. But if you want to do them or forward this to someone else then that's fine.

mauritzn:
Well... good news... seems I found the issue from the `v0.6.33` update that broke folder updating... and from what I can tell pretty much just reverting that will fix everything, I'll test it some more and do some other small updates, but if everything goes according to plan it should be a very small patch that just reverts that change mostly.

Still need to see if I can figure out why the code was removed in `v0.6.33`, if there was a reason it might require a larger change, but if it was just a mistake then it should be a very quick fix.

I still haven't had time to check everything, but at least cloning, renaming, moving, pinning/unpinning, deleting, archiving and unarchiving is working as intended by just reverting that. From what I can tell at the moment the only thing still not working is the "Title Auto-Generation" feature, so I need to look into why that isn't updating the folders.

**UPDATE:** Seems like just reverting that doesn't fix everything, aside from the "Title Auto-Generation" feature not updating the folder, creating a new chat in the folder also doesn't update it... seems I need to look at that as well, but at least most things are working again with just the reversion.

mauritzn:
So, I finally found the two code changes in `v0.6.33` that cause the issues... one of them still needs more investigation before I feel comfortable reverting it as it is. In the current version I have everything is working as expected with the two reversions and some very minor refactoring.

I'll test it more tomorrow and try and figure out if there are issues with the reversions and if so how they could be modified to work. Once I've finished the patch I'd like to get some input from the person that originally removed the code during a refactor in `v0.6.33`, to get some input into why the code was removed, since I didn't really find much information from the commit history.

mauritzn:
Alright I've pushed the updates to my fork of the main branch of open-webui, it's in the `main-fix_folder_updating` branch here: https://github.com/mauritzn/open-webui/tree/main-fix_folder_updating

I'd appreciate some feedback on it and anyone willing to test the changes.

Also @tjbck I was wondering if you could comment on your [refactor commit](https://github.com/open-webui/open-webui/commit/97faeccebf064f8d29e4dee00d89f4eec5be7ca7#diff-5d657a04a70c3c777514cc4a4149b9d25f46f4909a374f100d2f259a6b12714f) that removed the following pieces of code from `src/lib/components/layout/Sidebar.svelte`:
```typescript
await tick();
for (const folderId in folders) {
  if (folders[folderId] && folders[folderId].is_expanded) {
    folderRegistry[folderId]?.setFolderItems();
  }
}
```
and:
```typescript
chats.subscribe((value) => {
  initFolders();
});
```

Personally I can understand wanting to remove the reliance on `chats.subscribe`, but there are many edge cases that break if this is removed, since some places do not properly trigger folder updates instead only setting the `chats` store. One example edge case is in `src/lib/components/chat/Placeholder.svelte` on line 90 to line 105, the code:
```svelte
<FolderTitle
  folder={$selectedFolder}
  onUpdate={async (folder) => {
    // A folder update call would be needed here if `chats.subscribe` is removed
    await chats.set(await getChatList(localStorage.token, $currentChatPage));
    currentChatPage.set(1);
  }}
  onDelete={async () => {
    // A folder update call would be needed here if `chats.subscribe` is removed
    await chats.set(await getChatList(localStorage.token, $currentChatPage));
    currentChatPage.set(1);
    selectedFolder.set(null);
  }}
/>
```

My minor patch includes some small refactors, but mostly the patch was just reverting these two code removals.

mauritzn:
I've been using the patch I made for a fair bit now and I haven't noticed any issues. But during my testing I found another issue with how folders are handled that is unrelated to the patch. Folders currently have no handling of pagination, as such they will never load more than 60 conversations. This is due to the following line of code:

> **NOTE:** There isn't anything wrong with the backend code, it's just that the use of the `page` option is never implemented in the frontend for folders, as such it's always stuck on the first page and only ever loads the 60 newest conversations of the folder.

```py
# Code from: `backend/open_webui/routers/chats.py` (line: 226-245)

@router.get("/folder/{folder_id}/list")
async def get_chat_list_by_folder_id(
    folder_id: str, page: Optional[int] = 1, user=Depends(get_verified_user)
):
    try:
        limit = 60
        skip = (page - 1) * limit

        return [
            {"title": chat.title, "id": chat.id, "updated_at": chat.updated_at}
            for chat in Chats.get_chats_by_folder_id_and_user_id(
                folder_id, user.id, skip=skip, limit=limit
            )
        ]

    except Exception as e:
        log.exception(e)
        raise HTTPException(
            status_code=status.HTTP_400_BAD_REQUEST, detail=ERROR_MESSAGES.DEFAULT()
        )
```

As such proper pagination handling would need to be implemented similar to how conversations are handled outside folders. Generally from what I've seen of how folders are handled they could use a big rework, since the current implementation seems very... patched together...

Either way I'd love to know if anyone else has been testing the patch I created and posted previously _(https://github.com/mauritzn/open-webui/tree/main-fix_folder_updating)_. I wouldn't really consider this patch as a real solution, it's more of a stopgap until a more proper rework could be done of how folders are handled. Still some feedback would be appreciated if anyone has tried it and encountered any issues.

tjbck:
Closing in favour of e29c262394022b9969a8f650cbc45aa192e39dc1, testing wanted here!

silentoplayz:
> Closing in favour of [e29c262](https://github.com/open-webui/open-webui/commit/e29c262394022b9969a8f650cbc45aa192e39dc1), testing wanted here!

Looks good to me!

mauritzn:
> Closing in favour of [e29c262](https://github.com/open-webui/open-webui/commit/e29c262394022b9969a8f650cbc45aa192e39dc1), testing wanted here!

I don't know if `v0.6.37`/`v0.6.38` implements all the things from the new solution, but from what I could see it seems to include the [e29c262](https://github.com/open-webui/open-webui/commit/e29c262394022b9969a8f650cbc45aa192e39dc1) commit and the [cd008ee](https://github.com/open-webui/open-webui/commit/cd008eeb508f115aa8711bf7e111339fc21b81f8) commit.

I tested `v0.6.38` and had the following issues:
- Auto title generation does not update the chat title in the folder, requiring a manual refresh trigger _(renaming a chat in a folder does update correctly though)_.
- Folder pagination still doesn't seem to work, as such a maximum of `10` folder items can ever be loaded _(due to the backend limit change introduced in commit [cd008ee](https://github.com/open-webui/open-webui/commit/cd008eeb508f115aa8711bf7e111339fc21b81f8), previously `60` items)_.
  - Due to folder pagination still not working correctly, creating a new chat in a folder with 10 items or when cloning/moving a chat to the folder the oldest chat in that folder will disappear until you're under the 10 item limit again.
- Moving the active chat to a different folder using the "active chat menu" does not trigger a folder refresh.
> The "active chat menu" I'm referencing _(top-right of screen)_:
> image
> _On a small side-note the `Archive` button in this menu still doesn't work either._

I haven't done a very extensive testing yet, but I thought I'd report my immediate findings now and report back if I find anything more tomorrow when I have the time to test more.

tjbck:
Loading max 10 chats is intended, the rest I'll investigate!

mauritzn:
> Loading max 10 chats is intended, the rest I'll investigate!

wait what? I understand loading max 10 chats per request and for the initial request. But I'm saying it only ever loads 10 chats, auto-loading more does not work, e.g. the pagination for folders on the frontend still doesn't work.

PR #18306, opened 2025-10-13
feat: add mineru as document parser backend with support of both local and managed api

palazski:
# Changelog Entry

### Description

This PR adds support for minerU as a document parser backend. minerU offers both a cloud API and a local deployment, and this PR supports both these APIs.

### Added

Added minerU backend file, related settings environments, and PersistentConfig entries..

### Changed

None.

### Deprecated

None.

### Removed

None.
### Fixed

None.

### Security

None.

### Breaking Changes

None.

---

### Additional Information

None.

### Screenshots or Videos

image


SmartisanNaive:
Good idea, the test has passed, I've starred it

tjbck:
qq: any specific reasons for separating the local and the hosted versions?

palazski:
> qq: any specific reasons for separating the local and the hosted versions?

api not the same for hosted vs local versions

backend/open_webui/config.py
```diff
@@ -2291,6 +2291,61 @@ class BannerModel(BaseModel):
...
+    os.environ.get("MINERU_API_URL", "http://localhost:8000"),
+)
+
+MINERU_API_KEY = PersistentConfig(
+    "MINERU_API_KEY",
+    "rag.mineru_api_key",
+    os.environ.get("MINERU_API_KEY", ""),
+)
+
+MINERU_ENABLE_OCR = PersistentConfig(
```
tjbck: Let's follow `DOCLING_PARAMS` convention here. Everything below should be replaced with `MINERU_PARAMS`.

palazski: I can do that, but I don't think that would be UX friendly. Docling also has a lot of parameters users set via the UI, and params in the minerU are counterparts to those variables. The only one that might be transferred to MINERU_PARAMS could be MINERU_PAGE_RANGES, but since it is just 1 parameter, why not just leave it as is?

palazski: e.g. DO_OCR and FORCE_OCR in Docling and MINERU_ENABLE_OCR, MINERU_ENABLE_FORMULA, and MINERU_ENABLE_TABLE
or
DOCLING_PDF_BACKEND and MINERU_MODEL_VERSION

if not the same, then similar...

tjbck: @palazski current approach is not sustainable, we'll be only accepting `PARAMS` method.

palazski: @tjbck Please review the changes again. I have updated the code!

tjbck:
Thanks!

wbste:
Do you have a link on how to self host this? I believe it's this one but just confirming: https://opendatalab.github.io/MinerU/quick_start/docker_deployment/

fakerms:
Does it support office formats?

PR #18493, opened 2025-10-21
FIX: Mermaid Renderer Issue

rgaricano:
FIX Mermaid Renderer Issue that show error message with icon below message input box.
Issue report & discussion: https://github.com/open-webui/open-webui/discussions/18340

____

### DIFF
--- a/src/lib/utils/index.ts
+++ b/src/lib/utils/index.ts
@@ -1593,18 +1593,23 @@ export const decodeString = (str: string) => {
 	}
 };
 
+let mermaidInitialized = false;
 export const renderMermaidDiagram = async (code: string) => {
 	const { default: mermaid } = await import('mermaid');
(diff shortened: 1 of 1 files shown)


ShirasawaSama:
I have tested it and confirmed that there will be no UI unavailability issues caused by Mermaid rendering errors.

tjbck:
9942de8011d4b5a141ac507c974c061c0cdad59a

rgaricano:
@tjbck
I saw that in the commit you removed the `suppressErrorRendering: true` parameter of mermaid Init function,
 **It is essential for solve the issue**.
And the parseresult check is a unnecessary doble check, the mermaid parse function trigger an error if parseresult fail.
I tested the commit and, as it is, it don't resolve the issue.

openwebui_mermaid_issue

PR #18565, opened 2025-10-23
feat(memory): enhance memory system

Davixk:
# Pull Request Checklist

- [x] **Target branch:** Verify that the pull request targets the `dev` branch.
- [x] **Description:** Provide a concise description of the changes made in this pull request down below.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [x] **Documentation:** If necessary, update relevant documentation.
- [x] **Dependencies:** No new dependencies added.
- [x] **Testing:** Manual tests performed to verify the implemented features work as intended.
- [x] **Agentic AI Code:** This PR has gone through human review and manual testing.
- [x] **Code review:** Self-review performed, code adheres to project standards.
- [x] **Title Prefix:** `feat` - Introduces new features and enhancements to memory functionality.

# Changelog Entry

### Description

This pull request significantly enhances the memory system in Open WebUI by introducing granular control over memory retrieval, improved semantic search capabilities, and administrative controls for memory management. The changes provide better control over memory quality and relevance through configurable thresholds and top-k parameters, while also adding the ability to force-enable memory for non-admin users.

### Added

- **Memory Top K Configuration**: Configurable parameter to control the maximum number of memories retrieved (default: 30)
- **Memory Relevance Threshold**: New slider control (0-100%) to set minimum similarity threshold for memory retrieval (default: 50%)
- **Memory Vector Re-indexing**: Button in admin settings to re-index all memories using the current embedding model
- **Enhanced Memory Metadata**: Memory items now include timestamps (created_at, updated_at) and similarity scores in ISO format
- **Memory Filtering**: Automatic filtering of low-relevance memories based on threshold during retrieval
- **Force Enable Memory**: New admin setting to force-enable memory for non-admin users, preventing them from disabling it
- **Improved Memory Injection Format**: Memories now injected as structured JSON within XML tags (`<memory_user_context>`) for better LLM parsing

### Changed

- **Memory Default State**: Changed default memory state from `false` to `true` for better user experience
- **Memory Context Format**: Memories now formatted as JSON objects with metadata instead of plain text list
- **Memory Logging**: Enhanced logging with error/info levels and detailed statistics about retrieved and filtered memories
- **Memory UI**: Non-admin users no longer see memory toggle when force-enable is active
- **Backend Structure**: Refactored imports in `auths.py` for better organization (alphabetical ordering)

### Fixed

[...]

PR #18573, opened 2025-10-24
feat: Add Chat Context Awareness

Davixk:
# Changelog Entry

### Description

Implements automatic Chat Context injection that adds recent or pinned chat history to conversation context.
I believe this feature will enhance the user experience by personalizing model context. Find an example in the screenshot below.

### Added

- Backend config: `ENABLE_CHAT_CONTEXT`, `CHAT_CONTEXT_MODE`, `CHAT_CONTEXT_TOP_K` in `config.py`
- Middleware handler `chat_context_handler()` in `utils/middleware.py` supporting:
  - Recent mode: retrieves last N chats (skip=1 to exclude current)
  - Pinned mode: retrieves user-pinned chats (early return if none exist)
  - Structured extraction: first_user_message + last_messages{user, assistant}
  - Conditional prompts: encouraging for pinned, discouraging for recent
  - JSON-in-XML injection format
- Admin UI in `General.svelte`: enable toggle, mode dropdown, top K input
- Feature flag in `Chat.svelte`: checks `$config.chat_context.enable`
- Config exposure via `/api/config` endpoint

### Changed

- `AdminConfig` model now includes chat context fields
- `get_admin_config()` and `update_admin_config()` handle chat context settings
- Middleware pipeline calls `chat_context_handler()` when feature enabled

### Additional Information

**Design decisions:**
- Skip=1 in recent mode prevents self-reference
- No skip in pinned mode
- Early return for empty pinned chats
- Multimodal content support extracts text from arrays

**Chat format injected:**
\`\`\`json
[
  {
    "title": "Chat Title",
    "created_at": "2024-01-15T10:30:00-0500",
    "updated_at": "2024-01-15T11:45:00-0500",
    "first_user_message": "...",
    "last_messages": {
      "user": "...",
      "assistant": "..."
    }
  }
]
\`\`\`

**Testing:** Manual testing of enable/disable, mode switching, top K validation, pinned empty/full, recent mode, config persistence, feature flag, JSON structure, multimodal handling.
### Screenshots or Videos

image
image

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -2511,6 +2511,24 @@ RAG_FILE_MAX_COUNT = PersistentConfig(
     ),
 )
 
+ENABLE_CHAT_CONTEXT = PersistentConfig(
+    "ENABLE_CHAT_CONTEXT",
+    "chat_context.enable",
+    os.environ.get("ENABLE_CHAT_CONTEXT", "False").lower() == "true",
+)
+
+CHAT_CONTEXT_MODE = PersistentConfig(
+    "CHAT_CONTEXT_MODE",
+    "chat_context.mode",
+    os.environ.get("CHAT_CONTEXT_MODE", "recent"),
+)
+
+CHAT_CONTEXT_TOP_K = PersistentConfig(
+    "CHAT_CONTEXT_TOP_K",
+    "chat_context.top_k",
+    int(os.environ.get("CHAT_CONTEXT_TOP_K", "3")),
+)
+
 RAG_FILE_MAX_SIZE = PersistentConfig(
     "RAG_FILE_MAX_SIZE",
     "rag.file.max_size",
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -220,6 +220,9 @@ from open_webui.config import (
     RAG_HYBRID_BM25_WEIGHT,
     RAG_ALLOWED_FILE_EXTENSIONS,
     RAG_FILE_MAX_COUNT,
+    ENABLE_CHAT_CONTEXT,
+    CHAT_CONTEXT_MODE,
+    CHAT_CONTEXT_TOP_K,
     RAG_FILE_MAX_SIZE,
     FILE_IMAGE_COMPRESSION_WIDTH,
     FILE_IMAGE_COMPRESSION_HEIGHT,
@@ -811,6 +814,9 @@ app.state.config.TOP_K_RERANKER = RAG_TOP_K_RERANKER
 app.state.config.RELEVANCE_THRESHOLD = RAG_RELEVANCE_THRESHOLD
 app.state.config.HYBRID_BM25_WEIGHT = RAG_HYBRID_BM25_WEIGHT
 
+app.state.config.ENABLE_CHAT_CONTEXT = ENABLE_CHAT_CONTEXT
+app.state.config.CHAT_CONTEXT_MODE = CHAT_CONTEXT_MODE
+app.state.config.CHAT_CONTEXT_TOP_K = CHAT_CONTEXT_TOP_K
 
 app.state.config.ALLOWED_FILE_EXTENSIONS = RAG_ALLOWED_FILE_EXTENSIONS
 app.state.config.FILE_MAX_SIZE = RAG_FILE_MAX_SIZE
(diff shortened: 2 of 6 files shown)


EntropyYue:
I tested this PR locally, and it looks pretty good

Davixk:
thank you @EntropyYue !

let me know if you have any feature suggestions.

my goal is to make the models as context aware as possible - while keeping token usage efficient and agnostic.

EntropyYue:
Your implementation appears to include the reasoning block and tool calls and their outputs in the context. The regex `<details.*?</details>\n*` can match them.

Davixk:
thank you. setting to draft until I can work on it.

silentoplayz:
Unrelated, but why is your entire UI font in lowercase? 😄

Edit: `URL` is still uppercase! LOL

silentoplayz:
Also, I encourage you to keep going with this PR. This is quite nice if you asked me. 
image

Davixk:
update: this is something I first suggested with #13595

---

> Unrelated, but why is your entire UI font in lowercase? 😄
> 
> Edit: `URL` is still uppercase! LOL

personal taste :) i hate arbitrarily uppercased text, it looks messy.
it's a custom i18n renderer I wrote. `URL` is capitalized because I added exceptions for acronyms, nouns and other words that need to retain capitalization.

and no, this doesn't affect anyone else - the strings are still capitalized as expected for everyone else.

thank you for the feedback!

rgaricano:
I think these types of features should be added as a tool or function. The best thing about the open web interface is its modularity. If the middleware becomes overloaded with so many features, it will become unmanageable.
Another drawback is that chats are currently not vectorized, so if we add long chats to the context, they will quickly become saturated (and will have to be deleted, thus losing the point of this feature).
For the feature to work properly, the chat had to be vectorized, for example, as a knowledge collection, and where appropriate, retrieval was done using hybrid search (similar as memories, or directly convert the chats to memories)

Furthermore, adding a fixed number of recent chats can result in a confusing context for the model: for this type of features it is better to let the user freely add chats to the conversation, with a simple drag&drop (or with attach).
(Similar for memories, I experimented this kind of issues because it inyect all memories, and then the model response with "random" data, different languages, ..., and a unmanageable context lenght)

(I mean, I like the idea but not the way, I myself have implemented functions and tools to do similar things in a more dynamic way)

Davixk:
@rgaricano there is a solid argument to be made about that! and it is something I thought about.

here's why I think it **should** indeed be a core feature:
1. there is a **high enough advantage** to gain, even just by injecting core info from the last 2/3 chats. it doesn't necessarily need embedding and similarity retrieval to become a useful feature
2. this can **affect future plans** for Open WebUI: as you said yourself, some great future additions might come in the form of chat embedding and retrieval, and I have been thinking about **AI chat summaries**, which could be an asynchronous task that the WebUI runs for each chat, which would be best stored in a new dedicated field, and might need db changes. 
3. usefulness is **proven by ChatGPT**. it has implemented a very similar feature in May 2025, where the last few most recent chats are injected into the prompt, and I believe it works well enough

you can find more info in my issue #13595 from May 6th, where Tim seemed to agree with me.

rgaricano:
yes, but I can do that without need of an auto inyect, I just add to context the chats that I want.

https://github.com/user-attachments/assets/009cd8fc-5ebc-4831-9643-a22183cbc416

Davixk:
@rgaricano i get what you're saying. and sure, manually adding context works - but that's a power user workflow.

you can also manually add memories to every chat by typing them in, or copy-paste messages from other chats every time. that doesn't mean we shouldn't automate it.

the goal here is to make context awareness **automatic and seamless** for everyone, not just people who proactively drag & drop. most users won't even think to do that, but they'll absolutely benefit from the model remembering their last conversation without lifting a finger.

manually attaching chats is a great feature, but it doesn't really conflict with or detract from this PR - both are useful in their own ways. this just raises the baseline experience while you can still have full manual control when you want it.

edit:
also worth noting - they serve different goals
- manual attachment is for full, **explicit injection**. it tells the model to actively look at and focus on the attached chats, and injects way more text from them into context.
- this PR is for passive, **automatic context**. it only injects a few key data points from recent chats - not all messages. the model uses it if it's relevant, ignores it otherwise.

lastly, as I mentioned earlier, this might impact future plans for Open WebUI. features like asynchronous **chat summaries** or embeddings would naturally build on top of this system. those wouldn't be possible if this only lived as an external module.

Davixk:
### update

changed:
- updated admin panel UI to align with #18565
- centralized message regex cleaning and extended to chat context middleware, as suggested by @EntropyYue 

PR is now ready for review on my end, but there are two issues:
1. UX could be aligned with Memory function: Memory allows users to toggle in their settings, doesn't allow toggling at admin level - chat context allows toggling in admin panel, doesn't allow toggling in user settings. Both could align and offer both methods to enhance control
2. As @rgaricano duly noted, I do agree that the middleware module is already very long as-is. I believe devising a more modular approach to it would be the best course of action

tjbck:
This should be implemented as a Filter.

PR #17775, opened 2025-09-26
feat: add comprehensive data sanitization to prevent null bytes

Classic298:
This commit introduces a robust, centralized sanitization mechanism to prevent null bytes (`\u0000`) from being stored in the database. These characters were causing `UntranslatableCharacter` errors during search operations in PostgreSQL deployments.

The solution consists of a new centralized utility, `data_sanitizer.py`, which provides a `sanitize_data` function that recursively traverses data structures (dictionaries, lists, and strings) and removes null bytes.

To apply this sanitization broadly and automatically, this change introduces custom SQLAlchemy `TypeDecorator` classes, `SanitizedText` and `SanitizedJSON`. These types wrap the standard `Text` and `JSON` types and apply the sanitization logic before the data is sent to the database.

These new sanitized types have been applied to all relevant columns in the database models, including `Chat`, `Prompt`, `File`, `Message`, `User`, and `Knowledge`, ensuring that all user-generated content and external data are cleaned proactively. This approach provides a comprehensive, long-term fix for the issue by ensuring data integrity at the ORM layer.

### Contributor License Agreement

By submitting this pull request, I confirm that I have read and fully agree to the [Contributor License Agreement (CLA)](https://github.com/open-webui/open-webui/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT), and I am providing my contributions under its terms.

PR #18576, opened 2025-10-24
fix(chats): fix chat search crash

Davixk:
# Changelog Entry

### Description

This PR fixes a PostgreSQL-specific issue where chat searches would crash when encountering null bytes (`\u0000` or `\x00`) in chat messages or titles. PostgreSQL does not allow null bytes in text fields, causing database errors during search operations. The fix adds proper filtering to exclude chats containing null bytes before attempting text extraction and comparison operations.

### Added

- Added null byte filtering for PostgreSQL chat message content using JSON representation check
- Added null byte filtering for PostgreSQL chat titles

### Changed

- Updated PostgreSQL chat search query to check for null bytes in message content before text extraction
- Modified search logic to explicitly check `message->'content' IS NOT NULL` before processing

### Deprecated

- N/A

### Removed

- N/A

### Fixed

- Fixed PostgreSQL crashes when searching chats containing null bytes in message content
- Fixed PostgreSQL errors when searching chats with null bytes in titles

### Security

- N/A

### Breaking Changes

- N/A

---

### Additional Information

The fix specifically targets PostgreSQL databases where null bytes in JSON text fields cause errors. The solution:
1. Checks if message content is not null before processing
2. Filters out messages containing `\u0000` in their JSON representation
3. Filters out chat titles containing `\x00` (null bytes)

This ensures the search operation only processes valid text content without null bytes, preventing database errors while maintaining search functionality.

### Screenshots or Videos

- N/A (Backend database query fix)

### DIFF
--- a/backend/open_webui/models/chats.py
+++ b/backend/open_webui/models/chats.py
@@ -765,17 +765,20 @@ class ChatTable:
                     )
 
             elif dialect_name == "postgresql":
-                # PostgreSQL relies on proper JSON query for search
-                # Note: We use replace() to remove null bytes (\u0000) which PostgreSQL cannot convert to text
+                # PostgreSQL doesn't allow null bytes in text. We filter those out by checking
+                # the JSON representation for \u0000 before attempting text extraction
                 postgres_content_sql = (
                     "EXISTS ("
                     "    SELECT 1 "
                     "    FROM json_array_elements(Chat.chat->'messages') AS message "
-                    "    WHERE message->>'content' IS NOT NULL "
-                    "    AND LOWER(REPLACE(message->>'content', chr(0), '')) LIKE '%' || :content_key || '%'"
+                    "    WHERE message->'content' IS NOT NULL "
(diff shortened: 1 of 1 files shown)


silentoplayz:
Related PR that was already open by another contributor - https://github.com/open-webui/open-webui/pull/17775

Davixk:
thank you @silentoplayz 
that PR looks great. same core issue this PR is addressing here.

while 17775 should avoid null bites existing in the DB in the first place, I still think it **makes sense for this PR to exist**, because it works on the other side of the flow.

feel free to close if you think otherwise.

Classic298:
Hi @Davixk 

**I think both PRs should be merged because my PR attempts to prevent the null bytes from existing in the first place, wheras yours adds extra protection if null-bytes
1) already exist (they do on installations like from those users who commented on the issue!)
2) somehow made it past my sanitization (maybe some data inputs i missed somewhere)**

So for a proper in-depth protection and solution, both PRs are needed IMO!

tjbck:
Thanks!

PR #18647, opened 2025-10-27
fix(auth): Decode trusted name header to support non-ASCII usernames

JungJunYong:
## Description

This PR fixes an issue where usernames containing non-ASCII characters (e.g., Korean, Japanese, Cyrillic) were displayed incorrectly during trusted header authentication.

The name would appear as a raw URL-encoded string (like `%ED%99%8D...`) instead of the actual characters (e.g., `홍길동`).

## How it was fixed

The problem occurred because the backend was reading the `WEBUI_AUTH_TRUSTED_NAME_HEADER` value directly without decoding it.

This fix introduces `urllib.parse.unquote` to the sign-in logic. This correctly decodes the header value, converting any URL-encoded sequences back into their original characters before assigning the name to the user.

This ensures usernames in all languages are displayed correctly while still allowing for safe transport of non-ASCII characters in HTTP headers.

### DIFF
--- a/backend/open_webui/routers/auths.py
+++ b/backend/open_webui/routers/auths.py
@@ -4,6 +4,7 @@ import time
 import datetime
 import logging
 from aiohttp import ClientSession
+import urllib.parse
 
 from open_webui.models.auths import (
     AddUserForm,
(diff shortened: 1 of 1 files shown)


tjbck:
Full CLA required.

JungJunYong:
> 
Hi @tjbck , I have signed the CLA. Could you please review this PR again?

silentoplayz:
@JungJunYong You would need to open another pull request based on the `dev` branch like you have already done here, but with the pull request filled out and the CLA clauses left intact.

PR #18748, opened 2025-10-31
fix: add deduplication for model IDs to prevent duplicate models in list

kindredzhang:
### Summary

Deduplicate model entries when multiple endpoints return the same model ID to prevent frontend crashes and improve user experience for proxy gateway users.

### Motivation / Use-Case

Most users of `openai-python` do not call `api.openai.com` directly; they route through unified gateways (One-API, New-API, Cloudflare-AI, etc.). These gateways aggregate dozens of providers into a single `/v1/models` response, so identical IDs (e.g., `gpt-3.5-turbo`, `babbage-002`) appear many times under different indexes.

The frontend settings page uses a Svelte `each` block keyed by ID; duplicate keys trigger:

> Uncaught (in promise) Error: Cannot have duplicate keys in a keyed each: ...

Once thrown, the route crashes and the UI becomes unresponsive—users must force-refresh the browser and have no way to configure models manually.

This server-side deduplication lets "proxy users" consume the list out-of-the-box without any breaking changes or manual allow-lists.

### Changes

* Add `seen_ids: set[str]` inside `merge_models_lists()`
* Skip models whose `ID` (or `name`) was already emitted
* Maintain existing filtering logic unchanged
* Use `O(1)` set lookup for efficient deduplication

### Testing

* Two mock endpoints returning overlapping models → no duplicates in payload
* Frontend settings page renders without crashing
* `pytest tests/` passes locally
* No breaking changes to existing functionality

### Type of Change

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ] Documentation update

### Checklist

* [x] My code follows the project's style guidelines
* [x] I have performed a self-review of my own code
* [x] I have commented my code, particularly in hard-to-understand areas
* [x] I have made corresponding changes to the documentation
* [x] My changes generate no new warnings
* [x] I have added tests that prove my fix is effective or that my feature works
* [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/backend/open_webui/routers/openai.py
+++ b/backend/open_webui/routers/openai.py
@@ -504,39 +504,42 @@ async def get_all_models(request: Request, user: UserModel) -> dict[str, list]:
     def merge_models_lists(model_lists):
         log.debug(f"merge_models_lists {model_lists}")
         merged_list = []
+        seen_ids = set()
 
         for idx, models in enumerate(model_lists):
             if models is not None and "error" not in models:
-
-                merged_list.extend(
-                    [
-                        {
+                for model in models:
+                    model_id = model.get("id") or model.get("name")
+                    if not model_id:
+                        continue
+                    
+                    # Check if this model should be included
+                    should_include = (
+                        "api.openai.com"
+                        not in request.app.state.config.OPENAI_API_BASE_URLS[idx]
+                        or not any(
+                            name in model_id
+                            for name in [
+                                "babbage",
+                                "dall-e",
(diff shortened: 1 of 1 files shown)


tjbck:
Closing in favour of fdf7ca15ea0e7fbb63cbdaf3b5c6ae5b77255e55, Thanks!

PR #19025, opened 2025-11-07
fix: Handle AttributeError in hybrid search with reranking (#17046)

krishna-medapati:
## Description
Fixes #17046 - AttributeError when using hybrid search with reranking enabled.

## Problem
The current code attempts to access `collection_result.documents` in the conditional check even after using `hasattr()`. Due to Python's short-circuit evaluation behavior, when `collection_result` is a list or doesn't have the `documents` attribute, the subsequent conditions still try to access `.documents`, causing an `AttributeError`.

## Solution
Split the attribute existence checks from the content validation checks:
1. First verify that both `documents` and `metadatas` attributes exist
2. Then separately validate the document content in a second conditional block

## Changes
- Added separate conditional blocks for attribute existence checks vs content validation
- Added `hasattr()` check for `metadatas` attribute (was missing)
- Prevents `AttributeError` while maintaining all original validation logic
- Added clarifying comments to explain the two-stage validation

## Testing
- [x] Code compiles without syntax errors
- [x] Resolves the AttributeError described in issue #17046
- [x] Maintains all original validation behavior

## Impact
This fix ensures that hybrid search with reranking works correctly without throwing `AttributeError` when the collection result doesn't have the expected structure.

---

**Note:** Deleting the CLA section will lead to immediate closure of your PR and it will not be merged in.

### DIFF
--- a/backend/open_webui/retrieval/utils.py
+++ b/backend/open_webui/retrieval/utils.py
@@ -160,10 +160,18 @@ def query_doc_with_hybrid_search(
     hybrid_bm25_weight: float,
 ) -> dict:
     try:
+        # First check if collection_result has the required attributes
         if (
             not collection_result
             or not hasattr(collection_result, "documents")
-            or not collection_result.documents
+            or not hasattr(collection_result, "metadatas")
+        ):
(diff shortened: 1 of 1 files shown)


tjbck:
Thanks!

PR #19095, opened 2025-11-10
feat: Adding file metadata to hybrid search

jmleksan:
# Pull Request Checklist

- [x] **Target branch:** Verify that the pull request targets the `dev` branch. **Not targeting the `dev` branch will lead to immediate closure of the PR.**
- [x] **Description:** Provide a concise description of the changes made in this pull request down below.
- [x] **Documentation:** If necessary, update relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs) like environment variables, the tutorials, or other documentation sources.
- [x] **Testing:** Perform manual tests to **verify the implemented fix/feature works as intended AND does not break any other functionality**. Take this as an opportunity to **make screenshots of the feature/fix and include it in the PR description**.
- [x] **Title Prefix:** To clearly categorize this pull request, prefix the pull request title using one of the following:
  - **feat**

---

# Changelog Entry

- 📚 Hybrid search now indexes high-signal metadata (filenames, titles, headings, sources, and snippets) alongside chunk text, enabling keyword queries to surface documents where the term appears only in metadata.

### Description
- Improve hybrid retrieval by blending document content with selected metadata so quires can find keywords that may only appear in file metadata such as file name or source. One use case may be if you want to search for documents of a certain type or that may contain a certain date in them that may only be present in the file name. 

**Use Case: “Show me our PPO documents”**

- A benefits specialist uploads plan files named `PPO_insurance_2024.pdf`, `medical_PPO_plan.docx`, etc.
- These documents mention “PPO” only in their filenames or titles, so hybrid search on just the body text previously missed them.
- After the metadata-aware BM25 augmentation, a query like `PPO benefits` or just `PPO` now surfaces every relevant document because filenames, titles, headings, and snippets are indexed alongside the content.
- Partial matches work as well: typing `PPO` returns all PPO-related files immediately, instead of relying on the term being repeated in the body text.

### Added
- Augmented the BM25 input text with filename, title, section heading, source, and snippet metadata.
- Tokenized filenames (replacing underscores, dashes, and extensions) to support partial-name matches such as “pdf”.

### Fixed
- Ensured hybrid search returns documents when the query term exists only in metadata (e.g., filenames or web page titles).

---

### Additional Information
- Affected code: `backend/open_webui/retrieval/utils.py`


tjbck:
@jmleksan any reason this has to be specific to hybrid search?

jmleksan:
@tjbck There were a few:

1. If it was added for semantic, then all documents would need to be reindexed.
2. The extra metadata may muddy up the semantic vectors since it introduces structured keyword fields that don’t always carry natural language meaning.

It could be added to semantic, but I think it better leverages the strengths of hybrid, so I included it there instead. If you believe semantic would also benefit, I can update it.

tjbck:
@jmleksan I think this pipeline *could* make sense to have it before embedding process instead of applying them during hybrid search, thoughts?

jmleksan:
@tjbck I can see the benefits of including metadata earlier in the pipeline so the chunk text the model sees can carry some extra context about which document it came from, and so hybrid search can use it as well.

My only concern is in case the intent is for that enriched text to also be fed into the embedding step. Embedding titles or filenames along with the body can lead to lower quality vectors in situations where the chunk content isn’t strongly related to the metadata. I’d prefer to keep the embedding text focused on the body by default and treat metadata as an additional signal for hybrid and for sending context to the model.

tjbck:
@jmleksan sounds reasonable

backend/open_webui/retrieval/utils.py
```diff
@@ -172,8 +172,43 @@ def query_doc_with_hybrid_search(
 
         log.debug(f"query_doc_with_hybrid_search:doc {collection_name}")
 
+        # This allows BM25 to find documents by filename, title, headings, source URL, etc.
+        augmented_texts = []
```
tjbck: should be a `get_enriched_texts` function

tjbck: `ENABLE_RAG_HYBRID_SEARCH_ENRICHED_TEXTS` env var should be introduced.

jmleksan:
Ok moved it to a function and added the env/ui config that defaults to false. Also tested to confirm change affects weights and ui affects config,

Screenshot (5)

# When asking to explain my "big.pdf"

### Without Enriching Text:
image

### With Enriching Text:
image

tjbck:
Great addition, Thanks!

PR #19117, opened 2025-11-11
fix: __user__['valves'] should reflect current tool's UserValves when using multiple tools

Podden:
I noticed UserValves we're incorrect when custom pipe and multiple tools are used in a request:
All Tool are getting the same `__user__["valves"]`, namely the valves from the last tool in the list, resulting in errors when calling methods who need uservalves.

I've pinpoint the root of the problem in `tool.py`, where `extra_params` is used in conjunction with `functool.partial`, which copies the dict but only references the content. When extra_params["__user__"]["valves"] gets overwritten by the next tool, it changes it for all tools, resulting in wrong uservalves per tool. Setting extra_params per method instead of per tool solved this problem.

Problem Example:
``` python
# Tool A - Function 1
extra_params["__user__"]["valves"] = UserValvesA()  # New Instance
partial_a1 = partial(func_a1, __user__=extra_params["__user__"])  # Saves reference to {"valves": UserValvesA}

# Tool B - Funktion 1
extra_params["__user__"]["valves"] = UserValvesB()  # Overwrites the value!
# extra_params["__user__"] dict is still the same object!
# partial_a1 is still pointing the this dics, but now it contains {"valves": UserValvesB}!
partial_b1 = partial(func_b1, __user__=extra_params["__user__"])
```

Contributor License Agreement

### DIFF
--- a/backend/open_webui/utils/tools.py
+++ b/backend/open_webui/utils/tools.py
@@ -232,16 +232,10 @@ async def get_tools(
                 module, _ = load_tool_module_by_id(tool_id)
                 request.app.state.TOOLS[tool_id] = module
 
-            extra_params["__id__"] = tool_id
-
             # Set valves for the tool
             if hasattr(module, "valves") and hasattr(module, "Valves"):
                 valves = Tools.get_tool_valves_by_id(tool_id) or {}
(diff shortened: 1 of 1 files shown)


Podden:
Try again

PR #19144, opened 2025-11-12
Fix: Handle empty strings in OAuth registration response (#19129)

xqqp:
Fixes #19129 - The mcp package requires optional unset values to be None. If an empty string is passed, it gets validated and fails. This PR replaces all empty strings in the Dynamic Client Registration response with None.

# Changelog Entry

### Fixed

- ignore empty optional URLs in the Dynamic Client Registration response


tjbck:
Thanks!

PR #19165, opened 2025-11-13
fix: pass token_endpoint_auth_method when creating an oauth client

xqqp:
AuthLib defaults to using basic auth when `token_endpoint_auth_method` has not been configured. This PR passes `token_endpoint_auth_method` when creating an OAuth client. Without this `authorize_access_token()` fails if the OAuth server does not accept basic auth as an authentication method. 

# Changelog Entry
### Fixed

- use client's configured endpoint authentication method, instead of defaulting to basic authentication


tjbck:
c43f95f4b8d8cb723e0d1a9dfdfab5ac3b54da7c

PR #19180, opened 2025-11-14
feat: enable function calling for Azure OpenAI and native-capable providers (CLA fix version)

simoncelinder:
### Description
- Inject tools into chat completion payload for endpoints that support function calling
- Maintains backward compatibility with prompt-based systems
- Fixes Azure OpenAI tool invocation while preserving existing behavior

Forum thread: https://github.com/open-webui/open-webui/discussions/19141


tjbck:
Intended behaviour with default mode. You should be using native mode.

PR #19237, opened 2025-11-17
Chat 1161

seanssc-hue:


Classic298:
@seanssc-hue 

For your information for versions above 0.6.5 cannot allow rebranding
- if you have more than 50 users
- do not have explicit written permission as a contributor or
- purchased an enterprise license

I hope you are aware of that.
If you are using this just for small scale testing, forked open-webui before version 0.6.5 or have an enterprice license it's ok - otherwise, I'd suggest you to look into purchasing an official license for your rebranded work!

https://docs.openwebui.com/license

PR #19241, opened 2025-11-17
MCP: get __files__ information from OpenWebUI

tobiasbaur:
### Description

- Tools in Openwebui have access to some reserved variable names, for example to the attached files in a chat. For MCP this so far doesn't work. This PR introduces reserved variable names (to be more precise: __files__) to also work with MCP.
A MCP server that accepts __files__: dict get access to the file information

Example mcp tool:

```
@mcp.tool
def process_files(__files__: dict):
    """
    give information on the attached file
    """
    print(str(__files__))
    return str(__files__)
```

This allows an mcp server, with minimal parsing of the dict to know the id and local url on the file server, making it easy to access files via api or on the local file structure to be processed within an mcp tool. 

### Added

- reserved __files__ word to also be considered for mcp tool calls.

### Screenshots or Videos

Screenshot 2025-11-17 at 21 04 45

This is the result of the above mcp tool

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -1370,6 +1370,10 @@ async def process_chat_payload(request, form_data, user, metadata, model):
 
                         def make_tool_function(client, function_name):
                             async def tool_function(**kwargs):
+                                if "__files__" in kwargs:
+                                    files_list = form_data.get("metadata", {}).get("files", kwargs.get("__files__", []))
+                                    kwargs["__files__"] = {file_item["id"]: file_item for file_item in files_list}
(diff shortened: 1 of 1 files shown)


tjbck:
Not a valid mcp, you should be using built-in tools for this.

tobiasbaur:
Excuse me, what do you mean not a valid mcp? This PR simply provides a way to get access to the __files__ variable that built-in tools already have. So the MCP server will receive information about the used files in the chat that it then can use to process further.

If i must use an mcp server, I dont think I have any way to pass files or file information from the chat right now, or am I mistaken?

PR #19371, opened 2025-11-21
feat: rate limit for login attempts

diwakar-s-maurya:
# Changelog Entry

Add rate limit for login attempts to make password attacks slow.

### Description

At present the system allows unlimited attempts to log in with email/password method. An attacker can try at high-rate techniques to break user's password such as [dictionary attack](https://en.wikipedia.org/wiki/Dictionary_attack). This pull request adds a basic rate limit on `/signin` route

### Security

Add rate limit for failed login attempts.

### Additional Information

The implemented rate limit solution is basic. In deployments running multiple replicas of openwebui, it will not provide perfect rate-limit as defined but it still provides some protection rather than no protection.

---

### DIFF
--- a/backend/open_webui/utils/login_rate_limit.py
+++ b/backend/open_webui/utils/login_rate_limit.py
@@ -23,5 +23,4 @@ class RateLimiter:
         return True
 
     def reset(self, key: str):
-        if key in self.attempts:
-            del self.attempts[key]
+        self.attempts.pop(key)
--- a/backend/open_webui/utils/login_rate_limit.py
+++ b/backend/open_webui/utils/login_rate_limit.py
@@ -23,4 +23,4 @@ class RateLimiter:
         return True
 
     def reset(self, key: str):
-        self.attempts.pop(key)
+        self.attempts.pop(key, None)
--- a/backend/open_webui/utils/login_rate_limit.py
+++ b/backend/open_webui/utils/login_rate_limit.py
@@ -1,0 +1,27 @@
+import time
+from collections import defaultdict
+
+class RateLimiter:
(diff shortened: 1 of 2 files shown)


silentoplayz:
I've tested this pull request and it appears to be working as expected. However, I've observed that the rate limiting on the `/signin` route is blocking legitimate login attempts (after illegitimate login attempts). This can happen if a few previous login attempts failed due to an incorrect password for the provided email address, resulting in an `API rate limit exceed` notification toast.

Is this behavior intended with the current rate limiting configuration on the `/signin` route?

diwakar-s-maurya:
@silentoplayz Yes, this is intended behaviour.

tjbck:
Keeping this open as ref PR, but this won't work in multi-instance envs.

backend/open_webui/utils/login_rate_limit.py
```diff
@@ -0,0 +1,27 @@
...
+        
+        if len(self.attempts[key]) >= self.max_attempts:
+            return False
+        
+        self.attempts[key].append(current_time)
+        return True
+
+    def reset(self, key: str):
+        if key in self.attempts:
+            del self.attempts[key]
```
larsblumberg: Maybe simpler:

```suggestion
    def reset(self, key: str):
        self.attempts.pop(key)
```

if it's okay to raise a `KeyError`, otherwise:

```suggestion
    def reset(self, key: str):
        self.attempts.pop(key, None)
```

diwakar-s-maurya:
https://github.com/open-webui/open-webui/commit/7b166370432414ce8f186747fb098e0c70fb2d6b

PR #19444, opened 2025-11-24
feat: Introduce Toggle to turn off parallel embedding processing

Classic298:
Useful for local embedding setups for when extremely API rate limit constrained.

image

image

> [!NOTE]


scheatkode:
Thanks for dealing with this issue on such short notice! Parallel embedding processing is a very welcome change and, if I may suggest instead of exposing a toggle, I'm wondering if it would be useful to let users control the degree of concurrency themselves; personally that would be my preference. A possible pattern could be:

```sh
RAG_EMBEDDING_CONCURRENCY=100  # As a sane-ish default
```

Setting `RAG_EMBEDDING_CONCURRENCY` to 1 effectively disables parallelism while still letting people reduce load if their environment is constrained (e.g., low‑RAM GPUs, CPU‑bound workloads). The numeric scale also keeps the API clean with no extra boolean flag needed.

Any thoughts or concerns about this approach ?

tjbck:
Closing in favour of 2328dc284ead6814836b600ce0cfa8dfdcfba956, Thanks!

Issue #19480, opened 2025-11-25
issue:  v0.6.40 cannot start anymore with TLS redis configured (ValueError: Unsupported Redis URL schema: rediss)


nick-hicx:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.6.40

### Operating System

Linux

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

The application is up and running healthy with the same configuration I had before upgrade to latest version using latest image from `ghcr.io/open-webui/open-webui:main` 

### Actual Behavior

The application cannot start anymore with Redis TLS

### Steps to Reproduce

1. Configure redis with TLS and start the application with a new DB with the version [v0.6.36]
2. Upgrade to v0.6.40 

## Quick Description

The image `ghcr.io/open-webui/open-webui:main` failed for me with the error 

`ValueError: Unsupported Redis URL schema: rediss`

## Suspects: 

Upgrade to python-socketio==5.14.0 see  https://github.com/open-webui/open-webui/commit/363ef194d8c41057dd0156dbfc8415f759ea752b

## Full error
`INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
[...]

PR #19488, opened 2025-11-25
fix: update dependency to prevent rediss:// failure

Classic298:
Fixes https://github.com/open-webui/open-webui/issues/19480

> [!NOTE]


Classic298:
@tjbck

tjbck:
Thanks!

98h398hrpohpoai:
Not sure if this was being tracked, but it applies to "ValueError: Unsupported Redis URL schema: unix" as well.

icep87:
Do we know when this will be released as this prevents upgrading to 0.6.40 that has the fix for the excessive SQL queries.

PR #19572, opened 2025-11-28
feat: Add smart RAG token threshold for automatic bypass optimization

eliem-ark:
Description
Adds an intelligent token-based threshold to automatically decide between RAG (chunking + embedding) and full context mode for uploaded files, optimizing performance for small documents while maintaining RAG for large files.

Context: Following [discussion #19177](https://github.com/open-webui/open-webui/discussions/19177). Cannot be implemented as a filter due to performance requirements.

Problem: All files are currently processed through chunking/embedding regardless of size, wasting 3-60s for small documents that fit in context windows.

Solution: Files ≤ threshold skip RAG processing entirely. Evaluated per-file: uploading a 1KB and 100KB file with 50K threshold will bypass RAG for the 1KB file only.

Added
Configuration: RAG_TOKEN_THRESHOLD (integer, default: 0 = disabled)
Environment variable: RAG_TOKEN_THRESHOLD
Admin Panel > Settings > Documents
Token counting: Uses existing tiktoken with TIKTOKEN_ENCODING_NAME
Metadata flag: Files bypassing RAG marked with bypass_rag: true in metadata
UI field: Number input (step: 1000, min: 0) in Documents settings, hidden when global bypass enabled
Logging: Decision logged per file: "File 'doc.pdf': 25000 tokens (<= 50000), bypassing RAG"

Changed
backend/open_webui/config.py: Added RAG_TOKEN_THRESHOLD PersistentConfig
backend/open_webui/main.py: Imported and initialized RAG_TOKEN_THRESHOLD
backend/open_webui/routers/retrieval.py:
Added RAG_TOKEN_THRESHOLD to ConfigForm, GET/POST endpoints
Enhanced process_file() with token counting logic
Set bypass_rag metadata flag when threshold bypassed
backend/open_webui/retrieval/utils.py: Check bypass_rag flag in get_sources_from_items()
src/lib/components/admin/Settings/Documents.svelte: Added threshold input field

Fixed
Performance: Small files (< threshold) upload a lot faster by skipping unnecessary chunking/embedding
Implementation Notes
fallback: Token counting errors proceed with normal RAG processing
Respects global BYPASS_EMBEDDING_AND_RETRIEVAL
Compatible with existing RAG settings (hybrid search, reranking, etc.)

Screenshots
Admin Panel > Settings > Documents

Token Threshold setting in Documents UI

Logs when uploading 2 files (only one above treshold set to 2k tokens for the example) :

image

Contributor License Agreement


rbsn-cpu:
Nice job !

tjbck:
Any reason why this couldn't just be done with chunk size param?

eliem-ark:
@tjbck It could, chunk count would be slightly faster. However the token count is independant of CHUNK_SIZE configuration, changing CHUNK_SIZE to improve quality won't silently change bypass behavior. The overhead on large files is negligible in practice. 
I chose token count for precision and config stability, but I'm happy to switch to chunk estimation if you prefer. Just let me know your preference and I'll update the PR accordingly.

sir3mat:
Any update on this feature to be merged?

Classic298:
The question above was never answered and there has been no activity since 24 March 2026. This also conflicts with dev now. Closing it as stale. If chunk size does not cover your case, open a discussion explaining your case

Generally this should also be covered by the native path already

PR #19672, opened 2025-12-02
fix: add on:click and on:mouseleave to Generate Title button

bshi61646:
## Summary
Add `on:click` and `on:mouseleave` handlers to the "Generate Title" button to fix silent failure in Safari.

## Problem
The button relied solely on the blur event's `relatedTarget` to trigger title generation, which behaves differently in Safari and causes the button to fail silently.

## Solution
- Add `on:click` handler to explicitly call `generateTitleHandler()`
- Add `on:mouseleave` handler to reset `ignoreBlur` flag

Fixes #19616


tjbck:
Addressed in dev

PR #19680, opened 2025-12-02
perf: Update milvus_multitenancy remove flush

Classic298:
### Description

Calling `collection.flush()` after every upsert creates severe performance bottlenecks and causes 'Rate limit exceeded' errors on standard Milvus configurations. It also leads to the 'small file problem' in MinIO/S3 and puts unnecessary load on etcd. Milvus handles persistence automatically via WAL and auto-flush policies. Removing manual flush allows Milvus to manage segments more efficiently.

> [!NOTE]


Classic298:
@tjbck

PR #19748, opened 2025-12-04
fix: add missing env var parameter pass through for enable async embedding

Classic298:
My original PR which included this didn't get merged unfortunately.

User reported in #19725 that when setting this only from environment variables the value is not correctly passed through to the functions.

| Location | `enable_async` passed? | Result |
|----------|----------------------|--------|
| `main.py` initialization | ❌ No | Always defaults to `True` |
| `save_docs_to_vector_db` | ❌ No | Always defaults to `True` |
| `update_embedding_config` | ✓ Yes | Works correctly |

**If you update the embedding config through the UI after startup, it would work correctly but not if you do it only via environment variables.**

Therefore there were two missing locations where this must be added.
These are added here.

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1031,6 +1031,7 @@ app.state.EMBEDDING_FUNCTION = get_embedding_function(
         if app.state.config.RAG_EMBEDDING_ENGINE == "azure_openai"
         else None
     ),
+    enable_async=app.state.config.ENABLE_ASYNC_EMBEDDING,
 )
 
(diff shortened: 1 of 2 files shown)


Classic298:
@tjbck I believe this should solve the issue in https://github.com/open-webui/open-webui/discussions/19725 - **the issue seems to be that the value of the env var - if only the env var is ever touched by the administrator - is not propagates to the functions correctly because of missing parameters.**

**The value IS being correctly propagated when configured exclusively through the UI - but not if exclusively done via environment variables**

tjbck:
Thanks!

PR #19770, opened 2025-12-05
sync fork

dctrng:


Classic298:
@dctrng Do you have an enterprise license? Rebranding the software if you have more than 50 registered users is only allowed with an enterprise license

dctrng:
Thanks for the info. I’m just using OWUI for developing & testing within my homelab, no rebranding or large-scale deployment planned.

PR #19791, opened 2025-12-06
FIX: Pipeline save settings - Handle undefined valves property

rgaricano:
### FIX: Pipeline save settings - Handle undefined valves property

When a Pipeline valve have a null value the settings isn't saved. The error occurs because the code tries to call `.split()` on a `null` value when saving pipeline valves. This happens when you set a valve to "None" (null) and then click save.

This PR Fix this issue.

___

> [!NOTE]


tjbck:
Thanks!

PR #19799, opened 2025-12-07
feat: introduce REDIS_SOCKET_CONNECT_TIMEOUT with sane default (for smooth failover)

Ithanil:
# Pull Request Checklist

- [x] **Target branch:** Verify that the pull request targets the `dev` branch. **Not targeting the `dev` branch will lead to immediate closure of the PR.**
- [x] **Description:** Provide a concise description of the changes made in this pull request down below.
- [] **Documentation:** If necessary, update relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs) like environment variables, the tutorials, or other documentation sources.
- [x] **Testing:** Perform manual tests to **verify the implemented fix/feature works as intended AND does not break any other functionality**. Take this as an opportunity to **make screenshots of the feature/fix and include it in the PR description**.
- [x] **Title Prefix:** To clearly categorize this pull request, prefix the pull request title using one of the following:

# Changelog Entry

### Description

I'm the author of Open WebUI's initial Redis Sentinel implementation and while I would like to believe that I confirmed failover working as expected in all relevant cases back then, I found that at least with the current implementation (which saw several changes) there are major issues with failover in the case when a node that is currently *master* goes offline *permanently*:
1) Open WebUI will turn unresponsive and not recover
2) At least if the *first node in the list of Sentinels* is the affected, the application will even hang on startup

I have done extensive research and testing to find the minimal set of changes to prevent both manifestations of the issue, specifically achieving:
1) Short (seconds, not minutes) downtime of the application
2) Chats / response streaming working immediately
3) Application can be restarted normally

It turned out that all 3 points get addressed by merely setting the option `socket_connect_timeout` for Redis and Sentinel constructors of redis-py to a sane value (e.g. 5 seconds). Per default it is `None`, leading to the observed permanent blocking.
Of course I also tried to set `socket_timeout`, but it is not required to achieve the effect and does not play well with pubsub in python-socketio (see https://github.com/miguelgrinberg/python-socketio/pull/1532). This is still the case with python-socketio 5.15.0, which is used since https://github.com/open-webui/open-webui/commit/4df5b7eb2e4b39eb01d488de4c95444ce6dea88c.
Furthermore, it is not enough to set the timeouts just for the Sentinel connections inside `sentinel_kwargs` (but setting `socket_connect_timeout` on the top level will apply to both Redis and Sentinel).

Also, if the timeout is set everywhere *except* for `AsyncRedisManager` of python-socketio, you will see the following effect:
1) Application becomes responsive after short break
2) Chats won't work for several minutes

So setting the option also for `AsyncRedisManager` is definitely warranted. In the present implementation this is achieved by adding the value of `REDIS_SOCKET_CONNECT_TIMEOUT` as `socket_connect_timeout` into the options given by `WEBSOCKET_REDIS_OPTIONS`, while a `socket_connect_timeout` present in the latter will take precedence. This ensures a sane default, while maintaing the configurability.

It is still possible to set  `REDIS_SOCKET_CONNECT_TIMEOUT` to `None` to achieve the current configuration. I have confirmed that setting the option to `None` restores the broken failover. Per default it will be 5 seconds, which ensures a short downtime without being too aggressive.

### Added

- env `REDIS_SOCKET_CONNECT_TIMEOUT`, default 5 seconds

### Changed

- apply `REDIS_SOCKET_CONNECT_TIMEOUT` as `socket_connect_timeout` to all Redis/Sentinel connections

### Fixed

- several issues with failover in Redis Sentinel setups

### Breaking Changes

- although the default is changed I don't consider this breaking in any way

### Notes

I'm happy to make the required addition to the docs as soon as the PR is approved.

> [!NOTE]

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -395,6 +395,18 @@ try:
 except ValueError:
     REDIS_SENTINEL_MAX_RETRY_COUNT = 2
 
+# Socket/Connect timeout for connections to Redis/Sentinel nodes
+REDIS_SOCKET_CONNECT_TIMEOUT = os.environ.get(
+    "REDIS_SOCKET_CONNECT_TIMEOUT", "5"
+)
+try:
+    if REDIS_SOCKET_CONNECT_TIMEOUT.lower() == "none":
+        REDIS_SOCKET_CONNECT_TIMEOUT = None
+    else:
+        REDIS_SOCKET_CONNECT_TIMEOUT = max(float(REDIS_SOCKET_CONNECT_TIMEOUT), 0.0)
+except ValueError:
+    REDIS_SOCKET_CONNECT_TIMEOUT = 5.0
+
 ####################################
 # UVICORN WORKERS
 ####################################
--- a/backend/open_webui/socket/main.py
+++ b/backend/open_webui/socket/main.py
@@ -32,6 +32,7 @@ from open_webui.env import (
     WEBSOCKET_SENTINEL_PORT,
     WEBSOCKET_SENTINEL_HOSTS,
     REDIS_KEY_PREFIX,
+    REDIS_SOCKET_CONNECT_TIMEOUT,
     WEBSOCKET_REDIS_OPTIONS,
     WEBSOCKET_SERVER_PING_TIMEOUT,
     WEBSOCKET_SERVER_PING_INTERVAL,
@@ -62,17 +63,22 @@ REDIS = None
 SOCKETIO_CORS_ORIGINS = "*" if CORS_ALLOW_ORIGIN == ["*"] else CORS_ALLOW_ORIGIN
 
 if WEBSOCKET_MANAGER == "redis":
+    mgr_redis_options = {**WEBSOCKET_REDIS_OPTIONS} if WEBSOCKET_REDIS_OPTIONS else {}
+    if "socket_connect_timeout" not in mgr_redis_options:
+        mgr_redis_options["socket_connect_timeout"] = REDIS_SOCKET_CONNECT_TIMEOUT
+
     if WEBSOCKET_SENTINEL_HOSTS:
         mgr = socketio.AsyncRedisManager(
             get_sentinel_url_from_env(
                 WEBSOCKET_REDIS_URL, WEBSOCKET_SENTINEL_HOSTS, WEBSOCKET_SENTINEL_PORT
             ),
-            redis_options=WEBSOCKET_REDIS_OPTIONS,
+            redis_options=mgr_redis_options,
         )
     else:
         mgr = socketio.AsyncRedisManager(
-            WEBSOCKET_REDIS_URL, redis_options=WEBSOCKET_REDIS_OPTIONS
+            WEBSOCKET_REDIS_URL, redis_options=mgr_redis_options
         )
+
     sio = socketio.AsyncServer(
         cors_allowed_origins=SOCKETIO_CORS_ORIGINS,
         async_mode="asgi",
--- a/backend/open_webui/utils/redis.py
+++ b/backend/open_webui/utils/redis.py
@@ -11,6 +11,7 @@ from open_webui.env import (
     REDIS_SENTINEL_MAX_RETRY_COUNT,
     REDIS_SENTINEL_PORT,
     REDIS_URL,
+    REDIS_SOCKET_CONNECT_TIMEOUT,
 )
 
 log = logging.getLogger(__name__)
@@ -162,6 +163,7 @@ def get_redis_connection(
                 username=redis_config["username"],
                 password=redis_config["password"],
                 decode_responses=decode_responses,
+                socket_connect_timeout=REDIS_SOCKET_CONNECT_TIMEOUT,
             )
             connection = SentinelRedisProxy(
                 sentinel,
@@ -188,6 +190,7 @@ def get_redis_connection(
                 username=redis_config["username"],
                 password=redis_config["password"],
                 decode_responses=decode_responses,
+                socket_connect_timeout=REDIS_SOCKET_CONNECT_TIMEOUT,
             )
             connection = SentinelRedisProxy(
                 sentinel,


Ithanil:
1. Yes.
2. Multi-instance deployment against 3-node Redis+Sentinel setup, shutting down current masters and observing behavior/logs (specifically: will reloading the web page and then streaming chat responses work as expected). Monitored current master/failover with a separate script also using redis-py, to ensure shutting down the correct node and failover actually taking place.

backend/open_webui/env.py
```diff
@@ -395,6 +395,18 @@ def parse_section(section):
 except ValueError:
     REDIS_SENTINEL_MAX_RETRY_COUNT = 2
 
+# Socket/Connect timeout for connections to Redis/Sentinel nodes
+REDIS_SOCKET_CONNECT_TIMEOUT = os.environ.get(
+    "REDIS_SOCKET_CONNECT_TIMEOUT", "5"
```
tjbck, line 400: Should be "" and convert to None by default.

Ithanil: Half the point of the PR is to provide a decent default and `None` is problematic, as explained in my PR description.

tjbck:
@Ithanil why couldn't we just use `WEBSOCKET_REDIS_OPTIONS` and provide more documentation here?

Ithanil:
> @Ithanil why couldn't we just use `WEBSOCKET_REDIS_OPTIONS` and provide more documentation here?

`WEBSOCKET_REDIS_OPTIONS` applies only to `AsyncRedisManager` of `python-socketio` and we should aim for a default configuration that fundamentally works. Currently it doesn't work in basically the worst way possible (i.e. application potentially can't even be restarted).

From the PR description
>In the present implementation this is achieved by adding the value of REDIS_SOCKET_CONNECT_TIMEOUT as socket_connect_timeout into the options given by WEBSOCKET_REDIS_OPTIONS, while a socket_connect_timeout present in the latter will take precedence. This ensures a sane default, while maintaing the configurability.

and 

>It is still possible to set REDIS_SOCKET_CONNECT_TIMEOUT to None to achieve the current configuration. I have confirmed that setting the option to None restores the broken failover. Per default it will be 5 seconds, which ensures a short downtime without being too aggressive.

backend/open_webui/env.py
tjbck: @Ithanil everything stay as default unless explicitly configured.

Ithanil: You're the boss. But it should be highlighted in the docs then that a Redis Sentinel configuration strictly requires setting this value.
So you want the env variable to default to empty/not set and this to be interpreted as `None`, correct?

tjbck:
Closing in favour of ba158d378f18706cc85231504ede1610509cb7c4, testing needed here!

Ithanil:
> Closing in favour of [ba158d3](https://github.com/open-webui/open-webui/commit/ba158d378f18706cc85231504ede1610509cb7c4), testing needed here!

Besides that I don't agree with your stance on the default, your implementation has other downsides compared to mine: https://github.com/open-webui/open-webui/commit/ba158d378f18706cc85231504ede1610509cb7c4#r172307477

Ithanil:
And to be honest: Sorry, I basically spent a whole day on this matter and if you quickly decide to replace my final result with an objectively worse commit don't count on me testing it.

tjbck:
@Ithanil I understand where you’re coming from, and I genuinely appreciate the depth of testing and analysis you invested into this. The challenge, however, is that while a 5-second socket_connect_timeout may indeed behave sanely in the specific Sentinel configurations you evaluated, we can’t confidently assert that it represents a universally correct baseline across all deployment scenarios. Open WebUI is deployed in an unusually broad spectrum of environments: single-node Redis instances running locally on bare metal; containerized Redis clusters behind overlay networks; multi-AZ Sentinel setups with uneven zone latency; distributed Kubernetes deployments with aggressive pod eviction policies; hybrid edge/cloud architectures with periodic link degradation; on-prem installations with slow rotational storage and high I/O contention; enterprise firewalled networks where handshakes are delayed by inspection layers; and even hobbyist setups on low-power ARM boards where network responsiveness varies wildly. With such variability, any attempt to prescribe a default timeout, even a well-reasoned one, inevitably risks constraining cases that diverge from your testbed. This is why we lean toward respecting whatever the user explicitly configures rather than imposing assumptions at the framework level. That said, your detailed investigation absolutely has value, and your effort in isolating, documenting, and proposing a solution for the failover behavior hasn’t gone unnoticed.

Ithanil:
> @Ithanil I understand where you’re coming from, and I genuinely appreciate the depth of testing and analysis you invested into this. The challenge, however, is that while a 5-second socket_connect_timeout may indeed behave sanely in the specific Sentinel configurations you evaluated, we can’t confidently assert that it represents a universally correct baseline across all deployment scenarios. Open WebUI is deployed in an unusually broad spectrum of environments: single-node Redis instances running locally on bare metal; containerized Redis clusters behind overlay networks; multi-AZ Sentinel setups with uneven zone latency; distributed Kubernetes deployments with aggressive pod eviction policies; hybrid edge/cloud architectures with periodic link degradation; on-prem installations with slow rotational storage and high I/O contention; enterprise firewalled networks where handshakes are delayed by inspection layers; and even hobbyist setups on low-power ARM boards where network responsiveness varies wildly. With such variability, any attempt to prescribe a default timeout, even a well-reasoned one, inevitably risks constraining cases that diverge from your testbed. This is why we lean toward respecting whatever the user explicitly configures rather than imposing assumptions at the framework level. That said, your detailed investigation absolutely has value, and your effort in isolating, documenting, and proposing a solution for the failover behavior hasn’t gone unnoticed.

Thanks for taking the time to for a deep response, that's greatly appreciated.

And in fact I took so much time to really make only a minimal change, with minimal risk of unexpected regressions. Still, I understand that you err on the side of keeping the redis-py default.
Let me inform you though that this default has been questioned in several issues and particularly, it seems like the valkey Python client uses a 5 seconds default even for `socket_timeout` (not `socket_connect_timeout`), which indeed leads to side-effects (see discussion here https://github.com/miguelgrinberg/python-socketio/pull/1532).

Let's settle it like it is. But as I said, this mandates highlighting in the documentation.

PR #19985, opened 2025-12-16
feat: Model Fallbacks for custom models from workspace section

Classic298:
# Motivation:

Imagine a large open webui instance. Users create custom models.

One day, an announcement gets sent out: "in 3 weeks we will remove gpt-5, please adjust your custom models until then to use a different base model".

But some users may not see the announcement, some users may be on vacation, or sick, or else.
The created custom models may be mission critical for certain processes inside the company - this is what this PR solves.
**If the configured base model for a custom model is not found, normally a "404: Model not found" error would appear.
Instead, the code inside openai.py, ollama.py and main.py will now attempt to fall back to the configured DEFAULT chat model configured by the admin inside the admin panel > models > settings > Default Models - and there use the first configured model.**

**ONLY if that configured default model also does not work (or is not configured), then an error is thrown for "404: Model not found".**

**All other requests are not affected by this change. This ONLY applies to custom models from the workspace section.**

> [!NOTE]


Classic298:
Tested locally, worked

Classic298:
made it optional via env var

Classic298:
disabled by default

backend/open_webui/env.py
```diff
@@ -453,6 +453,10 @@ def parse_section(section):
     os.environ.get("BYPASS_MODEL_ACCESS_CONTROL", "False").lower() == "true"
 )
 
+ENABLE_MODEL_FALLBACK = (
```
tjbck: `ENABLE_CUSTOM_MODEL_FALLBACK`

backend/open_webui/main.py
```diff
@@ -1534,11 +1535,32 @@ async def chat_completion(
     metadata = {}
     try:
         if not model_item.get("direct", False):
+            # Get model info first to check if it's a custom model
+            model_info = Models.get_model_by_id(model_id)
+
+            # Try fallback to default model for custom models if base model not found
+            if (
```
tjbck: Let's create a utility function for this instead.

tjbck (changes requested):
.

tjbck:
Closing in favour of b35aeb8f46e0e278c6f4538382c2b6838e24cc5a!

PR #19990, opened 2025-12-16
prexisting customizations

plasmabiscuit:


Classic298:
@plasmabiscuit you can only rebrand if you have less than 50 registered users, obtained an enterprise license or have written permission by Tim to rebrand. Be aware.

PR #20074, opened 2025-12-21
fix: Fix handling of absolute paths for SQLCipher databases

Classic298:
Fixes: https://github.com/open-webui/open-webui/issues/20051#issuecomment-3678747823

Previously, the connection logic incorrectly stripped the leading slash from `sqlite+sqlcipher` paths, forcibly converting absolute paths (e.g., `sqlite+sqlcipher:////app/data.db`) into relative paths (which became `app/data.db`). This caused database initialization failures when using absolute paths, such as with Docker volume mounts.

This change removes the slash-stripping logic, ensuring that absolute path conventions (starting with `/`) are respected while maintaining support for relative paths (which do not start with `/`).

> [!NOTE]


tjbck:
💪

PR #20228, opened 2025-12-28
fix: normalize local CrossEncoder reranking scores for relevance threshold

Classic298:
**Description:**

**Fix: Normalize local CrossEncoder reranking scores to 0-1 range**

The relevance threshold setting (0-1 range in the UI) doesn't work correctly with local CrossEncoder reranking models because MS MARCO models (the most commonly used rerankers) return raw logits (roughly -10 to +10) instead of normalized scores. When a user sets a threshold of 0.5, practically everything passes because even poor matches score above 0.5 in logit space.

**Solution:** Pass `activation_fn=torch.nn.Sigmoid()` to the CrossEncoder constructor. This is the approach recommended by the [sentence-transformers documentation](https://www.sbert.net/docs/cross_encoder/pretrained_models.html) for MS MARCO models and has the following benefits:

- Normalization runs on-device during inference rather than post-hoc in Python
- Threshold behavior now works as expected (0.5 = 50% relevance confidence)
- External rerankers are unaffected (they already return normalized scores via their APIs)

> "Note: You can initialize these models with activation_fn=torch.nn.Sigmoid() to force the model to return scores between 0 and 1. Otherwise, the raw value can reasonably range between -10 and 10.

> Example code: `model = CrossEncoder("cross-encoder/ms-marco-MiniLM-L6-v2", activation_fn=torch.nn.Sigmoid())`"

Source: https://www.sbert.net/docs/cross_encoder/pretrained_models.html

Edge case: Models that already output 0-1 scores (like STS-B) will have sigmoid applied twice, compressing their output range. **However, STS-B models are designed for semantic similarity, not reranking and <ins>should not be used for reranking</ins>** - in practice, **reranking pipelines use MS MARCO or BGE models.** Even if someone did use an STS-B model, **ranking order is preserved (sigmoid is monotonic) and thresholds still function**, just with compressed score magnitudes.

RE: https://github.com/open-webui/open-webui/discussions/19999

> [!NOTE]

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -765,8 +765,8 @@ else:
 # Whether to apply sigmoid normalization to CrossEncoder reranking scores.
 # When enabled (default), scores are normalized to 0-1 range for proper
 # relevance threshold behavior with MS MARCO models.
-RAG_RERANKING_MODEL_NORMALIZE_SCORES = (
-    os.environ.get("RAG_RERANKING_MODEL_NORMALIZE_SCORES", "True").lower() == "true"
+SENTENCE_TRANSFORMERS_CROSS_ENCODER_SIGMOID_ACTIVATION_FUNCTION = (
+    os.environ.get("SENTENCE_TRANSFORMERS_CROSS_ENCODER_SIGMOID_ACTIVATION_FUNCTION", "True").lower() == "true"
 )
 
 ####################################
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -762,6 +762,13 @@ else:
     except Exception:
         SENTENCE_TRANSFORMERS_CROSS_ENCODER_MODEL_KWARGS = None
 
+# Whether to apply sigmoid normalization to CrossEncoder reranking scores.
+# When enabled (default), scores are normalized to 0-1 range for proper
+# relevance threshold behavior with MS MARCO models.
+RAG_RERANKING_MODEL_NORMALIZE_SCORES = (
+    os.environ.get("RAG_RERANKING_MODEL_NORMALIZE_SCORES", "True").lower() == "true"
+)
+
 ####################################
 # OFFLINE_MODE
 ####################################
--- a/backend/open_webui/retrieval/utils.py
+++ b/backend/open_webui/retrieval/utils.py
@@ -913,9 +913,15 @@ def get_reranking_function(reranking_engine, reranking_model, reranking_function
             [(query, doc.page_content) for doc in documents], user=user
         )
     else:
-        return lambda query, documents, user=None: reranking_function.predict(
-            [(query, doc.page_content) for doc in documents]
-        )
+        import math
+        def sigmoid(x):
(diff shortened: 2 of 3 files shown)


rgaricano:
This solution os leess effective than native activation.

**Important Considerations**
External Rerankers: External rerankers might already return normalized scores ( https://github.com/open-webui/open-webui/blob/6f1486ffd0cb288d0e21f41845361924e0d742b3/backend/open_webui/retrieval/models/external.py#L61-L63 ) . Applying sigmoid again could distort already-normalized values. Consider adding a check to avoid double-normalization if scores are already in 0-1 range.

Threshold Adjustment: Your RELEVANCE_THRESHOLD must be set to 0-1 range values.

Performance: This adds Python-level math operations to every prediction call, which is less efficient than model-level activation.

Classic298:
@rgaricano i specifically avoided external rerankers in my PR. How do you come to the conclusion i double normalize them?

rgaricano:
@Classic298
If it's used a local CrossEncoder model that natively outputs scores in the 0-1 range, it will cause double normalization.
It may need to detect the model's output range and conditionally apply normalization only to models that output raw logits (-10 to 10 range). This could be done by checking the model documentation or testing a sample prediction before applying normalization.

Classic298:
@rgaricano the PR already only applies it to local models, not external models. How did you come to the conclusion my PR affects external rerankers?

But thanks for the feedback regarding potentially double-normalizing local reranker scores.

Regarding your concern about models that already output 0-1 scores (like the STS-B models): **in practice, reranking in RAG pipelines almost exclusively uses MS MARCO models since they're specifically trained for passage relevance ranking**. The STS-B models are designed for semantic similarity tasks, not reranking. But even if someone did use one, sigmoid is monotonic so ranking order is preserved, and the threshold behavior would still work (just with slightly compressed score magnitudes).

Instead of applying sigmoid post-hoc in get_reranking_function, I'm now passing activation_fn=torch.nn.Sigmoid() directly to the CrossEncoder constructor. This is the approach recommended by the sentence-transformers docs for MS MARCO models and has the benefit of running the activation on-device rather than in Python.

Classic298:
So the STS models that do output already normalized 0 to 1 values should not be used for reranking in the first place and therefore are out of scope here

backend/open_webui/env.py
```diff
@@ -762,6 +762,13 @@ def parse_section(section):
     except Exception:
         SENTENCE_TRANSFORMERS_CROSS_ENCODER_MODEL_KWARGS = None
 
+# Whether to apply sigmoid normalization to CrossEncoder reranking scores.
+# When enabled (default), scores are normalized to 0-1 range for proper
+# relevance threshold behavior with MS MARCO models.
+RAG_RERANKING_MODEL_NORMALIZE_SCORES = (
```
tjbck: `SENTENCE_TRANSFORMERS_CROSS_ENCODER_SIGMOID_ACTIVATION_FUNCTION`

Classic298:
@tjbck renamed please review

tjbck:
💪

PR #20258, opened 2025-12-30
feat: make changelog modal X button persist dismissal like "Okay, Let's Go!" button

Classic298:
## Summary
The X (close) button on the "What's New" changelog modal now has the same behavior as the "Okay, Let's Go!" button, properly persisting the dismissal so the modal doesn't reappear until the next update.

This will kill some frustrations some users have who do not know these buttons behave differently, that keep on opening issues again and again because they do not know they need to click on "Okay, Let's Go!" for the modal to permanently disappear.

## Problem
Previously, clicking the X button to close the changelog modal only set the version in localStorage and closed the modal. This was inconsistent with the "Okay, Let's Go!" button which also synced the version to user settings on the server. This could cause the modal to reappear unexpectedly in certain scenarios (e.g., clearing localStorage, logging in from another device).

## Solution
Updated the X button click handler in ChangelogModal.svelte to match the "Okay, Let's Go!" button behavior:
- Set version in localStorage
- Update the settings store with the current version
- Sync settings to the server via updateUserSettings API
- Close the modal
This ensures both dismissal methods have consistent behavior.

> [!NOTE]

### DIFF
--- a/src/lib/components/ChangelogModal.svelte
+++ b/src/lib/components/ChangelogModal.svelte
@@ -23,6 +23,13 @@
 		changelog = await getChangelog();
 	};
 
+	const closeModal = async () => {
+		localStorage.version = $config.version;
+		await settings.set({ ...$settings, ...{ version: $config.version } });
+		await updateUserSettings(localStorage.token, { ui: $settings });
+		show = false;
+	};
+
 	$: if (show) {
 		init();
 	}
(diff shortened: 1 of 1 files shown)


src/lib/components/ChangelogModal.svelte
tjbck, line 1: Should create a function instead.

tjbck (changes requested):

src/lib/components/ChangelogModal.svelte
Classic298: ok!

Classic298:
@tjbck updated

tjbck:
💪

PR #20260, opened 2025-12-30
feat: expose ADMIN_EMAIL setting in Admin Panel

### LINKED ISSUE #12500: feat: Have a separate setting for admin contact
### Check Existing Issues

### Problem Description

Currently, when the option to `Show Admin Details in Account Pending Overlay` is enabled, it will use the first User in the db table.
This is also why the view for /api/v1/users/update/role has a condition that directly disallows changing the role of the first User in the db table.
This isn't desirable because people might create other users and want to show those as contact info instead.

Create an additional field for the User model (or otherwise store this flag) for whether this is the Contact Info user.

### Alternatives Considered

Alternative could be to change the info (email + name) for that User, but if you've been using it and have lots of data there, then this is definitely not ideal, as changing things like the name will affect your prompting and experience if you have references to User name in the prompt.

### Additional Context

I'll try to draft a PR for this.

Classic298:
# Description

Adds an "Admin Contact Email" input field to the Admin Settings → General page, allowing administrators to configure which email is displayed in the "Account Pending" overlay directly from the UI.

Previously, this setting could only be changed via the ADMIN_EMAIL environment variable. This PR exposes it in the admin panel for easier configuration.

## Changes

**Backend (routers/auths.py):**
- Added ADMIN_EMAIL to get_admin_config endpoint response
- Added ADMIN_EMAIL to AdminConfig Pydantic model
- Added ADMIN_EMAIL handling in update_admin_config endpoint

**Frontend (components/admin/Settings/General.svelte):**
- Added conditional "Admin Contact Email" input field that appears when "Show Admin Details in Account Pending Overlay" is enabled
- Placeholder text explains the fallback behavior: "Leave empty to use first admin user"

## Behavior

- If a value is entered: That email will be displayed in the pending account overlay
- If left empty: Falls back to the first user in the database (existing behavior)
- Supports any email address, including ones not in the system (matching env var behavior)

Closes #12500

image

Tested, works

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/auths.py
+++ b/backend/open_webui/routers/auths.py
@@ -977,6 +977,7 @@ async def get_admin_details(
 async def get_admin_config(request: Request, user=Depends(get_admin_user)):
     return {
         "SHOW_ADMIN_DETAILS": request.app.state.config.SHOW_ADMIN_DETAILS,
+        "ADMIN_EMAIL": request.app.state.config.ADMIN_EMAIL,
         "WEBUI_URL": request.app.state.config.WEBUI_URL,
         "ENABLE_SIGNUP": request.app.state.config.ENABLE_SIGNUP,
         "ENABLE_API_KEYS": request.app.state.config.ENABLE_API_KEYS,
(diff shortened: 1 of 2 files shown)


tjbck:
💪

PR #20317, opened 2026-01-01
fix: inject full context knowledge into system message for KV prefix caching

Classic298:
### Summary
Fixes #20301 - Knowledge provided as "entire document" / "bypass embedding" is now injected into the system message instead of the last user message, enabling KV prefix caching to work correctly.

### Problem
Previously, full document knowledge was always appended to the last user message. Since the user message changes every turn, the knowledge block constantly shifted positions in the prompt, invalidating the KV cache and forcing the model to re-process the entire document on every interaction.

### Solution
- Static knowledge (full context / entire document mode) is now injected into the **system message**, which remains stable at the start of the conversation
- Dynamic content (web search results, tool outputs) continues to be injected into the user message to avoid caching stale data
- Uses rag_template with an empty prompt parameter to preserve formatting instructions while keeping the system message static across turns

> [!NOTE]


Classic298:
tested here https://github.com/open-webui/open-webui/discussions/20301#discussioncomment-15398184

tjbck:
We should introduce a env var for this instead of trying to use full context detection.

backend/open_webui/config.py
```diff
@@ -2717,6 +2717,12 @@ class BannerModel(BaseModel):
     os.getenv("RAG_FULL_CONTEXT", "False").lower() == "true",
 )
 
+RAG_KV_PREFIX_CACHING = PersistentConfig(
```
tjbck: RAG_SYSTEM_CONTEXT

backend/open_webui/utils/middleware.py
```diff
@@ -1600,15 +1600,50 @@ async def tool_function(**kwargs):
             raise Exception("No user message found")
 
         if context_string != "":
-            form_data["messages"] = add_or_update_user_message(
-                rag_template(
+            # Determine if using full context mode (static knowledge)
+            files = metadata.get("files", []) or []
```
tjbck: We pretty much just need the original implemenation but swapping between `add_or_update_user_message` and `add_or_update_system_message`

tftranslate:
I have a humble question here. I work with a rather powerful but slow model for certain reasons. I have adopted a manual process to talk with it about a couple of documents as context like this:
```

Please answer a question about document 1 and document 2 which are attached as context.

<context document 1>
.... lots of markdown ....
</context document 1>

<context document 2>
.... lots of markdown ...
</context document2>
```

That is, I manually paste my documents into the first prompt. Let's say the first prompt processes 20000 input_tokens (the documents are long). When I then ask a follow-up question, the answer is quick *and*  thanks to a large context window the model still knows the context (I could ask "please recite the first paragraph of document 2" and I will get a correct answer. The model only processes like 900 input_tokens for such a follow_uprequest. Splendid.

I want to achieve the same with open-webui when attaching documents and toggling the "use entire document" toggle. Folks have suggested my request may be related to this #PR but I am not sure.

I think I have tested out all possible combinations with RAG_SYSTEM_CONTEXT set to either true or false (I verified it indeed gets picked up by manually hacking the python code), with "Bypass RAG" in the admin console being set to true or false, when "Bypass RAG" being set to false then "Full context" in admin console being set to "true" or "false".

In each case, I always also toggle the toggle  "use full document" for each document that I upload.

My outcome with open-webui is consistently this: In settings in which the model can still answer questions like "Please recite paragraph 2 of document 1" in the followup-question, it will process the full 20000 tokens again, which is slow. I found some combination of the above paramters in which the model will not re-process the full 20000 tokens (RAG_SYSTEM_PROMPT false, "bypass RAG" true) but also only like 900  But then it cannot answer "please recite paragraph x of document y" type questions in the follow-up, and complains that it only knows that the documents exist, but has no access to their textual content.

Whereas in my manual process, I also only have 900 input_tokens on the follow_up questions but the model can still recite the context that I pasted manually in the previous prompt.

This is on  0.9.1 and has been like this on 0.8.12 as well.

Any help would be super appreciated. I think what I want is the most basic use case. Open-Webui has a lot of advanced features for handling RAG and document context, but I want the plain case: Fire-and-forget attached documents into the prompt to which they are attached, but in full! and do nothing more.

Please let me know whether this relates to this discussion or whether I should start a new one.

Issue #20340, opened 2026-01-02
issue: Model Presets send undefined as model ID, causing 404/400 errors


peetuhr:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.6.43

### Ollama Version (if applicable)

0.13.5

### Operating System

Windows 11

### Browser (if applicable)

Arc 1.85.1.3

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

Selecting a Model Preset should send the configured base model ID to the backend and generate a response.

### Actual Behavior

When using a Model Preset, the model ID is sent as `undefined` to the backend, resulting in "model not found" errors (404 from OpenRouter, 400 from Ollama).

Direct model selection from the dropdown works perfectly with the same models.

### Steps to Reproduce

1. Existing Open WebUI installation via Docker Compose
2. Configure OpenRouter connection in Admin Settings > Connections (or Ollama)
3. Navigate to Workspace > Models > Create New Model
4. Select any base model from dropdown (e.g., `anthropic/claude-sonnet-4`)
5. Save the preset
6. Start a new chat
7. Select the preset from the model dropdown
8. Send any message
9. **Expected:** Response from model
10. **Actual:** "Model not found" error (404/400)

### Logs & Screenshots

[...]

PR #20344, opened 2026-01-03
fix: workspace models returning "Model not found" error

Classic298:
### Problem
Workspace models (custom models created via Workspace > Models) were returning "Model not found" (404) errors when attempting to chat with them.

https://github.com/open-webui/open-webui/issues/20340

### Root Cause
When the generate_chat_completion functions in openai.py and ollama.py are called internally from chat.py (not via HTTP), the db parameter defaults to a Depends(get_session) object rather than an actual Session. This Depends object is truthy, so get_db_context attempted to use it as a session, causing Models.get_model_by_id to fail silently and return None.

With model_info being None, the base_model_id resolution was skipped, and the workspace model ID was passed directly to the API instead of its underlying base model.

### Solution
Added a check in both openai.py and ollama.py to verify that db is an actual Session instance before using it. If not, db is set to None so get_db_context creates a fresh session.

### Changes
- openai.py: Added isinstance check for db parameter in generate_chat_completion
- ollama.py: Added isinstance check for db parameter in generate_chat_completion

@tjbck 

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/ollama.py
+++ b/backend/open_webui/routers/ollama.py
@@ -1278,6 +1278,9 @@ async def generate_chat_completion(
         del payload["metadata"]
 
     model_id = payload["model"]
+    # When called internally (not via HTTP), db may be a Depends object, not a Session
+    if not isinstance(db, Session):
+        db = None
     model_info = Models.get_model_by_id(model_id, db=db)
 
     if model_info:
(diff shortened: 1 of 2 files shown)


silentoplayz:
I've tested this PR locally on my end and I can confirm that it solves https://github.com/open-webui/open-webui/issues/20340 for both my Ollama AI and OpenAI API connections and allows for Workspace models to work as expected again. LGTM!

tjbck:
Should be fixed on dev.

PR #20243, opened 2025-12-29
feat: Show Prompts/Models/Knowledge/Tools in Workspace when Read only - add proper visual cues for read only access

Classic298:
## Read-Only visual indicators and access control enforcement for Workspace items

# TL;DR: What does this PR do?

1. If a prompt/model/knowledgebase/tool is shared to a group with READ permission, **<ins>it will now be shown to other members in that group in the workspace section with a blue [READ ONLY] icon to clarify this item is read-only</ins>**
2. If another member of the group CLICKS on a workspace items with read-only permission, **<ins>he can view it, but all fields to Save any changes are **greyed out** and there is a tooltip informing them they cannot save any changes because they have insufficient permissions</ins>**
3. **<ins>PUBLICLY SHARED ITEMS like prompts, models etc. will ALSO be shown now in the workspace section, notably, also as [READ ONLY] labled items</ins>**, so all other users receiving this publicly shared item will be able to view/see it but not edit nor save it.
4. Administrators, unless disabled via BYPASS_ADMIN_ACCESS_CONTROL env var, <ins>**still retain full write access to all items regardless of permissions being set (tested locally)**</ins>

This PR enhances the UI/UX for workspace items (Prompts, Tools, Models, Knowledge Bases) by providing clear visual cues for read-only access and enforcing frontend restrictions.

#### Frontend
- **Workspace Lists**: Added a "Read" badge to Prompts, Tools, Models, and updated Knowledge Bases to use a consistent indicator for items where the user lacks write permission.
- **Editors**:
    - Updated Prompt, Tool, Model, and Knowledge Base editors to accept and respect a `write_access` flag.
    - Disabled all input fields for read-only items.
    - Disabled "Save" buttons and added tooltips explaining the permission restriction.

#### Backend
- **API Models**: Updated response schemas for Prompts, Tools, and Models to include a `write_access` boolean.
- **Routers**: Modified list endpoints to return all items with read access (instead of just write access) and correctly calculate the `write_access` status based on user role and permissions.

# Addresses:

https://github.com/open-webui/open-webui/discussions/15315
https://github.com/open-webui/open-webui/issues/20048
https://github.com/open-webui/open-webui/discussions/19664
https://github.com/open-webui/open-webui/issues/19660

### Contributor License Agreement

[...]

PR #20368, opened 2026-01-04
feat: Prompts Atomic PR of #20243

Classic298:

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/prompts.py
+++ b/backend/open_webui/routers/prompts.py
@@ -98,7 +98,7 @@ async def create_new_prompt(
 ############################
 
 
-@router.get("/command/{command}", response_model=Optional[PromptModel])
+@router.get("/command/{command}")
 async def get_prompt_by_command(command: str, user=Depends(get_verified_user), db: Session = Depends(get_session)):
     prompt = Prompts.get_prompt_by_command(f"/{command}", db=db)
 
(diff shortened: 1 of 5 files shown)


src/routes/(app)/workspace/prompts/edit/+page.svelte
```diff
@@ -29,6 +31,11 @@
 	onMount(async () => {
 		const command = $page.url.searchParams.get('command');
 		if (command) {
+			// Get prompt list to check write_access
+			const promptList = await getPromptList(localStorage.token).catch(() => []);
```
tjbck: write_access should be passed from getPromptById

backend/open_webui/routers/prompts.py
```diff
@@ -97,15 +108,17 @@ async def get_prompt_by_command(command: str, user=Depends(get_verified_user), d
             or prompt.user_id == user.id
             or has_access(user.id, "read", prompt.access_control, db=db)
         ):
-            return prompt
-        else:
-            raise HTTPException(
-                status_code=status.HTTP_401_UNAUTHORIZED,
-                detail=ERROR_MESSAGES.ACCESS_PROHIBITED,
-            )
+            return {
```
tjbck: PromptAccessResponse

src/routes/(app)/workspace/prompts/edit/+page.svelte
Classic298: addressed

backend/open_webui/routers/prompts.py
Classic298: Thanks for the feedback! I've updated the get_prompt_by_command endpoint to return PromptAccessResponse instead of a raw dict. This properly uses the Pydantic model for type safety while still including the write_access field in the response.

tjbck:
💪

PR #20369, opened 2026-01-04
feat: Models Atomic PR of #20243

Classic298:

> [!NOTE]

### DIFF
--- a/src/lib/apis/index.ts
+++ b/src/lib/apis/index.ts
@@ -39,8 +39,7 @@ export const getModels = async (
 		throw error;
 	}
 
-	// Handle both old format (res.data) and new format (res.items for ModelAccessListResponse)
-	let models = res?.data ?? (res?.items ?? []);
+	let models = res?.data ?? [];
 
 	if (connections && !base) {
 		let localModels = [];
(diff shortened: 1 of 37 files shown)


Classic298:
@tjbck Applied the same fix as prompts/tools - getModelById now returns ModelAccessResponse with write_access included. The edit page extracts write_access directly from the response instead of fetching the entire model list.

src/routes/(app)/workspace/models/edit/+page.svelte
tjbck, line 1: redundant.

src/lib/components/workspace/Models/ModelEditor.svelte
tjbck, line 1: redundant.

src/lib/apis/index.ts
```diff
@@ -39,7 +39,8 @@ export const getModels = async (
 		throw error;
 	}
 
-	let models = res?.data ?? [];
```
tjbck, line 42: why is this changed?

Classic298: The backend was updated to return {items, total} format (ModelAccessListResponse) instead of {data}. This change handles both formats for backward compatibility - it checks res.data first (old format), then falls back to res.items (new format).

Classic298:
tested works

tjbck:
💪

PR #20382, opened 2026-01-05
fix: Add missing y-protocols dependency for collaboration features

Classic298:
#  Summary
Added y-protocols as an explicit dependency in package.json to fix Vite build errors with collaborative editing features.

# Problem
The y-prosemirror package has y-protocols as a peer dependency, but it wasn't explicitly listed in package.json. This caused 
Vite to fail with "Could not resolve y-protocols/awareness" during dependency optimization.

#  Solution
Added "y-protocols": "^1.0.7" to the dependencies section to ensure it's installed and available for Vite's dependency 
optimizer.

#  Changes
Added y-protocols dependency

> [!NOTE]


tjbck:
Should be addressed in dev

PR #20427, opened 2026-01-06
fix: resolve N+1 query pattern in users endpoint

Classic298:
## Summary

Fixed N+1 query pattern in the `/api/v1/users` endpoint where groups were being fetched for each user individually.

### Problem

The `GET /api/v1/users` endpoint called `Groups.get_groups_by_member_id()` for each user, resulting in:
- 1 query for users
- N queries for groups (one per user)

### Solution

Added a new `Groups.get_groups_by_member_ids()` method that fetches groups for multiple users in a single query using SQL `IN` clause and `JOIN`.

### Result

- Before: 1 + N queries
- After: 2 queries total (1 for users, 1 for all groups)

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/groups.py
+++ b/backend/open_webui/models/groups.py
@@ -271,6 +271,27 @@ class GroupTable:
                 .all()
             ]
 
+    def get_groups_by_member_ids(
+        self, user_ids: list[str], db: Optional[Session] = None
+    ) -> dict[str, list[GroupModel]]:
+        """Fetch groups for multiple users in a single query to avoid N+1."""
+        with get_db_context(db) as db:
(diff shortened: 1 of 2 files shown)


backend/open_webui/models/groups.py
```diff
@@ -271,6 +271,27 @@ def get_groups_by_member_id(
                 .all()
             ]
 
+    def get_groups_by_member_ids(
```
tjbck, line 274: return type required

Classic298: there already is a return type though - what do you mean?

tjbck:
💪

PR #20431, opened 2026-01-06
fix: stop TTS playback when leaving chat window

M61ss:
# Changelog Entry

This pull request has not been written by an AI Agent.

### Description

The TTS playback continued speaking even if the user left chat window, so now it stops.

### Added

- N/A

### Changed

- N/A

### Deprecated

- N/A

### Removed

- N/A

### Fixed

- Stop TTS playback when leaving chat window

### Security

- N/A

### Breaking Changes

- N/A

---

### Additional Information

- Related to https://github.com/open-webui/open-webui/issues/3023

### Screenshots or Videos

https://github.com/user-attachments/assets/f3fe0aa9-679d-40a8-a5e9-3f9e9657e294

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Messages/ResponseMessage.svelte
+++ b/src/lib/components/chat/Messages/ResponseMessage.svelte
@@ -198,6 +198,10 @@
 		}
 	};
 
+	window.navigation.addEventListener("navigate", (event) => {
+	    stopAudio();
+	})
+
 	const speak = async () => {
 		if (!(message?.content ?? '').trim().length) {
 			toast.info($i18n.t('No content to speak'));
--- a/src/lib/components/chat/Messages/ResponseMessage.svelte
+++ b/src/lib/components/chat/Messages/ResponseMessage.svelte
@@ -201,6 +201,11 @@
 	window.navigation.addEventListener("navigate", (event) => {
 	    stopAudio();
 	})
+	document.addEventListener("visibilitychange", (event) => {
+	  if (document.visibilityState != "visible") {
+	    stopAudio();
+	  }
+	});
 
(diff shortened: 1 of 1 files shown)


Classic298:
why would i want it to stop if i leave the window? I think this should be configurable with user interface setting

M61ss:
I fixed that behaviour because there is an opened issue flagged as PR welcome. Make it configurable could be a nice idea, but requires more effort. Tell me your opinions.

M61ss:
Maybe it could be nice to open a separated pull request that makes configurable the feature that I implemented here in order to respect the concept of atomic updates. I'm new, please be patient🙏🏻

tjbck:
This is already the case.

M61ss:
Sorry, I don't undersand to what do you refer to with "this is already the case". Can you explain me please?

Classic298:
@M61ss 
they mean TTS already stops playback when you move to a different tab. @silentoplayz tested it

silentoplayz:
> @M61ss 
> 
> they mean TTS already stops playback when you move to a different tab. @silentoplayz tested it 

TTS playback stops if I swap to a different chat or page within my Open WebUI instance*. The TTS doesn't stop from simply swapping to a different browser tab or if I have minimized the browser tab.

M61ss:
Ok thank you all. 
So, what I have to change in my pull request to correct it? Should I make TTS stop only when moving to another tab or minimizing the window? Should I make this feature configurable from settings? 

> [!NOTE]
> In my OpenWebui instance the playback stops when moving to other chats, but not when moving to workspace, settings or other locations of the webapp. I tested it before to say.

PR #20456, opened 2026-01-07
fix(files): optimize file search with SQL pagination and database-level filtering

Classic298:
Replaced in-memory pagination with efficient SQL queries for the file search endpoint.

Changes:
- Add Files.search_files() method with SQL ILIKE filtering
- Add LIMIT/OFFSET pagination to search endpoint (skip/limit params)
- Convert fnmatch wildcards to SQL LIKE patterns (* -> %, ? -> %)
- Remove unused fnmatch import from router

Performance improvement: ~400x faster (2000ms -> 5ms)
- Before: Fetched ALL files, then filtered in Python memory
- After: SQL WHERE clause with pagination at database level

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/files.py
+++ b/backend/open_webui/models/files.py
@@ -233,6 +233,44 @@ class FilesTable:
                 for file in db.query(File).filter_by(user_id=user_id).all()
             ]
 
+    def search_files(
+        self,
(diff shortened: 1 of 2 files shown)


Classic298:
tested locally before and after - works

PR #20457, opened 2026-01-07
fix(notes): eliminate N+1 query in get_notes endpoint

Classic298:
Replaced per-note user lookup with batch fetch using SQL IN clause.

Changes:
- Fetch all note user_ids in a single pass
- Use Users.get_users_by_user_ids() for batch lookup
- Build user mapping to avoid DB calls in loop

Performance: Reduces N+1 queries to 2 queries (notes + users)

> [!NOTE]


tjbck:
084f0ef6a5491e186bf6b71c6386973ba18ef2fa

PR #20463, opened 2026-01-07
feat: Implement a global default tools setting

filiptrplan:
**AI Disclaimer**: I used Gemini CLI to create this PR but tested it manually and read all the code it generated.

# Changelog Entry

### Added

- Option to configure "Default Tools" in Settings > Tools. Users can now select specific tools in their settings that will be automatically enabled for every new chat session, regardless of the selected model.

---

### Additional Information

- Closes discussion: https://github.com/open-webui/open-webui/discussions/3390

### Screenshots or Videos

image

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -248,6 +248,16 @@
 		onSelectedModelIdsChange();
 	}
 
+	let settingsLoaded = false;
+
+	$: if (settingsLoaded === false && $settings && Object.keys($settings).length > 0) {
+		settingsLoaded = true;
+
+		if ($chatId === '' && Object.keys(history.messages).length === 0) {
+			setDefaults();
+		}
+	}
+
 	const onSelectedModelIdsChange = () => {
 		resetInput();
 		oldSelectedModelIds = JSON.parse(JSON.stringify(selectedModelIds));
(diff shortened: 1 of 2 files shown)


Classic298:
"Closes discussion: Allow user settings to permanently enable "Web search" #3390"

invalid

This discussion can already be closed because in the Interface settings there is already a toggle to permanently enable web search

Classic298:
Placement of this feature inside "external tools" when not necessarily external is questionable

filiptrplan:
> "Closes discussion: Allow user settings to permanently enable "Web search" #3390"
> 
> invalid
> 
> This discussion can already be closed because in the Interface settings there is already a toggle to permanently enable web search

While this is technically true for the main topic of the discussion, a lot of the more recent comments are asking for the exact feature I am implementing.

filiptrplan:
> Placement of this feature inside "external tools" when not necessarily external is questionable

I am open to suggestions. Should I place it in "General" or "Interface"? Or rename "External Tools" to "Tools"?

tjbck:
Admin-level tools can be set as default from model editor.

filiptrplan:
> Admin-level tools can be set as default from model editor.

I think this should be re-opened, as the current settings must be set for *each* model and by the *admin*. 

This PR gives the *users* the ability to set a set of tools for *all* the models.

It is a common use case and it would be a huge QoL improvement for the users and also for the admins that don't have to set toolkits for each model.

PR #20480, opened 2026-01-08
fix: prevent system prompt duplication in native function calling

jvadura:
## Summary

Prevents system prompt content from being duplicated during native function calling with MCP tools, which was causing quadratic token growth and excessive API costs.

## Problem

When using native function calling mode with MCP tools, each tool call iteration triggers `update_message_content()` which prepends the system prompt to the existing system message. This causes the same prompt to be duplicated multiple times:

```
Tool call 1: "System prompt" (~20k tokens)
Tool call 2: "System prompt\nSystem prompt" (~40k tokens)
Tool call 3: "System prompt\nSystem prompt\nSystem prompt" (~60k tokens)
```

**Impact:** A 20k token conversation can balloon to 3M+ tokens over multiple tool call iterations, causing massive unnecessary API costs.

## Root Cause

The bug occurs in the agentic tool call loop:

1. Initial request applies system prompt via `apply_system_prompt_to_body()` with `replace=True`
2. Each tool call iteration calls `generate_chat_completion()` again
3. The router applies model system prompt via `apply_system_prompt_to_body()` with `replace=False` (default)
4. This calls `update_message_content()` with `append=False`, which **prepends** the content
5. The same system prompt gets prepended on every iteration

## Fix

Add a check in `update_message_content()` to skip the update if the content is already present at the start of the existing message. This prevents duplicate prepending while preserving the ability to append genuinely new content.

## Test Plan

- [x] Enable native function calling mode
- [x] Connect MCP tool server
- [x] Trigger multiple tool calls
- [x] Verify token count stays stable (not growing by ~system_prompt_size each iteration)

**Verified:** Token count remains stable at ~12k tokens after 3 tool calls (previously would have grown to ~36k+).

Related: #19656

---

By submitting this pull request, I confirm that I have read and fully agree to the [Contributor License Agreement](https://github.com/open-webui/open-webui/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT.md) (CLA), and I am providing my contributions under its terms.


jvadura:
The total message tokens balooned 8k -> 15k -> 23k -> 29k - I logged it in llama.cpp, on top of that cache got trashed each round trip causing full prompt processing each time. I made my own logs:

```
{"time": "10:15:54", "action": "INSERT", "sys_before": 0, "sys_after": 1, "msg0_role": "user", "append": false, "content_preview": "# SFPI Knowledge Assistant - System Prompt v2.0\n\nYou are a specialized assistant for the **SFPI** (S", "all_roles": ["system", "user"]}
{"time": "10:16:21", "action": "UPDATE", "sys_before": 1, "sys_after": 1, "msg0_role": "system", "append": false, "content_preview": "# SFPI Knowledge Assistant - System Prompt v2.0\n\nYou are a specialized assistant for the **SFPI** (S", "all_roles": ["system", "user", "assistant", "tool"]}
{"time": "10:17:44", "action": "UPDATE", "sys_before": 1, "sys_after": 1, "msg0_role": "system", "append": false, "content_preview": "# SFPI Knowledge Assistant - System Prompt v2.0\n\nYou are a specialized assistant for the **SFPI** (S", "all_roles": ["system", "user", "assistant", "tool", "assistant"]}
```

From this it was obvious the system message was not duplicated, content was being added to the message itself. Then Opus applied a patch and after calling tools now:

image

1) llama.cpp prompt caching works so no more forced full prefill
2) token growth has been fixed:

```
new prompt, n_ctx_slot = 131072, n_keep = 0, task.n_tokens = 8964
new prompt, n_ctx_slot = 131072, n_keep = 0, task.n_tokens = 9509
new prompt, n_ctx_slot = 131072, n_keep = 0, task.n_tokens = 9811
```

However I cannot tell if this fix could cause any side effects. But since this appears to be a critical bug in tool call system, I wanted to at least inform you about my fix.

jvadura:
Closed, does not meet Pull Request Checklist. Apply the patch if you need a quick fix and test.

rgaricano:
@jvadura 

This is because each tool call iteration triggers `update_message_content()` which prepends rather than replaces the system prompt.

Another fix:

add

```
# Re-apply system prompt with replace=True to prevent duplication
if system_message := get_system_message(form_data.get("messages", [])):
    new_form_data = apply_system_prompt_to_body(
        system_message.get("content"),
        new_form_data, 
        metadata, 
        user, 
        replace=True  # Key fix: replace instead of prepend
    )
```
after  (dev branch) https://github.com/open-webui/open-webui/blob/c1147578c073a8c7fa7e7f836149e1cdfec8f18d/backend/open_webui/utils/middleware.py#L3396-L3408
or after (main branch) https://github.com/open-webui/open-webui/blob/a7271532f8a38da46785afcaa7e65f9a45e7d753/backend/open_webui/utils/middleware.py#L3049-L3061

The key fix is to ensure replace=True is passed when applying system prompts during native function calling iterations. This prevents the system prompt from being prepended multiple times and instead replaces the existing system message content.

Classic298:
https://github.com/open-webui/open-webui/commit/9223efaff0db6e56bfa157ef214d9590005156d2

PR #20529, opened 2026-01-09
fix: prevent event loop blocking during file processing 

jk-f5:
## Description

Fixes event loop blocking during file upload and document processing that causes liveness probe failures and pod restarts in Kubernetes deployments.

### Problem

`save_docs_to_vector_db()` used `asyncio.run()` to call the async embedding function, which blocks the entire event loop. During large file uploads, this prevents uvicorn from responding to any requests (including health checks), causing Kubernetes to kill pods.

Additionally, CPU-bound text splitting operations (`split_documents()`) can block the event loop for extended periods on large files.

### Solution

1. Convert `save_docs_to_vector_db()` and `process_file()` to async functions
2. Replace `asyncio.run(embedding_function(...))` with `await embedding_function(...)`
3. Wrap CPU-bound text splitting in `run_in_threadpool()` to prevent blocking
4. Update all callers in `files.py` and `retrieval.py` to await the async functions, including the inner `_process_handler` function

### Testing

I initially implemented this from branching off the v0.6.43 tag. I've been running this in our prod environment for about a week. It has been modified further due to changes on the dev branch to apply cleanly and account for upstream changes.

### AI disclaimer

AI wrote some of this MR description and reviewed the code for this MR. 

### DIFF
--- a/backend/open_webui/routers/files.py
+++ b/backend/open_webui/routers/files.py
@@ -115,7 +115,7 @@ def has_access_to_file(
 ############################
 
 
-def process_uploaded_file(
+async def process_uploaded_file(
     request,
     file,
     file_path,
@@ -124,7 +124,7 @@ def process_uploaded_file(
     user,
     db: Optional[Session] = None,
 ):
-    def _process_handler(db_session):
+    async def _process_handler(db_session):
         try:
             if file.content_type:
                 stt_supported_content_types = getattr(
@@ -139,7 +139,7 @@ def process_uploaded_file(
                         request, file_path_processed, file_metadata, user
                     )
 
-                    process_file(
+                    await process_file(
                         request,
                         ProcessFileForm(
                             file_id=file_item.id, content=result.get("text", "")
@@ -150,7 +150,7 @@ def process_uploaded_file(
                 elif (not file.content_type.startswith(("image/", "video/"))) or (
                     request.app.state.config.CONTENT_EXTRACTION_ENGINE == "external"
                 ):
-                    process_file(
+                    await process_file(
                         request,
                         ProcessFileForm(file_id=file_item.id),
                         user=user,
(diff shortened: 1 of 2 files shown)


Classic298:
when does this occur exactly? I never observed this behaviour of blocking event loop even with gigantic file uploads.

jannikstdl:
Noticed the same behavior sometimes with high workloads on Kubernetes landscapes.

Classic298:
Is this PR tested?

jk-f5:
> Is this PR tested?

I have this patch applied to v0.7.2 and is currently running in our prod environment with hundreds of concurrent users.

Classic298:
Can you confirm this is still the case?

Classic298:
Closing this PR — the core issue it addressed has already been resolved (i believe in 0.8.2), and the remaining changes would introduce breakage.

**Original problem (fixed):**
The original asyncio.run() call inside save_docs_to_vector_db would fail when called from within an already-running event loop. **This has since been replaced with asyncio.run_coroutine_threadsafe()**, which safely schedules the embedding coroutine on the main event loop from a worker thread. **So the critical bug this PR targeted no longer exists.**

**Why the remaining async conversion can't be merged as-is:**
The PR converts process_file, save_docs_to_vector_db, upload_file_handler, and process_uploaded_file to async, but several upstream callers of these functions are still sync:

- add_file_to_knowledge_by_id (sync) calls process_file directly
- update_file_from_knowledge_by_id (sync) calls process_file directly
- update_file_data_content_by_id (sync) calls process_file directly
- upload_image in images.py (sync) calls upload_file_handler
- upload_audio in utils/files.py (sync) calls upload_file_handler

If merged, these sync callers would receive coroutine objects instead of actual results. The embedding/processing would silently not happen, with files appearing to upload but never getting embedded.

A proper async conversion would need to also convert all these upstream callers and their callers in turn, making this a significantly larger refactor than what's scoped here.

Thank you for the contribution — the intent was correct and the issue was real, it's just been addressed differently in the meantime.

PR #20545, opened 2026-01-10
fix: release database connections immediately after auth instead of holding during LLM calls

Classic298:
fix: release database connections immediately after auth instead of holding during LLM calls

Authentication was using Depends(get_session) which holds a database connection for the entire request lifecycle. For chat completions, this meant connections were held for 30-60 seconds while waiting for LLM responses, despite only needing the connection for ~50ms of actual database work.

With a default pool of 15 connections, this limited concurrent chat users to ~15 before pool exhaustion and timeout errors:

    sqlalchemy.exc.TimeoutError: QueuePool limit of size 5 overflow 10 reached,
    connection timed out, timeout 30.00

The fix removes Depends(get_session) from get_current_user. Each database operation now manages its own short-lived session internally:

```
    BEFORE: One session held for entire request
    ──────────────────────────────────────────────────
    │ auth │ queries │ LLM wait (30s) │ save         │
    │         CONNECTION HELD ENTIRE TIME            │
    ──────────────────────────────────────────────────

    AFTER: Short-lived sessions, released immediately  
    ┌──────┐ ┌───────┐                 ┌──────┐
    │ auth │ │ query │   LLM (30s)     │ save │
    │ 10ms │ │ 20ms  │  NO CONNECTION  │ 20ms │
    └──────┘ └───────┘                 └──────┘
```

This is safe because:
- User model has no lazy-loaded relationships (all simple columns)
- Pydantic conversion (UserModel.model_validate) happens while session is open
- Returned object is pure Pydantic with no SQLAlchemy ties

Combined with the telemetry efficiency fix, this resolves connection pool exhaustion for high-concurrency deployments, particularly on network-attached databases like AWS Aurora where connection hold time is more impactful.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/auth.py
+++ b/backend/open_webui/utils/auth.py
@@ -44,8 +44,6 @@ from open_webui.env import (
 
 from fastapi import BackgroundTasks, Depends, HTTPException, Request, Response, status
 from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
-from sqlalchemy.orm import Session
-from open_webui.internal.db import get_session
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


tjbck:
LGTM

PR #20547, opened 2026-01-10
fix(files): prevent connection pool exhaustion in file status streaming

Classic298:
Refactored the file processing status streaming endpoint to avoid holding a database connection for the entire stream duration (up to 2 hours). Changes:
- Each status poll now creates its own short-lived database session instead of capturing the request's session in the generator closure
- Increased poll interval from 0.5s to 1s, halving database queries with negligible UX impact This prevents a single file status stream from blocking a connection pool slot for hours, which could contribute to pool exhaustion under load.

The old code also passed the entire file object but only ever used file.id from it. Passing just the ID is:
- Cleaner - explicitly shows we only need the ID
- Avoids keeping a reference to the full file object in the closure
- Functionally identical - same behavior

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/files.py
+++ b/backend/open_webui/routers/files.py
@@ -495,32 +495,35 @@ async def get_file_process_status(
         if stream:
             MAX_FILE_PROCESSING_DURATION = 3600 * 2
 
-            async def event_stream(file_item):
-                if file_item:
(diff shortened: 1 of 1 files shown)


tjbck:
LGTM, Thanks!

PR #20620, opened 2026-01-12
Added support for connecting to self hosted weaviate deployments using connect_to_custom replacing connect_to_local, which is better suited for cases where HTTP and GRPC are hosted on different ingresses.

rohithshenoy:
Added support for connecting to self hosted weaviate deployments using connect_to_custom replacing connect_to_local, which is better suited for cases where HTTP and GRPC are hosted on different ingresses. 

Testing:

Tested using weaviate hosted in a Container App on Azure with ports enabled for both 443 and 50051. Confirmed connection to weaviate is made correctly, objects are created on the node. 

> [!NOTE]


tjbck:
is this backward compatible?

rohithshenoyg:
> is this backward compatible?

Yes, connect_to_custom supports both locally hosted and remote weaviate instances. The earlier connect_to_local was only able to connect to localhost and ended up with an infinite 302 response from weaviate when you try to connect to https endpoint. 

Looks like it accidentally landed in main branch.  

Also non localhost deployments will almost always use seperate host names for grpc and https since the ingress would have to be different without significant effort so this support is critical for most weaviate deployments. 

Actually another common case would be support for weaviate cloud which uses connect_to_cloud. I imagine most people wouldn't like to self host and would endup not being able to connect to their weaviate cloud instance. I can work on another PR for that soon.

tjbck:
Thanks!

PR #20628, opened 2026-01-12
fix: Use handler when model builtin tools disabled

EntropyYue:
# Changelog Entry

### Description

- As described in #20608, Handler are always skipped for models with native function calls, even if builtin_tools are set to False. This PR solved it.

### Fixed

- Use handler when model builtin tools disabled

> [!NOTE]


tjbck:
https://github.com/open-webui/open-webui/discussions/20608#discussioncomment-15503204

PR #20638, opened 2026-01-13
perf: defer profile_image_url in user list endpoints

Classic298:
## Summary

Optimizes the user list API endpoints by preventing large base64 profile image data from being fetched from the database and serialized to JSON responses.

## Problem

The /api/v1/users/ endpoint was taking 1.5-2 seconds to return just 30 users because each user's profile_image_url field contained embedded base64 image data (~35KB per user), resulting in ~1MB response payloads.

## Solution

- Added defer_profile_image parameter to Users.get_users() that uses SQLAlchemy's defer() to exclude the profile_image_url column from database queries
- Override profile_image_url in responses with a URL path that points to the existing profile image endpoint

## Changes

- models/users.py: Add defer_profile_image parameter with defer(User.profile_image_url)
- routers/users.py: Pass defer_profile_image=True and override with /api/v1/users/{id}/profile/image

## Impact

| Metric | Before | After |
|--------|--------|-------|
| Response size (30 users) | ~1 MB | ~5 KB |
| Database I/O | Fetches ~35KB per user | Skips column |

## Non-breaking

The frontend already uses the /api/v1/users/{id}/profile/image endpoint for displaying avatars, so this change is transparent to clients.

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/users.py
+++ b/backend/open_webui/models/users.py
@@ -330,16 +330,12 @@ class UsersTable:
         filter: Optional[dict] = None,
         skip: Optional[int] = None,
         limit: Optional[int] = None,
-        defer_profile_image: bool = False,
         db: Optional[Session] = None,
     ) -> dict:
         with get_db_context(db) as db:
-            # Join GroupMember so we can order by group_id when requested
-            query = db.query(User)
-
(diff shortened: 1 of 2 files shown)


Classic298:
bump

tjbck:
Addressed in dev!

PR #20664, opened 2026-01-14
fix: await uploadFileHandler and preserve error message

lorenzophys:
# Changelog Entry

### Description

When uploading files not allowed by the `RAG_ALLOWED_FILE_EXTENSIONS` variable, there should be a red error message in the top-right corner of the screen explaining why the upload has failed. Instead OpenWebUI fails silently, displaying only a generic violation of the CSP in the JS console.

I found that the error message should already be printed, but `uploadFileHandler` is not awaited, so the message doesn't actually shows. Moreover, the default error message is too generic and the user will only see `Error uploading file`, so I added an `except` block to propagate a more specific error.

### Fixed

- `uploadFileHandler` is now awaited
- Show a better error message when upload fails due to CSP

---

### Additional Information

Issue I opened: https://github.com/open-webui/open-webui/issues/20662

### TEST: Screenshots or Videos

I performed a manual test:

Built the image with the changes:

```bash
docker build -t oui:await-fix .
```

Ran a container

```bash
docker run -d \
  -p 3000:8080 \
  -e RAG_ALLOWED_FILE_EXTENSIONS="pdf,txt,docx" \
  -v open-webui:/app/backend/data \
  --name open-webui-test \
  oui:await-fix
```

Clicked the plus icon, then the attach icon, selected a zip file and this is what shows:

Screenshot 2026-01-14 at 13 42 40

### DIFF
--- a/backend/open_webui/routers/files.py
+++ b/backend/open_webui/routers/files.py
@@ -332,8 +332,9 @@ def upload_file_handler(
                     detail=ERROR_MESSAGES.DEFAULT("Error uploading file"),
                 )
 
-    except HTTPException:
+    except HTTPException as he:
         # Re-raise HTTPException to preserve original message
+        log.exception(he)
         raise
     except Exception as e:
         log.exception(e)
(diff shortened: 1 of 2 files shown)


lorenzophys:
@tjbck why did you close it? If it's about the atomicity I can open two separate PRs
1. For the `await` fix here's the PR: https://github.com/open-webui/open-webui/pull/20788
2. For the `exception` block  fix I see somebody already fixed it

PR #20705, opened 2026-01-15
fix: coerce LLM tool call parameters to correct types in knowledge query

Classic298:
### Description

Fixes an issue where knowledge base queries would intermittently fail with "slice indices must be integers" errors.

**Root Cause:** When the LLM makes tool calls to query_knowledge_files, it sometimes formats parameters as strings instead of their proper types:
- count as "5" instead of 5
- knowledge_ids as "None" instead of null

This caused the slice operation in merge_and_sort_query_results to fail since you cannot slice a list with a string index.

**Fix:** Added type coercion at the start of query_knowledge_files to:
- Convert count from string to int (with fallback to default)
- Handle knowledge_ids being the literal strings "None", "null", or empty by converting to actual None
- Parse knowledge_ids as JSON array if passed as a string representation

Closes #20704

> [!NOTE]

### DIFF
--- a/backend/open_webui/tools/builtin.py
+++ b/backend/open_webui/tools/builtin.py
@@ -1599,6 +1599,25 @@ async def query_knowledge_files(
     if not __user__:
         return json.dumps({"error": "User context not available"})
 
+    # Coerce parameters from LLM tool calls (may come as strings)
+    if isinstance(count, str):
+        try:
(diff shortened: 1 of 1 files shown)


tjbck:
`enh` prefix would be more appropriate here, LGTM!

Issue #20641, opened 2026-01-13
issue: Web Search and Builtin Tools permissions break search


HenkieTenkie62:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.7.2

### Operating System

Pop os 22.04

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

Web search should only be executed when the integration is selected in chat or if it is selected as a default Feature for the model.

### Actual Behavior

Web search only works if Builtin Tools is also selected for the model.
Also when Builtin Tools is selected Search can't be toggled off in the interface without turning off Builtin Tools.

### Steps to Reproduce

Change the capabilities of the model to:
Web search ✅
Builtin Tools ✅
Search will work even without being enabled in chat, because of "Builtin Tools"

Change the capabilities of the model to:
Web search ✅
Builtin Tools ❌
Search will never work, even with being enabled in chat.

I'm fine with builtin tools like timestamps and memory.
But I'd like to see separate toggles for searching:
Web,
Chat history,
Knowledge base,
Notes

Randomly starting a search in chat history or knowledge base can be very useless.
Also in some chats information absolutely needs to stay offline.
If I can't control the online search I need to disable it to prevent information from leaking.

### Logs & Screenshots

No errors.

[...]

PR #20729, opened 2026-01-16
fix: respect built-in tool toggles in native function calling mode

Classic298:
# fix: built-in tool toggles ignored in native function calling mode

Fixes: https://github.com/open-webui/open-webui/issues/20641

## Problem

When using Native function calling mode with Built-in Tools capability enabled on a model, the chat interface toggles (web search ON/OFF, image generation ON/OFF, code interpreter ON/OFF) were completely ignored. All built-in tools were injected into the model's tool list regardless of the toggle state.

### Impact
**User Control Broken:**
- Users explicitly turning OFF web search still had web search tools available to the model
- Users could not prevent the model from performing unwanted actions (web searches, image generation, code execution)
- The toggle UI in the chat interface was effectively non-functional in Native mode

**Admin Control Broken:**
- Admins using inlet() filters to programmatically disable tools (e.g., disabling web search when documents are attached) had no effect
- Security and compliance policies could not be enforced via tool control in Native mode

**Inconsistent Behavior:**
- Default mode correctly respected toggles (toggles control whether task model is prompted about each tool)
- Native mode ignored toggles entirely
- This created confusion where the same UI behaved differently depending on function calling mode

### Reproduction Steps
1. Configure a model with Native function calling enabled and Built-in Tools capability ON
2. Start a new chat
3. Turn OFF the web search toggle in the chat interface
4. Send a message like "Search the web for the latest AI news"
5. Observe that the model still has search_web in its tool list and can perform searches

## Root Cause
In the get_builtin_tools function, the pattern for checking whether to inject a tool was inconsistent:

**Memory tools (correct pattern):**
- Checked features.get("memory") — the toggle state from the chat interface
- Tools only injected when toggle is ON

**Web search, image generation, code interpreter (broken pattern):**  
- Only checked global config (e.g., ENABLE_WEB_SEARCH) and model capability
- Did NOT check the toggle state from features dict
- Tools injected whenever global config and model capability allowed, regardless of toggle
The features dict is populated from form_data after inlet() filters run, so this also meant inlet() filter modifications to toggle state had no effect on Native mode tool injection.

## Solution
Added the missing feature toggle checks to align with the memory tools pattern:
**Web search tools:**
- Added: `and features.get("web_search")`
**Image generation/edit tools:**
- Added: `and features.get("image_generation")`
**Code interpreter tool:**
- Added: `and features.get("code_interpreter")`
Now all built-in tools follow the same three-layer control model:
1. Global config must enable the feature (admin-level master switch)
2. Model capability must allow it (per-model configuration)
3. Per-chat toggle must be ON (user control / inlet() filter control)

## Behavior After Fix
| Global Config | Model Capability | Chat Toggle | Tool Injected? |
|---------------|------------------|-------------|----------------|
| OFF | Any | Any | No |
| ON | OFF | Any | No |
| ON | ON | OFF | No |
| ON | ON | ON | Yes |

## Backward Compatibility
- Default mode behavior is completely unchanged (this code path is only for Native function calling)
- No changes to API contracts or data structures
- Existing configurations will work — users who had toggles ON will continue to have tools available

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/tools.py
+++ b/backend/open_webui/utils/tools.py
@@ -434,27 +434,34 @@ def get_builtin_tools(
     if features.get("memory"):
         builtin_functions.extend([search_memories, add_memory, replace_memory_content])
 
-    # Add web search tools if enabled globally AND model has web_search capability
-    if getattr(
-        request.app.state.config, "ENABLE_WEB_SEARCH", False
-    ) and get_model_capability("web_search"):
+    # Add web search tools if enabled globally AND model has web_search capability AND toggle is ON
+    if (
+        getattr(request.app.state.config, "ENABLE_WEB_SEARCH", False)
+        and get_model_capability("web_search")
+        and features.get("web_search")
+    ):
         builtin_functions.extend([search_web, fetch_url])
 
-    # Add image generation/edit tools if enabled globally AND model has image_generation capability
-    if getattr(
-        request.app.state.config, "ENABLE_IMAGE_GENERATION", False
-    ) and get_model_capability("image_generation"):
+    # Add image generation/edit tools if enabled globally AND model has image_generation capability AND toggle is ON
+    if (
+        getattr(request.app.state.config, "ENABLE_IMAGE_GENERATION", False)
+        and get_model_capability("image_generation")
+        and features.get("image_generation")
+    ):
         builtin_functions.append(generate_image)
-    if getattr(
-        request.app.state.config, "ENABLE_IMAGE_EDIT", False
-    ) and get_model_capability("image_generation"):
+    if (
+        getattr(request.app.state.config, "ENABLE_IMAGE_EDIT", False)
+        and get_model_capability("image_generation")
+        and features.get("image_generation")
+    ):
         builtin_functions.append(edit_image)
 
-    # Add code interpreter tool if enabled globally AND model has code_interpreter capability
+    # Add code interpreter tool if enabled globally AND model has code_interpreter capability AND toggle is ON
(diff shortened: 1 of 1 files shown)


Classic298:
CC @MarouaneZhani

Classic298:
Docs PR: https://github.com/open-webui/docs/pull/1011

Classic298:
tested

tjbck:
Intended.

Classic298:
@tjbck why is it intended? Other UI like Claude also have a toggle in the chat level that allow or disallow Claude to search the web. It does not force a web search - it decides if the capability is available. Just like this PR does

tjbck:
You should disable features from the model editor instead (re: chatgpt pattern), toggles for builtin tools are wip.

HenkieTenkie62:
@tjbck Good to hear the possibility for the toggles is being worked on, but why not have it available at the model card and in chat?

Toggles for search, searching notes, etc. would follow, the "Default Features" section already present in the model card.
And these could only be enabled if the "Capability" is selected on the model card, as already implemented for web search, image generation and code interpreter.

Z.ai, Kimi, Claude and VSCode use similar tactics.
If you want to emulate the no-user-choice ChatGPT style, maybe include a "hide Built-in tools in chat" toggle to leave the possibility open for power-users?
I think that way everybody wins.

Classic298:
added in dev https://github.com/open-webui/open-webui/commit/f1a1e64d2e9ad953b2bc2a9543e9a308b7c669c8

PR #20751, opened 2026-01-17
fix: handle None value for capabilities in get_model_capability helper

YuriNachos:
### Description
- Fixes #20565 - AttributeError when model capabilities are None

### Added
- N/A

### Changed
- Added `or {}` to handle None value for capabilities in `get_model_capability()` helper function

### Deprecated
- N/A

### Removed
- N/A

### Fixed
- **Model capabilities check**: Fixed AttributeError when model metadata capabilities field is None
- Previously, when a model's `meta.capabilities` was `None`, calling `.get()` on it would crash
- Now the code properly converts None to empty dict before accessing capabilities
- Affects models imported from external OpenAI-compatible connections with incomplete metadata

### Security
- N/A

### Breaking Changes
- N/A

---

### Additional Information
- Issue: #20565
- Applies the same fix pattern used in `middleware.py` to the `get_model_capability()` helper in `tools.py`
- Example affected models: XiaomiMiMo/MiMo-V2-Flash and others with `capabilities: None` in metadata
- Workaround of re-saving models in Admin Panel is no longer needed

### DIFF
--- a/backend/open_webui/utils/tools.py
+++ b/backend/open_webui/utils/tools.py
@@ -400,7 +400,7 @@ def get_builtin_tools(
         return (
             model.get("info", {})
             .get("meta", {})
-            .get("capabilities", {})
+            .get("capabilities") or {}
             .get(name, default)
         )
 


Classic298:
Did you test it?

Classic298:
I tested it, because you did not.

It does not work.

Test your fixes before submitting them as ready-for-review PRs as explained above.

marc1107 (commented):
Won't work, youre missing opening and closing (). Like that:
```python
 return (
            (model.get("info", {})
            .get("meta", {})
            .get("capabilities") or {})
            .get(name, default)
        )
```

Classic298:
thats not the reason why it does not work - even if you fix that it will not work. But reviewing it now is pointless since it's been closed 2 days ago

PR #20861, opened 2026-01-22
fix: add TTS voice selector to match admin audio settings UI

tacha6613:
### Description
In Admin Panel > Settings > Audio, there's a voice selector dropdown, but in Model Editor you have to type the voice ID manually. Added the same dropdown to Model Editor.

Uses the same logic as Admin Panel - shows browser voices for Web API, fetches from backend for other engines.

### Screenshots
**Admin Panel > Settings > Audio (reference):**
image
image

**Model Editor > TTS Voice:**
workspace -> models  
| Before | After |
|--------|-------|
image
| image

### Added
- TTS voice selector in Model Editor
### Fixed
- UX inconsistency between Admin Settings and Model Editor
---
By submitting this pull request, I confirm that I have read and fully agree to the Contributor License Agreement (CLA), and I am providing my contributions under its terms.

### DIFF
--- a/src/lib/components/workspace/Models/ModelEditor.svelte
+++ b/src/lib/components/workspace/Models/ModelEditor.svelte
@@ -2,11 +2,12 @@
 	import { toast } from 'svelte-sonner';
 
 	import { onMount, getContext, tick } from 'svelte';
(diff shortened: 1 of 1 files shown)


tjbck:
Addressed in dev!

PR #20868, opened 2026-01-22
Fix idle in transaction leaks in Open WebUI

Classic298:
The HTTP middleware was calling ScopedSession.commit() but not ScopedSession.remove(), causing database connections to remain "checked out" from the pool indefinitely. This resulted in "idle in transaction" connections in PostgreSQL that could persist for 30-50+ minutes.

With SQLAlchemy's scoped_session:
- commit() commits but keeps the session active
- remove() is required to return the connection to the pool

This fix adds the missing remove() call, ensuring connections are properly returned after each HTTP request.

Also includes IDLE_TRANSACTION_ANALYSIS.md documenting the full root cause analysis and additional recommendations.

> [!NOTE]


tjbck:
💪

Issue #20801, opened 2026-01-19
issue: HTML Preview renders code blocks inside thinking blocks


IDEDARY:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.7.1

### Operating System

Ubuntu 24

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

The preview should render only code that is outside of thinking block

### Actual Behavior

Merges all codeblocks into one and renders them, does not matter if they are inside thinking block

### Steps to Reproduce

1. Have OpenWebUI
2. Have thinking model that can code
3. Click preview on the code and you will see that it also used code blocks inside thinking block

### Logs & Screenshots

<img width="1860" height="935" alt="Image" src="https://github.com/user-attachments/assets/a6486441-8e19-467b-b6e3-3545070f7345" />

tjbck:
Testing wanted with dev branch here!

adhusch:
> Testing wanted with dev branch here!

Tested using [git-3567054-slim] (dev-slim). I see a model dependent behaviour. Works for GPT-OSS, for Mistral Magistral see screeshot:

[...]

PR #20877, opened 2026-01-22
fix: Artifacts render when code in reasoning

EntropyYue:
# Changelog Entry

### Description

- This resolves the issue where HTML, CSS, and JavaScript code generated during model inference may be incorrectly rendered as Artifacts.
- Fixes: #20801

### Fixed

- Fixed Artifacts render when code in reasoning

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -52,6 +52,7 @@
 		createMessagesList,
 		getPromptVariables,
 		processDetails,
+		removeDetails,
 		removeAllDetails,
 		getCodeBlockContents,
 		isYoutubeUrl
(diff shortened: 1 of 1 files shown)


src/lib/components/chat/Chat.svelte
```diff
@@ -845,7 +846,7 @@
 					html: htmlContent,
 					css: cssContent,
 					js: jsContent
-				} = getCodeBlockContents(message.content);
+				} = getCodeBlockContents(removeDetails(message.content, ['reasoning']));
```
tjbck, line 849: shouldn't it remove ALL details?

EntropyYue: Deleting all details will cause issues with the HTML containing them

tjbck:
Should be addressed in dev!

PR #20880, opened 2026-01-22
feat: support initMessage query param for auto-submitting prompt on load

ragaslan:
# What does this PR do?
 This PR adds support for sending an initial chat input to OpenWebUI when it is loaded inside an iframe. The chat message input can be optionally initialized via a URL query parameter, allowing the host application to automatically submit a text or prompt as soon as the **iframe** is ready.

gercek-ezgif com-video-to-gif-converter

## Why is this needed? 
When OpenWebUI is embedded in a cross-origin iframe, the parent application currently has no reliable way to send an initial prompt without user interaction. **So we can give an initial message to open-webui iframe in the parent origin easily.** 

## How does it work? 
- Reads an optional initMessage query parameter from the iframe URL
- On mount, sends a postMessage event to the window with the provided prompt

## How was this tested?
 - Embedded OpenWebUI in a host application running on a different port and domain
 - Verified that the initial postMessage is received by the window on load 
 - Tested in Chrome

### Example of usage
{0232BD7B-7A3F-4A4A-BD90-059EC7A46623}

> [!NOTE]


tjbck:
Open to discussion.

PR #20886, opened 2026-01-22
perf: reduce TTFT by caching model lookups in chat completion

Classic298:
fix: reduce TTFT by caching model lookups in chat completion

<ins>**massive perf improvement**</ins>

Skip expensive get_all_models() calls when models are already cached in app.state. This significantly reduces Time To First Token (TTFT) for chat completions and embeddings requests.

Previously, every request called get_all_models() which fetches model lists from all configured backends. Now we check the cache first and only call get_all_models() on cache miss.

Affected endpoints:
- openai: generate_chat_completion, embeddings
- ollama: embed, embeddings

Fixes #20069

@mickeytheseal
 co author

fixed for ollama and openai endpoints while we are at it

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/ollama.py
+++ b/backend/open_webui/routers/ollama.py
@@ -997,13 +997,16 @@ async def embed(
     log.info(f"generate_ollama_batch_embeddings {form_data}")
 
     if url_idx is None:
-        await get_all_models(request, user=user)
-        models = request.app.state.OLLAMA_MODELS
-
         model = form_data.model
 
         if ":" not in model:
             model = f"{model}:latest"
+
+        # Check if model is already in app state cache to avoid expensive get_all_models() call
+        models = request.app.state.OLLAMA_MODELS
+        if not models or model not in models:
+            await get_all_models(request, user=user)
+            models = request.app.state.OLLAMA_MODELS
 
         if model in models:
             url_idx = random.choice(models[model]["urls"])
@@ -1079,13 +1082,16 @@ async def embeddings(
     log.info(f"generate_ollama_embeddings {form_data}")
 
     if url_idx is None:
-        await get_all_models(request, user=user)
-        models = request.app.state.OLLAMA_MODELS
-
         model = form_data.model
 
         if ":" not in model:
             model = f"{model}:latest"
+
+        # Check if model is already in app state cache to avoid expensive get_all_models() call
+        models = request.app.state.OLLAMA_MODELS
+        if not models or model not in models:
+            await get_all_models(request, user=user)
+            models = request.app.state.OLLAMA_MODELS
 
         if model in models:
             url_idx = random.choice(models[model]["urls"])
--- a/backend/open_webui/routers/openai.py
+++ b/backend/open_webui/routers/openai.py
@@ -858,8 +858,13 @@ async def generate_chat_completion(
                 detail="Model not found",
             )
 
-    await get_all_models(request, user=user)
-    model = request.app.state.OPENAI_MODELS.get(model_id)
+    # Check if model is already in app state cache to avoid expensive get_all_models() call
+    # This significantly reduces TTFT when models are already cached
+    model = request.app.state.OPENAI_MODELS.get(model_id) if request.app.state.OPENAI_MODELS else None
+    if not model:
+        await get_all_models(request, user=user)
+        model = request.app.state.OPENAI_MODELS.get(model_id)
+
     if model:
         idx = model["urlIdx"]
     else:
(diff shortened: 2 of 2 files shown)


ssealus:
I used this one for cache check:
```py
    if (
        not request.app.state.config.ENABLE_BASE_MODELS_CACHE
        or not request.app.state.BASE_MODELS
    ):
        await get_all_models(request, user=user)
```

But using `request.app.state.OPENAI_MODELS` and `request.app.state.OLLAMA_MODELS` totally makes sense. Somehow i missed theese vars. Huge thanks!

jfahrenkrug:
Thank you for this! I'm experiencing exactly the same issue with TTFT of over 5 seconds for each chat message I send. I have activated both `ENABLE_BASE_MODELS_CACHE` and set `MODELS_CACHE_TTL` to 600, thinking it would mean that all "get models" lookups would now be cached, but it didn't improve things.

tjbck:
@Classic298 @mickeytheseal is this tested? could you share your findings?

ssealus:
@tjbck i found screenshots of my previous tests with different load profiles, but for some reason i am unable to upload any files here right now

| version  | metric | avg    | min    | max     | p99     | p90     | p50    | std    |
|----------|--------|-------:|-------:|--------:|--------:|--------:|-------:|-------:|
| no-patch | TTFT (ms) | 934.13 | 592.85 | 6,146.35| 3,534.66| 1,214.22| 799.51 | 575.84 |
| patch    | TTFT (ms) | 373.56 | 254.21 | 5,437.95| 1,137.93| 474.59  | 311.13 | 314.53 |


| version  | metric    | avg     | min    | max      | p99      | p90      | p50    | std    |
|----------|-----------|---------:|-------:|---------:|---------:|---------:|-------:|-------:|
| no-patch | TTFT (ms) | 1,111.05 | 617.75 | 6,419.03 | 5,605.47 | 1,363.99 | 803.79 | 963.16 |
| patch    | TTFT (ms) | 565.07   | 313.85 | 3,461.55 | 2,845.08 | 665.73   | 458.00 | 425.56 |

this one is running on a dev cluster for a few weeks now and i didnt observe any issues regarding models list or cache.

jfahrenkrug:
@Classic298 @mickeytheseal Thank you for this! I'd love to see this get merged. In order to fix the failing pipeline, I think you have to run `npm run format:backend` on your code and then commit those changes. The pipeline checks that there are no code changes after running the formatter (in other words: it requires the code to be properly formatted). Thank you! See https://github.com/open-webui/open-webui/actions/runs/21266350736/job/61206357518?pr=20886#step:6:1

Classic298:
Don't worry that's not necessary here

jfahrenkrug:
I'm curious: Which model were you seeing the slow TTFT with? I realized yesterday that in my case, the slowness came from using gpt-5-mini, which is a reasoning model. It wasn't the fault of get_all_models at all (in my case). Switching to gpt-4.1-mini significantly reduced TTFT. More info: https://community.openai.com/t/gpt-5-is-very-slow-compared-to-4-1-responses-api/1337859

Classic298:
I am not seeing slowness but @mickeytheseal and others are - and they can even measure it as from their benchmarks above so it definitely makes sense

and I strongly assume that the slowness is only an issue if you have to fetch many models from the foreign endpoint anytime
So i suspect
1) if the foreign endpoint is very fast to respond and
2) if not many models must be fetched

then the operation is fast

but if the foreign endpoint takes i.e. 300ms to respond then that is an additional 300ms that gets added to your request TTFT time

and if the foreign endpoint returns many many models and open webui has to process all of them that's another couple hundred ms to the TTFT time for each and every single request

Classic298:
I have re-reviewed the PR again

I do not see any issues with it whatsoever

it does not worsen the cache-staleness (which already exists thanks to user level cache) and it also could not land on attribute errors and the global caches are initialized on startup too

Furthermore if the cache really does not exist OR the model is not inside the cache, then we try to fetch the model list again anyways, i.e. falling back to the old behaviour.

ssealus:
@jfahrenkrug in my case it was non-reasosning model being inferenced via vLLM. So i made measures of 3 cases:
- vLLM directly. Lowest latency
- Through OWUI API. Highest latency as shown on my bench above.
- Through OWUI API with patch. Mid latency. Still higher then direct vLLM, but not as high as without patch.

As soon as there is only one model and inference service it couldnt produce such latency difference. 
I also profiled OWUI with py-spy, which showed me a big number of get_all_models() calls.

Classic298:
@mickeytheseal what other functions did you find that cause latency - if any?

ssealus:
@Classic298 just get_all_models(), but in 1-2 weeks i will do some load testing and profiling on the latest OWUI version. I will share the results as soon as i find something.

Classic298:
feel free to create a new discussion by then once you have new insights

PR #20899, opened 2026-01-23
feat: Add support for model parsing in OpenAI proxy handler

gaby:
### Description
- Ensure proxy forwards requests to the correct upstream by parsing JSON proxy payloads and honoring the `model` field when present.
- Use the resolved model routing in `OPENAI_MODELS` to pick the correct `OPENAI_API_BASE_URLS`/keys/configs instead of always using index `0`.
- If the model is not found, default to using index `0`.

### Motivation
- My backend is vLLM which supports /skills and /v1/responses endpoints. When using multiple models in OpenWebUI the proxy routes only use the OpenAI URL defined via ENV, if not present it defaults to whatever last model is on the list of OpenAI urls.
- The proxy routes do not allow using the models defined in OpenWebUI.
- This pull request will allow users to define models in OpenWebUI and proxy to their backends to use routes defined there, ex: skills, responses.

### Note
- The dev branch is failing ruff formatting, but from my checks the code in this file is correctly formatted.
- This PR was created with the help of codex and copilot.

> [!NOTE]


gaby:
@tjbck Thoughts on this?

gaby:
@Classic298 Can you take a look at this, thanks!

Classic298:
Hi, what do you wish for?

gaby:
@Classic298 For someone in the OpenWebUI team to refiew this.

Classic298:
I am not Open WebUI Team but I can review if you want?

gaby:
@Classic298 Yes, that would be helpful. Especially since you have so many contributions to the project.

I've only done 3-4 PR's to Open-WebUI

backend/open_webui/routers/openai.py
```diff
@@ -1077,7 +1077,26 @@ async def proxy(path: str, request: Request, user=Depends(get_verified_user)):
 
     body = await request.body()
 
+    payload = None
+    if body:
+        try:
+            payload = json.loads(body)
+        except json.JSONDecodeError:
+            payload = None
+
```
Classic298, line 1222: Silent failure on JSON decode here may cause confusing behavior. If the body exists but isn't valid JSON:
- Non-Azure path: Silently uses idx=0 (first backend)
- Azure path: Raises HTTP 400

Consider raising 400 immediately on decode failure (consistent behavior)

backend/open_webui/routers/openai.py
```diff
@@ -1077,7 +1077,26 @@ async def proxy(path: str, request: Request, user=Depends(get_verified_user)):
...
     idx = 0
+    if model:
+        # First try to use an existing model mapping to avoid unnecessary
+        # calls to get_all_models on every proxy request.
+        model_entry = request.app.state.OPENAI_MODELS.get(model)
+        if not model_entry:
+            await get_all_models(request, user=user)
+            model_entry = request.app.state.OPENAI_MODELS.get(model)
+        if model_entry:
+            idx = model_entry["urlIdx"]
```
Classic298, line 1234: The double-check on model_entry is correct (model might not exist even after refresh), but could be made clearer.

And a note: if the model still isn't found after refresh, we silently fall back to backend 0. Is that the intended behavior, or should we log a warning/raise an error?

Classic298: could also raise

else:
    raise HTTPException(status_code=404, detail="Model not found")

Classic298 (commented):
,

backend/open_webui/routers/openai.py
gaby: @Classic298 The idea was to fallback to idx=0, so to defaults to current behavior which is to use the openai api base url defined via ENV.

tjbck:
Addressed with abc9b63093d65f4d74342db85b7d5df1809aa0f0 in dev!

gaby:
> Addressed with [abc9b63](https://github.com/open-webui/open-webui/commit/abc9b63093d65f4d74342db85b7d5df1809aa0f0) in dev!

Thank you! 💪

gaby:
@tjbck @Classic298 I notice that when using a model alias, the `/openai` proxy returns 404. My guess is the router is not resolving the alias back to the base model for these handlers.

PR #20913, opened 2026-01-23
fix: clean up all event listeners in +layout.svelte onMount

sjhddh:
## Summary

Fix memory leaks in `+layout.svelte` by properly cleaning up all event listeners added in `onMount`.

## Problem

The `onMount` hook adds event listeners for `message`, `touchstart`, `touchmove`, `touchend`, `visibilitychange`, and `resize`, but the cleanup function only removes the `resize` listener. The anonymous touch handlers cannot be removed because they have no reference. This causes memory leaks during navigation and hot-reloads.

## Solution

1. Extract anonymous touch event handlers into named functions (`touchstartHandler`, `touchmoveHandler`, `touchendHandler`)
2. Add `{ passive: false }` to `touchmove` listener to allow `preventDefault()`
3. Update the `onMount` cleanup function to remove **all** event listeners

### Fixed

- Memory leaks caused by unremoved event listeners during component unmount
- Touch event handlers now properly cleaned up on navigation

### Changed

- Refactored anonymous touch handlers into named functions for proper cleanup

---

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -629,12 +629,12 @@
 			return nav && (el === nav || nav.contains(el));
 		}
 
-		document.addEventListener('touchstart', (e) => {
+		const touchstartHandler = (e) => {
 			if (!isNavOrDescendant(e.target)) return;
 			touchstartY = e.touches[0].clientY;
-		});
+		};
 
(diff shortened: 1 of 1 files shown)


tjbck:
Thanks!

PR #20931, opened 2026-01-25
fix: remove invalid expunge call on Pydantic FileModel

Classic298:
fix: remove invalid expunge call on Pydantic FileModel Files.get_file_by_id() returns a Pydantic FileModel, not an SQLAlchemy ORM object. Calling db.expunge() on a Pydantic model fails with UnmappedInstanceError since it lacks _sa_instance_state. The expunge was also unnecessary because subsequent DB updates already use fresh sessions via get_db() context manager.
Fixes #20925

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/retrieval.py
+++ b/backend/open_webui/routers/retrieval.py
@@ -1766,13 +1766,12 @@ def process_file(
                 }
             else:
                 try:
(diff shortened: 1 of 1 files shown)


silentoplayz:
I've tested this PR and can confirm that it solves #20925; LGTM!

PR #20945, opened 2026-01-26
enh: prompts

tjbck:


martijnn:
@tjbck 

What a great addition!

We'd like to see version control for system prompts of models too. But I don't know if that's an easy change and/or quick win.
Perhaps system promps should be able to reference to prompts?

Let me take this opportunity to express my gratitude for your contributions to this project ❤️

PR #20993, opened 2026-01-28
fix: prevent empty embeds attribute from rendering as {} in tool calls UI

cannontrodder:
## Summary

Fixes #20991

When a tool call has no embeds, an empty `{}` is incorrectly rendered in the UI. This is caused by the default value for `embeds` being an empty string `""` instead of an empty list `[]`.

## Changes

- Changed default value for `tool_result_embeds` from `""` to `[]`
- Only include `embeds` attribute in HTML when non-empty

## Root Cause

The empty string default gets JSON-encoded to `'""'` and HTML-escaped to `'&quot;&quot;'`. The frontend's `parseJSONString()` then fails to parse this correctly, resulting in a JSON parse error and `{}` being rendered.

## Testing

- **Before:** Tool calls without embeds show `{}` in red box and `SyntaxError: JSON.parse: unexpected character at line 1 column 1` in console
- **After:** Tool calls without embeds render cleanly with no spurious content

### Test steps performed:
1. Deployed OpenWebUI with AWS Bedrock backend via LiteLLM
2. Triggered image generation tool call ("Generate an image of a fluffy cow")
3. Observed `{}` rendering and console error before fix
4. Applied fix and verified clean rendering

## Checklist

- [x] Code follows project style guidelines
- [x] Minimal change to fix the issue
- [x] No breaking changes to existing functionality
- [x] Personally tested all changes

> [!NOTE]


Classic298:
please rebase and stash the commits

PR #21002, opened 2026-01-28
perf: resolve N+1 query in models list access control check

Classic298:
## Summary
Eliminates N+1 database query in the get_models endpoint (/list). Previously, the has_access function would query the database for user groups on each model in the list.
## Changes
Pre-compute user_group_ids once from the already-fetched groups (which were being fetched anyway for the filter), then pass to has_access via the user_group_ids parameter.
## Performance Impact
Before: N+1 queries - Groups.get_groups_by_member_id called once per model
After: 1 query total - reuses the groups already fetched for filter construction
The has_access function already supported the user_group_ids parameter to skip the DB lookup, it just wasn't being used here.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/models.py
+++ b/backend/open_webui/routers/models.py
@@ -83,12 +83,17 @@ async def get_models(
     if direction:
         filter["direction"] = direction
 
+    # Fetch user's groups once - used for both filter and access control checks
+    groups = []
     if not user.role == "admin" or not BYPASS_ADMIN_ACCESS_CONTROL:
         groups = Groups.get_groups_by_member_id(user.id, db=db)
         if groups:
             filter["group_ids"] = [group.id for group in groups]
 
         filter["user_id"] = user.id
+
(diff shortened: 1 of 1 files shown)


Classic298:
works locally

Classic298:
## Benchmark Results
Tested locally using a Python benchmark script simulating access control checks for 20 models against a PostgreSQL database (20 iterations).
| **N+1 (get_groups_by_member_id per model)** | 11,311,005 ns (11.311 ms) | 11,319,400 ns (11.319 ms) |
| **Batch (fetch groups once, reuse user_group_ids)** | 542,665 ns (0.543 ms) | 539,300 ns (0.539 ms) |
| **Improvement** | **95% faster (21x speedup)** | |
✅ Confirmed: Pre-fetching user groups and reusing user_group_ids for access control checks eliminates N+1 queries and provides 21x performance improvement for the models list endpoint.

PR #21005, opened 2026-01-28
perf: resolve N+1 query in SCIM group_to_scim user lookup

Classic298:
## Summary
Eliminates N+1 database query in SCIM group_to_scim function. Previously, when converting a group to SCIM format, each member's user was fetched individually.
## Changes
routers/scim.py group_to_scim function:
- Replace loop calling get_user_by_id per member
- Use single get_users_by_user_ids batch call
## Performance Impact
Before: N+1 queries (1 per group member)
After: 1 query total using SQL IN clause

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/scim.py
+++ b/backend/open_webui/routers/scim.py
@@ -352,18 +352,17 @@ def user_to_scim(user: UserModel, request: Request, db=None) -> SCIMUser:
 def group_to_scim(group: GroupModel, request: Request, db=None) -> SCIMGroup:
     """Convert internal Group model to SCIM Group"""
     member_ids = Groups.get_group_user_ids_by_id(group.id, db) or []
-    members = []
 
-    for user_id in member_ids:
-        user = Users.get_user_by_id(user_id, db=db)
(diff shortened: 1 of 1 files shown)


Classic298:
## Benchmark Results
Tested locally using a Python benchmark script comparing the N+1 pattern (individual user lookups in a loop) vs the batch pattern (single query with get_users_by_user_ids) against a PostgreSQL database (20 users, 20 iterations).
| **N+1 (current router code)** | 12,261,110 ns (12.261 ms) | 12,419,000 ns (12.419 ms) |
| **Batch (this PR)** | 926,855 ns (0.927 ms) | 926,400 ns (0.926 ms) |
| **Improvement** | **92% faster (13x)** | **93% faster** |
Note: The batch method get_users_by_user_ids already exists - this PR changes the router to use it instead of looping with individual lookups.
✅ Confirmed: Switching from N+1 individual lookups to the existing batch method provides ~13x performance improvement for groups with 20 members.

tjbck:
🚀

PR #21007, opened 2026-01-28
perf: resolve N+1 query in prompts list access control check

Classic298:
## Summary
Eliminates N+1 database query in get_prompt_list endpoint (/prompts/list). Previously, the has_access function would query the database for user groups on each prompt in the list.
## Changes
routers/prompts.py get_prompt_list endpoint:
- Initialize groups before the condition block
- Pre-compute user_group_ids from already-fetched groups
- Pass user_group_ids to has_access to skip redundant DB lookups
## Performance Impact
Before: N+1 queries - Groups.get_groups_by_member_id called once per prompt
After: 1 query total - reuses groups already fetched for filter construction

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/prompts.py
+++ b/backend/open_webui/routers/prompts.py
@@ -99,12 +99,17 @@ async def get_prompt_list(
     if direction:
         filter["direction"] = direction
 
+    # Fetch user's groups once - used for both filter and access control checks
+    groups = []
     if not (user.role == "admin" and BYPASS_ADMIN_ACCESS_CONTROL):
         groups = Groups.get_groups_by_member_id(user.id, db=db)
         if groups:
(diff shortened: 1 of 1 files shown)


Classic298:
## Benchmark Results
Tested locally using a Python benchmark script simulating access control checks for 20 prompts against a PostgreSQL database (20 iterations).
| **N+1 (get_groups_by_member_id per prompt)** | 11,257,090 ns (11.257 ms) | 11,298,400 ns (11.298 ms) |
| **Batch (fetch groups once, reuse)** | 560,445 ns (0.560 ms) | 541,000 ns (0.541 ms) |
| **Improvement** | **95% faster (20x speedup)** | |
✅ Confirmed: Pre-fetching user groups and reusing them for access control checks eliminates N+1 queries and provides 20x performance improvement.

PR #21011, opened 2026-01-28
perf: eliminate redundant SELECT after UPDATE in users.py

Classic298:
## Summary
Eliminates redundant database queries in 5 update methods in users.py. Previously, each method called update() followed by a separate query().first() to return the updated object.
## Changes
models/users.py - 5 methods updated to use fetch-modify-refresh pattern:
- update_user_role_by_id
- update_user_status_by_id  
- update_user_profile_image_url_by_id
- update_last_active_by_id
- update_user_by_id
Each method now:
1. Fetches user once
2. Modifies in place
3. Commits and refreshes
4. Returns the same object
## Performance Impact
Before: 2 queries per update (UPDATE + SELECT)
After: 1 query + commit + refresh on same object

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/users.py
+++ b/backend/open_webui/models/users.py
@@ -530,9 +530,12 @@ class UsersTable:
     ) -> Optional[UserModel]:
         try:
             with get_db_context(db) as db:
-                db.query(User).filter_by(id=id).update({"role": role})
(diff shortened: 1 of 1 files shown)


tjbck:
💪

PR #21012, opened 2026-01-28
perf: eliminate N+1 query for is_user_active in channels

Classic298:
## Summary
Eliminates N+1 database queries when computing is_active status for DM channel users. Previously, Users.is_user_active was called per user, causing a separate database query for each.
## Changes
routers/channels.py get_channels endpoint:
- Compute is_active directly from user.last_active_at instead of calling is_user_active
- Added time import for the 3-minute threshold calculation
- Uses same logic as is_user_active (active if last_active_at within 3 minutes)
## Performance Impact
Before: N queries for N users in DM channels
After: 0 additional queries - uses already-fetched user data

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/channels.py
+++ b/backend/open_webui/routers/channels.py
@@ -2,6 +2,7 @@ import json
 import logging
 import base64
 import io
+import time
 from typing import Optional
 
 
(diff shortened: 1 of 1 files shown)


Classic298:
## Benchmark Results
Tested locally using a Python benchmark script comparing is_user_active calls vs inline computation against a PostgreSQL database (20 users, 20 iterations).
| **N+1 (is_user_active per user)** | 10,938,080 ns (10.938 ms) | 10,980,600 ns (10.981 ms) |
| **Inline (compute from user.last_active_at)** | 1,720 ns (0.002 ms) | 1,400 ns (0.001 ms) |
| **Improvement** | **~100% faster (6,359x speedup)** | |
✅ Confirmed: Computing is_active directly from already-fetched user data eliminates N database round-trips, providing massive performance improvement for DM channels.

Classic298:
1149x speedup on sqlite

PR #21014, opened 2026-01-28
perf: eliminate duplicate requests on Prompts page load

Classic298:
## Summary
Fixes duplicate API requests and content flash when visiting the Prompts workspace page.
## Root Cause
Two reactive statements were both triggering on mount:
1. Query debounce block (fires when query goes from undefined to empty string)
2. Page/filter block (fires when page is set to 1 in onMount)
## Changes
src/lib/components/workspace/Prompts.svelte:
- Added queryInitialized guard to skip initial empty query trigger
- Added loaded check to page/filter reactive block
- Reset page to 1 when search query changes (UX improvement)
## Result
Single API request on page load instead of duplicates.

> [!NOTE]

### DIFF
--- a/src/lib/components/workspace/Prompts.svelte
+++ b/src/lib/components/workspace/Prompts.svelte
@@ -58,17 +58,23 @@
 
 	let page = 1;
 
-	// Debounce only query changes
+	// Debounce only query changes (skip initial empty query)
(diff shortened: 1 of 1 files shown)


silentoplayz:
Commented to say I've tested this PR on my end and it LGTM!

PR #9673, opened 2025-02-09
feat: Add config to pass OAuth token to OpenAI service

imranq2:
add ENABLE_FORWARD_OAUTH_TOKEN environment variable.  If this is set then pass the OAUTH token to OpenAI service.  This allows our OpenAI compatible service to authenticate the user and control access based on the user.

# Pull Request Checklist

### Note to first-time contributors: Please open a discussion post in [Discussions](https://github.com/open-webui/open-webui/discussions) and describe your changes before submitting a pull request.

https://github.com/open-webui/open-webui/discussions/8957

**Before submitting, make sure you've checked the following:**

- [x] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [x] **Description:** Provide a concise description of the changes made in this pull request.
- [ ] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ ] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [x] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation? No additional dependencies.
- [ ] **Testing:** Have you written and run sufficient tests for validating the changes?
- [ ] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [ ] **Prefix:** To cleary categorize this pull request, prefix the pull request title, using one of the following:
  - **BREAKING CHANGE**: Significant changes that may affect compatibility
  - **build**: Changes that affect the build system or external dependencies
  - **ci**: Changes to our continuous integration processes or workflows
  - **chore**: Refactor, cleanup, or other non-functional code changes
  - **docs**: Documentation update or addition
  - **feat**: Introduces a new feature or enhancement to the codebase
  - **fix**: Bug fix or error correction
  - **i18n**: Internationalization or localization changes
  - **perf**: Performance improvement
  - **refactor**: Code restructuring for better maintainability, readability, or scalability
  - **style**: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc.)
  - **test**: Adding missing tests or correcting existing tests
  - **WIP**: Work in progress, a temporary label for incomplete or ongoing work

# Changelog Entry

### Description

[...]

PR #21048, opened 2026-01-29
feat: Enable OAuth token forwarding with configurable token type for OpenAI-compatible backends

pivu-sbab:
## ✅ Description 

Enhances the OAuth integration by allowing system OAuth tokens (`access_token` or `id_token`) to be forwarded to OpenAI-compatible backend services. This enables backend services to receive the OAuth token and implement custom authorization logic based on token claims (such as user groups, roles, etc.) for `/v1/models` and `/v1/chat/completions` endpoints.

## ✅ Dependencies

No new dependencies.

## ✅ Testing

- Verified backward compatibility with existing bearer token configurations
- Tested access_token forwarding (default behavior)
- Tested id_token forwarding with Microsoft Entra ID
- Confirmed environment variable configuration parsing
- Validated Admin Panel UI functionality (see the `ID Token` option)

admin-panel-ui

Configuration example, either as an environment variable:

```bash
OPENAI_API_CONFIGS='{"0": {"auth_type": "system_oauth", "token_type": "id_token"}}'
```

Or through the Admin Panel:
-  Settings » Connections » Add Connection (/admin/settings/connections)

I have confirmed that my OpenAI service receives either the access_token (by default) or the id_token (if configured like above).

## ✅ Agentic AI Code and Code Review

AI has been used, but all suggestion has been rewritten and modified by me (human). Everything has also been manually tested by me, and code reviewed by two other colleagues in my team.

We have strived to have the most minimal git diff that fulfills the needs for this feature, no more, no less.

### Description

(See PR description above)

### Added

- Added `token_type` configuration option to the `system_oauth` authentication, supporting both `access_token` (default) and `id_token`
- Added JSON parsing support for the `OPENAI_API_CONFIGS` environment variable, used in containerized deployments
- Added token_type dropdown in the connection configuration modal (Settings » Connections » Add Connection (/admin/settings/connections)) with "Access Token" and "ID Token" options

### Changed

- Refactored `send_get_request()` to re-use `get_headers_and_cookies()`, instead of using the hardcoded Bearer key. 
- Updated `system_oauth` authentication to support configurable token type selection, allowing backend OpenAI services to receive OAuth Access tokens or ID tokens

### Breaking Changes

- **None**: All changes maintain full backward compatibility. Existing configurations continue to work without modification, `token_type` defaults to `access_token` when not specified, and `OPENAI_API_CONFIGS` parsing gracefully falls back to empty configuration on errors

## ✅ Additional Information

- Related PR/discussions
    - https://github.com/open-webui/open-webui/pull/9673
    - https://github.com/open-webui/open-webui/discussions/8957

> [!NOTE]


pivu-sbab:
The PR title prefix has now been changed:
```
feat.
feat:
```

pivu-sbab:
I saw your "Possible Scope Creep PR Detected".

I agree that the primary intent is "Enhance System OAuth authentication to support configurable token types (Access Token vs ID Token)."

But I think refactoring `send_get_request` is needed as well, otherwise the oauth_id_token will not be sent from Open WebUI to the `/v1/models` and `/v1/chat/completions` in the OpenAI backend service. Our main intent is to forward oauth_id_token to the OpenAI service.

I also need to be able to configure `id_token` instead of `access_token` (the default backward compatible option). I therefore think that in this PR we should keep at least the Admin Panel changes for this new option, but would be great if we also could add support for the `OPENAI_API_CONFIGS` environment variable. We would really like to provision all of our configuration and avoid click-operations in the Admin Panel.

backend/open_webui/config.py
```diff
@@ -1075,10 +1075,17 @@ def feishu_oauth_register(oauth: OAuth):
     "OPENAI_API_BASE_URLS", "openai.api_base_urls", OPENAI_API_BASE_URLS
 )
 
+OPENAI_API_CONFIGS_ENV = os.environ.get("OPENAI_API_CONFIGS", "{}")
```
Classic298: not a good env var name imho

Classic298 (commented):
.

tjbck:
id_token should be retrieved from the tool server side using the access token

pivu-sbab:
I guess you mean to use `TOOL_SERVER_CONNECTIONS` according to https://docs.openwebui.com/getting-started/env-configuration/#tool_server_connections?

But as I've understood it, the access token will be forwarded only for the `/v1/chat/completions` endpoint when a user starts a chat, right? But not for the `/v1/models` endpoint?

What do you think about just a change in `send_get_request()`, so it re-uses `get_headers_and_cookies()`?

pivu-sbab:
Never mind my previous question, I will use a pipe function. 

Just a last final question, do you have any plans on adding support so that pipe functions have access to the OAuth id_token when loading the models (in my case from `/v1/models`)? Could this be a good candidate for a new PR?

Currently my pipe function only has access to the id_token when calling `/v1/chat/completions`.

PR #21050, opened 2026-01-29
fix: Handle unclosed reasoning tags across tool call iterations

Classic298:
### Summary
Fixes text duplication and raw tag rendering when models use inline reasoning tags (like <think>) but don't close them before tool calls.

image

### Problem
Some models (e.g., MiniMax M2.1, Kimi K2) output reasoning content using inline <think> tags but don't always close them with </think> before making tool calls. On subsequent iterations:

1. The closing </think> tag would appear at the start of a new text block, rendering as raw text
2. New <think> tags would match against stale tags from previous iterations in the accumulated content variable
3. This caused thinking content to appear as plain text and duplicate across iterations

### Solution

1. **Orphaned end tag handling**: When a text block starts with a closing tag like </think>, we now look backwards through content_blocks to find an unclosed reasoning block and close it properly
2. **Scoped start tag matching**: Changed start tag matching to search in the current text block's content instead of the global accumulated content variable, preventing stale tags from being matched

### Testing

Tested with MiniMax M2.1 using multiple tool call iterations with web search - reasoning blocks now render correctly and text no longer duplicates.

> [!NOTE]


Classic298:
WORKS

Tested on MiniMax M2.1, GPT-5.2, Claude Haiku

image

image

image

Classic298:
# Key Insight
The fix we implemented doesn't try to change model behavior - it makes Open WebUI resilient to this quirky output pattern by:

1. Recognizing that </think> at the start of new content should close a previous reasoning block
2. Not getting confused by stale <think> tags from previous iterations

silentoplayz:
Tested on my end and LGTM!

noahkiss:
@Classic298 did this fix get merged elsewhere? Or why was it closed?

Classic298:
@noahkiss https://github.com/open-webui/open-webui/pull/21355

Classic298:
Addressed in dev!

Issue #21054, opened 2026-01-30
issue: some requests are blocking (no timeouts, not async)


jannikstdl:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

v0.7.2

### Operating System

Kubernetes / Linux

### Browser (if applicable)

Chrome

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

HTTP requests to external services (TTS, Images, Pipelines, Ollama, etc.) should fail gracefully with a timeout error when the configured service is unreachable, allowing the UI to remain responsive.

### Actual Behavior

When external services are unreachable (e.g., misconfigured URL, service down, network issues), the HTTP requests hang indefinitely because they
   have no timeout parameter set. This causes:                                                                                                    
  - The entire UI to hang/freeze                                                                                                                  
  - Admin settings pages to become unresponsive                                                                                                   
  - No feedback to the user about what's happening                                                                                                
  - Server threads/connections to be held open indefinitely  

### Steps to Reproduce

  1. Open Admin Settings and navigate to the relevant section (e.g. Audio under TTS)          
  3. Configure an external (OpenAI) service URL to an unreachable address (e.g., http://10.255.255.1:9999)                                                            
[...]

PR #21055, opened 2026-01-30
fix:  timeout of audio requests

jannikstdl:
### Description

- Some requests are blocking. One of them are the ones in `audio.py` as mentioned in #21054.

### Added

-  Added proper timeouts to audio requests

### Fixed

- UI not responding for minutes when the TTS URL for example is not responding.

---

### Additional Information

- #21054

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/audio.py
+++ b/backend/open_webui/routers/audio.py
@@ -53,6 +53,7 @@ from open_webui.env import (
     ENV,
     AIOHTTP_CLIENT_SESSION_SSL,
     AIOHTTP_CLIENT_TIMEOUT,
+    AIOHTTP_CLIENT_TIMEOUT_MODEL_LIST,
     DEVICE_TYPE,
     ENABLE_FORWARD_USER_INFO_HEADERS,
 )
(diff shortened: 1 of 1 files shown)


backend/open_webui/routers/audio.py
```diff
@@ -704,6 +706,7 @@ def transcription_handler(request, file_path, metadata, user=None):
                     headers=headers,
                     params=params,
                     data=file_data,
+                    timeout=AIOHTTP_CLIENT_TIMEOUT or 120,
```
tjbck, line 709: `or 120` should not be specified

tjbck:
The rest LGTM

jannikstdl:
@tjbck i noticed that many routes throughout the project do not have any timeouts or running still synchronous, like web search endpoints, SST, etc.
Would it be a plan to change all of these in a refac?

Classic298:
I don't think one universal timeout for all endpoints is a good idea, but having a dozen different config variables for the different endpoint timeouts is probably also not so smart

Making everything async is in process. Slow process but it's being done.

But higher priority right now should probably be fixing other, more pressing topics. That some endpoints are still sync is not that urgent imho

tjbck:
Addressed in dev.

PR #21158, opened 2026-02-04
fix: prevent worker death during document upload by using run_coroutine_threadsafe

Classic298:
Best solution would be to fully propagate async/await the entire chain but this is a lot more work.

Replace asyncio.run() with asyncio.run_coroutine_threadsafe() in save_docs_to_vector_db() to prevent uvicorn worker health check failures.

The issue: asyncio.run() creates a new event loop and blocks the thread completely, preventing the worker from responding to health checks during long-running embedding operations (>5 seconds default timeout).

The fix: Schedule the async embedding work on the main event loop using run_coroutine_threadsafe(). This keeps the main loop responsive to health check pings while the sync caller waits for the result.

Changes:
- main.py: Store main event loop reference in app.state.main_loop at startup
- retrieval.py: Use run_coroutine_threadsafe() instead of asyncio.run()

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -589,6 +589,10 @@ https://github.com/open-webui/open-webui
 
 @asynccontextmanager
 async def lifespan(app: FastAPI):
+    # Store reference to main event loop for sync->async calls (e.g., embedding generation)
+    # This allows sync functions to schedule work on the main loop without blocking health checks
+    app.state.main_loop = asyncio.get_running_loop()
+
     app.state.instance_id = INSTANCE_ID
     start_logger()
 
--- a/backend/open_webui/routers/retrieval.py
+++ b/backend/open_webui/routers/retrieval.py
@@ -1578,14 +1578,17 @@ def save_docs_to_vector_db(
             enable_async=request.app.state.config.ENABLE_ASYNC_EMBEDDING,
         )
 
-        # Run async embedding in sync context
-        embeddings = asyncio.run(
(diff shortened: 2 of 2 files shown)


Classic298:
Explanation here: https://github.com/open-webui/open-webui/discussions/21151#discussioncomment-15693804

Classic298:
@lewcpe feel free to test this

Classic298:
The issue is that asyncio.run() creates a new event loop and blocks the thread entirely, preventing the worker from responding to uvicorn health checks during long embedding operations. run_coroutine_threadsafe schedules the work back onto the main event loop, which stays responsive.

A few notes from the review:

**All callers are covered:**
- process_file (sync endpoint): FastAPI runs it in a threadpool, so run_coroutine_threadsafe + future.result() correctly bridges back to the main loop
- process_text and process_files_batch (async endpoints): both call save_docs_to_vector_db via run_in_threadpool, same pattern works
- Multi-worker is safe since each worker process gets its own event loop stored in its own app.state

**On the 600s timeout:** This is fine as a hardcoded safety net. The original asyncio.run() had no timeout at all, so this is actually stricter. 10 minutes is generous enough for any realistic embedding operation. If it takes longer, something external is broken and you want it to fail.

**Sentence-transformer path:** When embedding_engine is empty (local), the async embedding function uses asyncio.to_thread internally, which creates one extra thread hop (worker thread → main loop → to_thread). Works correctly and is no worse than the original behavior.

tjbck:
Thanks!

PR #21209, opened 2026-02-06
fix: handle error dict in citation parser when web search fails

veeceey:
### Description

Fixes #21070

When builtin tools like `search_web` fail (timeout, API error, etc.), they return `{"error": "..."}` as a JSON dict. The citation parser in `get_citation_source_from_tool_result` expected a list and iterated over the dict keys (strings), causing `AttributeError: 'str' object has no attribute 'get'`.

This fix parses JSON once at the top of the function and returns an empty list when an error dict is detected, allowing the chat to continue gracefully instead of crashing.

### Changed

- `backend/open_webui/utils/middleware.py`: Parse JSON result once at the top of `get_citation_source_from_tool_result()` and check for error dict before processing

### Fixed

- Fixed `AttributeError: 'str' object has no attribute 'get'` crash when web search tool returns an error response
- Eliminated redundant `json.loads` calls across tool branches by reusing the parsed result

### Testing

I have personally tested all changes:

1. Traced through the code path: when `search_web` returns `{"error": "timeout"}`, `json.loads` parses it into a dict. Iterating over a dict yields its keys (strings). Calling `.get()` on a string raises `AttributeError`. The fix catches this by checking `isinstance(parsed, dict) and "error" in parsed` before any iteration.
2. Verified normal web search still produces citations correctly - the parsed result flows into the same branches as before.
3. Verified `view_knowledge_file` and `query_knowledge_files` branches still work correctly since they now use the same pre-parsed result.

#### Fixed
- Fixed crash when web search tool returns an error response (e.g. timeout, API error) - the citation parser now gracefully returns empty citations instead of raising `AttributeError` (#21070)

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -171,9 +171,14 @@ def get_citation_source_from_tool_result(
     Returns a list of sources (usually one, but query_knowledge_files may return multiple).
     """
     try:
+        # Handle error responses from tools - all builtin tools return {"error": ...} on failure
+        parsed = json.loads(tool_result)
+        if isinstance(parsed, dict) and "error" in parsed:
+            return []
+
         if tool_name == "search_web":
             # Parse JSON array: [{"title": "...", "link": "...", "snippet": "..."}]
-            results = json.loads(tool_result)
+            results = parsed
             documents = []
             metadata = []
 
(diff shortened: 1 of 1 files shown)


veeceey:
This PR is atomic — all changes serve the single purpose of fixing the `AttributeError` crash when tools return error dicts (#21070).

The "DRY optimization" flagged by the bot is not a separate refactoring decision. To check for error dicts **before** any branch processes the result, `json.loads` must be called once at the top of the function. Once the result is already parsed there, reusing `parsed` in the three branches below is the only sensible thing to do — calling `json.loads(tool_result)` again on the same string would be redundant and wasteful.

In other words: the error-dict guard **requires** parsing at the top, and reusing that parsed value is an unavoidable side effect of the fix, not a separate change.

veeceey:
Closing in favor of #21250 which supersedes this PR with proper CLA format, PR template, and detailed test evidence.

PR #21238, opened 2026-02-07
fix: validate and normalize model tags to prevent UI crash

veeceey:
# Pull Request Checklist

- [x] **Testing:** Manually tested with unit tests showing before/after behavior.

# Changelog Entry

### Description

When model tags are stored as plain strings (e.g., `["tag1", "tag2"]`) instead of the expected `[{"name": "tag1"}]` format via direct API calls, the `/api/v1/models/tags` endpoint crashes with `AttributeError: 'str' object has no attribute 'get'`, causing a 500 error on the models list API and leaving the UI stuck in an infinite loading state.

### Fixed

- Added a Pydantic `model_validator` to `ModelMeta` that normalizes string tags to `{"name": "..."}` dict format on input
- Added defensive type checking in `get_model_tags` endpoint to handle both string and dict tag formats gracefully

## Test Results

```python
# Test 1: String tags get normalized to dict format
meta = ModelMeta(tags=['tag1', 'tag2'])
assert meta.model_dump()['tags'] == [{'name': 'tag1'}, {'name': 'tag2'}]  # PASS

# Test 2: Dict tags pass through unchanged  
meta = ModelMeta(tags=[{'name': 'tag1'}, {'name': 'tag2'}])
assert meta.model_dump()['tags'] == [{'name': 'tag1'}, {'name': 'tag2'}]  # PASS

# Test 3: No tags - no crash
meta = ModelMeta()
assert 'tags' not in meta.model_dump()  # PASS

# Test 4: Mixed/invalid tags - invalid entries skipped, strings normalized
meta = ModelMeta(tags=['tag1', {'name': 'tag2'}, 123])
assert meta.model_dump()['tags'] == [{'name': 'tag1'}, {'name': 'tag2'}]  # PASS

# Test 5: get_model_tags handles both formats without crash
tags_set = set()
for tag in [{'name': 'a'}, 'b', 42]:
    if isinstance(tag, dict):
        tag_name = tag.get('name')
    elif isinstance(tag, str):
        tag_name = tag
    else:
        continue
    if tag_name:
        tags_set.add(tag_name)
assert sorted(tags_set) == ['a', 'b']  # PASS
```

**Before fix:** API call with string tags causes 500 Internal Server Error, UI shows infinite loading.  
**After fix:** String tags are auto-normalized to the correct format. Existing data with string tags is handled gracefully.

Fixes #20819

### DIFF
--- a/backend/open_webui/models/models.py
+++ b/backend/open_webui/models/models.py
@@ -9,7 +9,7 @@ from open_webui.models.groups import Groups
 from open_webui.models.users import User, UserModel, Users, UserResponse
 
 
-from pydantic import BaseModel, ConfigDict
+from pydantic import BaseModel, ConfigDict, model_validator
 
 from sqlalchemy import String, cast, or_, and_, func
 from sqlalchemy.dialects import postgresql, sqlite
@@ -48,7 +48,20 @@ class ModelMeta(BaseModel):
 
     model_config = ConfigDict(extra="allow")
 
-    pass
+    @model_validator(mode="before")
+    @classmethod
+    def validate_tags(cls, data):
+        if isinstance(data, dict) and "tags" in data:
+            tags = data["tags"]
+            if isinstance(tags, list):
+                normalized = []
+                for tag in tags:
+                    if isinstance(tag, str):
+                        normalized.append({"name": tag})
+                    elif isinstance(tag, dict) and "name" in tag:
+                        normalized.append(tag)
+                data["tags"] = normalized
+        return data
 
 
 class Model(Base):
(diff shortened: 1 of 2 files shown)


veeceey:
This PR is atomic — both changes address the same bug: model tags stored as strings instead of lists causing crashes.

- The **model validator in `models.py`** ensures new/updated data is validated at the schema level, preventing malformed tags from being persisted going forward.
- The **defensive handling in `routers.py`** is necessary because existing data in the database may already contain strings instead of dicts/lists. Without this, the validator alone wouldn't protect against crashes when reading legacy data.

Submitting the router fix separately wouldn't make sense — it would be an incomplete fix that only handles reads but still allows bad data to be written. Similarly, the validator alone would fix writes but leave existing malformed data to crash on read. The two changes are complementary halves of the same fix.

veeceey:
Closing in favor of #21249 which supersedes this PR with proper CLA format, PR template, and detailed test evidence.

PR #21272, opened 2026-02-09
enh.: Allow to disable icon inversion

Classic298:
### Problem
Icons were being inverted in dark mode based on whether the string `svg` 
appeared *anywhere* in the icon URL. This caused PNG icons to randomly 
get inverted if their base64 encoding happened to contain "svg" by coincidence.

### Solution
- Added configurable `icon_dark_invert` frontmatter property for actions
- Changed logic to:
  1. Use explicit `icon_dark_invert: true/false` if defined in docstring
  2. Fall back to checking actual MIME type (`startsWith('data:image/svg')`) 
     instead of `.includes('svg')`

### Usage
```python
"""
title: Export to PDF
icon_url: data:image/png;base64,...
icon_dark_invert: false
"""
```

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Messages/ResponseMessage.svelte
+++ b/src/lib/components/chat/Messages/ResponseMessage.svelte
@@ -1397,7 +1397,7 @@
 													<div class="size-4">
 														<img
 															src={action.icon}
-															class="w-4 h-4 {action.icon.includes('svg')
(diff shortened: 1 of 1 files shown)


Classic298:
noooooooooooooooooooooooooooooooooooo

PR #21294, opened 2026-02-10
feat: Enable Rich UI embed support for action functions

Classic298:
Action functions can now return HTMLResponse objects or (html, headers) tuples with Content-Disposition: inline to render rich UI iframes in chat, matching the existing tool behavior.

# Why?

<ins>**Actions and Tools are extremely similar - the only difference is who activates them**</ins>

Tools get activated by LLMs
Actions get activated by Humans

Otherwise they behave the same. They get called on-demand and do something

So they should have the same capabilities and abilities for developers.

If you want to display beautiful graphs and charts with Actions, you have to inject a codeblock with the HTML content into the assistant's message by modifying the body - and even then it is only shown as an Artifact and not as a rich UI embed like the tools have available.

---

How action authors can now use Rich UI
Option A — HTMLResponse:

```
from fastapi.responses import HTMLResponse

async def action(self, body, __event_emitter__=None):
    html = "<html><body><h1>Dashboard</h1></body></html>"
    return HTMLResponse(content=html, headers={"Content-Disposition": "inline"})
```

Option B — Tuple with headers:

```
async def action(self, body, __event_emitter__=None):
    html = "<h1>Interactive Chart</h1><script>...</script>"
    return (html, {"Content-Disposition": "inline", "Content-Type": "text/html"})
```

No frontend changes were needed — the socket handler at Chat.svelte:416 already receives "embeds" events generically and renders them via FullHeightIframe.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/chat.py
+++ b/backend/open_webui/utils/chat.py
@@ -448,6 +448,30 @@ async def chat_action(request: Request, action_id: str, form_data: dict, user: A
             else:
                 data = action(**params)
 
+            # Process action result for Rich UI embeds (HTMLResponse, tuple with headers)
+            # Deferred import to avoid circular dependency (middleware imports chat)
+            from open_webui.utils.middleware import process_tool_result
+
+            processed_result, _, action_embeds = process_tool_result(
+                request,
+                action_id,
+                data,
+                "action",
+            )
+
+            if action_embeds:
+                await __event_emitter__(
+                    {
+                        "type": "embeds",
+                        "data": {
+                            "embeds": action_embeds,
+                        },
+                    }
+                )
+                # Replace data with the processed status dict so we don't
+                # try to serialize the raw HTMLResponse / tuple back to the client
+                data = processed_result
+
         except Exception as e:
             raise Exception(f"Error: {e}")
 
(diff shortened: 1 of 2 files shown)


Classic298:
Our specific usecase is to trigger a download by pressing a button on iOS Open WebUI PWA app, which just does not allow downloads.

iOS does not permit downloads by just clicking a download button WHEN INSIDE a pwa app, because pwa apps use a sandboxed webkit environment.

But we found out that by triggering the download using a javascript blob, we can work around this and still have a download.

But for this to be displayed as a button with javascript executed when the button is pressed.... we need rich UI.

And also for other usecases like displaying graphs and charts we are working on. Tools can do this too, but we have some use cases where a human should trigger it, not the model.

Classic298:
Tested working

image

Action triggering rich UI embed

Classic298:
.@Fu-Jie i bet you could make good use of this

Fu-Jie:
@Classic298  Wow, this change clearly boosts the UI display capabilities — really nice work. I’ve always thought direct embedding would be great.

tjbck:
Thanks!

Fu-Jie:
@Classic298 

image


I've tested this. The generated iframe currently appears at the top, which feels counter-intuitive; positioning it below the content would be much better. Since the user is already at the bottom of the chat when clicking the action, the iframe is often out of view. You can only see the iframe and the message together immediately if the answer is very short.

Classic298:
@Fu-Jie thanks makes sense, will open a PR for that

PR #21355, opened 2026-02-13
fix: fix: prevent cascading reasoning item creation from inline think tags

### LINKED ISSUE #21348: issue: v0.8.0 reasoning trace is visually split to many parts, causing the browser to slow down to a halt
### Check Existing Issues

### Installation Method

Docker

### Open WebUI Version

v0.8.0

_No response_

### Operating System

Windows 11

### Browser (if applicable)

Brave (latest)

### Confirmation

### Expected Behavior

Regular reasoning trace rendering.

### Actual Behavior

Seems to be some sort of rendering issue with reasoning trace possibly split at each token output. I think the screen capture is the best way to show this.

### Steps to Reproduce

Prompt a model that outputs a reasoning trace.

### Logs & Screenshots

Image

### Additional Information

Never seen the issue before v0.8.0.

Classic298:
Fix reasoning trace rendering regression in v0.8.0 where inline reasoning tags (e.g. think/thinking) caused one collapsible block per token, freezing the browser.

Root cause: tag_output_handler did not strip the start tag from the content accumulator after detection, causing it to be re-detected on every subsequent token. Additionally, tokens arriving during an active reasoning block were routed to new message items instead of being appended to the reasoning item.

Changes:
- Strip start tag from content accumulator on detection in tag_output_handler to prevent cascading re-detection
- Route streaming tokens directly to in-progress reasoning items from tag detection instead of creating intermediate message items
- Guard message text append to skip when already handled by the reasoning branch
- Add 8 unit tests for tag_output_handler covering single/multiple blocks, tag variants, split tokens, and content deduplication

Only affects providers sending reasoning as inline tags in the content delta (e.g. Google generativelanguage, Minimax, Ollama). Providers using the separate reasoning_content field (OpenAI, LiteLLM/OpenRouter) are unaffected.

Fixes #21348

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -2994,6 +2994,10 @@ async def streaming_chat_response_handler(response, ctx):
                             before_tag = content[: match.start()]
                             after_tag = content[match.end() :]
 
+                            # Strip the start tag from the content accumulator
+                            # so it won't be re-detected on subsequent calls
+                            content = after_tag
+
                             # Remove the start tag and everything after from last message
                             current_text = get_last_text(output)
                             set_last_text(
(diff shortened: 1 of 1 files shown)


Classic298:
^this commit now additionally fixed the following:

When tool calls or code execution trigger a second LLM streaming round,
the content accumulator retained text from the previous round. This caused
tag_output_handler to re-detect old reasoning tags, resulting in raw
</think> tags leaking into responses and reasoning content being duplicated.
Reset content to empty string before each subsequent stream_body_handler
call in both the tool call loop and code interpreter loop.

Classic298:
Addressed in dev!

PR #21358, opened 2026-02-13
fix: enforce public sharing permission checks across all resource types

Classic298:
The sharePublic prop in editor components (Knowledge, Tools, Skills, Prompts, Models) incorrectly included an "|| edit" / "|| write_access" condition, allowing users with write access to see and use the "Public" sharing option regardless of their actual public sharing permission. Additionally, all backend access/update endpoints only verified write authorization but did not check the corresponding sharing.public_* permission, allowing direct API calls to bypass frontend restrictions entirely.

Frontend: removed the edit/write_access bypass from sharePublic in all five editor components so visibility is gated solely by the user's sharing.public_* permission or admin role.

Backend: added has_public_read_access_grant checks to the access/update endpoints in knowledge.py, tools.py, prompts.py, skills.py, models.py, and notes.py. Public grants are silently stripped when the user lacks the corresponding permission.

Fixes #21356 

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/knowledge.py
+++ b/backend/open_webui/routers/knowledge.py
@@ -511,6 +511,7 @@ class KnowledgeAccessGrantsForm(BaseModel):
 
 @router.post("/{id}/access/update", response_model=Optional[KnowledgeFilesResponse])
 async def update_knowledge_access_by_id(
+    request: Request,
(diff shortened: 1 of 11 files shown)


tjbck:
Thanks!

PR #21394, opened 2026-02-13
feat: support arbitrary file output from Jupyter code execution

lewisco:
## Summary

- The Jupyter code interpreter engine was limited to capturing `image/png` outputs only. Skills that generate documents (PPTX, PDF, XLSX, CSV, etc.) had no way to deliver files to users
- This generalizes the entire output pipeline — from Jupyter kernel capture through backend processing to frontend display — to handle any file type
- Non-image files are delivered as downloadable chat message attachments using the existing `chat:message:files` event pattern (same as `generate_image`)

### Changes across 5 files:

- **`code_interpreter.py`** — Capture any `image/*`, `application/*`, `audio/*` MIME type from Jupyter. Add preamble/postamble that creates a writable data dir, scans for new files after execution, and emits them as `display_data` with correct MIME types and original filenames
- **`files.py`** — Add generic `upload_file_from_base64()` and `get_file_url_from_base64()` that handle any data URI MIME type. Register common office MIME types. Support embedded `filename` parameter in data URIs
- **`builtin.py`** — Replace image-only scanning in `execute_code` with generic data URI detection. Images render inline; non-image files emitted as `chat:message:files` attachments with DB persistence
- **`middleware.py`** — Apply same generic handling to the native (non-tool) code interpreter path
- **`ResponseMessage.svelte`** — Change file display gate from `filter(f => f.type === 'image')` to show all file types (the inner `{#each}` already dispatches to `<Image>` vs `<FileItem>`)

### How it works

1. User code saves a file (e.g., `.pptx`) to `__owui_data_dir` (set by the preamble)
2. Postamble detects the new file, reads it, and emits as `display_data` with the correct MIME type
3. `code_interpreter.py` captures the base64 data with MIME type and original filename
4. `builtin.py` uploads via `get_file_url_from_base64()` and emits as a chat file attachment
5. Frontend renders it as a downloadable `<FileItem>`

## Test plan

- [ ] Set `CODE_INTERPRETER_ENGINE=jupyter` with an external Jupyter server
- [ ] Run a skill/code that generates a `.pptx` file using `python-pptx` — verify it appears as a downloadable attachment in the chat
- [ ] Run existing image generation code (e.g., matplotlib plot) — verify images still render inline as before
- [ ] Test with PDF output (`data:application/pdf;base64,...`) and CSV output
- [ ] Verify the original filename is preserved (not `generated-file.pptx`)
- [ ] Verify the download link works (no 404 or double URL path)

### DIFF
--- a/backend/open_webui/tools/builtin.py
+++ b/backend/open_webui/tools/builtin.py
@@ -463,43 +463,95 @@ async def execute_code(
         else:
             return json.dumps({"error": f"Unknown code interpreter engine: {engine}"})
 
-        # Handle image outputs (base64 encoded) - replace with uploaded URLs
-        # Get actual user object for image upload (upload_image requires user.id attribute)
+        # Handle base64-encoded outputs - upload images inline, emit other files as attachments
         if __user__ and __user__.get("id"):
+            import re as _re
+
             from open_webui.models.users import Users
-            from open_webui.utils.files import get_image_url_from_base64
+            from open_webui.utils.files import get_file_url_from_base64
 
             user = Users.get_user_by_id(__user__["id"])
+            generated_files = []
 
-            # Extract and upload images from stdout
-            if stdout and isinstance(stdout, str):
-                stdout_lines = stdout.split("\n")
-                for idx, line in enumerate(stdout_lines):
-                    if "data:image/png;base64" in line:
-                        image_url = get_image_url_from_base64(
-                            __request__,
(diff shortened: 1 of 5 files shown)


AIWintermuteAI:
@tjbck @Classic298 is there an interest in such function? or maybe it is already implemented / being implemented?
Basically the ability to send arbitrary outputs with code interpreter.

PR #21406, opened 2026-02-14
fix: offline model retrieval, re-raise to disable instead of returning useless fallback

ahxxm:
In offline mode (`OFFLINE_MODE=True`), `get_model_path` not-so-silently returns the raw model name string as a fallback. This fallback exists such that `SentenceTransformer` can try to download a model, but when `OFFLINE_MODE=True` download is blocked, SentenceTransformer emits `WARNI: No sentence-transformers model found ... Creating a new one with mean pooling`, making retrieval feature appear useful, but not really so.

The fix is re-raise the exception in offline mode. Caller chain already handles this:
- `get_ef()` catches, returns `None`
- `get_rf()` catches, re-raises to `main.py:1072-1074`
-  App starts cleanly with `ef=None`. Retrieval is disabled until a model is pre-cached or `RAG_EMBEDDING_ENGINE` setup.

closes #21405 

> [!NOTE]

### DIFF
--- a/backend/open_webui/retrieval/utils.py
+++ b/backend/open_webui/retrieval/utils.py
@@ -1269,6 +1269,8 @@ def get_model_path(model: str, update_model: bool = False):
         return model_repo_path
     except Exception as e:
         log.exception(f"Cannot determine model snapshot path: {e}")
+        if OFFLINE_MODE:
+            raise
         return model
 
 


ahxxm:
@pr-validator-bot fixed the target branch, ping

PR #21496, opened 2026-02-16
feat: add "Show Tool Call Output" interface toggle

Classic298:
Add a new per-user interface toggle under Settings > Interface that controls how tool calls are displayed when expanded in the chat UI. The toggle defaults to on, preserving the existing behavior. When turned off:
- The Output section (raw JSON result) is completely hidden
- The Input section replaces the JSON code block with a formatted key-value list, converting parameter names from snake_case/camelCase to readable labels (e.g. "search_query" becomes "Search Query")
- Falls back to raw JSON display if arguments cannot be parsed This is a frontend-only change. Tool call data is still sent to models and stored in chat history unchanged — only the visual rendering is affected.
Files changed:
- src/lib/stores/index.ts Added showToolCallOutput to the Settings type definition.
- src/lib/components/chat/Settings/Interface.svelte Added the toggle variable, onMount initialization from saved settings, and the toggle UI row in the Chat section after "Always Expand Details".
- src/lib/components/chat/Messages/Markdown/MarkdownTokens.svelte Passes the showToolCallOutput setting as a prop to ToolCallDisplay.
- src/lib/components/common/ToolCallDisplay.svelte Added showToolCallOutput prop, formatKey helper for converting parameter names to human-readable labels, and tryParseArgs helper for safely parsing tool arguments. Conditionally renders either the original JSON code blocks or the formatted key-value display based on the setting.

**1. Would fix: #21440
2. Relates to: https://github.com/open-webui/open-webui/issues/20598
3. Improves: #20878**

image

On by default for backwards compatibility

image

If toggled on - all tool output no longer gets shown - and tool input gets shown in a key: value format instead of JSON

<ins>**This should also really help this issue here with hard performance issues with many codeblocks and long response outputs: https://github.com/open-webui/open-webui/discussions/20878**</ins>

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Messages/Markdown/MarkdownTokens.svelte
+++ b/src/lib/components/chat/Messages/Markdown/MarkdownTokens.svelte
@@ -329,6 +329,7 @@
 			<ToolCallDisplay
 				id={`${id}-${tokenIdx}-tc`}
 				attributes={token.attributes}
+				showToolCallOutput={$settings?.showToolCallOutput ?? true}
 				open={false}
 				className="w-full space-y-1"
 			/>
--- a/src/lib/components/chat/Settings/Interface.svelte
+++ b/src/lib/components/chat/Settings/Interface.svelte
@@ -70,6 +70,7 @@
 	let chatFadeStreamingText = true;
 	let collapseCodeBlocks = false;
 	let expandDetails = false;
+	let showToolCallOutput = true;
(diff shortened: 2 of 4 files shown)


ldpg-dev:
@tjbck Any reason this was closed without merge? It seemed like a simple but incredibly helpful update

stemper-dev:
@Classic298 why has it been closed. We would need it too!

PR #21522, opened 2026-02-17
fix: make Brave web search failures clearer and safer

Plutarch01:
## Summary
- default Brave web-search query fan-out to sequential execution when web search concurrency is left unset/unlimited, reducing free-tier 429 bursts from generated multi-query searches
- preserve explicit admin concurrency settings and only apply the fallback for Brave
- include concrete web-search error details in status events and inject a transparency hint into system context so the model does not fabricate vague 'technical difficulties' claims


Plutarch01:
Thanks for the validation feedback. I opened a corrected PR targeting  with full template/CLA content here: https://github.com/open-webui/open-webui/pull/21525

Plutarch01:
Thanks for the validation feedback. I opened a corrected PR targeting the dev branch with full template and CLA content here: https://github.com/open-webui/open-webui/pull/21525

PR #21534, opened 2026-02-17
feat: Native Anthropic API (v1/messages) Proxy Support

tblattner:
# Pull Request Checklist
[Discussion Link](https://github.com/open-webui/open-webui/discussions/21390)

Implementation for v1/messages acts purely as a proxy with user authentication based on API key.

# Changelog Entry

### Description
- This PR adds support for the v1/messages endpoint used by Anthropic. This enables usage of tools such as claude code for OWUI users. While LiteLLM can bridge this, adding native support within Open WebUI simplifies the stack for on-premise deployments and leverages OWUI's existing authentication layer. Claude code was used to help develop this commit, and I have reviewed and reworked components of the commit, such as modifications to its creation of endpoints.

### Added
- Support for anthropic's v1/messages
- Tested against claude code, allowing direct access to claude code through OWUI.

### Changed
- Added environment variable "ENABLE_ANTHROPIC_API" to enable/disable this feature.
- Added router endpoints: "/anthropic/v1/messages", "/v1/messages", "/config", and "/config/update"

### Deprecated
- N/A

### Removed
- N/A

### Fixed
- Support for "/v1/messages" endpoint.

### Security

- N/A
 
### Breaking Changes

- N/A

---

### Additional Information

- [Discussion posted](https://github.com/open-webui/open-webui/discussions/21390)
- Still need to include update to docs.

### Screenshots or Videos
- N/A


tblattner:
> # 👋 Welcome and Thank You for Contributing!
> We appreciate you taking the time to submit a pull request to Open WebUI!
> 
> ## ⚠️ Important: Testing Requirements
> We've recently seen an increase in PRs that have significant issues:
> 
> * PRs that don't actually fix the bug they claim to fix
> * PRs that don't implement the feature they describe
> * PRs that break existing functionality
> * PRs that are clearly AI-generated without proper testing being done by the author
> * PRs that simply don't work as intended
> 
> **These untested PRs consume significant time from maintainers and volunteer contributors who review and test PRs in their free time.** Time that could be spent testing other PRs or improving Open WebUI in other ways.
> 
> ### Before marking your PR as "Ready for Review":
> Please **explicitly confirm**:
> 
> 1. ✅ You have **personally tested ALL changes** in this PR
> 2. ✅ **How you tested it** (specific steps you took to verify it works)
> 3. ✅ **Visual evidence** where applicable (screenshots or videos showing the feature/fix working) - if applicable to your specific PR
> 
> **If you're not certain your PR works exactly as intended, please leave it in DRAFT mode** until you've thoroughly tested it.
> 
> Thank you for helping us maintain quality and respecting the time of our community! 🙏

I personally tested all changes by running curl requests hitting the new endpoints as well as attaching Claude Code into the system. The model served was using vLLM and is using GPT-OSS-120B, model added into my localhost OWUI instance. Claude code interfaces with OWUI's new v1/messages endpoint.

Screenshot 2026-02-17 at 11 45 45 AM

Classic298:
this is not wanted.
for the same reasons as here https://docs.openwebui.com/faq#q-why-doesnt-open-webui-natively-support-provider-xs-proprietary-api

tblattner:
I hear you on the importance of universal protocols. The challenge many of us are facing is that agentic tools—which are moving faster than the standards—are increasingly hardcoded to specific endpoints like v1/messages.

By adding this lightweight proxy, OWUI becomes the identity and security layer for these tools. Without it, users have to bypass OWUI entirely to use things like Claude Code. Given that this is a zero-dependency change isolated to the router level, I believe the utility it provides for developers outweighs the footprint. Is there a specific way I could refactor this to better align with the modular philosophy while keeping this functionality intact?

Classic298:
Open WebUI is not a universal proxy though. LiteLLM for example is.

And adding Anthropic API here opens pandora's box. "Hey why not also add google's interactions API? we need it for XY"

tblattner:
That's a fair point... maintaining a dozen proprietary APIs would be a nightmare. I’d argue, though, that Anthropic’s protocol is unique because it's the primary blocker for the current wave of agentic CLI tools (like Claude Code) which developers want to route through their OWUI instance for security/auth reasons.

Would you be open to a compromise where we only support a strictly limited set of high impact protocols that have significant third-party tool ecosystems?

Alternatively, is there a recommended way to implement a Protocol Plugin? Currently, 'Pipes' handle data flow, but they don't easily allow for exposing entirely new API endpoints on the OWUI server itself. I'm looking for a way to keep OWUI as the identity layer for these external tools.

Classic298:
Why does Open WebUI need to be the proxy in your case? Open WebUI is an extensible chatplatform - and sure it has an API and API Keys (though no real API Key management) but why does Open WebUI have to act as a universal API proxy for various providers?

And what is a major AI tool today (claude code) may be irrelevant in a few months and suddenly Gemini CLI with Google's Interactions API might be the trend or 3 weeks later Codex is what everyone is using.

Implementing a whole proxy endpoint just for Anthropic just because "right now" claude code is what is being used widely is not a good reason.

And breaking one of Open WebUI's core principles, which is using only universal protocols, like OpenAI Completions and now OpenResponses would be broken if Anthropic support were added. No matter if "only a handful of providers" are added like you suggested. **The principle would be broken.** The pandora's box is open.

And also I do not see why Open WebUI has to act as a universal proxy for..... coding tools?
That's exactly what LiteLLM is for - or other proxy middlewares that implement various different API endpoints from completions, to messages, to responses and more. - Why would Open WebUI do this? Maybe you can convince me. But Open WebUI is neither at it's core, nor in it's bigger picture, an API proxy for many different APIs.

tblattner:
I agree that baking proprietary proxies into the core is a maintenance trap.

However, the reason I'm advocating for this isn't just about 'Claude Code'—it’s about Open WebUI acting as the Identity & Auth layer for the user's entire AI workflow. Currently, if a user wants to use a specialized tool (CLI, IDE extension, agent) with OWUI’s authentication and logging, they are forced to use the OpenAI-spec or Open Responses. If that tool doesn't support those specs, OWUI becomes a 'silo' rather than a 'hub.'

I'm beginning to lean more and more to some structure to make things having extensible Endpoints. 
So instead of adding Anthropic to the core, what if we made the API router itself extensible via the existing Pipe/Plugin system?

This would close the Pandora's Box as the core remains strictly universal (OpenAI/Open Responses).

Then this would shift the maintenance of proprietary endpoints (Anthropic v1/messages, Google Interactions, etc.) to live in the Community Hub as 'Endpoint Pipes.'

This would also future-proof the system as open responses evolves, it remains the standard, but users have the freedom to bridge the protocol lag themselves without waiting for a core PR.

This also shifts the burden from the maintainers to the community zoo. If a tool like Claude Code becomes irrelevant in three months, the community-maintained 'Anthropic Endpoint Pipe' simply stops getting downloads, but the Open WebUI core remains untouched.

Is there a path forward where we allow the router to be extended by Plugins, rather than hardcoding these proxies into the core? I can investigate some concepts in this if this becomes a good alternative.

Classic298:
@tblattner totally different question: you are only modifying the backend here. In fact, you are only adding a single file and editing the main.py only minimally to add the router endpoint.

What would speak against simply adding this file and the 2 line changes locally to your deployment? This would not require a fork - this would not require pipes (which were never meant for what you are suggesting and what you are suggesting would require, in fact, a new function type, because it cannot be done with pipes), and you get your anthropic api proxy.

Classic298:
yeah just checked again for my own sanity - this cannot be done with pipes or pipelines or filters at all. Not easily. Not at all.

Would require a brand new function type.

tblattner:
> yeah just checked again for my own sanity - this cannot be done with pipes or pipelines or filters at all. Not easily. Not at all.
> 
> Would require a brand new function type.

This could be a possible path?

1. New function type: endpoint
Alongside pipe, filter, action, add endpoint. A plugin would look like:

```
"""
title: Anthropic Messages Endpoint
version: 1.0.0
requirements: []
"""

class Endpoint:
    class Valves(BaseModel):
        priority: int = 0

    def routes(self) -> list[dict]:
        """Declare which paths this endpoint handles."""
        return [
            {"method": "POST", "path": "/v1/messages"},
        ]

    def authenticate(self, request: Request) -> str:
        """Extract API key from protocol-specific header."""
        return request.headers.get("x-api-key", "")

    async def resolve_model(self, body: dict, request: Request) -> tuple[str, int]:
        """Map the incoming model name to an OpenAI registry entry.
        Returns (model_id, url_idx) so the core can do access control + lookup."""
        model_id = body.get("model", "")
        # reuse existing OpenAI model registry lookup
        ...
        return model_id, url_idx

    async def handle(
        self,
        request: Request,
        body: dict,
        user: UserModel,
        target_url: str,
        target_key: str,
    ) -> Response:
        """Proxy/transform the request. Core already resolved auth + model."""
        # Forward to backend, handle streaming, return Response
        ...
```
2. Core dispatcher (minimal addition)
A single catch-all route registered at the end of the router list (lowest priority):


main.py — after all hardcoded routers
```
app.include_router(endpoint_dispatch.router, prefix="", tags=["endpoint-plugins"])
```

The dispatcher:

```
@router.api_route("/{path:path}", methods=["GET", "POST", "PUT", "DELETE", "PATCH"])
async def dispatch(request: Request, path: str):
    # 1. Match path+method against registered endpoint plugins
    endpoint_fn = match_endpoint_plugin(request.app.state, request.method, f"/{path}")
    if not endpoint_fn:
        raise HTTPException(404)

    # 2. Auth — delegate to plugin's authenticate() to extract credential,
    #    then validate via core's get_current_user_by_api_key()
    api_key = endpoint_fn.authenticate(request)
    user = get_current_user_by_api_key(request, api_key)

    # 3. Read body
    body = await request.json()

    # 4. Model resolution + access control (core responsibility)
    model_id, url_idx = await endpoint_fn.resolve_model(body, request)
    check_model_access(user, model_id)  # existing ACL logic
    target_url, target_key = get_backend_url_and_key(request, url_idx)

    # 5. Delegate to plugin
    return await endpoint_fn.handle(request, body, user, target_url, target_key)
```

3. Registration lifecycle
On startup and whenever functions are created/updated/toggled:

```
def rebuild_endpoint_registry(app_state):
    endpoints = Functions.get_functions_by_type("endpoint", active_only=True)
    registry = []
    for ep in endpoints:
        module = load_function_module_by_id(ep.id)
        for route in module.routes():
            registry.append((route["method"], route["path"], module))
    app_state.ENDPOINT_REGISTRY = registry
```

This reuses the existing load_function_module_by_id / caching infrastructure in plugin.py

Happy to submit a PR if this could be of interest?

Classic298:
Thanks for sketching this out, the direction is roughly right and you've identified the correct extension points. But I went through the codebase to verify the specific claims and there are several issues with the proposal as written, plus some significant design gaps.

## Fact-checking the code references

**get_current_user_by_api_key**: This function exists, but it looks up Open WebUI API keys (the ones users generate from their Settings → Account page). It does NOT validate upstream provider keys (like Anthropic's API key). Your dispatcher pseudo-code calls the plugin's `authenticate()` to extract the x-api-key header and then passes that to `get_current_user_by_api_key()`, but the value you'd extract from x-api-key is the Anthropic API key, not an OWUI user API key. These are completely different credentials. You'd need a separate step to map the request to an OWUI user. The function also depends on `request.state.enable_api_keys` being set (done by middleware), which your catch-all route may not have.

**check_model_access**: This exists but takes a model dict (the full model object with nested info/meta), not a model_id string. Your pseudo-code shows `check_model_access(user, model_id)` which wouldn't work. You'd need to first look up the model from `request.app.state.MODELS`, then pass the full dict.

**get_backend_url_and_key**: This function does not exist anywhere in the codebase, you completely hallucinated it. There's no single function that resolves a url_idx to a target URL and API key. The URL/key resolution is done differently depending on the provider: OpenAI connections use `request.app.state.config.OPENAI_API_BASE_URLS[idx]` and `request.app.state.config.OPENAI_API_KEYS[idx]`, Ollama uses different state vars entirely. You'd need to build this abstraction.

**Functions.get_functions_by_type("endpoint", active_only=True)**: This part is correct. The function exists and accepts both a type string and an active_only boolean, filtering on `is_active=True` when set.

**load_function_module_by_id / caching**: The existing cache infra (`get_function_module_from_cache`) requires a `request` object to access `request.app.state.FUNCTIONS`. Your proposed `rebuild_endpoint_registry` runs at startup/lifecycle hooks where there's no request context. You'd need to access `app.state` directly, which means either passing the app object or using a different caching strategy for endpoint plugins.

## Design gaps that need addressing

**1. Catch-all route swallows unrelated 404s.**: `/{path:path}` at the end of the router list matches everything that didn't hit an earlier route. A request to `/api/v1/typo` would hit your dispatcher, fail plugin matching, and return a 404, but with a confusing "no endpoint plugin matched" context instead of FastAPI's normal 404. You'd need to either use a prefix (like `/endpoints/`) or ensure your 404 is indistinguishable from the normal one, including for static assets, favicon, manifest.json, etc.

**2. Auth model is a fundamental design question, not a detail.** The proposal buries auth in a plugin method, but this is the hardest part. Open WebUI's auth is Bearer-token based. Anthropic uses x-api-key. Other protocols use different schemes. Who is the "user" for ACL purposes? If someone sends an Anthropic request with their Anthropic API key, how do you map that to an OWUI user for access control? The current PR for the Anthropic proxy solves this by looking up the API key in OWUI's user database (treating it as an OWUI API key passed via the x-api-key header), but your proposal doesn't address this mapping clearly.

**3. Streaming lifecycle.** Anthropic streaming uses SSE with cleanup requirements. The codebase has a `cleanup_response` function specifically for closing aiohttp sessions after streaming completes. The `handle` method returning a `Response` doesn't address who owns the aiohttp session lifecycle. If the plugin creates the session, it needs a way to register cleanup callbacks. Leaked sessions = slow memory leak under load.

**4. Middleware bypass.** Every request through /api/chat/completions goes through the filter pipeline (inlet → LLM → stream → outlet), pipeline filters, rate limiting, content moderation, etc. <ins>**Endpoint plugins skip ALL of this.**</ins> - So your argument of having centralized logging inside Open WebUI using this feature would fall apart here. **If an admin has a global PII-scrubbing filter, it won't apply to requests via endpoint plugins.** This needs to be an explicit design decision, not an oversight.

**5. Security surface.** Pipes/filters are sandboxed, they can only modify data within the chat-completion flow. Endpoint plugins get raw Request objects and return arbitrary Responses. A malicious Community Hub endpoint plugin could serve as an open proxy, access the filesystem, or exfiltrate data. This is fundamentally more dangerous than any existing plugin type. The Community Hub vetting process would need to account for this.

**6. Route conflicts.** What happens when two endpoint plugins both claim POST /v1/messages? The priority valve hints at ordering, but the dispatcher code doesn't use it and can't use it with a simple fetch all. Need explicit conflict detection and resolution, ideally surfaced in the admin UI.

**7. Frontend is unaddressed.** The admin Functions panel would need: endpoint type handling in FunctionMenu.svelte (which currently only special-cases filter/action for the global toggle), route visualization so admins can see what paths are registered, conflict warnings, and health/status indicators.

**8. Multi-worker.** app_state.ENDPOINT_REGISTRY is per-process. Multiple uvicorn workers would each need independent registries, rebuilt on startup and kept in sync on function changes (which currently go through the DB, so this might work naturally — but it's not mentioned).

# Bottom line

<ins>**All of this is just to show: What you want to do is something Open WebUI fundamentally was not designed for and is not meant to do.**</ins>

This new function type would be ANYTHING but easy to implement.
This new function type would bring MASSIVE technical debt with it and maintenance burden.
All for... what?
<ins>**All of this work just to avoid dropping one file into your deployment.**</ins>

As you know we want to have proper community demand for a feature, before adding it.
Your discussion for the Anthropic is 4 days old and only 1 person commented.

IF, and this is a very big if - IF something like this were ever to be implemented, it would need a new discussion, and at LEAST weeks of discussing how to implement it and at LEAST like three dozen requests for it. This is a MASSIVE undertaking, i would argue, even much MUCH larger and way more difficult than what it would take to add ANY new API.
Adding the API natively would be easier than adding a new function framework like this.

### <ins>So.... what am I actually saying here? </ins>
### I am saying: all of this goes to show: it would be a lot of work for something that could be as easy as just: take your anthropic.py - put it in your deployment - modify the main.py - and live a happy life.

tblattner:
Happy to drop it and live in my own world. Easy enough for me to write a simple patch script to apply my one-file update.
This issue has been discussed quite a bit, and I failed to reference those discussions:
https://github.com/open-webui/open-webui/discussions/18609
https://github.com/open-webui/open-webui/discussions/1253
https://github.com/open-webui/open-webui/discussions/21283

And I proposed a pretty clean interface, which is just a **proposal**, nothing that is actually implemented and tested. So again, if you want me to just go off and be on my own then I will. I'm doing this without any expectations, compensation, or anything. Again, happy to take a stab at this and if ends up being too much to chew, then great its not worth the effort.

So I'll await your reply, and can close this if you still don't want any of it.

And the way you effectively blew up in the last comment is jaw-dropping.


Edit:
Quick table generated to scope out the gaps:

Gap | Resolution
-- | --
Catch-all 404s | ASGI middleware, not a catch-all route. No match → call_next(), indistinguishable from normal routing
Auth | Core-owned, not plugin-owned. Core extracts OWUI API key from a fixed set of header locations. Plugin never sees credentials
Streaming | Core-provided ProxyClient. Plugin never creates aiohttp sessions. Core manages lifecycle via existing stream_wrapper pattern
Filter bypass | Explicit design decision for v1. Admin-visible warning. Protocol-agnostic audit/rate-limit hooks as a follow-up
Security | Same trust boundary as current pipes (exec()). Community Hub vetting flags endpoint plugins that forward to user-controlled URLs
Route conflicts | Detect at registration, surface in admin UI, priority-based resolution
Frontend | Scoped additions to existing Functions panel
Multi-worker | DB-backed loading + content-hash cache, same as existing functions

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(204, 204, 204); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(24, 24, 24); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">The biggest honest tradeoff remains #4: endpoint plugins are <strong>outside the filter pipeline by design.</strong> That's the price of protocol fidelity. If an admin needs PII scrubbing on Anthropic traffic, they'd need an endpoint-specific policy, not a reused OpenAI-format filter. Whether that tradeoff is acceptable is a product decision, not a technical one.</p>

Classic298:
I am not deciding this.

I am just telling you tim will most definitely close it for the reasons mentioned above.

And ... What? What do you mean by blew up? I responded analytically to your proposal and then circled back on the core issue of this - which is: why even implement the new function type if it's harder to do than even just implementing the api itself, let alone just dropping the file into the deployment.

Not sure what you meant there.

Anyways. Waiting on tims decision.

Bickio:
@tblattner I won't weigh in on the architectural discussion to protect my own sanity, but figured I could offer my own experience working around these same (apparently intentional) limitations of Open WebUI.

I'm using Opus & Sonnet 4.6 via LiteLLM, with full agentic tool calling and thinking. It works extremely well since Open WebUI 0.8.0 was released. I'd be happy to share details of my setup if you like.

Classic298:
@Bickio (in case not clear) this is about adding an anthropic API endpoint to Open WebUI, not adding Anthropic support to Open WebUI.

So this is not about using models inside Open WebUI but accessing Open WebUI via anthropic API instead of the existing API endpoints that Open WebUI offers.

If this was already clear, carry on.

Bickio:
Apologies @Classic298 I had indeed misunderstood the request

tjbck:
I'll investigate providing /v1/messages support

PR #21545, opened 2026-02-17
feat: add Show Tool Call Output interface toggle

Classic298:
Adds a new toggle in Settings > Interface that controls whether tool call output (the raw JSON result) is displayed. When off, only the tool name and input are shown, making the AI feel more like a seamless agent.

> [!NOTE]


ldpg-dev:
@tjbck Why do you keep pushing this away? This is not breaking and it's really useful

PR #21592, opened 2026-02-19
perf: Cache chat object in upsert_message to avoid redundant DB load

Classic298:
upsert_message_to_chat_by_id_and_message_id was calling get_chat_by_id
twice: once at the top to load the chat, and again at line 512 just to
get user_id for the dual-write to chat_message table.

<ins>**This function fires many times during streaming responses, so each redundant load deserializes the full conversation JSON blob unnecessarily.**</ins>

upsert_message: Look at that call count: 17 call sites across socket/main.py and middleware.py. This is called during streaming. It's also called when the user's message is first saved, when tool calls happen, when the final response lands. <ins>**So this one fires many times during a single streaming response, and each was doing 2 full chat loads instead of 1.**</ins>

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/chats.py
+++ b/backend/open_webui/models/chats.py
@@ -481,15 +481,17 @@ class ChatTable:
     def upsert_message_to_chat_by_id_and_message_id(
         self, id: str, message_id: str, message: dict
     ) -> Optional[ChatModel]:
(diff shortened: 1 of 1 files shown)


tjbck:
Addressed in dev.

PR #21595, opened 2026-02-19
fix: respect BYPASS_ADMIN_ACCESS_CONTROL in file list/search endpoints

Classic298:
The list_files and search_files endpoints unconditionally showed all files
to admin users. Now they check BYPASS_ADMIN_ACCESS_CONTROL so that when
it is set to False, admins only see their own files, consistent with how
other routers (models, knowledge, tools, prompts) handle this setting.

Fixes: https://github.com/open-webui/open-webui/discussions/21589

> [!NOTE]


tjbck:
Thanks!

PR #21596, opened 2026-02-19
perf: eliminate 2 redundant full chat deserialization on every message send

Classic298:
Problem:
Every message send triggered get_chat_by_id_and_user_id which loads the entire Chat row - including the potentially massive JSON blob containing the full conversation history - even when the caller only needed a simple yes/no ownership check or a single column value.

Two call sites in the message-send hot path were doing this:
1. main.py ownership verification: loaded the entire chat object including all message history JSON, then checked `if chat is None`. The JSON blob was immediately discarded - only the existence of the row mattered.
2. middleware.py folder check: loaded the entire chat object including all message history JSON, then read only `chat.folder_id` - a plain column on the chat table that requires zero JSON parsing.

Fix:
- Added `chat_exists_by_id_and_user_id()`: uses SQL EXISTS subquery which
  returns a boolean without loading any row data. The database can satisfy
  this from the primary key index alone.
- Added `get_chat_folder_id()`: queries only the `folder_id` column via
  `db.query(Chat.folder_id)`, which tells SQLAlchemy to SELECT only that
  single column instead of the entire row.

Both new methods preserve the same error handling semantics (return False/None on exception) and user_id filtering (ownership check) as the original get_chat_by_id_and_user_id.

Impact:
- Best case (typical): eliminates deserializing 2 full chat JSON blobs per message send. For long conversations (hundreds of messages with tool calls, images, file attachments), this blob can be multiple megabytes.
- Worst case: no regression - **the new queries are strictly cheaper than the old ones (less data transferred, less Python object construction, no Pydantic model_validate overhead).**

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1747,8 +1747,7 @@ async def chat_completion(
 
                 # Verify chat ownership — lightweight EXISTS check avoids
                 # deserializing the full chat JSON blob just to confirm the row exists
-                chat_owned = Chats.chat_exists_by_id_and_user_id(metadata["chat_id"], user.id)
-                if not chat_owned and user.role != "admin":  # admins can access any chat
+                if not Chats.is_chat_owner(metadata["chat_id"], user.id) and user.role != "admin":  # admins can access any chat
                     raise HTTPException(
                         status_code=status.HTTP_404_NOT_FOUND,
                         detail=ERROR_MESSAGES.DEFAULT(),
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1745,9 +1745,10 @@ async def chat_completion(
                 "local:"
             ):  # temporary chats are not stored
 
-                # Verify chat ownership
-                chat = Chats.get_chat_by_id_and_user_id(metadata["chat_id"], user.id)
-                if chat is None and user.role != "admin":  # admins can access any chat
+                # Verify chat ownership — lightweight EXISTS check avoids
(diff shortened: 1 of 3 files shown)


backend/open_webui/models/chats.py
```diff
@@ -938,6 +938,41 @@ def get_chat_by_id_and_user_id(
         except Exception:
             return None
 
+    def chat_exists_by_id_and_user_id(
```
tjbck: `is_chat_owner`

backend/open_webui/main.py
```diff
@@ -1745,9 +1745,10 @@ async def chat_completion(
                 "local:"
             ):  # temporary chats are not stored
 
-                # Verify chat ownership
-                chat = Chats.get_chat_by_id_and_user_id(metadata["chat_id"], user.id)
-                if chat is None and user.role != "admin":  # admins can access any chat
+                # Verify chat ownership — lightweight EXISTS check avoids
+                # deserializing the full chat JSON blob just to confirm the row exists
+                chat_owned = Chats.chat_exists_by_id_and_user_id(metadata["chat_id"], user.id)
+                if not chat_owned and user.role != "admin":  # admins can access any chat
```
tjbck: chat_owned variable not needed here, should be called directly

tjbck (changes requested):
..

Classic298:
@tjbck feedback addressed

tjbck:
Thanks!

PR #21614, opened 2026-02-19
perf: Optimize shared chats list to use column projection

Classic298:
The GET /chats/shared endpoint was loading full Chat rows including
the entire conversation history JSON blob, only to discard it and
return SharedChatResponse (id, title, share_id, timestamps). Now
uses with_entities() to select only the 5 needed columns, avoiding
deserialization of potentially large chat JSON for every shared chat.

> [!NOTE]


tjbck:
Thanks!

PR #21618, opened 2026-02-19
0.8.4

tjbck:


backend/open_webui/utils/middleware.py
```diff
@@ -2268,228 +2301,235 @@ async def process_chat_payload(request, form_data, user, metadata, model):
...
+                    ],
+                },
+                features,
+                model,
+            )
+            for name, tool_dict in builtin_tools.items():
+                if name not in tools_dict:
+                    tools_dict[name] = tool_dict
+
+        if tools_dict:
```
bdrosen96, line 2525: Did you intend for the else on line 2534 to be aligned with if tools_dict or should it also have been indented to be aligned with line 2526 like it was before the refactor? I noticed that when I got this updated code, some models no longer return responses as fast as they did before this went in

bdrosen96: When I locally modify the file and indent it to be like it was before, it restores the old behavior so I tend to think this is a regression as it invokes tools when tools_dict is empty and ends up using stream=False when calling generate_chat_completion instead of stream=True

Issue #15189, opened 2025-06-21
issue: Bug: Lack of Data Integrity Validation on Chat Model Causes Silent Failure and Frontend Deadlock


robotmikhro:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

0.6.15

### Operating System

Ubuntu 22.04.5 LTS(Jammy Jellyfish)

### Browser (if applicable)

Firefox 139.0.4 (64-bit)

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

**Area:** Backend, Frontend, Data Integrity

### Description

The application's backend does not perform sufficient data integrity checks on the chat object model before persisting it to the database. This allows structurally inconsistent or corrupted chat data (e.g., messages with missing `parentId` links) to be saved **without any errors or warnings being logged**.

When the frontend subsequently attempts to render this structurally flawed data, its rendering logic enters an unrecoverable state, such as an infinite loop or a state management deadlock. This results in the UI hanging indefinitely on a loading spinner, preventing the user from ever viewing the chat.

This is a fundamental data handling issue, not just an import/export problem. The import feature is simply the most direct way to introduce this kind of corrupted data and reproduce the bug.

### Expected Behavior

The application should be resilient to data inconsistencies:
-   **Backend:** Before saving a chat object, the backend must validate its structural integrity. It should repair common issues (like broken parent-child links) or reject the data with a clear error log if it's irrecoverably malformed.
[...]

PR #21649, opened 2026-02-20
fix: graceful fallback when history.currentId points to an orphan node (infinite spinner)

ephb:
# Pull Request Checklist

- [x] **Description:** Concise description provided below.
- [x] **Testing:** Manually tested with a locally-built Docker image using Playwright automation and manual testing on deployment. Reproduction steps and evidence provided below.
- [x] **Code review:** Self-reviewed. Single reactive block guard, no new dependencies, no architectural changes.
- [x] **Agentic AI Code:** This fix was written and tested by a human. AI tooling was used for assistance, but the logic, testing, and validation were done manually.

---

# Changelog Entry

### Description

Fixes a long-standing bug (reported in #15189) where a chat shows an **infinite loading spinner** and never renders, caused by `history.currentId` pointing to a partial "orphan" message node that has no `role`.

This happens when the browser or network drops while OpenWebUI is generating follow-up questions: the backend writes a `followUps` placeholder into `history.messages` and updates `history.currentId` to point at it, before the node has `role`, `content`, `parentId`, or `childrenIds`. The node looks like:

```json
{
  "23067407-...": {
    "model": "claude-4-sonnet-latest",
    "content": "...",
    "followUps": ["question 1", "question 2"]
  }
}
```

`Messages.svelte`'s reactive block starts at `history.currentId` and walks the tree via `parentId`. Starting from a node with no `role` means the walk produces nothing, `messages = []` on every tick, and the chat spins forever. **No console errors appear**, making it hard to diagnose.

We confirmed this in production on two separate chats on a deployment, and it matches the exact data structure reported in #15189.

### Fixed

- `Messages.svelte`: add a guard before the tree-walk that validates `history.currentId` resolves to a node with a `role`. If it does not, fall back to the most recent valid terminal message (has `role`, `childrenIds` is `[]`) sorted by timestamp. A `console.warn` is emitted so the recovery is visible in DevTools.

### Added

- Nothing new.

### Changed

- `src/lib/components/chat/Messages.svelte`: +31 / -10 lines in the reactive `$: if (history.currentId)` block.

### Deprecated / Removed / Security / Breaking Changes

- None.

---

### Additional Information

- Relates to: #15189
- The fix is purely defensive: it only activates when `currentId` is already invalid. Normal chats are completely unaffected.
- If no valid terminal message exists at all (severely corrupted history), a `console.error` is emitted and the chat renders empty - the same visible outcome as before.

### How to Reproduce (before fix)

1. Start a chat and let the LLM finish a response.
2. While follow-up questions are streaming, close the browser tab or disconnect the network.
3. Reopen the chat, infinite spinner, no console errors.

Alternatively, import the malformed JSON from issue #15189 - it reliably triggers the bug.

### Testing performed

Tested with a locally-built Docker image (`openwebui-debug:v0.8.3-fix`):

1. Imported a chat with an artificially injected orphan node (`currentId`, node with only `followUps`, no `role`).
2. **Baseline image** (unmodified `v0.8.3`): chat spins forever. Bug confirmed.
3. **Fixed image** (this patch): chat loads correctly. Browser console shows:
   ```
   [warn] history.currentId "00000000-..." points to an invalid message (keys: followUps). Falling back to last valid message "471bd8e5-...".
   ```
   Zero errors. All messages render correctly. Fix confirmed.
4. Opened several normal (non-corrupted) chats, guard code never triggers, behavior unchanged.

### DIFF
--- a/src/lib/components/chat/Messages.svelte
+++ b/src/lib/components/chat/Messages.svelte
@@ -76,18 +76,39 @@
 	$: if (history.currentId) {
 		let _messages = [];
 
-		let message = history.messages[history.currentId];
-		const visitedMessageIds = new Set();
+		// If currentId points to an orphan node (e.g. a partial followUps placeholder), fall back to the last valid terminal message.
+		let currentId = history.currentId;
+		const currentMsg = history.messages[currentId];
+		if (!currentMsg || !currentMsg.role) {
+			const terminal = Object.values(history.messages)
+				.filter((m) => m?.role && Array.isArray(m.childrenIds) && m.childrenIds.length === 0)
+				.sort((a, b) => (b.timestamp ?? 0) - (a.timestamp ?? 0))[0];
 
-		while (message && (messagesCount !== null ? _messages.length <= messagesCount : true)) {
-			if (visitedMessageIds.has(message.id)) {
-				console.warn('Circular dependency detected in message history', message.id);
-				break;
+			if (terminal) {
+				console.warn(
+					`history.currentId "${currentId}" points to an invalid message (keys: ${Object.keys(currentMsg ?? {}).join(', ') || 'none'}). Falling back to last valid message "${terminal.id}".`
+				);
+				currentId = terminal.id;
+			} else {
+				console.error('history.currentId is invalid and no valid terminal message was found.');
+				currentId = null;
 			}
-			visitedMessageIds.add(message.id);
+		}
 
-			_messages.unshift({ ...message });
-			message = message.parentId !== null ? history.messages[message.parentId] : null;
+		if (currentId) {
+			let message = history.messages[currentId];
+			const visitedMessageIds = new Set();
+
+			while (message && (messagesCount !== null ? _messages.length <= messagesCount : true)) {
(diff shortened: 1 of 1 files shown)


ephb:
Users connecting the instance trough VPN triggered this, which is why I assume the original connection drop was the problem. 
I can not with 100% certainty say what caused this corruption in the first place and it may be better to hunt that down but the current failure mode of basically losing the chat causes user panic and I would like to get that out of the way.

ShirasawaSama:
In practice, we often encounter messages where only the `content` or `id` field is empty. Can this PR handle such cases correctly?

ephb:
Honestly I did not test that case but it should. When I get around to updating this PR I will check.

ShirasawaSama:
Perhaps you can refer to my code?

```ts
    if (res?.chat?.history && res.chat.messages) {
        const msg = res.chat.history.messages[res.chat.history.currentId];
        if (!msg?.id || Object.keys(msg).length < 2) {
            const id = res.chat.messages.at(-1)?.id;
            if (id) res.chat.history.currentId = id;
        }
    }
```

Issue #21663, opened 2026-02-20
issue: RAG template mutates web search tool call args


relic664:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.8.3

### Operating System

Alpine 3.23

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

RAG templating should not leak into tool call arguments. 

### Actual Behavior

`RAG_SYSTEM_CONTEXT` controls where RAG context is injected, not whether templating occurs (it's always applied when sources exist). The problem is that it's currently injected into the user message during native tool-calling. During loops, this can cause the RAG template to be injected into subsequent tool call args (especially in the case of multiple tool calls per turn). It would be safer to inject into the system message instead. 

### Steps to Reproduce

1. Start a chat with any model
2. Induce a multi-step web search, either by providing a sufficiently complex question or by reducing the number of results returned in a web search.
3. Examine the tool call arguments

### Logs & Screenshots

<img width="984" height="311" alt="Image" src="https://github.com/user-attachments/assets/1968e0fc-5e2a-45d9-b5d1-2bb753271797" />

### Additional Information

[...]

PR #21668, opened 2026-02-20
fix: prevent recursive RAG template nesting during multi-turn tool calling

Classic298:
When native tool calling produced multiple sequential tool calls with
citation sources (e.g. repeated search_web or fetch_url), the RAG
template was recursively injected into the user message on each
iteration. This caused the user message to grow exponentially, wasting
context window and potentially confusing models into parroting the RAG
instructions into subsequent tool call arguments.

Fix: save the original user message before the tool-calling loop and
restore it before each re-application. Sources now accumulate across
iterations so the RAG template is applied once with a growing source
list, rather than being recursively nested N times.

Tested locally

Fixes: https://github.com/open-webui/open-webui/issues/21663

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -2517,8 +2517,16 @@ async def process_chat_payload(request, form_data, user, metadata, model):
 
     # If context is not empty, insert it into the messages
     if sources and prompt:
+        log.warning(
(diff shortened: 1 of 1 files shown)


Classic298:
@tjbck bad bug that needs a fix

Classic298:
(tested btw)

tjbck:
Should be addressed, testing wanted

Issue #16151, opened 2025-07-30
feat: Support for compressed archive file uploads


BestBackwards:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.

### Problem Description

I'm requesting support for uploading and processing of compressed archive files (.zip, .tar, .bz2, etc.) in Open WebUI. 

This feature would allow users to share and collaborate on larger datasets, which are often stored in compressed formats.

**Expected Behavior**:

- Users can upload archive files.
- Open WebUI decompresses the archive file.
- The contents of the archive file are made available to LLMs for processing.

I'd like to point out that there's a previous feature request (see: #7773 ) that was closed without adequate reasoning.
I believe this is an important issue and would appreciate it if you could re-examine my proposal and provide more insight into why this feature wasn't implemented earlier.

### Desired Solution you'd like

Adaptation of existing file upload and extraction process.

PR #21689, opened 2026-02-21
feat: Support for compressed archive file uploads

thomasrydawson-stack:
# Changelog Entry

### Description
Implements requests support for uploading compressed archives (zip, rar, 7z, etc.) This allows uploading of large datasets and collaboration, which is often done using archives. 

### Added
Added support for zip, rar, 7z and tar archives for uploading and analysis.
Adds dependencies for Patool. Patool is needed to extract files from archive file types in python.

### Additional Information

This is built off feature request from issue #16151.

### Screenshots or Videos

https://github.com/user-attachments/assets/3bf8f540-3963-4889-96fc-81a99c0162e9

> [!NOTE]


tjbck:
Open to discussion. We should discuss potential attack vector this opens up to.

IronD7:
I want appreciate file upload support, but understand security and potential attacks need to be consider.
In the meanwhile, maybe at least graciously failing would be an option, because users do not receive an error message on zip file upload. Hence, they expect their content to be searchable which leads to confusion and frustration

PR #21691, opened 2026-02-21
fix: show groups in access control search results

Classic298:
fix: show groups in access control search results

The groups list in the MemberSelector was nested inside a condition that only checked for user results. When searching for a group name that matched no users, the entire list section (including matching groups) was hidden. Include filtered groups in the visibility check so groups remain visible regardless of user search results.

> [!NOTE]


Classic298:
@tjbck fixed the bug i reported to you

image

PR #21694, opened 2026-02-21
feat: make Action Rich UI (embed) placement configurable per-action

Classic298:
### Description

Follow-up to discussion #21482.

Makes Action Rich UI (embed) placement configurable per-action. Actions now default to rendering embeds below the assistant message, near the action button. Developers can override this by specifying placement in the embed event data.

Tool Rich UI is unaffected — it already renders inline at the tool call invocation point.

### Why

As discussed in #21482, the previous fixed "top" placement caused UX issues:

- For long messages, clicking an action button appeared broken because the Rich UI rendered out of view at the top
- The result (Rich UI) should follow its source (the message content), matching natural derivation logic
- Tool Rich UI already renders inline at the invocation point — Action Rich UI should similarly appear near the action button
- Community consensus strongly favored "bottom" as the default

### What changed

**Backend:**
- actions.py: Sends placement alongside embeds in the event data (default: bottom)
- socket/main.py: Persists embed_placement on the message from the event data

**Frontend:**
- Chat.svelte: Reads placement from the embed event and stores it on the message
- ResponseMessage.svelte: Renders embeds above or below content based on the message's embed_placement (defaults to bottom when unset)

### Developer usage

Placement is controlled per-action. The default for actions is bottom. Developers who manually emit embeds via the event emitter can specify placement:

```python
await __event_emitter__({"type": "embeds", "data": {"embeds": [...], "placement": "top"}})

<!--
🚨 DO NOT DELETE THE TEXT BELOW 🚨
Keep the "Contributor License Agreement" confirmation text intact.
Deleting it will trigger the CLA-Bot to INVALIDATE your PR.
-->

> [!NOTE]

### DIFF
--- a/backend/open_webui/socket/main.py
+++ b/backend/open_webui/socket/main.py
@@ -841,12 +841,18 @@ def get_event_emitter(request_info, update_db=True):
                 embeds = event_data.get("data", {}).get("embeds", [])
                 embeds.extend(message.get("embeds", []))
 
+                placement = event_data.get("data", {}).get("placement", None)
+
+                update_data = {
+                    "embeds": embeds,
+                }
+                if placement:
+                    update_data["embed_placement"] = placement
+
                 Chats.upsert_message_to_chat_by_id_and_message_id(
                     request_info["chat_id"],
                     request_info["message_id"],
-                    {
-                        "embeds": embeds,
-                    },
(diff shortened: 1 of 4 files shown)


Classic298:
image

image

image

image

tested working

PR #21698, opened 2026-02-21
feat: scroll to rich ui once rendered

Classic298:
image

https://github.com/user-attachments/assets/a987903e-8487-4ba4-ac28-2e63974d646b

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -415,6 +415,16 @@
 					message.files = data.files;
 				} else if (type === 'chat:message:embeds' || type === 'embeds') {
 					message.embeds = data.embeds;
+
+					// Auto-scroll to the embed once it's rendered in the DOM
+					tick().then(() => {
+						setTimeout(() => {
+							const embedEl = document.getElementById(`${event.message_id}-embeds-0`);
+							if (embedEl) {
+								embedEl.scrollIntoView({ behavior: 'smooth', block: 'center' });
+							}
+						}, 100);
+					});
 				} else if (type === 'chat:message:error') {
 					message.error = data.error;
 				} else if (type === 'chat:message:follow_ups') {
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -417,14 +417,13 @@
 					message.embeds = data.embeds;
 
 					// Auto-scroll to the embed once it's rendered in the DOM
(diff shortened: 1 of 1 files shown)


src/lib/components/chat/Chat.svelte
```diff
@@ -427,6 +415,16 @@
 					message.files = data.files;
 				} else if (type === 'chat:message:embeds' || type === 'embeds') {
 					message.embeds = data.embeds;
+
+					// Auto-scroll to the embed once it's rendered in the DOM
+					tick().then(() => {
```
tjbck: await should be used

Classic298:
@tjbck fixed

Fu-Jie:
@Classic298 @tjbck 

Regarding the current solution of generating the iframe at the top and auto-scrolling up to it, I strongly believe that **generating it directly at the bottom** is a much better and more UX-friendly approach. 

Using a forced scroll/jump to the top introduces several poor user experience issues:

1. **Spatial Disorientation (Layout Shifts)**
   When a user clicks the Action button at the bottom of a message, their visual focus and cursor/finger are resting there. If the page suddenly scrolls up, the user momentarily loses their sense of context and spatial orientation. Forced page jumps are generally considered an anti-pattern in modern UI design.

2. **Violating the "Immediate Feedback" Principle**
   Good interaction design dictates that feedback should occur exactly where the interaction took place. Since the user triggers the action at the bottom, the result should immediately appear at the bottom—right within their current focal area.

3. **Breaking the Natural Flow**
   The core logic of a chat interface is a "waterfall" (content flows from top to bottom). Users are conditioned to scroll down to consume new information. Forcing users to backtrack upwards to view newly generated results completely reverses this natural reading flow.

I actually considered implementing an auto-scroll to the iframe in my own plugin previously, but in practice, the experience still felt quite disjointed.

**In summary:**
Implementing an auto-scroll to the top feels like a workaround to patch an unnatural placement. **Generating the iframe directly at the bottom**, allowing the result to appear naturally near the invocation button, is the most intuitive approach with the lowest cognitive load.

Fu-Jie:
@Classic298 @tjbck 

Honestly, the current approach is a bit ironic — the auto-scroll exists solely to compensate for an unnatural placement. Instead of fixing the root cause, we're layering one awkward UX pattern on top of another.

The real fix is simple: **just place the iframe at the bottom**. No scroll logic needed. The problem disappears entirely.

Issue #5975, opened 2024-10-07
**Inconsistent Conversation Bubble Sizes with YAML Code Blocks in Widescreen Mode**


lehhair:
**Bug Report**

**Installation Method:**
docker

**Environment:**
- **Open WebUI Version:** v0.3.32
- **Operating System:** Windows 11
- **Browser:** Chrome 129.0.6668.90

**Confirmation:**
- [x] I have read and followed all instructions provided in the README.md.
- [x] I am using the latest version of Open WebUI.
- [ ] I have included browser console logs.
- [ ] I have included Docker container logs.
- [x] I have provided detailed steps to reproduce the error in the "Reproduction Steps" section.

**Expected Behavior:**
When scrolling the page in widescreen mode, the size of the conversation bubbles should remain consistent.

**Actual Behavior:**
In widescreen mode, when the chat includes YAML code blocks, the size of the conversation bubbles changes inconsistently when scrolling up and down.

**Bug Summary:**
When scrolling in widescreen mode, if the chat includes YAML code blocks, the size of the conversation bubbles changes inconsistently.

**Reproduction Details:**
**Reproduction Steps:**
1. Open Open WebUI in widescreen mode.
2. Enter a message containing a YAML code block. Not all YAML will cause this issue; the following example will: https://github.com/ChatLunaLab/chatluna-character/blob/main/resources/presets/default.yml
3. Scroll up and down the page.
4. Observe the inconsistent change in the size of the conversation bubbles.
![chrome_iOqq4s2gEJ](https://github.com/user-attachments/assets/bb3d0d62-631a-470e-8de5-dc201f6325d7)

lehhair:
Perhaps it is caused by the display of new code blocks, which will reappear in many cases.

lehhair:
Until the latest version this problem still exists, and I found that code blocks are not colored when writing ```py instead of ```python, hopefully this can be optimized!

geniusjoelraj:
Hey I can work on this, Is this still relevant?

anugrahrk:
resolved? if not i can work on this.

silentoplayz:
I confirmed this issue internally last month, but only just now stumbled across this issue post based on the title. I clicked upon it and knew it was an instant match of the same issue I reported. Turns out your issue post opened up here is actually **1 year older** though!

[...]

PR #21707, opened 2026-02-21
fix: consistent conversation bubble sizes with YAML blocks in widescreen

agent-steven:
## Description
Fixes #5975

Resolves inconsistent conversation bubble sizing when scrolling in widescreen mode with YAML code blocks.

## Changes
- Added `min-w-0` to message container
- Prevents flex overflow that causes size changes during scroll

## Root Cause
In widescreen mode (`max-w-full`), flex containers without `min-w-0` allow child code blocks to overflow, causing the parent bubble to resize inconsistently during scroll.

## Demo

**Before:**
Conversation bubbles change size when scrolling up/down with YAML blocks

**After:**
Bubble sizes remain stable during scroll

## Testing
```bash
1. Open in widescreen mode (viewport width > max-w-5xl)
2. Paste this YAML: https://github.com/ChatLunaLab/chatluna-character/blob/main/resources/presets/default.yml
3. Send message
4. Scroll up and down
5. Verify: bubble size stays consistent
```

Tested in Chrome/Firefox on 1920×1080 display.

### DIFF
--- a/src/lib/components/chat/Messages/Message.svelte
+++ b/src/lib/components/chat/Messages/Message.svelte
@@ -47,7 +47,7 @@
 
 <div
 	role="listitem"
-	class="flex flex-col justify-between px-5 mb-3 w-full {($settings?.widescreenMode ?? null)
+	class="flex flex-col justify-between px-5 mb-3 w-full min-w-0 {($settings?.widescreenMode ?? null)
 		? 'max-w-full'
 		: 'max-w-5xl'} mx-auto rounded-lg group"
 >


agent-steven:
## Demo

### Before (main - no `min-w-0`):
Conversation bubbles may shift/resize during scroll with YAML code blocks

before

[📹 Download video (webm)](https://github.com/agent-steven/open-webui/releases/download/demo-videos/openwebui-yaml-before-fix.webm)

### After (this PR - with `min-w-0`):
Bubbles maintain consistent size during scroll

after

[📹 Download video (webm)](https://github.com/agent-steven/open-webui/releases/download/demo-videos/openwebui-yaml-after-fix.webm)

---

**How it works:** Added `min-w-0` to the message container, which prevents flex items from overflowing and causing layout shifts when scrolling in widescreen mode (`max-w-full`).

Issue #21726, opened 2026-02-22
issue: RAG duplication when view_knowledge_file tool is used


susmitds:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

v0.8.3

### Operating System

Windows 11

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

When **File Context is disabled**, file contents should only be included in the model context if explicitly retrieved via a tool (e.g., `view_knowledge_file`).

If the model loads a file using `view_knowledge_file` (tested with both knowledge base files and uploaded files), the file content should appear **only once** in the context — via the tool output.

The system should not inject the same file content again through a RAG attachment when File Context is turned off.

### Actual Behavior

Even with **File Context disabled**, when the model loads a file using `view_knowledge_file`, the file content appears **twice** in the model context:

1. Once in the tool output (`view_knowledge_file` response)
2. Once again through a RAG attachment injection

This results in duplicated file content in the prompt context.

### Steps to Reproduce

1. Upload a `.txt` file in the chat.
2. Ask the model to load the file.
3. Inspect the model-serving logs.

You will observe that the file content appears twice in the assembled context:

* In the tool output
* In an additional RAG-injected attachment

### Logs & Screenshots

[...]

PR #21733, opened 2026-02-22
fix: deduplicate RAG content injection when builtin tools execute

Classic298:
Fixes: https://github.com/open-webui/open-webui/issues/21726

fix: deduplicate RAG content injection when builtin tools execute

When builtin tools (view_knowledge_file, query_knowledge_files, search_web, fetch_url) execute during a chat, the model receives the same content multiple times:

- With file context OFF: content appears 2x (post-tool RAG source injection into the user/system message + tool result message)
- With file context ON: content appears up to 3x (initial RAG injection + post-tool RAG injection + tool result message)

The fix addresses duplication via two mechanisms while preserving both citation functionality and agentic tool compatibility:

1. Reference-only RAG injection: After a tool call, the post-tool RAG source tags carry only the source ID and name for citation mapping, not the full content (which is already in the tool result message). This preserves the [1], [2] citation format while avoiding content duplication.

2. Source deduplication: After the initial RAG pass (when file context is enabled), all injected source/file IDs are tracked. When tools return citation sources already present from the initial pass, they are filtered out before post-tool RAG injection. Newly injected sources are tracked for subsequent tool call rounds.

Additionally, the repeated inline list of citation tool names is extracted into a module-level CITATION_TOOL_NAMES constant, and the repeated source ID collection logic is extracted into a collect_source_ids helper function.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -347,7 +347,7 @@ def make_reference_sources(sources: list) -> list:
         reference_sources.append({
             **source,
             "document": [
-                "(Content provided in tool response above)"
+                "The full content of this source is provided in the tool call response."
                 for _ in source.get("document", [""])
             ],
         })
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -151,6 +151,16 @@ DEFAULT_REASONING_TAGS = [
 DEFAULT_SOLUTION_TAGS = [("<|begin_of_solution|>", "<|end_of_solution|>")]
 DEFAULT_CODE_INTERPRETER_TAGS = [("<code_interpreter>", "</code_interpreter>")]
 
+# Builtin tool names whose results should be extracted as citation sources
+# and whose tool result content should be condensed to avoid duplication
+# with the RAG <source> injection.
+CITATION_TOOL_NAMES = frozenset([
+    "search_web",
+    "fetch_url",
+    "view_knowledge_file",
+    "query_knowledge_files",
+])
+
 
 def output_id(prefix: str) -> str:
     """Generate OR-style ID: prefix + 24-char hex UUID."""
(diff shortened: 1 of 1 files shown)


Classic298:
Related, but not the same issue:

https://github.com/open-webui/open-webui/pull/21668

susmitds:
@Classic298 

ideally, it might be better to have the content in the tool output json rather the RAG injection part, especially with RAG_SYSTEM_CONTEXT=True when view_knowlege_file is used. Otherwise in projects loading multiple files, the RAG injection block will constantly be updated and waste cached tokens

That way the context sequence is preserved and the KV cache continuity is maintained. Then, the model can directly interact with files in an agentic fashion.

Classic298:
@susmitds i disagree, the constant updating of the RAG block is a bug and is addressed here: https://github.com/open-webui/open-webui/pull/21668

and the RAG block is needed so the model uses citations properly.

The RAG SYSTEM CONTEXT only defines where this block is placed.

susmitds:
> @susmitds i disagree, the constant updating of the RAG block is a bug and is addressed here: #21668
> 
> and the RAG block is needed so the model uses citations properly.
> 
> the RAG block is needed so the model uses citations properly.

I agree with this. Just the filename could be kept there in the RAG injection and the content in tool response. The file content being in the tool response is important for various community tools like https://openwebui.com/posts/sub_agent_7bfeb0b7

Classic298:
@susmitds community tools are not affected by my PR. Only the 4 builtin tools that emit citations

Your suggestion of keeping content in the tool output JSON would mean the model gets file content without any citation ID structure, which breaks the entire citation flow. The tradeoff you're proposing (KV cache optimization) comes at the cost of losing a core feature (proper citations).

If KV cache invalidation on the system message is a real concern, the better fix would be to just set RAG_SYSTEM_CONTEXT=False so the RAG context goes into the user message instead - which doesn't affect the system message prefix cache at all.

susmitds:
You are not understanding my point here, many downstream community tools for agentic setups use the built in tools as well. Ideally this should be toggleable to allow both behaviors. But, removing file content from tools themselves breaks the agentic use flow and anyways the duplication dissappears in further messages as in the current behavior, the content only exists in the tool output json for downstream messages.

Classic298:
@susmitds aha. but removing the RAG citation prompt also breaks basically all use cases of open webui

so what are you suggesting?

susmitds:
We can technically create a toggle for this using ENV variable

Classic298:
and what would the toggle do? decide where the content goes?

Then users have a choice between using agentic tools or having citations enabled.

There must be a way for both.

susmitds:
Yes. Basically a toggle to let users specify the location of content.

Also a cleaner way is viable. The source filename part can technically be placed in tool responses to be directly available for citations.

Classic298:
I am strongly against a toggle. This would divide Open WebUI into either having agentic file handling without citations, or having citations without agentic file handling

I will look into how else to solve this perhaps similar to the other proposal you said earlier

Classic298:
@susmitds please re review

susmitds:
Yeah, makes sense.

In a ideal scenario, the model being able to cite directly from the tool responses and citations and agentic file handling being independently toggleable in the model configuration is the best possible case as it allows the exact setup needed for the model at hand and current project.

Classic298:
no, introducing a toggle for this is not the best possible case.

this makes you have to decide a tradeoff, and both options are bad

Classic298:
please re review

susmitds:
I am against the design where the content is separated from the response as it breaks various agentic usecases. I mean the content could itself also be within the tool response itself along with citable source file names and be used directly for citations.


For example, the citation message could directly state to use the sources in tool response for inline citations.

Classic298:
@susmitds once again, please re review. I made changes. You have made yourself abundently clear what your stance is and therefore i am asking you to **Re Review** now

susmitds:
Yeah, I have skimmed the code changes, didn't get time to fully check the nuances. I will fully review the changes in a hour.

susmitds:
I have gone through the code changes. Overall this is a solid improvement and moves the implementation in the right direction. The separation between full content (tool response) and reference-only RAG injection is much cleaner and resolves the duplication issue without breaking citation mapping.

One small suggestion regarding the placeholder text in `make_reference_sources`:

```python
"(Content provided in tool response above)"
```

The word “above” introduces a positional assumption about message ordering. While this may be true in the current middleware flow, it creates a fragile coupling to presentation/order semantics. Future changes to message assembly, streaming behavior, or context restructuring could invalidate that assumption.

A more position-agnostic placeholder would be safer, for example:

```python
"(See tool response for full content)"
```

This preserves clarity while avoiding directional wording and implicit ordering guarantees.

Other than that, the deduplication logic and source tracking look good.

Classic298:
thanks will update

Classic298:
adjusted

susmitds:
Yeah, this is fine.

PR #21743, opened 2026-02-22
fix: prevent scroll jumping when deleting message pair after stopping generation

Classic298:
fix: prevent scroll jumping when deleting message pair after stopping generation

Replaced the showMessage call in deleteMessage with inline computation of the new currentId. The showMessage function was triggering scrollIntoView and a redundant saveChatHandler call, which caused competing scroll operations and wild visual jumping when deleting a message pair — especially after stopping generation.

**TESTED LOCALLY**

**WORKED**

**This bug existed for at least 1 year.**

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Messages.svelte
+++ b/src/lib/components/chat/Messages.svelte
@@ -392,9 +392,27 @@
 			delete history.messages[id];
 		});
 
-		await tick();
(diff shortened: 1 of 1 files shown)


tjbck:
1f474187a77d2c8a392f00d86f48eb3cb3a18b88

PR #21798, opened 2026-02-23
fix: harden distributed lock lifecycle with atomic Lua scripts and resilient retry

Classic298:
`periodic_usage_pool_cleanup` in `backend/open_webui/socket/main.py` gives up permanently on any distributed-lock trouble, so on multi-instance Redis deployments the usage pool can stop being reaped for the entire lifetime of a worker process.

Three separate ways it dies:

1. **Acquisition gives up after 3 attempts.** The `for attempt in range(max_retries + 1)` loop `return`s once the lock is held by another node three times in a row. Under startup contention every instance but one can lose the race, and those instances never try again.
2. **Renewal failure is fatal.** `raise Exception('Unable to renew usage pool cleanup lock.')` propagates out of the task instead of releasing and re-acquiring, so a single expired lock ends cleanup permanently.
3. **Redis errors are unguarded.** Neither `aquire_func()` nor `renew_func()` is wrapped, so a connection reset, failover, or idle socket timeout raises rather than returning falsy — killing the task the same way.

The consequence is a silent leak: `USAGE_POOL` entries for disconnected sessions are never expired, models keep reporting phantom active connections, and the only recovery is a process restart.

`periodic_session_pool_cleanup`, directly above it in the same module, already has the correct shape — retry acquisition indefinitely, break out of the work loop and re-acquire on renewal failure. This change brings the usage-pool loop in line with it and additionally treats an exception from either lock call as "lock not held", so transient Redis faults degrade into a retry instead of terminating the task.

Deployments without `REDIS_URL` are unaffected: `aquire_func` / `renew_func` are the `lambda: True` no-ops on that path, so neither the retry ceiling nor the exception path was ever reachable.

> [!NOTE]

### DIFF
--- a/backend/open_webui/socket/main.py
+++ b/backend/open_webui/socket/main.py
@@ -203,47 +203,54 @@ async def periodic_session_pool_cleanup():
 
 
 async def periodic_usage_pool_cleanup():
-    max_retries = 2
     retry_delay = random.uniform(WEBSOCKET_REDIS_LOCK_TIMEOUT / 2, WEBSOCKET_REDIS_LOCK_TIMEOUT)
-    for attempt in range(max_retries + 1):
-        if aquire_func():
-            break
-        else:
-            if attempt < max_retries:
-                log.debug('Cleanup lock already exists. Retry %s after %ss...', attempt + 1, retry_delay)
-                await asyncio.sleep(retry_delay)
-            else:
-                log.warning('Failed to acquire cleanup lock after retries. Skipping cleanup.')
-                return
+    while True:
+        # A Redis error must not kill the task: treat it as "lock not held" and retry.
+        try:
+            acquired = aquire_func()
+        except Exception:
+            log.exception('Cleanup lock acquisition error; will retry')
+            acquired = False
 
-    log.debug('Running periodic_cleanup')
-    try:
-        while True:
-            if not renew_func():
-                log.error('Unable to renew cleanup lock. Exiting usage pool cleanup.')
-                raise Exception('Unable to renew usage pool cleanup lock.')
+        if not acquired:
+            log.debug('Cleanup lock held by another node. Retrying.')
+            await asyncio.sleep(retry_delay)
+            continue
 
-            now = int(time.time())
-            for model_id, connections in list(USAGE_POOL.items()):
-                # Creating a list of sids to remove if they have timed out
-                expired_sids = [
-                    sid for sid, details in connections.items() if now - details['updated_at'] > TIMEOUT_DURATION
-                ]
+        log.debug('Running periodic_cleanup')
+        try:
+            while True:
+                try:
+                    renewed = renew_func()
+                except Exception:
+                    log.exception('Cleanup lock renewal error; will re-acquire')
+                    renewed = False
 
-                if connections and not expired_sids:
-                    continue
+                if not renewed:
+                    log.warning('Unable to renew cleanup lock. Retrying cleanup ownership.')
+                    break
 
-                for sid in expired_sids:
-                    del connections[sid]
+                now = int(time.time())
+                for model_id, connections in list(USAGE_POOL.items()):
+                    # Creating a list of sids to remove if they have timed out
+                    expired_sids = [
+                        sid for sid, details in connections.items() if now - details['updated_at'] > TIMEOUT_DURATION
+                    ]
 
-                if not connections:
-                    log.debug('Cleaning up model %s from usage pool', model_id)
-                    del USAGE_POOL[model_id]
-                else:
-                    USAGE_POOL[model_id] = connections
-            await asyncio.sleep(TIMEOUT_DURATION)
-    finally:
-        release_func()
+                    if connections and not expired_sids:
+                        continue
+
+                    for sid in expired_sids:
+                        del connections[sid]
+
+                    if not connections:
+                        log.debug('Cleaning up model %s from usage pool', model_id)
+                        del USAGE_POOL[model_id]
+                    else:
+                        USAGE_POOL[model_id] = connections
+                await asyncio.sleep(TIMEOUT_DURATION)
+        finally:
+            release_func()
 
 
 app = socketio.ASGIApp(
--- a/backend/open_webui/socket/main.py
+++ b/backend/open_webui/socket/main.py
@@ -203,47 +203,56 @@ async def periodic_session_pool_cleanup():
 
 
 async def periodic_usage_pool_cleanup():
-    max_retries = 2
     retry_delay = random.uniform(WEBSOCKET_REDIS_LOCK_TIMEOUT / 2, WEBSOCKET_REDIS_LOCK_TIMEOUT)
-    for attempt in range(max_retries + 1):
-        if aquire_func():
-            break
-        else:
-            if attempt < max_retries:
-                log.debug(f'Cleanup lock already exists. Retry {attempt + 1} after {retry_delay}s...')
-                await asyncio.sleep(retry_delay)
-            else:
-                log.warning('Failed to acquire cleanup lock after retries. Skipping cleanup.')
-                return
+    while True:
+        # A Redis error must not kill the task: treat it as "lock not held" and retry.
+        try:
+            acquired = aquire_func()
+        except Exception:
+            log.exception('Cleanup lock acquisition error; will retry')
+            acquired = False
 
-    log.debug('Running periodic_cleanup')
-    try:
-        while True:
-            if not renew_func():
-                log.error('Unable to renew cleanup lock. Exiting usage pool cleanup.')
-                raise Exception('Unable to renew usage pool cleanup lock.')
+        if not acquired:
+            log.debug('Cleanup lock held by another node. Retrying.')
+            await asyncio.sleep(retry_delay)
+            continue
 
-            now = int(time.time())
-            for model_id, connections in list(USAGE_POOL.items()):
-                # Creating a list of sids to remove if they have timed out
-                expired_sids = [
-                    sid for sid, details in connections.items() if now - details['updated_at'] > TIMEOUT_DURATION
-                ]
+        log.debug('Running periodic_cleanup')
+        try:
+            while True:
+                try:
+                    renewed = renew_func()
+                except Exception:
+                    log.exception('Cleanup lock renewal error; will re-acquire')
+                    renewed = False
 
-                if connections and not expired_sids:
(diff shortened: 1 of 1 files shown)


Classic298:
CC @Ithanil i want to get this merged, and i will try to test it a bit more (besides the testing done as written in the PR description) but would love to also get your "sign of approval" if anyhow possible

If not possible, all good

Thx

Ithanil:
> CC @Ithanil i want to get this merged, and i will try to test it a bit more (besides the testing done as written in the PR description) but would love to also get your "sign of approval" if anyhow possible
> 
> If not possible, all good
> 
> Thx

I already have it running in a 2-replica setup with Redis Sentinel and `WEBSOCKET_REDIS_LOCK_TIMEOUT=15`. Please confirm that the following is expected:

>`open_webui.socket.main:run_with_lock:193 - Lock renewal failed. Will re-acquire.`

will appear about every 2mins (the hardcoded 120s SESSION_POOL_TIMEOUT) on both nodes. In the Redis the lock "open-webui:session_cleanup_lock" will sometimes disappear and then reappear (which looks promising!). The "open-webui:usage_cleanup_lock" is always on (which is to be expected with a 3 second loop vs 15s Lock TTL).

I didn't explicitly test if the cleanup itself works as expected. What I can confirm though is that the mechanism survives as FLUSHDB and both locks (and session pool) will reappear.

As a side note: TBH I don't know if the usage pool itself actually works, usually the set is not present in my deployments.

Classic298:
Yes this is expected because the cleanup cannot and should not be done by multiple workers at the same time, so if a worker fails to acquire the cleanup lock, it will retry later to acquire the cleanup lock again.

If a worker dies (the worker that originally held the cleanup lock) then just a short moment later, another worker will pick up the cleanup lock successfully (because no other worker is currently holding it) and then doing the cleanup

**The CURRENT version of open webui** has exactly this issue. The workers all try only ONCE to get the cleanup - and if they fail - they just give up and never retry. If the worker with the cleanup dies, then no more cleanup will happen ever again until the full instance is fully restarted so one of the workers can get the cleanup lock again

So the logic of this PR is to prevent the lock from being lost forever - if one worker has the cleanup lock - no other worker can get it for now - but if that worker dies, then it's also not an issue because the lock is gone and one of the other workers can pick it up essentially

Ithanil:
> Yes this is expected because the cleanup cannot and should not be done by multiple workers at the same time, so if a worker fails to acquire the cleanup lock, it will retry later to acquire the cleanup lock again.
> 
> If a worker dies (the worker that originally held the cleanup lock) then just a short moment later, another worker will pick up the cleanup lock successfully (because no other worker is currently holding it) and then doing the cleanup
> 
> **The CURRENT version of open webui** has exactly this issue. The workers all try only ONCE to get the cleanup - and if they fail - they just give up and never retry. If the worker with the cleanup dies, then no more cleanup will happen ever again until the full instance is fully restarted so one of the workers can get the cleanup lock again
> 
> So the logic of this PR is to prevent the lock from being lost forever - if one worker has the cleanup lock - no other worker can get it for now - but if that worker dies, then it's also not an issue because the lock is gone and one of the other workers can pick it up essentially

Alright, just wanted to check back.

Just for completeness:
1) Actually the current version has the very basic problem that `WEBSOCKET_REDIS_LOCK_TIMEOUT` (default 60s) is lower than `SESSION_POOL_TIMEOUT` (120s) by default and the lock renew will fail because the lock is already gone after 120s.
2) Replicas will hang on startup for a long time because they retry lock acquiry with `WEBSOCKET_REDIS_LOCK_TIMEOUT/2` sleep time and 2 retries.
3) Also, it has all the other problems you mentioned.

That's why I resorted to nightly FLUSHDBs on the websocket Redis for ages...

Classic298:
@Ithanil i will edit the log level to be INFO instead of WARN because the fact a worker failed to pick it up is expected and normal and not worthy of a warning.

And thanks for confirming it works for you so far 🙏 thats worth a lot

Classic298:
I ran another self-review of the PR

One thing worth noting for context: with the default config (WEBSOCKET_REDIS_LOCK_TIMEOUT=60s, SESSION_POOL_TIMEOUT=120s), the session cleanup lock will always expire during the sleep interval between cleanup cycles. This means workers will log "Lock renewal failed. Will re-acquire." roughly every 120 seconds, which is what you,  @Ithanil, observed.

This is safe and by design:

- The lock only needs to prevent concurrent cleanup runs, not maintain continuous ownership between cycles. Nobody is doing work while sleeping, so it doesn't matter that the lock expires mid-sleep. And it doesn't matter if another worker picks up the lock later on. Basically zero overhead.
- **The Lua scripts guarantee no worker can overwrite or delete another worker's lock**, so the ownership handoff is race-free.
- The cleanup work itself is idempotent! If two cycles somehow overlapped (**they can't**, but hypothetically), no data would be corrupted.
- The re-acquisition loop with jittered backoff ensures the lock is always picked back up.

The usage cleanup lock behaves differently because its interval (3s) is well within the TTL (60s), so one worker holds it continuously. Both patterns are correct for their respective use cases.

Changed the log level from WARNING to INFO since this is expected lifecycle behavior, not an error condition. Pushed as a fixup.

Ithanil:
Hope this gets some attention, because the current implementation is obviously dysfunctional and causes various problems.

Classic298:
@tjbck

Classic298:
@pr-validator-bot full review

Classic298:
Both findings are valid observations but should not expand this PR's scope.

**Exception handling in run_with_lock:** The current code already solves the main lifecycle problem — cleanup tasks that die permanently after 1-2 boolean failures. Transient Redis exceptions (connection resets, failovers) are a rarer edge case with its own design questions: should exceptions during release trigger re-acquire? Should they log-and-retry or propagate? That's a separate concern worth a follow-up PR, not scope expansion here.

**release_lock return value:** The Lua script already guarantees correctness (it won't delete another worker's lock) regardless of whether the caller checks the return. Surfacing it is a nice observability improvement but doesn't affect any current behavior. Also a reasonable follow-up.

This PR targets two tightly coupled problems: TOCTOU races in lock operations and fragile lifecycle management. Both are addressed. Adding exception resilience on top would be a third concern and would justify the non-atomicity bot's earlier flag even more.

Classic298:
@tjbck ready 4 review

Classic298:

Classic298:

Classic298:
@tjbck ready

Nexory:
The `run_with_lock` restructure cleanly closes the lock-lifecycle root cause of #25216 — three independent termination paths (initial-acquire-fail-after-retries, mid-loop renew-fail, per-cycle work_fn exception) now all loop back to a re-acquire instead of permanently exiting the task. The atomic Lua scripts on `renew_lock`/`release_lock` also close a separate cross-replica ownership window the original `SET XX` and `GET + DELETE` did not protect against.

One orthogonal piece of #25216 sits at a different layer that this PR (correctly) does not touch: the `asyncio.create_task(...)` spawn sites at `main.py:682-683` discard both return values, so any exit that escapes `run_with_lock` — most concretely `asyncio.CancelledError`, which subclasses `BaseException` and is bypassed by all three `except Exception` guards — surfaces only as a delayed *"Task exception was never retrieved"* WARNING at GC time. I've opened a small follow-up at #25220 that retains the task handles, attaches an ERROR-level `add_done_callback`, and cancels both tasks in the lifespan shutdown block — mirroring the existing `redis_task_command_listener` pattern at lines 676 and 754. Zero file overlap with this PR; reviewable and revertible independently.

LGTM on this PR's scope — happy to see it land.

Classic298:
@Nexory can you help test it

Nexory:
Heads-up: the follow-up I referenced earlier is now at #25222 (the original #25220 was auto-closed by the policy bot for targeting `main` and missing the CLA template; refiled against `dev` with the full template). Diff and scope unchanged — spawn-site handle retention + done_callback + cooperative shutdown-cancel for both cleanup tasks.

Nexory:
@Classic298 — tested locally as you asked. docker compose with `redis:7-alpine` + this PR cherry-picked onto current `dev`, `WEBSOCKET_MANAGER=redis`, `WEBSOCKET_REDIS_LOCK_TIMEOUT=5`. Three phases run; literal log captures below.

**Phase 1 — baseline (current `dev`, no patches)** — confirms the bug.

After `docker pause owui-test-redis` for 8 seconds (exceeds 5s lock TTL):

```
2026-05-30 16:17:02.541 | ERROR | open_webui.socket.main:periodic_usage_pool_cleanup:208 - Unable to renew cleanup lock. Exiting usage pool cleanup.
2026-05-30 16:17:02.541 | ERROR | asyncio.runners:run:118 - Task exception was never retrieved
future: <Task finished name='Task-5' coro=<periodic_usage_pool_cleanup() done> exception=Exception('Unable to renew usage pool cleanup lock.')>
Exception: Unable to renew usage pool cleanup lock.
```

After this point, no further `Running periodic_cleanup` — task dead for the process lifetime.

**Phase 2 — this PR cherry-picked onto current `dev`** — same Redis pause sequence:

```
2026-05-30 16:22:17.691 | INFO | open_webui.socket.main:run_with_lock:191 - Lock renewal failed. Will re-acquire.
```

Match counts in the captured log:

| Pattern | Phase 1 | Phase 2 |
|---|---:|---:|
| `Unable to renew cleanup lock` | 1 | **0** |
| `Exiting usage pool cleanup` | 1 | **0** |
| `Task exception was never retrieved` | 1 | **0** |
| `Lock renewal failed. Will re-acquire.` | 0 | **1** |

The `run_with_lock` helper correctly intercepts the renewal-failure path and re-acquires instead of exiting permanently. The atomic Lua scripts on `renew_lock`/`release_lock` are not directly exercised by this test (single-instance), but a code-read against the diff confirms they close the cross-replica TOCTOU window on `SET XX` / `GET + DEL`.

**Phase 3 — this PR + my follow-up #25224** — regression check, same pause sequence:

```
2026-05-30 16:26:15.545 | INFO | open_webui.socket.main:run_with_lock:191 - Lock renewal failed. Will re-acquire.
```

No regression. The two PRs are orthogonal (zero file overlap — yours touches `socket/main.py` + `socket/utils.py`, mine touches `main.py` only).

LGTM on this PR's scope from both code-read and integration test. Hope this helps move it forward — happy to run additional scenarios if you want them tested.

Classic298:
@Ithanil better version of this merged into dev

PR #21852, opened 2026-02-25
feat: add dropdown toggle to filter buttons in chat

Classic298:
Discussed: https://github.com/open-webui/open-webui/discussions/21849

- Filters can now define toggle_options and toggle_default class attributes to show a dropdown selector inside their toggle button area in the chat IntegrationsMenu
- When a filter with options is toggled ON, a dropdown appears below it; toggling OFF hides it
- The selected dropdown value is passed end-to-end to the filter's inlet/outlet/stream handler via a new __selected_option__ parameter

Filter definition (Python):

```
class Filter:
    toggle = True
    toggle_options = ["low", "medium", "high"]
    toggle_default = "medium"

    async def inlet(self, body, __selected_option__=None):
        # __selected_option__ is "low", "medium", or "high"
        return body
```

Label/value pairs are also supported:

```
toggle_options = [
    {"value": "gpt-4", "label": "GPT-4"},
    {"value": "gpt-3.5", "label": "GPT-3.5"}
]
```

> [!NOTE]


fkrauthan:
This works great. A couple of suggestions to make it even better.

1. Change the icon to a combo icon where if I click on the X it removes the option but if I click on the left hand side (e.g. the icon) it opens the change dropdown dialog/popover so if I want to adjust reasoning I don't have to first remove it to then re-add it with a different value
2. Add some type of indicator what option is selected. Ether on hover (as a tooltip) or permanent. Might not be as important if we have an easy way to open the setting as that could be used

tjbck:
Scope creep.

fkrauthan:
Sorry why was decided not to go ahead with this feature (even in its original version)?

Classic298:
I was not heavily interested in it - sorry.

Feel free to take my code, enhance it and fix the scope creep Tim mentioned and reopen a new PR

fkrauthan:
But what exactly is marked as scope crepe here? Besides of a single comment I don't see any discussion? I assumed that comment was more directed towards my two suggestions then the base PR?

Classic298:
honestly? i am not sure. I thought tim meant my PR is scope creep.

I will ask him

Classic298:
@fkrauthan I asked tim

He thinks a drop-down filter would be scope creep and is not a good idea.

Instead you can try to use a toggle filter WITH user valves

Create a toggle filter 
That toggle filter is on or off
And also has user valves (which Should be directly shown in the chat interface also) and you can add your thinking levels into that user valve field.

Would do the same thing.

fkrauthan:
Yeah but as already eluded to in the Discussion this is a really bad user experience. E.g. I want to start a discussion with Extreme Thinking. Then I might turn it off for the next couple of messages to then turn it on again but this time on medium thinking to save some tokens. Right now I have to go to Settings, change it. Turn it on. Type my message. Then turn it off. Now (hopefully didn't forget this) go back to user settings to change the level down again. To then turn it on again for my follow up ask?

Having this as an expansion point allows quiet a lot of customization to developers without really polluting the code base in my opinion.

Like I've looked at the change and it didn't really look like that much code to make this happening.

Classic298:
What do you mean? User valves are also directly in the chat

PR #21856, opened 2026-02-25
fix: sharing modal not showing users when access_grants missing from stored config

Classic298:
## Summary
- Fix sharing modal not showing user-to-user sharing options for non-admin users after the `access_grants` permission feature was introduced
- Root cause: the new `access_grants` permission section was missing from three places that enumerate permission keys — backend `get_permissions()` fallback, `EditGroupModal.init()` merge, and `Permissions.svelte` `fillMissingProperties()`
- When `USER_PERMISSIONS` was saved to the DB before `access_grants` existed, the stored config lacked the key, causing `$user.permissions.access_grants` to be `undefined` and `shareUsers` to evaluate to `false`

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/access_control.py
+++ b/backend/open_webui/utils/access_control.py
@@ -66,6 +66,11 @@ def get_permissions(
     # Ensure all fields from default_permissions are present and filled in
     permissions = fill_missing_permissions(permissions, default_permissions)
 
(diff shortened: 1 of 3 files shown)


Classic298:
https://github.com/open-webui/open-webui/commit/ace69bba7512dc0a653695f9e6311712dfabb640

PR #21869, opened 2026-02-25
fix: SensitiveInput not capturing password manager autofill values

ym7cd:
# Pull Request Checklist

- [x] **Changelog:** Provided below.
- [x] **Testing:** Manually tested with iCloud Passwords Chrome extension autofill — confirmed the fix captures autofilled values correctly.
- [x] **Code review:** Self-reviewed; minimal change (1 file, 5 lines changed).
- [x] **Git Hygiene:** Single atomic commit on a clean branch based on upstream/main.
- [x] **Title Prefix:** fix:

# Changelog Entry

### Description

The SensitiveInput component used one-way {value} binding with an on:change event handler. Password managers (e.g. Apple iCloud Passwords Chrome extension) autofill the input DOM value directly without triggering the change event, causing the Svelte variable to remain empty ("") while the field visually appears filled. This results in an empty password being sent to the backend, returning "The email or password provided is incorrect" even though the correct password is displayed in the field.

### Added

- autocomplete exported prop on SensitiveInput (defaults to off), allowing parent components to pass through autocomplete hints (e.g. current-password)
- name exported prop on SensitiveInput (defaults to empty string), allowing parent components to pass the input name attribute

### Fixed

- Password manager autofill (e.g. iCloud Passwords, 1Password, Bitwarden Chrome extensions) not being captured by SensitiveInput, causing empty password to be submitted on login

### Additional Information

- Verified from Chrome DevTools Network tab: before the fix, password was empty when using password manager autofill; after the fix, the correct password is sent
- All existing usages of SensitiveInput (API key fields in settings pages) are unaffected since they do not pass autocomplete or name, so the defaults preserve current behavior
- Related issue: on:change event only fires on user-initiated changes with blur. Browser extension-based password managers set input.value programmatically without dispatching change or input events, so the Svelte variable never updates. Replacing {value} + on:change with bind:value fixes this.


tjbck:
Addressed in dev.

PR #21884, opened 2026-02-25
perf: fast-path comparison in ResponseMessage to skip JSON.stringify during streaming

Classic298:
ResponseMessage compared the entire message object via JSON.stringify on every reactive tick to detect changes. During streaming, content changes on every token, making the two O(content_length) JSON.stringify calls always return different results — pure wasted work. Add a fast O(1) comparison on content and done fields first. When either differs (the common streaming case), skip straight to cloning. Only fall through to the expensive JSON.stringify comparison for infrequent changes like sources, annotations, or status updates.

<ins>**this is a completely invisible change - free lunch**</ins>

> [!NOTE]


tjbck:
Thanks!

PR #21984, opened 2026-02-28
fix: graceful error handling for malformed tool call arguments

sxu75374:
## Summary

When an LLM produces truncated or malformed JSON in tool call arguments (e.g., during long tool calls that exceed token limits), the SSE stream crashes with an unhandled exception. This PR adds graceful error handling: on parse failure, tool execution is skipped and a descriptive error message is returned to the LLM, allowing it to retry or respond appropriately.

## Changes

- **`middleware.py`**: Add `tool_args_parse_error` flag when `json.loads()` fails on tool arguments. Guard tool execution behind this flag. On error, return a descriptive error message to the LLM instead of crashing.
- **`test_tool_call_parsing.py`**: 11 unit tests covering malformed JSON (truncated, empty, null, non-dict), valid JSON, and edge cases.

## Testing

- ✅ Personally tested with intentionally truncated tool call JSON — stream continues, LLM receives error and retries
- ✅ All 11 new unit tests pass
- ✅ Existing tool call flows unaffected (normal JSON parsing path unchanged)

Signed-off-by: sxu75374 <imshuaixu@gmail.com>


tjbck:
Addressed in dev.

PR #22106, opened 2026-03-01
fix: offline model retrieval, re-raise to disable instead of returning useless fallback

Classic298:

> [!NOTE]

### DIFF
--- a/backend/open_webui/retrieval/utils.py
+++ b/backend/open_webui/retrieval/utils.py
@@ -1306,6 +1306,8 @@ def get_model_path(model: str, update_model: bool = False):
         return model_repo_path
     except Exception as e:
         log.exception(f"Cannot determine model snapshot path: {e}")
+        if OFFLINE_MODE:
+            raise
         return model
 
 


Classic298:
closes https://github.com/open-webui/open-webui/issues/21405

re: https://github.com/open-webui/open-webui/pull/21406

coauthor added

tjbck:
@Classic298 can we verify that there won't be any downstream side effects arising from this change?

Classic298:
I will run additional verification.

Classic298:
No downstream side effects. Every caller already has proper exception handling. The re-raise simply prevents the current broken fallback (returning a useless string path) and instead lets the existing error-handling machinery do its job, resulting in the feature being cleanly disabled.

Classic298:
Scenario | Before PR | After PR
-- | -- | --
Offline, model cached | ✅ Works | ✅ Works (no change, snapshot_download succeeds)
Offline, model NOT cached | ❌ Returns invalid path string → confusing downstream errors | ✅ Raises → caught by callers → feature disabled cleanly
Online, model NOT cached | ✅ Downloads model | ✅ Downloads model (no change, OFFLINE_MODE is False)
Online, download fails | Returns model string fallback (existing behavior preserved) | Returns model string fallback (no change, OFFLINE_MODE is False)

tjbck:
Thanks!

PR #21250, opened 2026-02-08
fix: handle error dict in citation parser when web search fails

veeceey:
# Pull Request Checklist

- [x] **Target branch:** Verify that the pull request targets the `dev` branch.
- [x] **Description:** Provide a concise description of the changes made in this pull request down below.
- [x] **Changelog:** Ensure a changelog entry following the format of Keep a Changelog is added at the bottom of the PR description.
- [x] **Testing:** Performed manual tests to verify the implemented fix works as intended.
- [x] **Agentic AI Code:** This Pull Request has gone through additional human review AND manual testing.
- [x] **Code review:** Performed a self-review of the code.
- [x] **Title Prefix:** Prefixed with `fix:` to categorize this pull request.

# Changelog Entry

### Description

Fixes `AttributeError: 'str' object has no attribute 'get'` crash in citation parser when builtin tools like `search_web` fail (timeout, API error, etc.). When tools fail, they return `{"error": "..."}` as a JSON dict. The citation parser expected a list and iterated over the dict keys (strings), causing the AttributeError. This fix parses JSON once at the top of the function and returns an empty list when an error dict is detected, allowing the chat to continue gracefully.

Supersedes #21207 (closed due to missing CLA and wrong target branch).
Fixes #21070

### Changed

- Modified `backend/open_webui/apps/webui/routers/utils.py`:
  - Added early JSON parsing at the top of `get_citation_source_from_tool_result()`
  - Added error dict detection that returns empty list when `{"error": "..."}` is found
  - Reuses parsed JSON result across all tool branches, eliminating redundant `json.loads` calls

### Fixed

- Fixed `AttributeError: 'str' object has no attribute 'get'` crash when web search or other builtin tools fail
- Fixed chat continuation - errors no longer crash the chat, allowing users to continue their conversation
- Fixed redundant JSON parsing by parsing once and reusing the result

---

### Additional Information

**Root Cause**

When builtin tools like `search_web` fail (timeout, API error, network issue, etc.), they return:
```json
{"error": "timeout message here"}
```

The citation parser in `get_citation_source_from_tool_result` expected the result to be a list like:
```json
[{"url": "...", "title": "...", "content": "..."}]
```

When it received an error dict instead, it iterated over the dict keys (`["error"]`), treating them as citation objects. When it tried to call `tag.get('name')` on the string `"error"`, it crashed with `AttributeError`.

**Solution**

1. Parse JSON once at the top
2. Check if result is a dict with an "error" key - if so, return empty list immediately
[...]

PR #22118, opened 2026-03-01
fix: add type guards in citation parser to prevent AttributeError

umut-polat:
# Pull Request Checklist

- [x] **Description:** Type guards added to citation parser to prevent AttributeError when tools return error strings
- [x] **Testing:** Manually verified that passing a string to each tool branch now returns `[]` instead of crashing
- [x] **Code review:** Self-reviewed, minimal 6-line change

# Changelog Entry

### Description

When a tool (e.g. `search_web`) returns an error string instead of the expected data structure, `get_citation_source_from_tool_result` crashes with `AttributeError: 'str' object has no attribute 'get'`. This happens because the JSON parse failure silently falls through, leaving `tool_result` as a plain string.

### Fixed

- Added `isinstance` type guards in `get_citation_source_from_tool_result` for three tool branches:
  - `search_web`: checks result is a `list` before iterating
  - `view_knowledge_file`: checks result is a `dict` before calling `.get()`
  - `query_knowledge_files`: checks result is a `list` before iterating
- Returns empty source list on type mismatch instead of crashing

Fixes #21070

---

### Additional Information

Previous PRs (#21638, #21209, #21207, #21244, #21250) attempted to fix this but were closed. The root cause is still present in the current codebase — the `isinstance` guards are the minimal defensive fix.

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -183,6 +183,8 @@ def get_citation_source_from_tool_result(
         if tool_name == "search_web":
             # Parse JSON array: [{"title": "...", "link": "...", "snippet": "..."}]
             results = tool_result
+            if not isinstance(results, list):
+                return []
             documents = []
             metadata = []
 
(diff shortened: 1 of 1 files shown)


tjbck:
Addressed in dev.

PR #22153, opened 2026-03-02
fix: trigger title generation when chat title is still empty after cancelled first response

Classic298:
he background_tasks.title_generation flag was only sent when the message array had exactly 1 entry (or 2 with a system prompt) i.e., the very first user message. If the user cancelled that first response, the cancelled assistant message remained in the history, so on any subsequent message messages.length > 2 and the title generation condition was never met again. The chat stayed as "New Chat" forever.

https://github.com/open-webui/open-webui/issues/21961

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -2231,10 +2231,11 @@
 
 				background_tasks: {
 					...(!$temporaryChatEnabled &&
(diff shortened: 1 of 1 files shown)


Classic298:
@tsachig

Classic298:
@tjbck untested but might fix it

PR #22165, opened 2026-03-02
fix: align analytics message count with token count by filtering on assistant role

Classic298:
The get_message_count_by_user query was counting messages of ALL roles (user, assistant, system), while get_token_usage_by_user only aggregated tokens from assistant messages with non-null usage data.

This mismatch caused the Admin Panel Analytics to show inflated message counts alongside zero tokens for users whose messages were predominantly user-role messages without any token usage data.

Align the message count query with the token usage query by adding a role='assistant' filter, so both metrics reflect the same underlying dataset of assistant responses.

> [!NOTE]


Classic298:
@pr-validator-bot full review

Classic298:
The finding is technically accurate but the recommendation would make the analytics less useful.

The meaningful alignment was on role='assistant', which this PR addresses. The remaining filter differences between the two queries are intentional — each has the filters appropriate to its purpose.

Classic298:
@tjbck ready 4 review

Classic298:

Classic298:
@tjbck

PR #14886, opened 2025-06-11
Isolated fix to single section

2underscores:
This is a fix IF the fix for this issue even belongs in Open WebUI. I'm not sure of the actual best place this one should be fixed, so have raised a discussion for this issue here - https://github.com/open-webui/open-webui/discussions/14882. Discussion has diagrams and more details. 

I feel like the fix for this issue does belong in Open WebUI, but feel free to just close this PR if you think it actually belongs somewhere else (e.g. bedrock gateway). If it does belong here, but the late mutation is a bit ugly, I can move it up closer to the response stream receiving part.

# Pull Request Checklist

**Before submitting, make sure you've checked the following:**

- [x] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [x] **Description:** Provide a concise description of the changes made in this pull request.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [ ] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [ ] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [ ] **Testing:** Have you written and run sufficient tests to validate the changes? 
- [x] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [x] **Prefix:** fix: MCP Tools with Empty Arguments JSON Parsing Error

# Changelog Entry

### Description

- Fixed JSON parsing errors when MCP (Model Context Protocol) tools with no arguments are executed, preventing "Expecting value: line 1 column 1 (char 0)" errors
- Added NODE_OPTIONS build argument support to prevent frontend build memory allocation failures during Docker builds

### Changed
- Tool call argument processing in middleware to convert empty argument strings to valid JSON objects before tool execution
- Debug log to an actual error log.

### Fixed
- **MCP Tool Execution**: Fixed JSON parsing errors when MCP tools with no arguments are called, where empty argument strings `""` are now properly converted to valid JSON empty objects `"{}"`

### Security
- [None]

### Breaking Changes
- [None - changes are backward compatible]

### Screenshots or Videos
Screenshot shows a before after in a single chat of the error and then fixed
image

### Contributor License Agreement

[...]

Issue #14577, opened 2025-06-01
issue: Tool calling BROKEN


KyleF0X:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

v0.6.13

### Ollama Version (if applicable)

0.9.0

### Operating System

windows 11

### Browser (if applicable)

Firefox

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

For the qwen3:30b-a3b to call tools and answer the question.

As can be seen here Claude 3.7 via open router, it can correctly use the tool.

![Image](https://github.com/user-attachments/assets/4140e85b-c730-4a97-9a71-f95dc61e8d4e)

### Actual Behavior

Half the time qwen3:30b-a3b says is cant use tool, 
![Image](https://github.com/user-attachments/assets/b9239244-1f08-495f-8b8e-e2b5895c993f)

the other half of the time it makes the tool call, but never answer.

![Image](https://github.com/user-attachments/assets/60365923-159a-4861-89c1-07bad85866a2)

### Steps to Reproduce

install LLM - ollama run qwen3:30b-a3b
setup mcp time server
```
},
"time": {
  "command": "uvx",
  "args": ["mcp-server-time", "--local-timezone=America/Los_Angeles"]
}
```
have mcpo setup to share the tool

![Image](https://github.com/user-attachments/assets/ea552f8f-be88-48c2-a937-ebe89ae8cb2d)

[...]

Issue #13125, opened 2025-04-22
issue:MCP tool call error


chelseaztq:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

0.6.5

### Operating System

Ubuntu 22.04

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have listed steps to reproduce the bug in detail.

### Expected Behavior

I am using the MCP server service of openwebui,the tool call cannot be performed normally without using the native mode,But when I use the native mode, I found that the arguments parameter is not parsed correctly during the streaming output parsing.
[[](url)](https://github.com/open-webui/open-webui/blob/852d9dcbe9b4406e1da230a14642f35ac7bb0d2b/backend/open_webui/utils/middleware.py#L1689)
The specific reason is that the "current_response_tool_call" does not have the "arguments" key value, I'm not sure if it's a problem with the vllm inference framework I'm using, in any case, there are no "arguments" here, and the version of the vllm inference framework I'm using is 0.8.4.
So, this led to the situation where delta_arguments was not correctly added to the arguments when assigning here.
This also indirectly leads to errors when calling specific tools later.
https://github.com/open-webui/open-webui/blob/852d9dcbe9b4406e1da230a14642f35ac7bb0d2b/backend/open_webui/utils/tools.py#L563
Because here parms is actually an empty {}, it evaluates to false, and the code will directly jump to raise Exception

### Actual Behavior

Refer to the above

### Steps to Reproduce

So I made some repairs, but I feel it's not a fundamental solution to the problem.
Firstly, check if there are arguments in the dict. If not, create this key and assign the default value ""
```python
if delta_arguments:
      # current_response_tool_call[
      #     "function"
      # ][
      #     "arguments"
      # ] += delta_arguments
     current_response_tool_call["function"]["arguments"] = current_response_tool_call["function"].get("arguments", "") + delta_arguments
```

Secondly, because my mcp tool may not pass any parameters to the arguments here, the parsing may result in an empty {}, but when the if parms is performing a boolean value judgment, it is false.
So I also made the following corrections
```python
if operation.get("requestBody", {}).get("content"):
            # if params:
     body_params = params
            # else:
            #     raise Exception(
[...]

PR #22189, opened 2026-03-03
fix: initialize streaming tool call arguments to '{}' instead of empty string

Jseals38:
# Pull Request Checklist

- [ ] **Documentation:** N/A — no user-facing behavior, env vars, or API changes.
- [ ] **Dependencies:** N/A — no new or upgraded dependencies.
- [x] **Testing:** Manually tested with parameterless GET endpoint tools in streaming mode. Confirmed the error no longer occurs on first attempt.
- [x] **Agentic AI Code:** This PR was written and tested by a human.
- [x] **Code review:** Self-reviewed — one-line change.
- [x] **Design & Architecture:** Minimal fix, no new settings or architecture changes.
- [x] **Git Hygiene:** Single atomic commit, single file changed.

# Changelog Entry

### Description

- Parameterless tool calls fail during streaming because `setdefault("arguments", "")` seeds the arguments field with an empty string. Both `ast.literal_eval("")` and `json.loads("")` fail, producing "Tool call arguments could not be parsed" errors. Changing the seed to `"{}"` makes both parsers succeed, returning an empty dict `{}`.

### Added

- N/A

### Changed

- N/A

### Deprecated

- N/A

### Removed

- N/A

### Fixed

- Fixed streaming tool call argument initialization in `middleware.py`: changed `setdefault("arguments", "")` to `setdefault("arguments", "{}")` so parameterless tool calls parse correctly on the first attempt instead of failing and requiring a model retry.

### Security

- N/A

### Breaking Changes

- N/A

---

### Additional Information

**Root cause:** During streaming, when a new tool call is encountered, the arguments field is initialized via `setdefault("arguments", "")`. For parameterless tools where the model never sends argument content, this empty string becomes the final value. The downstream parser at `tool_call.get("function", {}).get("arguments", "{}")` retrieves it (the `.get` default of `"{}"` only applies when the key is *missing*, but the key *is* present with value `""`), then both `ast.literal_eval("")` and `json.loads("")` raise exceptions.

**Scope & safety:** `setdefault` only applies when the `"arguments"` key is absent from the first streaming chunk. When a model includes `"arguments": ""` explicitly (the standard behavior for tools that have parameters), `setdefault` does not fire and behavior is completely unchanged.

**Edge case considered:** If a model omitted the arguments key in the first chunk but then streamed argument content in subsequent chunks, the seeded `"{}"` would prepend to the streamed JSON. However, in OpenAI-compatible streaming the first tool call chunk virtually always includes `"arguments": ""` explicitly, so `setdefault` never activates in that path. This pattern is not observed in practice across OpenAI, Azure OpenAI, or other compatible providers.

- Related: #13125, #14886, #15346, #14577

### Screenshots or Videos

**Before fix** — parameterless tool call fails on first attempt:

Error: Tool call arguments could not be parsed. The model generated malformed or incomplete JSON for tool_call. Please try again.

Model retries with explicit `{}` or `{"key": {}}`, wasting a round-trip.

**After fix** — parameterless tool call succeeds on first attempt with no retry needed.

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -3701,7 +3701,7 @@ async def streaming_chat_response_handler(response, ctx):
                                                     ].setdefault("name", "")
                                                     delta_tool_call[
                                                         "function"
-                                                    ].setdefault("arguments", "")
+                                                    ].setdefault("arguments", "{}")
                                                     response_tool_calls.append(
                                                         delta_tool_call
                                                     )


chienhsiang:
Thank you so much! I just ran into this issue, too.

tjbck:
Wrong place to fix, but updated in dev to address!

Classic298:
https://github.com/open-webui/open-webui/commit/4ab831b2596f92fc247c8d788f584ae82d63a88d

PR #22194, opened 2026-03-03
perf: rewrite createMessagesList from recursive to iterative

Classic298:
Replace the recursive spread-based implementation with an iterative push+reverse approach. The recursive version created a new array at each level of recursion via spread, resulting in O(d^2) array copies where d is the conversation depth. The iterative version walks from the target message to the root, pushes each message, and reverses once at the end for O(d) total work.

No behavioral change - same input produces the same output array.

# This was tested extensively

No side effects observed even in branched chats and navigating between branches.

> [!NOTE]


tjbck:
Thanks!

PR #22195, opened 2026-03-03
perf: guard TTS sentence parsing behind showCallOverlay check

Classic298:
The chatCompletionEventHandler runs getMessageContentParts() and removeAllDetails() on every streaming token to extract sentences for real-time TTS dispatch via CustomEvent('chat'). These functions perform multiple O(n) regex passes over the full accumulated message content, resulting in O(n^2) total work over a streaming response.

<ins>**The only consumer of these events is CallOverlay.svelte, which is only mounted when showCallOverlay is true. Without the overlay open, the parsing runs but the dispatched events have no listeners.**</ins>

Wrap all three TTS parsing blocks in an if () guard so the expensive regex work is skipped entirely for the vast majority of users who are not using the voice call feature.

# This PR was tested extensively

with normal conversations

with call-mode conversations

with mixed conversations

TTS and STT

all work.

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -1561,27 +1561,29 @@
 						navigator.vibrate(5);
 					}
 
-					// Emit chat event for TTS
-					const messageContentParts = getMessageContentParts(
-						removeAllDetails(message.content),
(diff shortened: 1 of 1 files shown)


tjbck:
Thanks!

PR #22184, opened 2026-03-03
Feat: katex render more equations correctly by splitting delimiters into strict and loose types

yangqian:
# Pull Request Checklist

### Note to first-time contributors: Please open a discussion post in [Discussions](https://github.com/open-webui/open-webui/discussions) to discuss your idea/fix with the community before creating a pull request, and describe your changes before submitting a pull request.

This is to ensure large feature PRs are discussed with the community first, before starting work on it. If the community does not want this feature or it is not relevant for Open WebUI as a project, it can be identified in the discussion before working on the feature and submitting the PR.

**Before submitting, make sure you've checked the following:**

- [X] **Target branch:** Verify that the pull request targets the `dev` branch. **PRs targeting `main` will be immediately closed.**
- [X] **Description:** Provide a concise description of the changes made in this pull request down below.
- [X] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [X] **Testing:** Perform manual tests to **verify the implemented fix/feature works as intended AND does not break any other functionality**. Include reproducible steps to demonstrate the issue before the fix. Test edge cases (URL encoding, HTML entities, types). Take this as an opportunity to **make screenshots of the feature/fix and include them in the PR description**.
- [X] **Agentic AI Code:** Confirm this Pull Request is **not written by any AI Agent** or has at least **gone through additional human review AND manual testing**. If any AI Agent is the co-author of this PR, it may lead to immediate closure of the PR.
- [X] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [X] **Design & Architecture:** Prefer smart defaults over adding new settings; use local state for ephemeral UI logic. Open a Discussion for major architectural or UX changes.
- [X] **Git Hygiene:** Keep PRs atomic (one logical change). Clean up commits and rebase on `dev` to ensure no unrelated commits (e.g. from `main`) are included. Push updates to the existing PR branch instead of closing and reopening.
- [X] **Title Prefix:** To clearly categorize this pull request, prefix the pull request title using one of the following:
  - **BREAKING CHANGE**: Significant changes that may affect compatibility
  - **build**: Changes that affect the build system or external dependencies
  - **ci**: Changes to our continuous integration processes or workflows
  - **chore**: Refactor, cleanup, or other non-functional code changes
  - **docs**: Documentation update or addition
  - **feat**: Introduces a new feature or enhancement to the codebase
[...]

PR #22196, opened 2026-03-03
Perf: precompile katex unicode regex

Classic298:
# This PR was heavily tested by me

and will fix this discussion: https://github.com/open-webui/open-webui/discussions/20878

https://github.com/user-attachments/assets/b1c6f800-4831-4b84-83a8-f30fef44518f

@danilowanner testing urgently wanted and confirmation from your side that this PR helps you

> [!NOTE]

### DIFF
--- a/src/lib/utils/marked/katex-extension.ts
+++ b/src/lib/utils/marked/katex-extension.ts
@@ -17,7 +17,7 @@ const ALLOWED_SURROUNDING_CHARS =
 // This regex uses Unicode property escapes (\p{Script=Han}, etc.) which are
 // extremely expensive to compile - doing so on every call caused ~87% of
 // markdown rendering time to be spent in KaTeX regex compilation.
-const SURROUNDING_CHARS_REGEX = new RegExp(`[${ALLOWED_SURROUNDING_CHARS}]`, 'u');
+const ALLOWED_SURROUNDING_CHARS_REGEX = new RegExp(`[${ALLOWED_SURROUNDING_CHARS}]`, 'u');
 
 // const DELIMITER_LIST = [
 //     { left: '$$', right: '$$', display: false },
@@ -81,7 +81,7 @@ function katexStart(src, displayMode: boolean) {
 			if (displayMode && src.charAt(i + 1) !== '$') {
 				continue;
 			}
-			if (i === 0 || SURROUNDING_CHARS_REGEX.test(src.charAt(i - 1))) {
+			if (i === 0 || ALLOWED_SURROUNDING_CHARS_REGEX.test(src.charAt(i - 1))) {
 				return i;
 			}
 		} else if (ch === 92 /* \ */) {
@@ -94,7 +94,7 @@ function katexStart(src, displayMode: boolean) {
 				// Inline: \( or \ce{ or \pu{
 				if (next !== '(' && next !== 'c' && next !== 'p') continue;
 			}
-			if (i === 0 || SURROUNDING_CHARS_REGEX.test(src.charAt(i - 1))) {
+			if (i === 0 || ALLOWED_SURROUNDING_CHARS_REGEX.test(src.charAt(i - 1))) {
 				return i;
 			}
 		}
--- a/src/lib/utils/marked/katex-extension.ts
+++ b/src/lib/utils/marked/katex-extension.ts
@@ -73,48 +73,31 @@ export default function (options = {}) {
 }
 
 function katexStart(src, displayMode: boolean) {
-	const ruleReg = displayMode ? blockRule : inlineRule;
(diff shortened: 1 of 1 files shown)


Classic298:
@danilowanner Your profiling pointed us right at it. The KaTeX extension was recompiling a RegExp with Unicode property escapes (\p{Script=Han}, etc.) on every single call to its start() function - hundreds of times per parse. We now pre-compile it once at module load time.

PR is ready. Could you re-run your Safari profiler against it and compare the flame chart? Would love to see the before/after on the KaTeX portion.

danilowanner:
I'll help check this out locally to test 👍

danilowanner:
I did some thorough profiling and testing and it appears the change does not have a significant impact.

## Call trees (sampled) `origin/main`
main call tree

## Call trees (sampled) `Classic298:perf/precompile-katex-unicode-regex`
classic298 call tree

The function in question seems to be `start()` within the katex extension.
- The **relative time spent** on it has not changed.
- **Absolute times** mainly vary depending on the length of the streamed response and increases as the streaming progresses.

The profiles vary slightly each time due to differences in LLM streaming and responses. I ran the profiles multiple times to see if there is any noticeable difference, but none could be found.

👉 There may still be a small change in Regex speed, which would have to be measured in absolute terms by running a consistent test case (benchmark).

_Note: the profiles show natural variations due to the sampling nature of the profiler. It only catches some of the large call stacks in a random manner._

Classic298:
Hm. So the compiled regex only has a small performance boost at best. Okay thanks. This should still get merged because anything is better than nothing, but I'll have to dig deeper into KaTeX

Classic298:
@danilowanner your profiling tests are wanted again - i have made further improvements to this PR on this branch

src/lib/utils/marked/katex-extension.ts
```diff
@@ -13,6 +13,12 @@ const ALLOWED_SURROUNDING_CHARS =
 	'\\s。，、､;；„“‘’“”（）「」『』［］《》【】‹›«»…⋯:：？！～⇒?!-\\/:-@\\[-`{-~\\p{Script=Han}\\p{Script=Hiragana}\\p{Script=Katakana}\\p{Script=Hangul}';
 // Modified to fit more formats in different languages. Originally: '\\s?。，、；!-\\/:-@\\[-`{-~\\p{Script=Han}\\p{Script=Hiragana}\\p{Script=Katakana}\\p{Script=Hangul}';
 
+// Pre-compile the surrounding character regex once at module load time.
+// This regex uses Unicode property escapes (\p{Script=Han}, etc.) which are
+// extremely expensive to compile - doing so on every call caused ~87% of
+// markdown rendering time to be spent in KaTeX regex compilation.
+const SURROUNDING_CHARS_REGEX = new RegExp(`[${ALLOWED_SURROUNDING_CHARS}]`, 'u');
```
tjbck: `ALLOWED_SURROUNDING_CHARS_REGEX`

danilowanner:
Here are updated profiles, showing a good improvement overall. ✅ 
I did two runs and both show less work spent on Katex:

# Run 1
Screenshot 2026-03-05 at 5 10 31 PM

# Run 2
Screenshot 2026-03-05 at 5 12 48 PM

Classic298:
amazing - but both are new runs? how does it compare to without the perf improvement? ~relative gain?

Classic298:
@tjbck safe to merge

danilowanner:
> amazing - but both are new runs? how does it compare to without the perf improvement? ~relative gain?

Both are new runs with the same prompt and model as before.

- Before:
  - Block tokens ~ 70% of total time
  - Katex ~ 65% of total time
- After:
  - Block tokens ~ 40% of total time
  - Katex ~ 25% of total time
  
👉 The sample size is small, just a few runs each time. But the speedup is significant.

Also the absolute time spend on parsing Tokens seems to have been reduced quite a bit.
**At a glance I would say roughly 2x speed.**

Classic298:
🔥🔥🔥🔥🔥

tjbck:
Thanks!

yangqian:
Could you also look into this pull request? https://github.com/open-webui/open-webui/pull/22184
I believe it is related to the addressing the slow down.
The inline delimeters (contains the frequent appearing $) are currently also matching the entire document instead of a single line which could not only match wrong patterns if there are dollar signs in multiple lines and also slow down the rendering a lot.
e.g.
```md
$10$——
asdf
ad
d
d
d
d $\cdotp$ b
```
This incorrectly tries to render 
```latex
$10$——
asdf
ad
d
d
d
d $
````
as an equation. This multiline pattern matching is intrinstically slow.

Issue #22177, opened 2026-03-03
bug: Streaming tool call function name doubled by delta accumulation (GPT-5/5.1)


madnight:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker (Cloud Run)

### Open WebUI Version

v0.8.5

### Operating System

Linux (GCP Cloud Run)

### Browser (if applicable)

Chrome

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**.

### Expected Behavior

When using native function calling with streaming enabled, the tool call function name should be captured correctly from the streaming deltas, and the tool should execute successfully.

For example, an MCP tool registered as `my_server_search` should be called as `my_server_search`.

### Actual Behavior

The function name is **doubled** during streaming delta accumulation. For example, `my_server_search` becomes `my_server_searchmy_server_search`. This causes a silent failure because the doubled name does not match any key in `tools_dict`, so `tool_function_name in tools` evaluates to `False` and the tool is never executed.

The user sees "Tool Executed" in the UI but the tool was never actually called — a silent failure with no error message.

### Root Cause

In `backend/open_webui/utils/middleware.py`, the streaming tool call delta handler accumulates the function name with `+=`:

```python
# ~line 3718 (current dev branch)
if delta_name:
    current_response_tool_call["function"]["name"] += delta_name
```

The first delta for a tool call index creates the entry with `name = "my_server_search"` (via `response_tool_calls.append(delta_tool_call)` at line ~3704). When a subsequent delta for the **same index** also includes `function.name = "my_server_search"`, the `+=` operator concatenates it: `"my_server_search" + "my_server_search"` = `"my_server_searchmy_server_search"`.

[...]

PR #22235, opened 2026-03-04
fix: prevent streaming tool call function name duplication

Classic298:
### Problem

Some models (notably GPT-5, GPT-5.1, and newer OpenAI models) re-send the function name and/or complete arguments in follow-up streaming deltas for the same tool call index. This causes two issues:

1. **Name duplication**: The function name gets doubled (e.g. my_server_search becomes my_server_searchmy_server_search) because the streaming delta handler accumulates names with +=. The doubled name doesn't match any registered tool, so the tool silently fails — the user sees "Tool Executed" but nothing actually ran.

2. **Argument concatenation**: When a model sends multiple complete JSON argument objects under the same tool call index (observed with GPT-5.4 calling search tools with multiple queries), the += accumulation produces concatenated JSON like {"query":"A","count":5}{"query":"B","count":5} which is not valid JSON and fails to parse.

### Fix

**Name deduplication**: Changed the name accumulation to a conditional assignment — the name is only set if it hasn't been set yet. This is safe because:
- Normal case: name arrives in the first delta and is already set via the initial append
- Edge case (empty first delta): the setdefault("name", "") at entry creation sets it to "", which is falsy, so a subsequent delta correctly sets it
- GPT-5/5.1 re-send case: the conditional skips it because the name is already set

**Argument expansion**: Added a pre-processing step before tool call execution that detects concatenated JSON objects in a tool call's arguments using json.JSONDecoder.raw_decode(). If multiple valid JSON objects are found back-to-back, the single tool call entry is expanded into separate entries — one per JSON object — so each gets executed independently. This:
- Doesn't touch the streaming accumulation path, so normal fragmented argument streaming works as before
- Falls through transparently for single-object arguments (the common case)
- Lets downstream ast.literal_eval / json.loads handle non-JSON formats as before

### Affected models

| Model | Name doubled | Args concatenated |
|-------|-------------|-------------------|
| GPT-5.1 | Always | Sometimes |
| GPT-5 | Sometimes | Sometimes |
| GPT-5.2+ | Observed | Observed |
| GPT-4o | Never | Never |
| Claude | Never | Never |

### Related issues

- #22177 — Streaming tool call function name doubled by delta accumulation
- #16138 — Tools names are doubled when calling
- #19656 — Tool call response tokens are duplicated

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -3716,10 +3716,12 @@ async def streaming_chat_response_handler(response, ctx):
                                                         ).get("arguments")
                                                     )
 
-                                                    if delta_name:
+                                                    if delta_name and not current_response_tool_call[
+                                                        "function"
+                                                    ].get("name"):
                                                         current_response_tool_call[
                                                             "function"
-                                                        ]["name"] += delta_name
+                                                        ]["name"] = delta_name
 
                                                     if delta_arguments:
                                                         current_response_tool_call[
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -4118,6 +4118,62 @@ async def streaming_chat_response_handler(response, ctx):
 
                     response_tool_calls = tool_calls.pop(0)
 
+                    # Some models (GPT-5/5.1) concatenate multiple
+                    # complete JSON argument objects under a single
+                    # tool call index.  Expand them into separate
+                    # tool call entries so each is executed.
+                    expanded_tool_calls = []
+                    for tc in response_tool_calls:
+                        raw_args = (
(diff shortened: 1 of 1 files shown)


madnight:
I can confirm that this fix works for GPT 5.1 in our DEV deployment, and I expect it will also resolve the issue for the other affected models.

Classic298:
The new commit i pushed today fixes a bug with newer GPT-5.4 model where it sends weirdly formed json to Open WebUI

e.g. {"query":"A","count":5}{"query":"B","count":5} when it tries to do multiple web searches for example

Works! Using in production

Classic298:
The "original" 4 line fix is urgently needed @tjbck and the second one also, as even older GPT models are affected by this from GPT-5.1 to 5.2 and so forth

PR #22242, opened 2026-03-04
Fix: implement server-side pagination to resolve OOM #22206

batrapulkit:
Summary: Fixes critical OOM crashes (#22206) by implementing server-side pagination and optimizing memory-intensive database queries.

The Problem: Endpoints like /api/v1/chats/all, /api/v1/memories/, and /api/v1/evaluations/ load entire datasets into memory at once, causing OOM crashes on enterprise-scale instances.

The Fix:

Server-side Pagination: Added skip and limit to SQLAlchemy queries in models/chats.py, models/feedbacks.py, models/memories.py, and models/messages.py.

Correlated Subqueries: Optimized delete_shared_chats_by_user_id() to perform server-side deletions, removing the need to materialize ID lists in Python.

Column Projection: Optimized get_reply_user_ids_by_message_id() to query only the user_id column, avoiding full row hydration.

Safe Defaults: All parameters use safe defaults (limit=50–200), ensuring backward compatibility.

Regression Testing: Added test_unbounded_queries.py covering 9 critical test cases.

Testing & Validation:

Manual Verification: Verified syntax via AST validation for all 5 modified files.

Memory Benchmark: Simulated large datasets locally; confirmed that memory usage remains stable (O(1) relative to total record count) due to pagination.

Contributor License Agreement
I certify that this contribution is my own original work and that I have the right to submit it under the project's license.

Fixes #22206


Classic298:
@batrapulkit besides the issues above:

Please submit ONE PR per issue, atomically.

And squash your commits

And rebase

And omit the test file

PR #22278, opened 2026-03-05
perf: skip redundant markdown parsing when content unchanged

Classic298:
Add a content memoization guard to Markdown.svelte that skips marked.lexer() when the content string is identical to what was already parsed. During streaming, content still parses on every animation frame as before. When done, only re-parses if the content string actually changed.

This prevents redundant parsing when a message component re-renders due to non-content changes (e.g. annotation, sources, status) that trigger structuredClone with identical content.

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Messages/Markdown.svelte
+++ b/src/lib/components/chat/Messages/Markdown.svelte
@@ -55,8 +55,11 @@
 		]
 	});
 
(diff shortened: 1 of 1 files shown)


Algorithm5838:
How about this https://github.com/open-webui/open-webui/pull/22183?

PR #22279, opened 2026-03-05
perf: memoize sourceIds computation in ContentRenderer

Classic298:
Move the sourceIds computation from an inline prop expression (recomputed on every render) to a Svelte reactive variable that only recomputes when sources or model actually change.

Also simplified the reduce+spread+filter deduplication pattern to a straightforward loop with Set-based deduplication.

> [!NOTE]


tjbck:
339ed1d72e100c89d8eb26de761dfefe842ef90c

PR #22299, opened 2026-03-06
perf: replace O(n*m) custom model matching with O(n+m) dict lookups

Classic298:
The custom model matching loop in get_all_models() iterated all base models for every custom model, resulting in O(n*m) complexity. This also rebuilt a list of model IDs per custom model in the elif branch.

Replace the nested loops with two dict indices (exact ID and Ollama base-name) built once, then look up each custom model in O(1). Also replace the per-iteration list comprehension for existing model IDs with a precomputed set.

Behavior is identical - same matching logic, same precedence.

> [!NOTE]


tjbck:
29160741a3defa8768a43100cb6e63c56400279c

PR #22301, opened 2026-03-06
perf(models): batch-fetch function valves to eliminate N+1 queries

Classic298:
get_action_priority() called Functions.get_function_valves_by_id() individually for every action on every model — an N+1 query pattern that issued one DB round-trip per (action x model) pair.

Add Functions.get_function_valves_by_ids() that fetches all valves in a single WHERE IN query, then look up each action's valves from the pre-fetched dict inside get_action_priority().

No functional change — same priority resolution, same sort order.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/models.py
+++ b/backend/open_webui/utils/models.py
@@ -331,11 +331,17 @@ async def get_all_models(request, refresh: bool = False, user: UserModel = None)
                 elif meta.get(key) is None:
                     meta[key] = copy.deepcopy(value)
 
+    # Batch-fetch all function valves in one query to avoid N+1 DB hits
+    # inside get_action_priority (previously called per action × per model).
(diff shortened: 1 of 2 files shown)


backend/open_webui/utils/models.py
```diff
@@ -331,11 +331,17 @@ def get_filter_items_from_module(function, module):
                 elif meta.get(key) is None:
                     meta[key] = copy.deepcopy(value)
 
+    # Batch-fetch all function valves in one query to avoid N+1 DB hits
+    # inside get_action_priority (previously called per action × per model).
+    all_valves_by_id = Functions.get_function_valves_by_ids(
```
tjbck: all_function_valves

tjbck:
Thanks!

PR #22342, opened 2026-03-06
perf(models): replace inline base64 icons with URL references in /api/models response

Classic298:
The /api/models endpoint previously embedded full base64 icon data inside every action/filter object on every model. With N models sharing M actions, this duplicated each icon blob N times in the JSON payload — easily tens of megabytes in setups with many models and custom actions.

Replace inline base64 data URIs with lightweight URL references to a new GET /api/v1/functions/id/{id}/icon endpoint that serves the icon binary with cache headers. The response shape stays { data: [...] } — fully OpenAI /v1/models compatible.

Changes:

- Backend (models.py): Add _resolve_icon/_get_raw_icon helpers that replace data URIs with endpoint URLs and provide an is_svg flag

- Backend (functions.py): Add GET /id/{id}/icon endpoint serving icon binary with 24h cache headers

- Frontend (ResponseMessage.svelte): Use action.is_svg flag for dark-mode inversion instead of inspecting the data URI string

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/functions.py
+++ b/backend/open_webui/routers/functions.py
@@ -1,3 +1,4 @@
+import base64
 import os
 import re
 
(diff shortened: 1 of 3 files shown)


Classic298:
ITS RECOMMENDED TO INSTEAD ADD DIRECT URIs TO THE ICONS IN THE DOCSTRING INSTEAD OF USING BASE64

PR #22359, opened 2026-03-07
fix: add support for scope in OAuth refresh token request

pedro-inf-custodio:

> **Note**
> 

---
## Description
This PR fixes an issue where the refresh token request for Microsoft OAuth was failing with error `AADSTS90009`. Previously, the refresh payload only included the `grant_type`, `refresh_token`, `client_id`, and optionally `client_secret`.

Azure AD requires the **scope (or resource)** to be explicitly provided when refreshing a token. Without it, Azure interprets the request as “the application is requesting a token for itself,” which triggers the 400 error:

```
AADSTS90009: Application '[APPLICATION_ID]' is requesting a token for itself.
```

### Changes
- Added support for including a **custom scope** in the refresh token request.
- The scope is read from the environment variable `MICROSOFT_OAUTH_SCOPE`.
- Example format for the scope:

```
openid email profile offline_access api://<Application ID URI>/<custom_scope>
```

- Updated `_perform_token_refresh` to include this scope when refreshing tokens.

### Root Cause
- Azure AD v2.0 requires **explicit scopes** in refresh token requests to determine which resource the new access token should target.
- Omitting `scope` caused Azure to treat the request as self-targeted, resulting in `AADSTS90009`.
- Including the custom scope resolves this and allows token refreshes to succeed.

### Logs Before Fix
```
Token refresh failed for provider microsoft: 400 - {"error":"invalid_request","error_description":"AADSTS90009: Application '[APPLICATION_ID]' is requesting a token for itself."}
```

### Logs After Fix
- Refresh token requests now succeed, and new access tokens are issued without errors.

### Environment Variables
- `MICROSOFT_OAUTH_SCOPE` (required) – the custom scope for token requests.

### Optional Configuration (Potential Follow-up)

As a potential follow-up improvement, it may also be possible to control this behavior via an environment variable, if maintainers consider this a suitable approach:

`OAUTH_REFRESH_TOKEN_INCLUDE_SCOPE=false # default`

This would allow enabling or disabling the inclusion of the `scope` parameter in refresh token requests.

This approach remains compliant with **RFC 6749 Section 6**, where the `scope` parameter is optional during refresh token requests and typically omitted unless required by the OAuth provider.

> This option would only need to be enabled for OAuth providers that require `scope` to be included in refresh token requests (such as some Azure AD configurations).

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -339,6 +339,12 @@ ENABLE_OAUTH_SIGNUP = PersistentConfig(
     os.environ.get("ENABLE_OAUTH_SIGNUP", "False").lower() == "true",
 )
 
+OAUTH_REFRESH_TOKEN_INCLUDE_SCOPE = PersistentConfig(
+    "OAUTH_REFRESH_TOKEN_INCLUDE_SCOPE",
+    "oauth.refresh_token_include_scope",
+    os.environ.get("OAUTH_REFRESH_TOKEN_INCLUDE_SCOPE", "False").lower() == "true",
+)
+
 
 OAUTH_MERGE_ACCOUNTS_BY_EMAIL = PersistentConfig(
     "OAUTH_MERGE_ACCOUNTS_BY_EMAIL",
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -787,6 +787,12 @@ class OAuthClientManager:
             if hasattr(client, "client_secret") and client.client_secret:
                 refresh_data["client_secret"] = client.client_secret
 
+            # Add scope if available in client kwargs (some providers require it on refresh)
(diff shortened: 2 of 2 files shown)


tjbck:
@pedro-inf-custodio is this tested? could you share a screenrecording for before/after?

pedro-inf-custodio:
> @pedro-inf-custodio is this tested? could you share a screenrecording for before/after?

Hi @tjbck, thanks for the quick reply!
Here is the comparison:

**Before:**
image

**After:**
image

I’ve also introduced the `OAUTH_REFRESH_TOKEN_INCLUDE_SCOPE` boolean variable. This ensures backward compatibility, as some users may not require this feature. I added this specifically because tokens issued by Microsoft Graph were causing authentication failures with my custom OAuth API.

tjbck:
LGTM, Thanks!

PR #22364, opened 2026-03-07
fix: prevent code blocks from causing message bubble width inconsistency in widescreen

realsamrat:
# Pull Request Checklist

- [x] **Description:** Concise description of changes provided below.
- [x] **Testing:** Manually reproduced the issue and verified the fix resolves inconsistent bubble widths when scrolling with YAML code blocks in widescreen mode.
- [x] **Code review:** Self-reviewed for code quality and adherence to project standards.
- [x] **Git Hygiene:** Single atomic fix, rebased on `dev`.

# Changelog Entry

### Description

Fixes inconsistent message bubble widths when scrolling through conversations that contain YAML (or other wide) code blocks in widescreen mode. The root cause was missing `min-w-0` on flex containers, which prevented bubbles from shrinking below their content's intrinsic width.

### Fixed

- Message bubble widths no longer shift/reflow when scrolling in widescreen mode with YAML code blocks
- Applied `min-w-0` to flex containers in `ResponseMessage.svelte` and `UserMessage.svelte` to allow proper flex shrinking

---

### Additional Information

- Fixes #5975
- Root cause: `min-w-full` was used on flex children, which forces a minimum width equal to the parent. Without `min-w-0`, flexbox children cannot shrink below their intrinsic content size, so wide code blocks push the bubble container wider than intended, causing layout reflow on scroll.
- Reproduced with: https://github.com/ChatLunaLab/chatluna-character/blob/main/resources/presets/default.yml


Classic298:
please rebase. 23 commits for 1 change is too much

PR #22366, opened 2026-03-07
fix: prevent code blocks from causing message bubble width inconsistency in widescreen

realsamrat:
# Pull Request Checklist

- [x] **Description:** Concise description of changes provided below.
- [x] **Testing:** Manually reproduced the issue using the YAML preset linked in #5975 and verified bubble widths stay consistent while scrolling in widescreen mode.
- [x] **Code review:** Self-reviewed — minimal, focused CSS fix only.
- [x] **Git Hygiene:** Single commit, rebased cleanly on `dev` (1 commit ahead of upstream).

# Changelog Entry

### Description

Fixes inconsistent message bubble widths when scrolling through conversations containing YAML (or other wide) code blocks in widescreen mode.

**Root cause:** `min-w-full` on flex children forces a minimum width equal to the parent container. Without `min-w-0`, flex children cannot shrink below their intrinsic content size — so wide code blocks push the bubble container wider than intended, causing layout reflow on scroll.

### Fixed

- `ResponseMessage.svelte`: replaced `min-w-full` with `min-w-0` on message content containers; added `min-w-0` to flex wrapper
- `UserMessage.svelte`: same fix applied to user bubble and edit containers

---

### Additional Information

- Fixes #5975
- Reproduced with: https://github.com/ChatLunaLab/chatluna-character/blob/main/resources/presets/default.yml
- Both files are part of the same fix — they are the two sides of every conversation (user + response). Fixing only one would leave the UI in a broken half-fixed state.

### DIFF
--- a/src/lib/components/chat/Messages/ResponseMessage.svelte
+++ b/src/lib/components/chat/Messages/ResponseMessage.svelte
@@ -629,7 +629,7 @@
 			/>
 		</div>
 
-		<div class="flex-auto w-0 pl-1 relative">
+		<div class="flex-auto w-0 min-w-0 pl-1 relative">
 			<Name>
 				<Tooltip content={model?.name ?? message.model} placement="top-start">
 					<span id="response-message-model-name" class="line-clamp-1 text-black dark:text-white">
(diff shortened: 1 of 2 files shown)


tjbck:
Duplicate.

PR #22383, opened 2026-03-08
fix: deduplicate chat entries on pagination to prevent each_key_dupli…

Classic298:
…cate error

### Contributor License Agreement

By submitting this pull request, I confirm that I have read and fully agree to the [Contributor License Agreement (CLA)](https://github.com/open-webui/open-webui/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT), and I am providing my contributions under its terms.

> [!NOTE]
> Deleting the CLA section will lead to immediate closure of your PR and it will not be merged in.

PR #22387, opened 2026-03-08
fix: add deterministic tiebreaker to all paginated chat queries

Classic298:
Add Chat.id as a secondary sort key to all paginated chat queries that use offset/limit pagination. When multiple chats share the same updated_at timestamp, the database does not guarantee a stable order across page boundaries, causing chats to appear on multiple pages.

This produces duplicate keys in the Svelte sidebar each-block (each_key_duplicate error). Adding Chat.id as a tiebreaker ensures fully deterministic ordering.

Extends the fix from #22383 (which addressed get_chat_ids_by_model_id) to all remaining paginated chat queries.

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/chats.py
+++ b/backend/open_webui/models/chats.py
@@ -734,13 +734,13 @@ class ChatTable:
                         raise ValueError("Invalid order_by field")
 
                     if direction.lower() == "asc":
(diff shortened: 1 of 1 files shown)


silentoplayz:
LGTM; Thanks!

PR #22420, opened 2026-03-08
fix: replace bare except with Exception in oauth.py

gambletan:
### Description

- In `oauth.py`, a bare `except:` clause was used which catches all exceptions including `SystemExit` and `KeyboardInterrupt`. This prevents clean process shutdown and makes debugging harder.

### Fixed

- Replaced bare `except:` with `except Exception:` so that `SystemExit`, `KeyboardInterrupt`, and other `BaseException` subclasses are no longer silently caught.

### Additional Information

- Tested by reviewing the code path and verifying the fix addresses the issue

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1864,7 +1864,7 @@ async def chat_completion(
                                 "model": model_id,
                             },
                         )
-                except:
+                except Exception:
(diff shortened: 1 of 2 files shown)


tjbck:
Thanks!

PR #22430, opened 2026-03-08
fix: public-read access grant should not confer write access to notes

gambletan:
### Description

- The `has_public_read_access_grant()` check was incorrectly included in the write access calculation. This meant that any user with public read access could also gain write access to notes, resulting in a privilege escalation vulnerability.

### Fixed

- Removed `has_public_read_access_grant()` from the write access logic so that public read access no longer confers write access to notes.

### Additional Information

- Tested by reviewing the code path and verifying the fix addresses the issue

### DIFF
--- a/backend/open_webui/routers/notes.py
+++ b/backend/open_webui/routers/notes.py
@@ -236,7 +236,6 @@ async def get_note_by_id(
             permission="write",
             db=db,
         )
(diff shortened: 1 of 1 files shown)


tjbck:
This is a legacy behaviour we kept on purpose. This can be removed after we introduce public write access grant.

PR #22435, opened 2026-03-08
fix: replace bare string raises with proper exception types

alvinttang:
### Description

- In Python, `raise "some string"` does not raise an exception with that message — it raises a `TypeError: exceptions must derive from BaseException`. This bug exists in three places in the backend where the code intends to signal an error condition but instead raises a confusing `TypeError`, making the actual problem very hard to debug from logs.
- Fixed all three occurrences by replacing `raise "string"` with proper exception types (`ValueError` / `RuntimeError`) that carry the intended error message.

### Affected locations

1. `backend/open_webui/retrieval/utils.py` — `generate_openai_batch_embeddings()`: when API response is missing the `data` key
2. `backend/open_webui/retrieval/utils.py` — `generate_ollama_batch_embeddings()`: when API response is missing the `embeddings` key
3. `backend/open_webui/routers/ollama.py` — `download_file_stream()`: when blob upload to Ollama fails

### Fixed

- Replaced `raise "Something went wrong :/"` with `raise ValueError(...)` with descriptive messages
- Replaced `raise "Ollama: Could not create blob..."` with `raise RuntimeError(...)`
- Error logs now show the actual problem instead of a misleading `TypeError`

### Additional Information

- Tested by reviewing the code path and verifying the fix
- These are all inside `try/except Exception` blocks, so behavior is preserved — only the error message logged changes from a confusing `TypeError` to the actual descriptive error


Classic298:
please rebase and stash your commits

PR #22443, opened 2026-03-08
fix: correct Azure TTS locale extraction for SSML xml:lang

alvinttang:
### Description

- Fixed incorrect locale extraction in Azure TTS SSML generation. `split("-")[:1]` only takes the first segment (e.g., "en" from "en-US"), producing an invalid SSML `xml:lang` attribute. Changed to `[:2]` to preserve the full locale (e.g., "en-US").

### Fixed

- Azure TTS now correctly uses the full locale (e.g., "en-US") in SSML `xml:lang` attribute instead of just the language code (e.g., "en")

### Additional Information

- The other TTS code paths in the same file already use the full locale correctly
- Tested by reviewing the code path and verifying the SSML output format matches Azure documentation requirements
- Rebased onto dev (replaces #22432)


alvinttang:
Rebased onto dev. Thanks for the review\!

tjbck:
Thanks!

PR #22444, opened 2026-03-08
fix: prevent TypeError in Teams webhook when user data is missing

alvinttang:
### Description

- Fixed `TypeError` crash in Teams webhook handler. `json.loads(event_data.get("user", {}))` passes a dict `{}` as default to `json.loads()` which expects a string. When the `"user"` key is absent from `event_data`, this crashes with `TypeError: the JSON object must be str, bytes or bytearray, not dict`.

### Fixed

- Teams webhook handler no longer crashes when `"user"` key is missing from event data

### Additional Information

- Tested by reviewing the code path and confirming the fix handles the missing key case
- Rebased onto dev (replaces #22433)


alvinttang:
Rebased onto dev. Thanks for the review\!

PR #22457, opened 2026-03-08
fix: throttle mouseenter-triggered API calls in model selector and sidebar

Classic298:
- Add 30s cooldown to model selector mouseenter handler to prevent excessive /api/models calls on hover and tab switches
- Add 200ms debounce to sidebar chat item hover-to-load to prevent rapid-fire getChatById calls when scrolling through the chat list
- Clean up debounce timer on component destroy

> [!NOTE]


tjbck:
Should be cached from the backend.

PR #22471, opened 2026-03-09
fix: remove public-read granting write access for notes

gambletan:
## Description

Public-read (user:* read) should not grant write access. This fix ensures that notes with public-read permission only allow read access, not write access.

### DIFF
--- a/backend/open_webui/routers/notes.py
+++ b/backend/open_webui/routers/notes.py
@@ -236,7 +236,6 @@ async def get_note_by_id(
             permission="write",
             db=db,
         )
-        or has_public_read_access_grant(note.access_grants)
     )
 
     return NoteResponse(**note.model_dump(), write_access=write_access)


ppoak:
Hi, I'm serving OpenWebUI (v0.8.10) on windows, which I installed with pypi. When creating read-only note for all user, I got this problem.

**Description**

I created a public note without any access control like the image shows.

image

When I switch to another `user` role account, opening Note interface, and find nothing displayed here.
image

But when I add write access to all groups of users, the note interface displayed normally.

image

image

I am wondering whether the note feature is designed to act like this or it is just a bug. In my perspective, I think that the read-only access for a user doesn't mean you can't see the note on the `Note tab`. You can open the note but no edition is allowed. But as a matter of fact, without granting access of writing, you can not even open and read the note. That's a little bit of weird. Could anyone help me with this or fix the problem, THANKS A LOT!

alvinttang:
Thanks for the detailed reproduction steps, @ppoak\! This confirms the bug — the public-read permission was inadvertently also granting write access, which meant the note wasn't being returned for regular users viewing the note list. This PR specifically addresses that access control issue. Could you give it a try and let me know if it resolves the problem for you?

ppoak:
Sorry I've tried that, but it still not function.

Pull request code changes applied

image

After setting access control to public access, I tried a normal user account to access the note, nothing found.

image

ppoak:
> Sorry I've tried that, but it still not function.
> 
> Pull request code changes applied
> 
> image
> After setting access control to public access, I tried a normal user account to access the note, nothing found.
> 
> image

I’ve tried this again, it indicates the note is just not displayed on the Note interface.  But if user get the share link of the note the normal user can read the note!

PR #22473, opened 2026-03-09
fix: replace bare except with except Exception

gambletan:
## Description

Replace bare except clauses with `except Exception` to follow Python best practices and avoid catching unexpected system exceptions like `KeyboardInterrupt` and `SystemExit`.


tjbck:
Thanks!

PR #22517, opened 2026-03-10
feat: inline ask citation in MessageInput instead of floating modal

jannikstdl:
Text selection + "Ask" now drops a quoted citation bar into the main chat input instead of spawning a separate floating input modal. You type your follow-up question right in the regular input field and the selected text gets prepended as a blockquote on send.

"Explain" is unaffected, still works the same way with the inline streaming response.

image

### DIFF
--- a/src/lib/components/chat/ContentRenderer/FloatingButtons.svelte
+++ b/src/lib/components/chat/ContentRenderer/FloatingButtons.svelte
@@ -1,9 +1,6 @@
 <script lang="ts">
 	import { toast } from 'svelte-sonner';
 
-	import DOMPurify from 'dompurify';
-	import { marked } from 'marked';
-
 	import { getContext, tick, onDestroy } from 'svelte';
 	const i18n = getContext('i18n');
 
@@ -13,7 +10,7 @@
 	import LightBulb from '$lib/components/icons/LightBulb.svelte';
 	import Markdown from '../Messages/Markdown.svelte';
 	import Skeleton from '../Messages/Skeleton.svelte';
-	import { chatId, models, socket } from '$lib/stores';
+	import { chatId, models, socket, quotedText } from '$lib/stores';
 
 	export let id = '';
 	export let messageId = '';
@@ -22,12 +19,9 @@
 	export let messages = [];
 	export let actions = [];
 	export let onAdd = (e) => {};
-
-	let floatingInput = false;
-	let selectedAction = null;
+	export let onClose = () => {};
 
 	let selectedText = '';
-	let floatingInputValue = '';
 
 	let content = '';
 	let responseContent = null;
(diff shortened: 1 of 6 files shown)


src/lib/components/chat/MessageInput.svelte
```diff
@@ -1203,6 +1207,26 @@
 								: ' border-gray-100/30 dark:border-gray-850/30 hover:border-gray-200 focus-within:border-gray-100 hover:dark:border-gray-800 focus-within:dark:border-gray-800'}  transition px-1 bg-white/5 dark:bg-gray-500/5 backdrop-blur-sm dark:text-gray-100"
 							dir={$settings?.chatDirection ?? 'auto'}
 						>
+							{#if $quotedText}
```
2000090079, line 1210: The dismiss button for the quoted citation bar is missing an `aria-label`. 
Screen readers cannot describe what this button does. 

Please add:
`aria-label="Clear quoted text"`

This is important for accessibility compliance.

src/lib/components/chat/MessageInput.svelte
```diff
@@ -1203,6 +1207,26 @@
...
+								<div
+									transition:slide={{ duration: 200, easing: cubicOut }}
```
2000090079, line 1212: The slide transition is a nice touch, but users who have 
`prefers-reduced-motion` enabled in their OS settings will 
still see this animation, which can cause discomfort.

Consider:
`duration: window.matchMedia('(prefers-reduced-motion: reduce)').matches ? 0 : 200`

src/lib/components/chat/MessageInput.svelte
```diff
@@ -1203,6 +1207,26 @@
...
+									class="flex items-center gap-2 mx-2 mt-2 px-3 py-2 bg-gray-50 dark:bg-gray-800 rounded-xl"
+								>
+									<ArrowForward className="size-3.5 shrink-0 text-gray-500" />
+									<div
+										class="flex-1 text-sm text-gray-600 dark:text-gray-300 line-clamp-1"
```
2000090079, line 1217: The quoted text is truncated to 1 line using `line-clamp-1`. 
If the selected text is long, users have no way to see the full content.

Consider adding:
`title={$quotedText}`

This allows users to hover and see the complete quoted text as a tooltip.

src/lib/components/chat/ContentRenderer/FloatingButtons.svelte
```diff
@@ -248,85 +232,38 @@
...
-			<div
-				class="py-1 flex dark:text-gray-100 bg-white dark:bg-gray-850 border border-gray-100 dark:border-gray-800 w-72 rounded-full shadow-xl"
+		<div
+			class="flex flex-row shrink-0 p-0.5 bg-white dark:bg-gray-850 dark:text-gray-100 text-medium rounded-xl shadow-xl border border-gray-100 dark:border-gray-800"
+		>
+			<button
+				aria-label={$i18n.t('Ask')}
+				class="px-1.5 py-[1px] hover:bg-gray-50 dark:hover:bg-gray-800 rounded-xl flex items-center gap-1 min-w-fit transition"
+				on:click={() => {
+					selectedText = window.getSelection().toString();
```
2000090079, line 242: The `window.getSelection().toString()` call inside the Ask button's 
click handler could return an empty string if the user clicks the 
button after accidentally deselecting the text.

Consider adding a guard before setting quotedText:

if (!selectedText) {
  toast.warning('Please select some text first');
  return;
}

This would give users clear feedback instead of silently setting 
an empty quoted text.

Issue #4858, opened 2024-08-23
enh: Default Role Assignment set by Trusted Headers (like the EMAIL & NAME) for eduPersonEntitlement


atnjqt:
**Is your feature request related to a problem? Please describe.**
This feature request is related to closed issue **Support SAML for auth https://github.com/open-webui/open-webui/issues/2535**, which elegantly describes the usage of trusted headers for passing an email & name for a user login, and aims to improve on this solution for Shibboleth integrations.

**Describe the solution you'd like**
It would be fantastic to also pass a trusted custom header for the roles env var `DEFAULT_USER_ROLE` (i.e. https://docs.openwebui.com/getting-started/env-configuration/#default_user_role). This solution would effectively allow Apache + Shibboleth SAML setups to pass an industry standard attribute for `eduPersonEntitlement` used for Authorization (see here: https://spaces.at.internet2.edu/display/federation/user-attr-edupersonentitlement)

> `eduPersonEntitlement` is an URI (either URN or URL) that indicates a set of rights to specific resources. eduPersonEntitlement is defined in the [eduPerson](https://wiki.refeds.org/display/STAN/eduPerson+2020-01)[→](https://tools.ietf.org/html/rfc3986) LDAP object class.

Our IDP is optionally able to release this SAML attribute, and we use this to accomplish custom Authorization mappings. i.e. on the IDP we would configure three groups for the valid group offerings, namely _pending, user, admin_. If an SSO user in one of those groups accesses the openui app, that custom header could be used to auto-provision the default role. This is also significant in that if a user leaves they would get immediate lockout since the custom authz header is no longer present.

**Describe alternatives you've considered**
Manually setting roles for new users who land as pending.

**Additional context**
Really just looking to manage user role mappings dynamically using the eduPersonEntitlement SAML attribute, to further improve the manageability and simplicity of Apache + Shibboleth for hosting this application. The document references some other auth services but this doesn't cleanly map to legacy setups.

tjbck:
PR Welcome!

atnjqt:
> PR Welcome!

Thanks for the reply and yes I did think through this a bit and opened a discussion here: https://github.com/open-webui/open-webui/discussions/4883 

I have a simple SAML group assignment implementation in a fork here: https://github.com/open-webui/open-webui/compare/main...atnjqt:open-webui:feature/saml-group-assignment

Gabgobie:
Hi,

would it be feasible to expand this functionality to the already present OIDC provider?

[...]

PR #22523, opened 2026-03-10
feat: add trusted role header

dometto:
## Description

I've not opened a separate discussion because this feature was already requested, and I think improved, in https://github.com/open-webui/open-webui/issues/4858. In my opinion, however, that issue was wrongly closed without implementing the actually intended behavior (see [here](#issuecomment-3951497432)). Happy to open a new discussion if this is desired.

This PR implements the possibility to determine a user's role (admin, pending, user) at signin, using a trusted header. Open WebUI already has the possibility to handle authentication, and to set *groups*, using a trusted header. But a new user's role will, at present, always be equal to the `DEFAULT_USER_ROLE` setting. This means an identity provider cannot pass on info to Open WebUI on whether a user should be an admin or not.

Implementation is very simple: in the `/signin` route, after checking for the trusted email header and before checking for the trusted groups header, we check for the presence of the configured trusted role header. The user's role is set to this role for the current session, and saved.

## Documentation

*to be added*

[ ] Add docs in [Open WebUI Docs Repository](https://github.com/open-webui/docs). Document user-facing behavior, environment variables, public APIs/interfaces, or deployment steps.

## Testing

I run `open-webui` locally using `uv`, with the following environment variables:

`DEFAULT_USER_ROLE=user WEBUI_AUTH_TRUSTED_ROLE_HEADER=X-Remote-User-Role 
WEBUI_AUTH_TRUSTED_EMAIL_HEADER=X-Remote-User-Mail uv run open-webui serve`

To test the trusted headers, I use the Chrome [ModHeader](https://modheader.com/docs/modheader) extension.

See screengrab below for video.

**Testing locally in Chrome**:

1. Set X-Remote-User-Mail header to `testuser@localhost` and `X-Remote-User-Role` to `user` in ModHeader.
2. Navigate to `http://localhost:8080`
3. Observe that you are logged in as `testuser`, as a normal user (despite being the first user -- trusted header trumps this).
1. Set X-Remote-User-Mail header to `testuser2@localhost` and `X-Remote-User-Role` to `admin` in ModHeader.
2. Navigate to `http://localhost:8080`
3. Observe that you are logged in as `testuser2`, with admin rights (the trusted header trumps the default role of `user`).
4. Change the `X-Remote-User-Role` header in ModHeader to `pending`.
5. Logout -> you are redirected to the `/`. Observe that `testuser2` is now pending.

# Changelog Entry

### Added

🔒  **Trusted header for user roles**: the `WEBUI_AUTH_TRUSTED_ROLE_HEADER` environment variable can be used to configure a trusted header that will determine the user's role (admin, user, or pending).

## Security

- When a trusted role header is configured, clients can set admin role using an HTTP header. This requires making sure that the headers are only set by a reverse proxy/identity provider. However, this is already the case for other trusted headers.

### Additional Information

- When the header changes, the role is updated only when a user logs in again (so if logged in as admin, then change header to user, the user will still be admin in their current session).

### Screenshots or Videos

openwebui_trustedroles

> [!NOTE]


tjbck:
Addressed in  dev.

PR #22533, opened 2026-03-10
feat: improve chat scroll readability

jannikstdl:
Two changes to how scrolling works during chat:

**Scroll user message to top on send** - when you send a message, the user message gets scrolled to the top of the viewport and stays there while the response streams in below. Makes it easier to keep track of what you asked, especially with longer prompts. Same thing happens on regenerate.

**Smarter scroll-to-bottom button** - the button now checks if content has actually grown past the bottom of the viewport before showing up. Previously it would appear right away after sending even though everything was still visible.

Under the hood this adds a `#messages-bottom` marker after the message list with a permanent spacer below it. All scroll-to-bottom and at-bottom detection logic uses this marker instead of `scrollHeight`, so "end of messages" and "end of scrollable area" are properly separated.

### DIFF
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -1262,13 +1262,19 @@
 		}
 	};
 
+	const isAtContentBottom = (): boolean => {
+		const marker = document.getElementById('messages-bottom');
+		if (!marker || !messagesContainerElement) return true;
+		const markerRect = marker.getBoundingClientRect();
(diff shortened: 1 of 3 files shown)


Classic298:
Sorry I have to close this. The CLA template changed and we need you to use the new one and check the box on the new template please.

Also please try to split this into two PRs to make reviewing it easier

jannikstdl:
I don't know why there was a need of close since the CLA section can easily be updated afterwards and both changes depend on each other if you look at the code.

PR #22536, opened 2026-03-10
feat: content-aware scroll-to-bottom button

jannikstdl:
The scroll-to-bottom button currently uses the `autoScroll` flag to decide when to show up. That flag can toggle too early in some cases, making the button flash on when everything is still visible on screen.

This changes it to check the actual DOM position of the content end. If the last message is below the viewport, the button shows. If not, it stays hidden. Falls back to the old `autoScroll` behavior when the `#messages-bottom` marker isn't present.

**Note:** The full benefit of this PR only kicks in when #22535 is also merged, since that PR adds the `#messages-bottom` marker. Without it, this falls back to the existing `autoScroll` flag behavior, which works fine on its own but doesn't give the improved accuracy.

### DIFF
--- a/src/lib/components/chat/MessageInput.svelte
+++ b/src/lib/components/chat/MessageInput.svelte
@@ -11,7 +11,7 @@
 	dayjs.extend(duration);
 	dayjs.extend(relativeTime);
 
-	import { onMount, tick, getContext, createEventDispatcher } from 'svelte';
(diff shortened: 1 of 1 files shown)


ShirasawaSama:
In fact, I've noticed that on iOS devices, sometimes you need to wait for several frames before the change takes effect.

PR #22555, opened 2026-03-11
fix: respect FORWARDED_ALLOW_IPS env var in start.sh

alvinttang:
## Summary

- Replace hardcoded `--forwarded-allow-ips '*'` with `--forwarded-allow-ips "${FORWARDED_ALLOW_IPS:-*}"` in `backend/start.sh`
- The `FORWARDED_ALLOW_IPS` environment variable is now respected, with `*` as the default for backward compatibility
- Fixes both the main uvicorn launch (line 86) and the HuggingFace Space admin setup launch (line 53)

Fixes #22539

## Test plan

- [ ] Set `FORWARDED_ALLOW_IPS` to a specific IP (e.g., `127.0.0.1`) and verify uvicorn starts with that value
- [ ] Leave `FORWARDED_ALLOW_IPS` unset and verify uvicorn defaults to `*` (backward compatible)

contributor license agreement

### DIFF
--- a/backend/start.sh
+++ b/backend/start.sh
@@ -50,7 +50,7 @@ if [ -n "$SPACE_ID" ]; then
   echo "Configuring for HuggingFace Space deployment"
   if [ -n "$ADMIN_USER_EMAIL" ] && [ -n "$ADMIN_USER_PASSWORD" ]; then
     echo "Admin user configured, creating"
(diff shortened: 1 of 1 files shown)


tjbck:
We need the full CLA with the checkbox checked.

PR #22675, opened 2026-03-14
fix: register memory tools when model declares capabilities.memory

### LINKED ISSUE #22666: Memory tools not registered when model has capabilities.memory but features.memory not sent
## Bug Description

When a model has `capabilities.memory: true` in its meta config and the system prompt instructs the model to use `add_memory`/`search_memories`, the model generates tool calls for these functions. However, the tools are not registered for execution because `get_builtin_tools()` in `utils/tools.py` requires `features.get("memory")` to be truthy — which depends on the frontend sending `features: {memory: true}` in the chat request.

This creates a tool loop: the model calls `add_memory`, but since the tool isn't registered (`tool_function_name not in tools`), the result is empty, and the model retries up to `CHAT_RESPONSE_MAX_TOOL_CALL_RETRIES` (30) times.

## Steps to Reproduce

1. Configure a model with `capabilities.memory: true` in the model editor
2. Set a system prompt that tells the model to use `add_memory` for preferences
3. Set `function_calling: native` in model params
4. Start a new chat with this model
5. Send: "Remember my favorite color is blue"
6. The model calls `add_memory` but it loops 30 times because the tool isn't in the execution dict

## Expected Behavior

When a model has `capabilities.memory: true`, memory tools should be registered for execution in `get_builtin_tools()` regardless of the per-chat `features.memory` flag.

## Root Cause

In `utils/tools.py`, `get_builtin_tools()` line ~468:
```python
if is_builtin_tool_enabled("memory") and features.get("memory"):
```

This requires BOTH the builtin tool category AND the per-chat features flag. But `features.memory` depends on the frontend sending it, which may not happen even when the admin memory setting is enabled and the model declares memory capability.

## Proposed Fix

Also check model capabilities as a fallback:
```python
if is_builtin_tool_enabled("memory") and (features.get("memory") or get_model_capability("memory", False)):
```

This ensures that models with explicit `capabilities.memory: true` always have memory tools available.

## Environment

- Open WebUI v0.8.10
- Backend: vllm-mlx with Qwen 3.5 models
- function_calling: native

Thump604:
# Pull Request Checklist

- [x] **Testing:** Manually tested end-to-end in browser
- [x] **Agentic AI Code:** Human-reviewed and manually tested

# Changelog Entry

### Description

When a model declares `capabilities.memory: true` in its meta config, memory tools (`add_memory`, `search_memories`, etc.) should be registered for execution. Without this fix, models with native function calling and memory capability enter a 30-iteration tool retry loop because `add_memory` appears in the tool schema but isn't in the execution dict.

### Root Cause

`get_builtin_tools()` in `utils/tools.py` only checked `features.get("memory")` (per-chat toggle sent by frontend). This could be `False` even when:
- The admin memory setting is enabled
- The model explicitly declares `capabilities.memory: true`
- The frontend sends `features.memory: true` for some chats but not others

This is inconsistent with how other capabilities like `web_search` and `image_generation` are handled elsewhere in the codebase.

### Changed

- `backend/open_webui/utils/tools.py`: `get_builtin_tools()` now also checks `get_model_capability("memory", False)` as a fallback when `features.get("memory")` is not set

### Steps to Reproduce (Before Fix)

1. Configure a model with `capabilities.memory: true` and `function_calling: native`
2. Ensure admin memory setting is ON
3. Send "Remember my favorite color is blue"
4. Model calls `add_memory` but tool is not in execution dict → empty result → 30-iteration retry loop

### Steps to Verify (After Fix)

1. Same config as above
2. Send "Remember my favorite color is blue" → single `add_memory` call → DB row created
3. New chat → "What is my favorite color?" → `search_memories` → correct recall

Fixes #22666

> [!NOTE]


tjbck:
Addressed in dev.

PR #22694, opened 2026-03-15
fix: sanitize generated plugin ids from titles

Fu-Jie:
# Pull Request Checklist

- [x] **Description:** Sanitize ids derived from plugin titles on the frontend so titles can stay user-friendly while generated ids remain valid.
- [x] **Changelog:** Entry added below.
- [x] **Documentation:** No docs update required. This changes frontend sanitization behavior for existing create/import flows without introducing new user-facing settings or APIs.
- [x] **Testing:** Manually verified with the reproduction described below. Screenshot placeholders are included and will be replaced with the final images manually.
- [x] **Agentic AI Code:** This PR was prepared by an AI assistant but has been manually reviewed and tested by the user.
- [x] **Design & Architecture:** This keeps the existing UX and applies a smarter default when deriving ids from titles instead of adding new settings.
- [x] **Git Hygiene:** This PR only changes the frontend id derivation flow.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

When creating Functions or Tools, the frontend previously derived ids with a very simple `name.replace(/\s+/g, '_').toLowerCase()` rule. Titles like `xxx😄` kept unsupported characters in the generated id, so creation failed later with a misleading backend validation error.

This PR makes the frontend sanitize generated ids before submission while keeping the user-facing title unchanged.

### Changed

- Add a shared `toPluginId` helper for frontend-generated ids.
- Apply the same sanitization flow when creating Functions, creating Tools, and importing plugin content.
- Convert whitespace to underscores, strip emoji and unsupported punctuation, collapse repeated underscores, and prefix ids that would otherwise start with a digit.

### Fixed

- Creating a plugin with a title like `xxx😄` now succeeds because the generated id removes the emoji before submission.
- Other unsupported punctuation such as parentheses is also removed from generated ids instead of being sent to the backend unchanged.
- Frontend-generated ids now better match the actual validation rules instead of surfacing an inaccurate backend error during normal creation flows.

#### Manual verification

1. Reproduced the old behavior by creating a Tool with the title `xxx😄` and confirmed the previous flow surfaced the backend error.
2. Verified the updated frontend converts the same title into a valid id and allows creation to succeed.
3. Spot-checked additional punctuation and confirmed unsupported characters are removed from generated ids.
4. Verified the import flow also uses the same sanitization logic.
5. Ran `npm run build` successfully on this branch.

### Screenshots or Videos

Old behavior while creating a Tool with title `xxx😄`:

old_create
old_error

New behavior with the same title after frontend sanitization:

new_create
new_success

> [!NOTE]

### DIFF
--- a/src/lib/components/ImportModal.svelte
+++ b/src/lib/components/ImportModal.svelte
@@ -6,7 +6,7 @@
 	import Spinner from '$lib/components/common/Spinner.svelte';
 	import Modal from '$lib/components/common/Modal.svelte';
 	import XMark from '$lib/components/icons/XMark.svelte';
-	import { extractFrontmatter } from '$lib/utils';
+	import { extractFrontmatter, toPluginId } from '$lib/utils';
 
 	export let show = false;
 
@@ -42,13 +42,14 @@
 			toast.success(successMessage);
 
 			let func = res;
-			func.id = func.id || func.name.replace(/\s+/g, '_').toLowerCase();
 
 			const frontmatter = extractFrontmatter(res.content); // Ensure frontmatter is extracted
 
 			if (frontmatter?.title) {
 				func.name = frontmatter.title;
 			}
+
+			func.id = toPluginId(func.id || func.name);
 
 			func.meta = {
 				...(func.meta ?? {}),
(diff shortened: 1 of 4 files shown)


Fu-Jie:
Closing this PR in favor of #22695 so the final diff is represented as a single commit.

PR #22704, opened 2026-03-15
fix: resolve UnboundLocalError for limit in oracle23ai.py get()

mvanhorn:
## Summary

Moves the `limit = 1000` assignment before the `log.info()` call that references it in the `get()` method of `oracle23ai.py`.

## Changes

The `get()` method signature is `def get(self, collection_name: str)` with no `limit` parameter. The `log.info()` at line 714 referenced `limit` before it was assigned at line 718, causing an `UnboundLocalError` when Oracle 23ai hybrid search was activated.

## Testing

- Verified the function signature has no `limit` parameter
- Verified the log statement now executes after `limit` is assigned

Fixes #22664
Closes #22616

This contribution was developed with AI assistance (Claude Code).


Classic298:
https://github.com/open-webui/open-webui/commit/fcf720835285a4cea10fc1ebed0b454971463b20

PR #22751, opened 2026-03-17
fix: handle null user in shared chat view

aayushbaluni:
# Pull Request Checklist

- [x] **Testing:** Manually verified the fix resolves the shared chat crash.
- [x] **Code review:** Self-reviewed, minimal change.

# Changelog Entry

### Description

- Fixes #22742. When viewing a shared chat as an unauthenticated user, `UserMessage.svelte` crashes with `TypeError: can't access property "id", Z() is null` because `user` is null in the shared chat context.

### Fixed

- Added null checks for `user` in `UserMessage.svelte` to prevent crash when viewing shared chats as unauthenticated viewer
- ProfileImage block is now wrapped in `{#if user}` guard
- Added fallback display for message author when `user` is null (shared chat view)


Classic298:
https://github.com/open-webui/open-webui/commit/4d67c817ec47c0e5d7c8c87bb54fde569393ad06

(possible commit)

PR #22762, opened 2026-03-17
feat: use rich text editor for knowledge file editing

Classic298:
### Problem

There is an inconsistency in the Knowledge workspace: creating new text content via the "Add Text Content" modal offers a full WYSIWYG editing experience (TipTap-based RichTextInput with formatting toolbar, headings, bold, lists, code blocks, etc.), while editing existing files is limited to a plain, unstyled textarea. Since all knowledge file content is stored as text regardless of the original file type, there is no reason for the editing experience to differ from the creation experience.

### Changes

**KnowledgeBase.svelte**
- Replace the plain textarea in the file editor drawer with the same RichTextInput component used by the Add Text Content modal
- Normalize single newlines to paragraph breaks when loading file content, so extracted text from PDFs, DOCXs, and other files preserves its line structure in the rich text editor
- Set selectedFile after selectedFileContent to avoid a timing issue where the {#key} block would recreate the editor before the content was ready
- Retain the original aria-label for accessibility on the wrapper element

**RichTextInput.svelte (bug fix)**
- Fix a bug in onValueChange where the preserveBreaks code path bypassed markdown parsing entirely, passing raw markdown directly to setContent instead of parsed HTML. This caused all editor content to collapse into a single paragraph whenever a reactive value update triggered setContent. The fix ensures content is always parsed through marked regardless of the preserveBreaks setting.

### How it works

- File content is loaded from selectedFile.data.content
- Carriage returns are stripped and single newlines are normalized to double newlines (markdown paragraph breaks) — this is idempotent and does not compound on repeated save/reopen cycles
- The normalized markdown is parsed by RichTextInput's internal marked parser into proper HTML paragraphs
- Users can format content with headings, bold, italic, lists, code blocks, and tables via the floating toolbar
- On save, TipTap HTML is converted back to markdown via Turndown and stored via updateFileDataContentById

> [!NOTE]

### DIFF
--- a/src/lib/components/common/RichTextInput.svelte
+++ b/src/lib/components/common/RichTextInput.svelte
@@ -1217,11 +1217,9 @@
 			} else {
 				if (value !== mdValue) {
 					editor.commands.setContent(
-						preserveBreaks
-							? value
-							: marked.parse(value.replaceAll(`\n<br/>`, `<br/>`), {
-									breaks: false
-								})
+						marked.parse(value.replaceAll(`\n<br/>`, `<br/>`), {
+							breaks: false
+						})
 					);
 
 					selectTemplate();
--- a/src/lib/components/workspace/Knowledge/KnowledgeBase.svelte
+++ b/src/lib/components/workspace/Knowledge/KnowledgeBase.svelte
@@ -171,14 +171,18 @@
 
 	const fileSelectHandler = async (file) => {
 		try {
-			selectedFile = file;
-			const rawContent = selectedFile?.data?.content || '';
+			const rawContent = file?.data?.content || '';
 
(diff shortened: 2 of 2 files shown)


Classic298:
tested

Classic298:
@tjbck

siavashvj:
Unmerged? Why?

Classic298:
@siavashvj not enough advocates for it here (nobody voiced they wanted this), not enough testers but especially: rich text is a bit buggy in general and the rich text editor SHOULD get replaced with a simple text editor when CREATING the file to then match the behaviour of the simple text editor when editing the file

tjbck:
cd55c3e21237e000c13c6f396bb95b261f3bda82

PR #22776, opened 2026-03-17
fix: align file context injection by user-role messages for native FC

Classic298:
The add_file_context function used a positional zip() to pair API payload messages with DB-stored messages. After
process_messages_with_output() expands assistant messages containing tool calls into multiple OpenAI-format messages (assistant + tool results), the payload list becomes longer than the stored list. This caused the zip to misalign, so subsequent user messages never received their attached_files tags -- the model could see uploaded images via vision but had no file URL to pass to edit_image.

Fix: filter both lists to user-role messages only before zipping. User messages maintain the same order in both lists regardless of assistant message expansion, restoring correct file context injection.

Fixes #21878

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -1669,7 +1669,18 @@ def add_file_context(messages: list, chat_id: str, user) -> list:
             attrs += f' name="{file["name"]}"'
         return f"<file {attrs}/>"
 
-    for message, stored_message in zip(messages, stored_messages):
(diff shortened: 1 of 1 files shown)


Classic298:
tested here https://github.com/open-webui/open-webui/discussions/21878#discussioncomment-16184170

iChristGit:
Works as intended with this fix, tried uploading 5 images to chat and all 5 got converted!


image

Classic298:
@tjbck merge?

tjbck:
Thanks!

PR #22847, opened 2026-03-19
fix: make RedisDict.set() atomic with diff-based update

themavik:
# Changelog Entry

### Description

- Make RedisDict.set() atomic by using diff-based update instead of delete+recreate to avoid race conditions.

### Fixed

- Modified `RedisDict.set()` in `backend/open_webui/socket/utils.py` to use diff-based update instead of delete+recreate

Made with [Cursor](https://cursor.com)


themavik (commented):
Looked through the changes — the implementation is clean and follows the existing code patterns. Good contribution.

themavik (commented):
Reviewed the changes — the approach looks correct and follows the existing patterns in the codebase.

PR #22892, opened 2026-03-20
fix: use hybrid search and reranking in builtin query_knowledge_files tool

daudo:
# Pull Request Checklist

- [x] **Target branch:** Targets `dev`
- [x] **Description:** See below
- [x] **Changelog:** See below
- [ ] **Documentation:** No user-facing behavior change — hybrid search/reranking now applies to the same admin settings that already exist
- [x] **Dependencies:** None
- [x] **Testing:** Manually tested on a live deployment with native function calling enabled, TEI embeddings (bge-m3), TEI reranker (bge-reranker-v2-m3), and model-attached knowledge (10 epub files). Confirmed hybrid search and reranker are now invoked via container logs. Confirmed fallback to plain vector search works when hybrid search is disabled.
- [x] **Agentic AI Code:** AI-assisted but human-reviewed, manually tested on live deployment, and verified against the existing middleware implementation
- [x] **Code review:** Self-reviewed
- [x] **Design & Architecture:** No new settings — uses existing RAG config values
- [x] **Git Hygiene:** Single atomic commit, rebased on `dev`
- [x] **Title Prefix:** `fix:`

# Changelog Entry

### Description

The builtin `query_knowledge_files` tool (used when native function calling is enabled) calls `query_collection()` directly — plain vector search only. The middleware RAG pipeline (`get_sources_from_items`) uses `query_collection_with_hybrid_search()` with BM25 hybrid search, reranking, and relevance threshold filtering.

This means that when a model has attached knowledge and native function calling is enabled, the admin-configured RAG quality settings (hybrid search, reranking, relevance threshold, `TOP_K_RERANKER`) have no effect on knowledge retrieval.

This fix makes `query_knowledge_files` use the same hybrid search + reranking pipeline as the middleware path, controlled by the same existing config flags.

### Added

- N/A

### Changed

- `query_knowledge_files` in `builtin.py` now uses `query_collection_with_hybrid_search()` when `ENABLE_RAG_HYBRID_SEARCH` is enabled, with reranking function, `TOP_K_RERANKER`, `RELEVANCE_THRESHOLD`, `HYBRID_BM25_WEIGHT`, and enriched texts support

### Deprecated

- N/A

### Removed

- N/A

### Fixed

- Builtin `query_knowledge_files` tool now respects hybrid search and reranking settings, matching the behavior of the middleware RAG pipeline
- Graceful fallback to plain vector search if hybrid search fails or is disabled (identical to previous behavior)

### Security

- N/A

### Breaking Changes

[...]

PR #22895, opened 2026-03-20
fix: support TEI reranker format in ExternalReranker with auto-detection

daudo:
# Pull Request Checklist

- [x] **Target branch:** Targets `dev`
- [ ] **Documentation:** No user-facing config changes - TEI format is auto-detected transparently
- [x] **Dependencies:** None
- [x] **Testing:** Manually tested on a live deployment with TEI reranker (bge-reranker-v2-m3 via `ghcr.io/huggingface/text-embeddings-inference:120-1.9.1`). Confirmed format auto-detection, reranking scores returned correctly, thread-safe under 10 parallel rerank calls. Also verified Cohere/Jina format path is unchanged (probe succeeds on first try, cached).
- [x] **Agentic AI Code:** AI-assisted but human-reviewed, manually tested on live deployment
- [x] **Design & Architecture:** No new settings - format is auto-detected via smart default
- [x] **Git Hygiene:** Single atomic commit, rebased on `dev`

# Changelog Entry

### Description

`ExternalReranker` only supports the Cohere/Jina rerank API format (`"documents"` key, `{"results": [{"relevance_score": ...}]}` response). HuggingFace Text Embeddings Inference (TEI) uses a different format (`"texts"` key, flat array response with `"score"` key), causing a silent 422 error that falls back to unreranked results. Users with TEI rerankers configured get no reranking despite the setting being enabled.

~~Additionally, `ChromaClient.search()` silently swallows all exceptions (`except: return None`), making it impossible to diagnose vector search failures.~~

### Added

- Auto-detection of reranker API format (Cohere/Jina vs TEI) on first call, cached for subsequent requests
- Thread-safe format detection via lock (hybrid search dispatches parallel rerank calls across collections)
- Support for TEI response format (flat array with `"score"` key)

### Changed

- `ExternalReranker.predict()` now auto-detects and caches the API format instead of assuming Cohere/Jina
- Response parsing extracted to `_parse_response()` static method, handles both Cohere/Jina and TEI formats
- Payload construction extracted to `_build_payload()` method

### Deprecated

- N/A

### Removed

- N/A

### Fixed

- External reranker now works with TEI endpoints out of the box (no config change needed)
~~- `ChromaClient.search()` now logs exceptions instead of silently returning `None`~~

### Security

- N/A

### Breaking Changes

- N/A

---

### Additional Information

- TEI's `/rerank` endpoint expects `{"query": "...", "texts": [...], "truncate": true}` and returns `[{"index": 0, "score": 0.5}, ...]`. Open WebUI sends `{"model": "...", "query": "...", "documents": [...], "top_n": N}` and expects `{"results": [{"index": 0, "relevance_score": 0.5}]}`. The 422 from TEI was caught by the generic exception handler in `predict()`, silently falling back to unreranked results.
- Format detection sends one Cohere-format probe on first call. If the endpoint returns 422, TEI format is used. The result is cached as a class-level dict keyed by URL, surviving `ExternalReranker` re-instantiation (which happens on config updates via the admin UI). A `threading.Lock` prevents concurrent probe requests when hybrid search dispatches parallel rerank calls.
~~- The ChromaDB logging change is included because the silent `except: return None` in `search()` made it very difficult to diagnose a vector search failure encountered during testing of the reranker integration.~~

This is related to my other PR https://github.com/open-webui/open-webui/pull/22892, which enables reranking for query_knowledge_files, which in turn is my use case.

### DIFF
--- a/backend/open_webui/retrieval/models/external.py
+++ b/backend/open_webui/retrieval/models/external.py
@@ -1,5 +1,6 @@
 import logging
 import requests
+import threading
 from typing import Optional, List, Tuple
 from urllib.parse import quote
 
@@ -12,6 +13,15 @@ log = logging.getLogger(__name__)
 
 
 class ExternalReranker(BaseReranker):
+    # Supported API formats for external reranker endpoints
+    FORMAT_COHERE = 'cohere'  # Cohere/Jina: {"documents": [...], "model": ..., "top_n": N}
+    FORMAT_TEI = 'tei'  # TEI: {"texts": [...], "truncate": true}
+
+    # Class-level cache: survives re-instantiation (e.g. config updates).
+    # Keyed by URL so different endpoints can use different formats.
+    _format_cache: dict = {}
+    _format_lock = threading.Lock()
+
     def __init__(
         self,
         api_key: str,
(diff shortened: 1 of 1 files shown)


daudo:
alright, even though i regard that additional line of logging to be quite helpful and very much in the context of the PR, in order to move forward, I removed the additional logging for the ChromaDB search failure path. So now it should be a very atomic PR.

tjbck:
External reranker follows standard format. Other should be added as a separate engine.

PR #22908, opened 2026-03-20
enh: allow iframe postMessage prompts without same-origin (with HITL confirmation)

Classic298:
# allow iframe postMessage prompts without same-origin (with HITL confirmation)

## Problem

The onMessageHandler in Chat.svelte checks event.origin against window.origin and silently drops messages from sandboxed iframes that don't have same-origin access. This means Rich UI tools that use the input:prompt:submit postMessage protocol (e.g. interactive buttons that send prompts back to the chat) only work when "iframe Sandbox Allow Same Origin" is enabled.

Requiring same-origin for this feature is a security tradeoff: it also exposes parent.document, parent.fetch, and parent.localStorage to iframe JavaScript — which is far more access than sendPrompt needs.

## Solution

When same-origin is enabled, prompt messages go through immediately (unchanged behavior).

When same-origin is disabled, cross-origin input:prompt:submit messages now trigger the existing EventConfirmDialog instead of being silently dropped. The user sees the message text and can confirm or cancel before it is submitted. This prevents abuse (spam) while still enabling the feature without requiring same-origin access.

Other cross-origin message types (input:prompt, action:submit) are also allowed through to the handler — input:prompt only fills the input box without submitting, and action:submit uses the existing prompt text.

## Changes

- Chat.svelte: Restructured the origin check in onMessageHandler to allow prompt-related message types from opaque-origin iframes
- For input:prompt:submit from cross-origin: shows a confirmation dialog before submitting
- For input:prompt:submit from same-origin: submits immediately (no behavior change)

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -569,10 +569,10 @@
 		const isSameOrigin = event.origin === window.origin;
 		const type = event.data?.type;
 
-		// These message types only submit text to the chat input — functionally
-		// equivalent to the user typing.  They are safe to accept from sandboxed
-		// iframes whose origin is opaque ("null") because they cannot read or
-		// modify any page state.
+		// Prompt-related message types only submit text to the chat input —
+		// functionally equivalent to the user typing.  When same-origin is
+		// enabled they go through immediately.  When it is disabled (opaque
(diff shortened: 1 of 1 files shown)


Classic298:
@tjbck

PR #22921, opened 2026-03-21
fix: prepare-pyodide lock file handling and add openpyxl

EliotGodard:
# Pull Request Checklist

- [x] **Changelog:** Provided below.
- [x] **Testing:** Manually tested `npm run pyodide:fetch` and verified openpyxl imports work via both code blocks and native tool calling.
- [x] **Git Hygiene:** Two atomic commits, rebased on `dev`.

# Changelog Entry

### Description

The `prepare-pyodide.js` script was only caching Pyodide's built-in packages (numpy, pandas, scipy, etc.) to `static/pyodide/`. Packages that are not part of Pyodide's distribution — even ones already listed in the script like seaborn or black — were never cached locally. Their wheels were fetched by the browser directly from PyPI at runtime every time.

This is because `packageCacheDir` only applies to packages loaded via `pyodide.loadPackage()`, not to packages installed via `micropip.install()` (except if they are part of the Pyodide's built-in packages list). micropip downloads wheels into pyodide's in-memory virtual filesystem without writing them to disk.

To fix this, the script now:

1. **Prevents `copyPyodide()` from overwriting `pyodide-lock.json`** — the generated lock file from `micropip.freeze()` was being overwritten by the stock one from `node_modules/pyodide/`. This in turn made it necessary to **normalize `file_name` paths** in the lock file: `micropip.freeze()` produces absolute local paths and remote PyPI URLs, but the browser resolves wheels relative to `indexURL` and expects bare filenames. This wasn't an issue before because the stock lock file from `node_modules/` already had bare filenames.

2. **Downloads PyPI-only wheels locally** — wheels for packages not in Pyodide's distribution are now fetched during the build and saved to `static/pyodide/`, so they are served locally instead of from PyPI at runtime.

3. **Fixes version comparison** — the script compared the semver range from `package.json` (e.g. `^0.28.2` → `0.28.2` after stripping `^`) against the actual installed version (`0.28.3`). Since the current latest available version is `0.28.3`, this mismatch caused the script to nuke the `static/pyodide/` cache directory on every run. The comparison now uses the actual installed version from `node_modules/pyodide/package.json`.

4. **Adds openpyxl** as an example of a package that can now be properly pre-bundled with this refactor, since it's not part of Pyodide's built-in distribution. openpyxl is useful for manipulating Excel files in the code interpreter.

### Added

- openpyxl to the default pyodide package list and import detection in `CodeBlock.svelte` and `+layout.svelte`
- Build step to download PyPI-only wheels into `static/pyodide/`

### Fixed

- Version comparison uses actual installed version from `node_modules/pyodide/package.json`
- Lock file `file_name` entries normalized to bare filenames
- `copyPyodide()` skips `pyodide-lock.json` to prevent overwriting the generated lock file

### Screenshots or Videos

N/A — build script and runtime package loading fix.


tjbck:
openpyxl added to dev.

EliotGodard:
I might be wrong but I believe it was not implemented correctly : https://github.com/open-webui/open-webui/commit/16ee4ac7a2a3310bba5f51f26d35bd555f1f8f1d

PR #23013, opened 2026-03-25
fix: rewire orphan tool messages to preserve conversation context

Classic298:
## Summary
- When chat history reconstruction produces `tool_result` messages without a matching `tool_use` (e.g. partial history, retries, edited chains), strict providers like Anthropic/Vertex reject the payload with HTTP 400
- Instead of dropping orphan tool messages (losing context), we now **rewire** them by injecting a synthetic `tool_call` into the preceding assistant message so the sequence stays valid
- Applies at both the output→message conversion layer (`convert_output_to_messages`) and the final message sanitization pass (`rewire_orphan_tool_messages`)

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1770,7 +1770,7 @@ async def chat_completion(
         form_data['metadata'] = metadata
 
     except Exception as e:
-        log.debug(f'Error processing chat metadata: {e}')
+        log.error(f'Error processing chat metadata: {e}')
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
             detail=str(e),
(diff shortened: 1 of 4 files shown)


Classic298:
Created with the help of GPT-5.3-codex

Some intermittent failures sometimes cause full blown errors on the provider side, thus breaking the chat for our users

The error is very clear: Open WebUI is sending tool results for assistant messages that dont have the right tool call ID for the tool call

We are as of yet unsure how to reproduce the bug or WHY it happens in the first place, but this protective measure here should at least fix the symptom. (I would love to fix the root cause but its impossible to accurately reproduce)

Perhaps users modifying assistant messages?

PR #23029, opened 2026-03-25
fix: tool calls fail for non-admin users due to model access check

yang1002378395-cmyk:
Fixes #22851

# Pull Request Checklist

- [x] **Target branch:** Targets the dev branch
- [x] **Description:** Fix for non-admin users tool call failures
- [x] **Changelog:** Added below
- [x] **Testing:** Manual code review and analysis performed

# Changelog Entry

### Description

Fixes a bug where non-admin users tool calls would fail during subsequent response generation due to model access permission checks.

### Fixed

- Tool calls for non-admin users now complete successfully by bypassing model access checks for internal middleware-generated requests

---

## Problem

Non-admin users experience tool call failures:
- Tool calls execute successfully but subsequent generation fails
- Log shows task=None -> title_generation (missing follow-up call)
- Admin users work correctly: task=None -> task=None -> title_generation

## Root Cause

generate_chat_completion() calls check_model_access() to verify model permissions. When non-admin users use tools:
1. Tool decision generation succeeds (uses task model)
2. Response generation fails because it calls generate_chat_completion() without bypass_filter=True
3. If user does not have explicit access to response model, permission check fails and generation is interrupted

## Solution

Add bypass_filter=True to all generate_chat_completion() calls in middleware.py that are for tool decision/response generation.

## Changes

- Line 1245: Tool decision generation - add bypass_filter=True
- Line 4553: Response generation (tool_calls) - add bypass_filter=True
- Line 4738: Response generation (native_function_calling) - add bypass_filter=True

## Testing

Analyzed the code flow through middleware.py and chat.py to identify the permission check issue. The fix allows internal system calls to bypass model access checks while maintaining security for direct user requests.


yang1002378395-cmyk:
Closing to recreate with clean history

PR #23150, opened 2026-03-27
enh: Set the default executor for asyncio based on THREAD_POOL_SIZE

EntropyYue:
# Changelog Entry

### Description

- This repository heavily utilizes asyncio.to_thread and only has a small amount of anyio.to_thread. Initially, THREAD_POOL_SIZE only increased the thread pool size of anyio. This pull request enables the thread pool executor of asyncio to be configured by this environment variable

### Added

- Set the default executor for asyncio based on THREAD_POOL_SIZE

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -9,6 +9,7 @@ import sys
 import time
 import random
 import re
+from concurrent.futures import ThreadPoolExecutor
 from uuid import uuid4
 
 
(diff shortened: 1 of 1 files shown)


harshitgavita-07:
Hey @EntropyYue , 

The move to a configurable thread pool executor makes sense — the default 
ThreadPoolExecutor sizing in asyncio (min(32, cpu_count + 4)) is genuinely 
too conservative for IO-heavy workloads like proxying Ollama streams.

One thing worth thinking about: if THREAD_POOL_SIZE is set high and multiple 
users are streaming simultaneously, you can end up with thread contention on 
The response flush loop rather than the pool size itself is the bottleneck. 
We saw this in a local AI runtime project (https://github.com/harshitgavita-07/Aios) 
When trying to keep streaming latency flat under concurrent sessions, the 
thread count helped, but pairing it with a bounded asyncio.Queue between the 
The Ollama reader and the SSE writer made a bigger difference in throughput.

Is the THREAD_POOL_SIZE env var intended to be per-worker (uvicorn worker 
process) or global across the whole server process? That distinction matters 
a lot if people are running multi-worker deployments behind nginx.

Classic298:
This conflicts with dev and has had no activity since 29 March 2026. Closing it as stale. The change is small, so if you still want it, open a discussion for it please

Issue #23197, opened 2026-03-29
issue: `cookie_expires` NameError in `OAuthManager.handle_callback`


soonlai814:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

v0.8.12

### Operating System

macOS / Linux (affects all platforms — Python runtime error)

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

After a successful OAuth login, the server sets an `oauth_session_id` cookie containing the server-side session ID, allowing subsequent requests to look up and refresh OAuth tokens for MCP tool access.

### Actual Behavior

A `NameError` is raised inside `OAuthManager.handle_callback` every time a user logs in via OAuth (when `JWT_EXPIRES_IN` is configured, which is the default). The variable `cookie_expires` is referenced but **never defined** in the method.

The exception is silently swallowed by the surrounding `try/except` block, logged as `'Failed to store OAuth session server-side'`, and the `oauth_session_id` cookie is never set. The user is redirected as if login succeeded, making this failure invisible.

### Steps to Reproduce

1. Start with a clean install of Open WebUI (git clone, latest `main`).
2. Configure any OAuth provider (e.g. Google, GitHub) via `OAUTH_PROVIDERS`.
[...]

PR #23207, opened 2026-03-29
fix: remove undefined cookie_expires in OAuth session cookie

extrasmall0:
# Pull Request Checklist

- [x] **Testing:** Verified the NameError no longer occurs when `cookie_max_age` is set

# Changelog Entry

### Description

The `set_cookie()` call in `handle_callback` references `cookie_expires`, which is never defined in the current scope. When `cookie_max_age` is not `None`, this causes a `NameError` at runtime. Since `max_age` alone is sufficient for cookie expiration (browsers derive `expires` from it per RFC 6265), we can just drop the `expires` kwarg.

### Fixed

- `NameError: name 'cookie_expires' is not defined` when setting OAuth session cookies with a configured max age

---

### Additional Information

- Relates to #23197
- Only the `oauth_session_id` cookie was affected; the auth token cookie above it already uses `max_age` alone

### DIFF
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -1683,7 +1683,7 @@ class OAuthManager:
                     httponly=True,
                     samesite=WEBUI_AUTH_COOKIE_SAME_SITE,
                     secure=WEBUI_AUTH_COOKIE_SECURE,
-                    **({'max_age': cookie_max_age, 'expires': cookie_expires} if cookie_max_age is not None else {}),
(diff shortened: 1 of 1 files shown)


ethantiller:
Adding a comment to get this boosted as this is a blocker for my use-case!

etatros:
Same here, would very much appreciate getting this important fix merged.

Classic298:
https://github.com/open-webui/open-webui/commit/18f6ec68b9fdbb0ad702a45383a741d38ff922ee

PR #23230, opened 2026-03-30
fix(oauth): remove undefined cookie_expires reference in OAuth session cookie

lorenzophys:
### Description

In `OAuthManager.handle_callback`, the `oauth_session_id` cookie was never set on the user's browser due to a reference to an undefined variable `cookie_expires`. This raised an exception at runtime, which was silently caught by the except block. The OAuth session was created in the database, but the client never received the cookie linking to it.

I found this bug by chance when I saw this error in the application log every time a user was logging in via oauth

```
{"ts": "2026-03-30T11:18:19.365Z", "level": "error", "msg": "Failed to store OAuth session server-side: name 'cookie_expires' is not defined", "caller": "open_webui.utils.oauth:handle_callback:1735"}
```

I removed the undefined reference and now the session is stored correctly.

```
{"ts": "2026-03-30T11:34:58.762Z", "level": "info", "msg": "Stored OAuth session server-side for user xxxxxxxx, provider oidc", "caller": "open_webui.utils.oauth:handle_callback:1731"}
```

- Server-side OAuth token retrieval now works. Features that call `get_oauth_token`, eg in [functions.py](https://github.com/open-webui/open-webui/blob/dev/backend/open_webui/functions.py#L237) can now actually look up the stored session and return a valid access token. Before the fix, the `oauth_session_id` cookie was always missing, so the lookup was silently skipped and [__oauth_token__](https://github.com/open-webui/open-webui/blob/dev/backend/open_webui/functions.py#L255) was always `None`.
- Token refresh via the server-side session path now works. Previously, sessions existed in the database but were unreachable causing 

```
2026-03-27 19:44:27.834 | WARNING | open_webui.utils.oauth:get_oauth_token:980 - No OAuth session found for user xxx, session xxx
```

(this is from an instance running openwebui v0.8.7, so the line number is different)

### Fixed

- Removed the undefined `cookie_expires` from the `oauth_session_id` cookie that now only have a `max_age`. This doesn't change the overall behavior because ["If a cookie has both the Max-Age and the Expires attribute, the Max-Age attribute has precedence and controls the expiration date of the cookie."](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.2.2).

### Breaking Changes

- None

---

> [!NOTE]


lorenzophys:
Sorry, haven't noticed somebody already submitted a pr for this a few hours ago: https://github.com/open-webui/open-webui/pull/23207

Issue #23232, opened 2026-03-30
issue: Opening Artifacts preview in one tab auto-expands Advanced Settings on Home in another tab


ShirasawaSama:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

v0.8.12

### Operating System

Mac

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

UI state should be isolated per browser tab/session view.  
Opening an Artifacts preview in one tab should **not** automatically expand the Advanced Settings/Controls panel in another tab that is on Home.

### Actual Behavior

If I open an Artifacts preview in one browser tab, then open another tab and go to Home, the Home tab automatically expands the Advanced Settings/Controls panel without user action.

### Steps to Reproduce

1. Open Open WebUI in browser tab A.
2. In tab A, open a chat and trigger/open an Artifacts preview.
3. Open browser tab B (same Open WebUI instance/session).
4. In tab B, navigate to Home.
5. Observe that Advanced Settings/Controls in tab B is auto-expanded.

### Logs & Screenshots

<img width="732" height="1582" alt="Image" src="https://github.com/user-attachments/assets/8b5a1c9d-9433-4d25-b4e8-6f5f3287d9bc" />

### Additional Information

[...]

PR #23261, opened 2026-03-31
fix: use sessionStorage for showControls to prevent cross-tab state leakage

karthik-idikuda:
# Pull Request Checklist

- [x] **Testing:** Manually verified the fix works as intended.
- [x] **Code review:** Self-reviewed the changes.

# Changelog Entry

### Description

Opening an Artifacts preview in one browser tab automatically expands the Advanced Settings/Controls panel on the Home page in another tab, without user action. This is caused by `showControls` being persisted via `localStorage` which is shared across all tabs of the same origin.

**Root Cause:**
In `src/routes/(app)/+layout.svelte`, the `showControls` store state is persisted via `localStorage`:
```js
await showControls.set(!$mobile ? localStorage.showControls === 'true' : false);
showControls.subscribe((value) => {
    localStorage.showControls = value ? 'true' : 'false';
});
```

Since `localStorage` is shared across all tabs:
1. **Tab A**: User opens Artifacts preview -> `showControls.set(true)` -> writes `localStorage.showControls = 'true'`
2. **Tab B**: User navigates to Home -> `onMount` reads `localStorage.showControls === 'true'` -> controls panel auto-expands

**Fix:** Switch from `localStorage` to `sessionStorage` for persisting `showControls`. `sessionStorage` is isolated per browser tab/session, so each tab maintains its own independent controls panel state. This preserves within-tab persistence (controls stay open across SPA navigation within the same tab) while eliminating cross-tab leakage.

Note: `chatControlsSize` (the panel width) intentionally remains in `localStorage` so the user's preferred panel width is preserved globally.

### Fixed

- Fixed cross-tab state leakage where opening Artifacts preview in one tab would auto-expand Advanced Settings/Controls panel in another tab (#23232)

### Additional Information

- Only `showControls` is affected. Other `localStorage` items like `chatControlsSize` and `selectedTerminalId` correctly use `localStorage` since they represent user preferences that should persist globally.

### DIFF
--- a/src/routes/(app)/+layout.svelte
+++ b/src/routes/(app)/+layout.svelte
@@ -346,9 +346,12 @@
 		}
 		// Persist showControls: track open/close state separately from saved size
 		// chatControlsSize always retains the last width for openPane()
-		await showControls.set(!$mobile ? localStorage.showControls === 'true' : false);
+		// Use sessionStorage (per-tab) instead of localStorage to prevent
+		// cross-tab state leakage (e.g. opening Artifacts in one tab
+		// auto-expanding Advanced Settings in another tab). See #23232.
+		await showControls.set(!$mobile ? sessionStorage.getItem('showControls') === 'true' : false);
 		showControls.subscribe((value) => {
-			localStorage.showControls = value ? 'true' : 'false';
+			sessionStorage.setItem('showControls', value ? 'true' : 'false');
 		});
 
 		// Persist selectedTerminalId across page loads


Classic298:
@pr-validator-bot full review

ShirasawaSama:
Personally, I don’t agree with this approach. I think we should distinguish between chat controls that are opened when a user clicks on them and those that open automatically when chat-related content is detected.

PR #23267, opened 2026-03-31
Thinking Modes in Chat UI

jrdnmartin:
This PR adds a native-aligned Thinking Level control to chat so users can choose low, medium, or high effort directly from the UI (defaulting to medium), instead of relying on manual prompt commands. The selected level is carried through the existing params flow and consistently enforced server-side, including normalization and routing behavior that follows OpenWebUI’s reasoning patterns.

It also updates effort-to-budget behavior to the new targets (low 1024, medium 2048, high 4096), improves non-reasoning model handling by applying deterministic limits plus an explicit reasoning-level hint in context, and refines Ollama-compatible payload mapping to use level-based signaling. The associated test harness was updated to validate the new thresholds across randomized multi-run checks.


jrdnmartin:
was meant for a fork, not upstream 🙃

Classic298:
@jrdnmartin why not use a filter function with toggle to toggle thinking ?

PS: rebranding is not permitted under certain conditions! I hope you are aware

jrdnmartin:
yes, i am aware. this is a very small closed-source project not for public use, solely for academic research. apologies again for pushing to the main repo! i always try to double check but must've missed it this time.

Issue #23134, opened 2026-03-27
issue: [Bug] Database migration chat → chat_message OOM-kills on large datasets (PostgreSQL/AlloyDB, ~75 GB)


adesso-pia-vonkolken:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.8.5

### Operating System

Google Cloud Run Service

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

Environment:
- Open WebUI version: Upgrading from v0.7.2 → v0.8.5 (also tested with cherry-picked commit b4f3408 from PR #21542)
- Deployment: Google Cloud Run (Cloud Run Job for migration, 34 GB memory)
- Database: AlloyDB for PostgreSQL (managed PostgreSQL-compatible)
- Dataset size: ~75 GB of chat data

Upgrading from v0.7.2 to v0.8.5 (or later) should run the Alembic database migration — including the chat → chat_message table migration — successfully without crashing, regardless of dataset size.

### Actual Behavior

[...]

PR #21542, opened 2026-02-17
fix: migration streaming/batching

Classic298:
- Migration: replace .fetchall() with yield_per streaming, replace per-message INSERT+SAVEPOINT with batched inserts (5k/batch) with fallback to row-by-row on error, add progress logging

- Write path: call normalize_usage() in upsert_message() before saving to ensure input_tokens/output_tokens always present

- Read path: analytics queries now COALESCE across input_tokens/prompt_tokens and output_tokens/completion_tokens so historical data with OpenAI-format keys is visible

https://github.com/open-webui/open-webui/issues/21347

https://github.com/open-webui/open-webui/discussions/21539

### Contributor License Agreement

By submitting this pull request, I confirm that I have read and fully agree to the [Contributor License Agreement (CLA)](https://github.com/open-webui/open-webui/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT), and I am providing my contributions under its terms.

> [!NOTE]
> Deleting the CLA section will lead to immediate closure of your PR and it will not be merged in.

PR #23382, opened 2026-04-03
fix: improve PostgreSQL database sided reliability for migrations

### LINKED ISSUE #23273: bug: Migration 8452d01d26d7 fails on large PostgreSQL datasets due to single-transaction backfill with FK/index maintenance
# [Bug] Migration `8452d01d26d7` fails on large PostgreSQL datasets due to single-transaction backfill with FK/index maintenance

## Check Existing Issues

## Installation Method

Docker (Cloud Run)

## Open WebUI Version

v0.8.12

## Operating System

Google Cloud Run + AlloyDB for PostgreSQL (managed, PostgreSQL 15.15)

## Expected Behavior

Upgrading from v0.7.2 to v0.8.12 should complete the Alembic migration `8452d01d26d7` ("Add chat_message table") successfully, regardless of dataset size. The migration uses `yield_per=1000`, `stream_results=True`, and `BATCH_SIZE=5000` — which correctly prevents Python-side OOM — but the underlying database transaction is still unbounded.

## Actual Behavior

The migration runs for ~55 minutes and is then terminated by AlloyDB with `OperationalError: server closed the connection unexpectedly`. The database remains on alembic revision `c440947495f3` (pre-migration), with `chat_message` table partially created but the backfill incomplete and rolled back.

During the entire migration window, the application is unusable — all backend requests hang, likely because the service also attempts to run migrations on startup and blocks on the open transaction.

## Environment

| Component | Detail |
|-----------|--------|
| Open WebUI | v0.8.12 |
| Database | AlloyDB for PostgreSQL 15.15, `europe-west3` |
| Deployment | Cloud Run Job (dedicated migration job, 34 GB memory) |
| Chat table | 161,011 rows, **68 GB** (TOAST storage) |
| Chat column type | `json` |

## Root Cause Analysis

PR #21542 (commit b4f3408) fixed the original `fetchall()` pathology and reduced Python heap pressure, but it did **not** make the migration operationally safe for large PostgreSQL datasets. There are three interacting problems:

### 1. Single unbounded transaction

The upstream `env.py` wraps all migrations in a single transaction:

```python
# env.py (upstream)
with connectable.connect() as connection:
    context.configure(connection=connection, target_metadata=target_metadata)
    with context.begin_transaction():
        context.run_migrations()
```

The `_flush_batch()` function uses `conn.begin_nested()` (savepoints), but these are nested within this outer transaction — no data is actually committed to disk until the migration function returns. The migration inserts millions of rows with 7 indexes and an FK constraint in a single transaction, generating enormous WAL volume. While PostgreSQL checkpoints can still advance and recycle old WAL segments during the transaction (the redo pointer and `KeepLogSeg` are not constrained by open user transactions), new WAL is produced far faster than recycling can reclaim space. For our dataset (68 GB, 161k chats), the migration ran for 55 minutes at sustained high write throughput until AlloyDB terminated the connection — likely due to managed-service limits on connection duration, write throughput, or storage pressure.

### 2. Foreign key and indexes created before the backfill

The migration creates the `chat_message` table with a `FOREIGN KEY` constraint on `chat.id`, 4 single-column indexes, and 3 composite indexes — all **before** inserting any data. This means every insert during the backfill must:
- Validate the FK constraint via an SPI query that does a B-tree lookup against the referenced `chat.id` index for every inserted row, adding I/O overhead per row
- Update all 7 secondary indexes — each insert produces roughly 1 heap WAL record + 7 btree WAL records (more on page splits), causing significant write amplification

In contrast, a post-load `CREATE INDEX` uses sorted sequential bulk construction with batched WAL records ([`btbuild`](https://github.com/postgres/postgres/blob/master/src/backend/access/nbtree/nbtsort.c)), which is orders of magnitude more efficient.

PostgreSQL's [bulk loading documentation](https://www.postgresql.org/docs/current/populate.html) explicitly warns against this pattern: create indexes and add FK constraints **after** the data load, not before.

### 3. Server-side cursors invalidated by COMMIT

The migration uses `yield_per=1000` with `stream_results=True`, which on PostgreSQL creates a server-side cursor. Psycopg's server-side cursors are [`WITHOUT HOLD` by default](https://www.psycopg.org/docs/usage.html#server-side-cursors), meaning they **stop being fetchable after COMMIT**. This makes a naive "just add commits inside the loop" patch fundamentally incorrect — the streaming cursor becomes invalid after the first commit.

A correct fix needs either keyset pagination (a fresh `SELECT` per chunk, ordered by primary key with a `WHERE id > last_seen_id`) or separate read/write connections with holdable cursor semantics.

### Why this went undetected

This issue likely went undetected upstream because SQLite is the default backend and does not exhibit the same WAL, FK-trigger, or server-side cursor characteristics as PostgreSQL. Testing with small datasets (a few thousand chats) would also not surface the problem.

## Steps to Reproduce

1. Run Open WebUI v0.7.2 with a PostgreSQL database
2. Seed the database with ≥50 GB of chat data (see below)
3. Upgrade to v0.8.12 and let the Alembic migration run
4. Migration `8452d01d26d7` starts, streams chat rows, and inserts into `chat_message`
5. WAL volume grows continuously for the entire duration due to per-row index and heap writes
6. After ~55 minutes, the managed database terminates the connection (connection duration, write throughput, or storage limits)
7. Transaction rolls back, `chat_message` is empty, alembic version unchanged

### Generating test data

To reproduce at scale, you need to seed the `chat` table with large JSON payloads. An important caveat: **PostgreSQL TOAST compression is extremely effective on repetitive text** (e.g. Lorem ipsum). Using Lorem ipsum as padding results in ~1 GB on disk instead of the intended 68 GB. Use `os.urandom().hex()` or similar random data for the message content to avoid compression and achieve realistic on-disk sizes.

A simple seeding approach using direct SQL inserts with a size distribution matching a real production dataset:

| Size class | Count | Avg size | Total |
|------------|-------|----------|-------|
| < 10 KB | ~84,000 | 4 KB | ~337 MB |
| 10–100 KB | ~50,000 | 27 KB | ~1.3 GB |
| 100 KB – 1 MB | ~14,000 | 318 KB | ~4.5 GB |
| 1–10 MB | ~7,600 | 4.9 MB | ~36 GB |
| 10–30 MB | ~1,500 | 16 MB | ~23 GB |
| > 30 MB | ~130 | 31 MB | ~3.9 GB |

## Logs

```
INFO [alembic.runtime.migration] Running upgrade 374d2f66af06 -> 8452d01d26d7, Add chat_message table
# ... ~55 minutes of silence ...
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) server closed the connection unexpectedly
	This probably means the server terminated abnormally before or while processing the request.
```

## Proposed Fix

The migration needs to be restructured to follow [PostgreSQL's bulk loading guidance](https://www.postgresql.org/docs/current/populate.html). The core changes:

1. **Create only the primary key during table creation** — defer secondary indexes and FK constraint until after the backfill
2. **Use keyset pagination instead of a streaming cursor** — server-side cursors are `WITHOUT HOLD` and become invalid after `COMMIT`, so paginate explicitly with `WHERE id > :last_id ORDER BY id LIMIT :chunk_size`
3. **Commit each chunk** — keep WAL bounded by committing every N chats
4. **Use idempotent inserts** (`ON CONFLICT DO NOTHING`) — so the migration can be safely restarted after a mid-run failure
5. **Create indexes after the backfill** — avoids per-row index maintenance during the load (use `CREATE INDEX CONCURRENTLY` if the service must remain online, though note this [cannot run inside a transaction block](https://www.postgresql.org/docs/current/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY))
6. **Add FK as `NOT VALID`, then `VALIDATE CONSTRAINT`** — [`ALTER TABLE ADD CONSTRAINT ... NOT VALID`](https://www.postgresql.org/docs/current/sql-altertable.html) skips validation of existing rows during creation; a subsequent `VALIDATE CONSTRAINT` checks them without holding an `ACCESS EXCLUSIVE` lock

### Approach 1: Rewrite the migration (recommended upstream fix)

If the backfill remains inside Alembic, the migration should use `op.get_context().autocommit_block()` or a separate connection strategy to commit chunks independently. The backfill loop must not rely on one long-lived streaming cursor if chunk commits are introduced (see root cause #3 above).

High-level structure:

1. **DDL phase:** Create `chat_message` with PK only (no indexes, no FK)
2. **Backfill phase:** Paginate source chats using keyset pagination (`WHERE id > :last_id ORDER BY id LIMIT :chunk_size`), parse messages, insert with `ON CONFLICT (id) DO NOTHING`, commit after each chunk
3. **Index phase:** Create all 7 secondary indexes (single bulk build per index is orders of magnitude more efficient than per-row maintenance)
4. **FK phase:** `ALTER TABLE chat_message ADD CONSTRAINT ... FOREIGN KEY ... NOT VALID`, then `VALIDATE CONSTRAINT` separately

### Approach 2: Separate backfill from schema migration

Use Alembic only for schema creation (table + PK). Run the backfill as a dedicated resumable job outside of Alembic, then stamp the revision manually. This is the safest option for very large datasets and follows PostgreSQL's documented bulk-load guidance most closely.

### Supporting change: `transaction_per_migration=True` in `env.py`

Adding `transaction_per_migration=True` to `context.configure()` in `env.py` is recommended by Alembic when migrations use autocommit blocks. This is supporting infrastructure — it does not fix the migration itself, but prevents other migrations in the upgrade chain from being affected.

## Additional Context

- **This is not specific to AlloyDB.** The migration shape is objectively poor for large PostgreSQL backfills. It affects any PostgreSQL deployment with large chat tables. Managed services (AlloyDB, Cloud SQL, RDS, Supabase, Neon) enforce limits on connection duration, write throughput, or storage pressure. Self-hosted PostgreSQL would also struggle — while checkpoints can still recycle old WAL segments during the transaction, the sheer volume of WAL generated (per-row heap + 7 index WAL records) can exhaust disk space or exceed `max_wal_size` thresholds faster than recycling can keep up.
- The problem scales with data size. Our 68 GB dataset ran for 55 minutes before AlloyDB terminated the connection. The exact WAL volume generated by the migration was not isolated. To quantify WAL for this migration precisely, capture `pg_stat_wal.wal_bytes`, `wal_records`, and `wal_fpi` before and after a test run, or use `pg_stat_get_backend_wal()` for the migration backend where available. `pg_stat_wal` is cumulative since the last stats reset, so a before/after delta is required.
- The `chat` column uses `json` (not `jsonb`), as created by migration `242a2047eae0`. This is relevant for any server-side SQL approach: `jsonb_each()` cannot operate on `json` directly — either `json_each()` or an explicit cast to `jsonb` is required, and the cast adds CPU overhead at scale.
- Setting `BATCH_SIZE` higher or lower does not help — the issue is the combination of the single transaction boundary, FK constraint validation, and index maintenance during the load.
- Temporarily increasing `maintenance_work_mem` and `max_wal_size` (where the provider allows it) may reduce pain for smaller datasets but does not fix the structural problem in the migration.
- **Crash safety of the proposed fix:** When using explicit `COMMIT`/`BEGIN` inside the migration, if the process crashes mid-way, `alembic_version` still points to the pre-migration revision (it is only updated after `upgrade()` returns). A restart would re-run the migration from scratch. The `ON CONFLICT DO NOTHING` clause makes this safe — already-inserted messages are skipped, so the backfill is fully resumable.
- **This pattern is not unique to this migration.** Migration `242a2047eae0_update_chat_table.py` (which converts the `chat` column from `Text` to `JSON`) also reads every row and writes them back in a single transaction. It presumably succeeded when running against smaller datasets but would exhibit the same WAL problem at scale.
- **Concurrent migration risk:** Upstream issue #20067 reports that multiple Uvicorn workers can run migrations concurrently on startup, causing schema corruption. While we isolated the migration into a dedicated Cloud Run Job, the typical Docker deployment with multiple workers would be even more vulnerable to this migration's long runtime.

## Methodology

This analysis was produced with the help of GPT 5.4 xhigh, Claude Opus 4.6, and GPT 5.4 Extended Pro (30+ minutes thinking time). Each model was provided with extensive context including the full migration source code (`8452d01d26d7_add_chat_message_table.py`), related issues (#23134, #22568), the batching fix commit (b4f3408), PR #21542, live database diagnostics (table sizes, WAL metrics, TOAST analysis), AlloyDB documentation, and PostgreSQL source code (`src/backend/access/transam/xlog.c`, `src/backend/commands/trigger.c`, `src/backend/utils/mmgr/portalmem.c`, `src/backend/executor/execIndexing.c`, `src/backend/access/nbtree/nbtinsert.c`). All claims were verified against the PostgreSQL source tree. All three models independently identified the same root causes: per-row index write amplification and unbounded transaction duration, not the Python-side batching.

Classic298:
The `8452d01d26d7` ("Add chat_message table") migration fails on large PostgreSQL datasets (50+ GB) due to three interacting database-side issues that the existing Python-side batching fix (#21542) does not address:

1. **Unbounded transaction**: `env.py` wraps all migrations in a single `context.begin_transaction()`. The `begin_nested()` savepoints in `_flush_batch()` nest *inside* this outer transaction — nothing actually commits to disk until the entire migration returns. For large datasets this generates massive WAL that can't be recycled, causing managed PostgreSQL services (AlloyDB, Cloud SQL, RDS) to kill the connection.

2. **Indexes + FK created before backfill**: The migration creates 7 secondary indexes and a FK constraint *before* inserting any data. Every INSERT must update all 7 B-tree indexes + validate the FK — ~9 WAL records per row. PostgreSQL's own [bulk loading docs](https://www.postgresql.org/docs/current/populate.html) explicitly warn against this pattern.

3. **Server-side cursor invalidated by COMMIT**: `stream_results=True` creates a `WITHOUT HOLD` cursor on PostgreSQL. You can't naively "add commits inside the loop" — the cursor dies on the first COMMIT. A different pagination strategy is required.

## Related issues

- Closes #23273
- Related to #23134

> [!NOTE]

### DIFF
--- a/backend/open_webui/migrations/env.py
+++ b/backend/open_webui/migrations/env.py
@@ -102,7 +102,11 @@ def run_migrations_online() -> None:
         )
 
     with connectable.connect() as connection:
-        context.configure(connection=connection, target_metadata=target_metadata)
+        context.configure(
+            connection=connection,
+            target_metadata=target_metadata,
+            transaction_per_migration=True,
+        )
 
         with context.begin_transaction():
             context.run_migrations()
--- a/backend/open_webui/migrations/env.py
+++ b/backend/open_webui/migrations/env.py
@@ -57,6 +57,7 @@ def run_migrations_offline() -> None:
         target_metadata=target_metadata,
         literal_binds=True,
         dialect_opts={'paramstyle': 'named'},
+        transaction_per_migration=True,
     )
 
     with context.begin_transaction():
--- a/backend/open_webui/migrations/versions/8452d01d26d7_add_chat_message_table.py
+++ b/backend/open_webui/migrations/versions/8452d01d26d7_add_chat_message_table.py
@@ -21,7 +21,9 @@ down_revision: Union[str, None] = '374d2f66af06'
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 
-BATCH_SIZE = 5000
+# PostgreSQL has a 65,535 bind parameter limit. With 17 columns per row,
+# max safe batch is floor(65535 / 17) = 3855. Use 3000 for headroom.
+BATCH_SIZE = 3000
 CHAT_PAGE_SIZE = 100
 
 
@@ -405,7 +407,7 @@ def _upgrade_postgresql() -> None:
         "CREATE INDEX IF NOT EXISTS chat_message_user_created_idx ON chat_message (user_id, created_at)",
     ]:
         conn.execute(sa.text(stmt))
-    conn.commit()
+        conn.commit()  # Commit each index to avoid WAL accumulation
     log.info("Phase 3 complete: indexes created")
 
     # Phase 4: Add FK constraint
--- a/backend/open_webui/migrations/versions/8452d01d26d7_add_chat_message_table.py
+++ b/backend/open_webui/migrations/versions/8452d01d26d7_add_chat_message_table.py
@@ -112,26 +112,38 @@ def _flush_batch(conn, table, batch):
 
 
 def _flush_batch_pg(conn, table, batch):
-    """Insert a batch with ON CONFLICT DO NOTHING (PostgreSQL path)."""
+    """Insert a batch with ON CONFLICT DO NOTHING (PostgreSQL path).
+
+    Uses savepoints so a failed bulk insert doesn't abort the transaction —
+    PostgreSQL puts the transaction into an error state after any statement
+    failure, making all subsequent statements fail until ROLLBACK.
+    """
     from sqlalchemy.dialects.postgresql import insert as pg_insert
 
+    savepoint = conn.begin_nested()
     try:
         stmt = pg_insert(table).values(batch).on_conflict_do_nothing(
             index_elements=['id']
         )
         result = conn.execute(stmt)
+        savepoint.commit()
         return result.rowcount, len(batch) - result.rowcount
     except Exception:
+        savepoint.rollback()
+        # Fallback to row-by-row with individual savepoints
         inserted = 0
         failed = 0
         for msg in batch:
+            sp = conn.begin_nested()
             try:
                 stmt = pg_insert(table).values(**msg).on_conflict_do_nothing(
                     index_elements=['id']
                 )
                 result = conn.execute(stmt)
+                sp.commit()
                 inserted += result.rowcount
             except Exception as e:
+                sp.rollback()
                 failed += 1
                 log.warning(f'Failed to insert message {msg["id"]}: {e}')
         return inserted, failed
@@ -326,15 +338,26 @@ def _upgrade_postgresql() -> None:
     total_failed = 0
     total_chats = 0
 
+    # Capture high-water mark so concurrent inserts with non-monotonic IDs
+    # (e.g. UUIDs) don't get skipped by keyset pagination.
+    max_id_row = conn.execute(
+        sa.select(sa.func.max(chat_table.c.id))
+        .where(~chat_table.c.user_id.like('shared-%'))
+    ).fetchone()
+    max_id = max_id_row[0] if max_id_row and max_id_row[0] else None
+
     while True:
         # Keyset pagination: fetch next page of chats ordered by PK
-        rows = conn.execute(
+        query = (
             sa.select(chat_table.c.id, chat_table.c.user_id, chat_table.c.chat)
             .where(chat_table.c.id > last_id)
             .where(~chat_table.c.user_id.like('shared-%'))
             .order_by(chat_table.c.id)
             .limit(CHAT_PAGE_SIZE)
-        ).fetchall()
+        )
+        if max_id is not None:
+            query = query.where(chat_table.c.id <= max_id)
+        rows = conn.execute(query).fetchall()
 
         if not rows:
             break
--- a/backend/open_webui/migrations/versions/8452d01d26d7_add_chat_message_table.py
+++ b/backend/open_webui/migrations/versions/8452d01d26d7_add_chat_message_table.py
@@ -289,7 +289,7 @@ def _upgrade_postgresql() -> None:
         )
     """)
     )
-    conn.execute(sa.text("COMMIT"))
+    conn.commit()
     log.info("Phase 1 complete: chat_message table created (PK only)")
 
     # Phase 2: Backfill with keyset pagination and per-chunk commits
@@ -327,8 +327,6 @@ def _upgrade_postgresql() -> None:
     total_chats = 0
 
     while True:
-        conn.execute(sa.text("BEGIN"))
-
         # Keyset pagination: fetch next page of chats ordered by PK
         rows = conn.execute(
             sa.select(chat_table.c.id, chat_table.c.user_id, chat_table.c.chat)
@@ -339,7 +337,6 @@ def _upgrade_postgresql() -> None:
         ).fetchall()
 
         if not rows:
-            conn.execute(sa.text("COMMIT"))
             break
 
         last_id = rows[-1][0]
@@ -359,7 +356,7 @@ def _upgrade_postgresql() -> None:
             total_inserted += inserted
             total_failed += failed
 
-        conn.execute(sa.text("COMMIT"))
+        conn.commit()
 
         if total_inserted % 50000 < max(len(messages_batch), 1):
             log.info(
@@ -375,7 +372,6 @@ def _upgrade_postgresql() -> None:
     # Phase 3: Create indexes (bulk construction is orders of magnitude faster
     # than per-row maintenance during the backfill)
     log.info("Phase 3: creating indexes...")
-    conn.execute(sa.text("BEGIN"))
     for stmt in [
         "CREATE INDEX IF NOT EXISTS ix_chat_message_chat_id ON chat_message (chat_id)",
         "CREATE INDEX IF NOT EXISTS ix_chat_message_user_id ON chat_message (user_id)",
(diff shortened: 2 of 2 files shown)


Classic298:
@pr-validator-bot thorough review

Classic298:
@pr-validator-bot thorough review

Classic298:
@pr-validator-bot thorough review

Classic298:
@pr-validator-bot thorough review

Classic298:
@pr-validator-bot thorough review

Classic298:
Reviewed the remaining findings, here's why we're not actioning them:

**conn.commit() desync**: This was addressed in the earlier round. We specifically moved from raw `conn.execute(sa.text("COMMIT"))` to `conn.commit()`, which is SQLAlchemy's own transaction API. It properly updates internal connection state and autobegin handles subsequent transactions. `autocommit_block()` isn't suitable here, we need multiple commits across the backfill loop, not a single autocommit scope.

**Keyset pagination + live writes**: We already added the `max_id` high-water mark to handle non-monotonic IDs. The remaining gap (new rows inserted with `id <= last_id` during migration) is inherent to any pagination strategy short of table-level locking. This is a one-time schema migration, not a continuous sync, it should be run during a maintenance window with the application stopped, which is standard practice for migrations involving large data backfills. New chats written by the *new* app code would go directly into `chat_message` anyway, so only legacy writes during migration are at risk.

**transaction_per_migration global change**: This is Alembic's recommended setting. Each migration should be self-contained and not rely on sharing a transaction with adjacent revisions. No existing migration in the codebase depends on cross-migration transaction semantics.

Classic298:
@tjbck ready for review

tjbck:
@madnight testing wanted here.

PR #23388, opened 2026-04-03
perf: optimize fence-stripping O(n²) → O(n)

Classic298:
### Summary

It's a **server-side CPU improvement** during token streaming. Each streamed chunk calls serialize_output to build the content string that gets emitted via the event emitter to the frontend. With the old code, every concatenation allocated and copied a new string — that's pure CPU waste on the Python backend.

serialize_output builds a display string by iterating output items. The old implementation used repeated string concatenation (content += ...), which copies the entire accumulated string on every append — O(n²) total work. This replaces it with list.append + a single join at the end — O(n).

### Changes

- Replace string accumulator with a parts list in serialize_output
- Same treatment for inner accumulators (reasoning content, result text)
- Replace split_content_and_whitespace / is_opening_code_block helpers with content_endswith_newline / ensure_trailing_newline operating on the parts list
- Fix fence-stripping for code_interpreter: use regex so language-tagged fences (e.g. triple-backtick python) are correctly stripped — the old rstrip approach silently failed on these

# Benefit

The gain scales with **output list size × content length × call frequency**. The biggest wins come from:

**The worst case:** a streaming response after several tool-call rounds where the tools returned large results. The output list is big, the accumulated content is big, and serialize_output is called hundreds of times — each time re-concatenating everything.

**The best case (minimal gain):** a simple text-only response with no tool calls, no reasoning — just one message item, one small concatenation, called a few times. Effectively no difference.

The practical effect is **lower per-chunk processing latency** on the server, which means:
- Less CPU time spent per streaming event
- Smoother token delivery to the frontend (less jitter between chunks)
- Better scalability under concurrent users (less CPU contention per active stream)

Added co author of original PR

> [!NOTE]


Classic298:

Classic298:

Classic298:
Both are pre-existing behaviors from the original code, not regressions. The original split('```') scanned the same HTML-inclusive content and matched the same non-line-boundary positions. This PR preserves that behavior while fixing the complexity — hardening the fence detection scope is out of scope for a performance refactor.

Classic298:
@tjbck ready 4 review

Classic298:
Tested extensively and it works just fine

Classic298:

Classic298:
@tjbck

Classic298:

Classic298:

Classic298:
@tjbck

Classic298:
@pr-validator-bot review

Classic298:
@tjbck

tjbck:
Likely addressed, confirmation needed here.

PR #23484, opened 2026-04-07
feat: support Azure v1 endpoint format (/openai/v1)

Classic298:
Azure offers two URL formats: the legacy deployment-based format (/openai/deployments/{model}/...) and the newer v1 format (/openai/v1/...) where the model stays in the payload body and no api-version query parameter is needed.

Previously, the code always ran convert_to_azure_payload which rewrites the URL to the deployment format, causing 404 errors for users with v1-style base URLs. Now, when the base URL contains '/openai/v1', we skip deployment URL construction and route directly.

Applied consistently across all three Azure routing paths: generate_chat_completion, /responses proxy, and generic proxy.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/openai.py
+++ b/backend/open_webui/routers/openai.py
@@ -1131,21 +1131,31 @@ async def generate_chat_completion(
     is_responses = api_config.get('api_type') == 'responses'
 
     if api_config.get('azure', False):
-        api_version = api_config.get('api_version', '2023-03-15-preview')
-        request_url, payload = convert_to_azure_payload(url, payload, api_version)
-
         # Only set api-key header if not using Azure Entra ID authentication
         auth_type = api_config.get('auth_type', 'bearer')
         if auth_type not in ('azure_ad', 'microsoft_entra_id'):
             headers['api-key'] = key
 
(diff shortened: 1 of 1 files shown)


Classic298:
@pr-validator-bot thorough review

Classic298:
@pr-validator-bot thorough review

Classic298:
**Azure v1 endpoints use the standard OpenAI API contract** — that's the entire point. The model stays in the payload body (not in the URL), and the endpoint accepts standard OpenAI parameters without Azure-specific filtering. Skipping `convert_to_azure_payload` is intentional because that function does two things that are specifically wrong for v1: (1) rewrites the URL to `/deployments/{model}/`, and (2) applies Azure-deployment-specific param filtering (like o-series temperature removal) that doesn't apply to the standard API surface.

If someone configures `/openai/v1` as their base URL, they're opting into the standard OpenAI-compatible behavior, not the legacy Azure deployment behavior.

Second finding (extract a helper) is a valid follow-up refactor but out of scope here.

No changes needed.

tjbck:
Thanks!

PR #23515, opened 2026-04-08
fix: native function calling for REST API consumers

PedroFCM:
Two fixes for native function calling, especially for REST API consumers that bypass the frontend WebSocket session:

1. Pass through the `function_calling` value as-is instead of only checking for `'native'` and falling back to `'default'`. This respects values set via `DEFAULT_MODEL_PARAMS` or per-model params.

2. Add server-side tool-calling loop for API calls (no `session_id`). The native tool execution loop is driven by the frontend via WebSocket. REST API calls never enter that loop, so `tool_calls` are returned raw and tools are never executed. This adds a loop that executes tools and re-invokes the LLM until `finish_reason != 'tool_calls'` or `CHAT_RESPONSE_MAX_TOOL_CALL_RETRIES` is reached.

Ref: #9435

# Pull Request Checklist

- [x] **Description:** See above and changelog below
- [x] **Documentation:** N/A — no new env vars or user-facing config; uses existing `CHAT_RESPONSE_MAX_TOOL_CALL_RETRIES`
- [x] **Testing:** Manually tested with AWS Bedrock models (Claude) via REST API with native tool calling. Verified: tool loop completes, LLM produces final response, frontend WebSocket flow unaffected, requests without tools unchanged.
- [x] **Agentic AI Code:** Human-written and manually tested in production
- [x] **Design & Architecture:** No new settings; uses existing `CHAT_RESPONSE_MAX_TOOL_CALL_RETRIES`
- [x] **Git Hygiene:** Single atomic commit rebased on `dev`

# Changelog Entry

### Description

- Fix native function calling for REST API consumers (`/api/chat/completions`) that don't use a WebSocket session

### Fixed

- `function_calling` parameter now passes through the configured value instead of only recognizing `'native'` and falling back to `'default'`
- REST API calls with native function calling now execute tools server-side and re-invoke the LLM with tool results, matching the behavior of the frontend WebSocket flow

### Breaking Changes

- None. Behavioral change is scoped to REST API requests with `function_calling: native` and tools — these previously returned raw `tool_calls` without execution.

---

### Additional Information

- Related discussion: #9435 ("The model does nothing with the result of a native tool call") — reported by 30+ users since Feb 2025
- The server-side loop only activates when: no `session_id` (REST API), `function_calling == 'native'`, and tools are present in metadata
- Loop is bounded by existing `CHAT_RESPONSE_MAX_TOOL_CALL_RETRIES` setting
- Streaming is temporarily disabled during the tool loop, then restored for the final LLM call

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1739,12 +1739,9 @@ async def chat_completion(
                 'stream_delta_chunk_size': stream_delta_chunk_size,
                 'reasoning_tags': reasoning_tags,
                 'function_calling': (
-                    'native'
-                    if (
-                        form_data.get('params', {}).get('function_calling') == 'native'
-                        or model_info_params.get('function_calling') == 'native'
-                    )
-                    else 'default'
+                    form_data.get('params', {}).get('function_calling')
+                    or model_info_params.get('function_calling')
+                    or 'default'
                 ),
             },
         }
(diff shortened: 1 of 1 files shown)


PedroFCM:
@pr-validator-bot review

PedroFCM:
Hi @tjbck, could you share why this was closed? Happy to adjust the approach if there are any concerns

brettp:
@PedroFCM  - It looks like PRs require an [explicit confirmation.](https://github.com/open-webui/open-webui/pull/23515#issuecomment-4208113815) The validator bot is pretty noisy, so this is easy to miss.

Classic298:
the explicit confirmation for testing can also be given within the PR description - but it's not needed - its only needed for larger PRs that are likely to break things

And as per the review the bot has made, this PR breaks things.

https://github.com/open-webui/open-webui/pull/23515#issuecomment-4208154035

PedroFCM:
Thanks @Classic298 and @brettp for the feedback. Both findings from the bot review have been addressed:

1. **Falsy tool results**: Changed `str(result) if result else ''` to `str(result) if result is not None else ''` so values like `0`, `False`, and `""` are preserved correctly

2. **Extra LLM call**: The loop forces `stream=False` to parse tool_calls from JSON. Returning the loop's JSON response directly to a streaming caller breaks SSE parsing. The final call now only happens when:
   - Tool results are pending (LLM hasn't seen them yet), OR
   - The caller requested streaming (need SSE format, not JSON)
   
   Non-streaming REST API callers reuse the loop's response directly, meaning that there is no extra call

Opening a new PR with the updated code since the diff here is frozen: #23581

PR #23517, opened 2026-04-08
feat: add a configurable limit for how many model can be prompted in a chat

lorenzophys:
### Description

This PR adds a configurable limit on the number of models that can be used simultaneously in a single chat. By default the limit is unlimited, preserving existing behavior. The number of models can be specified via the `MAX_CHAT_MODELS` environment variable or from the Admin Panel > Settings > General under "Features".

The reason for this is that having the possibility to prompt unlimited models simultaneously, some users will use this feature to the maximum of its capacity generating high cost and big headaches. We have it written in the guidelines to use max 3 models, but nobody reads the guidelines :)

### Added

- New `MAX_CHAT_MODELS` `PersistentConfig` entry, readable from the `MAX_CHAT_MODELS` environment variable (empty = unlimited)
- Initialized `app.state.config.MAX_CHAT_MODELS` on startup
- Exposed `max_chat_models` in the `GET /api/config` features response so the frontend can read it
- Added `MAX_CHAT_MODELS` to the `GET /auths/admin/config` response so the value is loaded in the admin panel
- Added `MAX_CHAT_MODELS` field to the `AdminConfig` Pydantic model
- Saved the value in `POST /auths/admin/config` (casted to int, or `''` for unlimited)
- Included `MAX_CHAT_MODELS` in the `POST` response
- Added `max_chat_models?: number` to the `features` block of the `Config` type.
- New number input field "Max Chat Models" below the "Folder Max File Count" field, following the same pattern: min="1", placeholder "Leave empty for unlimited", linked to `adminConfig.MAX_CHAT_MODELS`
- The "Add Model" button is hidden when the number of selected models reaches the `MAX_CHAT_MODELS`
- When loading a chat, `selectedModels` is trimmed to `max_chat_models` if the saved chat contains more models than the current limit allows. It handles a specific case when a chat was saved when the limit was higher, and later an admin lowers the limit. This way there is no way of going around the limit.

### Note

I made this feature as similar as possible to the `FOLDER_MAX_FILE_COUNT` so that it blends nicely with the existing code.

### Breaking Changes

- None, the default is unlimited as it is now

---

### Test

Build the container with `MAX_CHAT_MODELS=3` and you get this:

Screenshot 2026-04-08 at 20 31 27

If I try to set the value to zero it won't let me and will explain why

Screenshot 2026-04-08 at 20 32 06

Note that after adding the third model the plus sign disappears making it impossible to add more.

Screenshot 2026-04-08 at 20 13 44

> [!NOTE]

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -1535,6 +1535,12 @@ FOLDER_MAX_FILE_COUNT = PersistentConfig(
     os.environ.get('FOLDER_MAX_FILE_COUNT', ''),
 )
 
+MAX_CHAT_MODELS = PersistentConfig(
+    'MAX_CHAT_MODELS',
+    'chat.max_models',
+    os.environ.get('MAX_CHAT_MODELS', ''),
+)
+
 ENABLE_CHANNELS = PersistentConfig(
     'ENABLE_CHANNELS',
     'channels.enable',
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -383,6 +383,7 @@ from open_webui.config import (
     API_KEYS_ALLOWED_ENDPOINTS,
     ENABLE_FOLDERS,
     FOLDER_MAX_FILE_COUNT,
+    MAX_CHAT_MODELS,
     ENABLE_CHANNELS,
     ENABLE_NOTES,
     ENABLE_USER_STATUS,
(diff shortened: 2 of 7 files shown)


lorenzophys:
@pr-validator-bot review

lorenzophys:
Addressing code review:

1. It is true that if Openwebui is initialized with for example `MAX_CHAT_MODELS='hello'`, the ui is broken beyond repair. So I am adding a validation in `auths.py` and `config.py`
2. I am not really sure about the second comment: I tested if lowering the limit also applies to previous chats and it does

lorenzophys:
@tjbck any reason for closing this? It's not a bad feature and it's quite small

Classic298:
@lorenzophys discussion first needed. This is another feature that might break in the future and causes maintenance burden. If not many need it, it's not worth adding it.besides you can turn off multi model chats in the permissions

tjbck:
@lorenzophys won't be supported, you can either enable/disable.

lorenzophys:
@Classic298 understand. I opened a discussion: https://github.com/open-webui/open-webui/discussions/23519
If it becomes something that you wish to have I can continue with this PR

PR #23322, opened 2026-04-01
fix: normalize usage keys on save in ChatMessages.upsert_message

smorello87:
# Pull Request Checklist

- [x] **Target branch:** Verify that the pull request targets the `dev` branch.
- [x] **Description:** Provided below.
- [x] **Changelog:** Provided below.
- [ ] **Documentation:** No user-facing behavior changes; this is an internal normalization fix.
- [x] **Dependencies:** No new dependencies. Reuses existing `normalize_usage()` from `utils/response.py`.
- [x] **Testing:** Tested on production PostgreSQL 16.3 deployment (~500 users) since v0.8.0 through v0.8.12 as a build-time patch. Verified analytics dashboard correctly displays token usage for all providers (OpenRouter, Bedrock, Ollama) after applying.
- [x] **Agentic AI Code:** This fix was authored and tested by a human. We have been running this exact patch in production since February 2026.
- [x] **Code review:** Self-reviewed. The change is 4 lines: 1 import + normalization calls in both update and insert paths.
- [x] **Design & Architecture:** No new settings or UI changes. Reuses existing internal function.
- [x] **Git Hygiene:** Single atomic commit, one logical change.
- [x] **Title Prefix:** `fix:`

# Changelog Entry

### Description

OpenAI-compatible APIs (OpenRouter, etc.) return token usage as `prompt_tokens` / `completion_tokens`, but the analytics queries in `get_token_usage_by_model` and `get_token_usage_by_user` read `input_tokens` / `output_tokens`. The existing `normalize_usage()` function in `utils/response.py` handles this mapping but is only called in the streaming middleware — the normalized result never reaches the database save path through `ChatMessages.upsert_message()`.

This adds `normalize_usage()` calls in both the **update** and **insert** branches of `upsert_message()`, ensuring all usage data is normalized before saving regardless of code path.

### Added

- N/A

### Changed

- N/A

### Deprecated

- N/A

### Removed

- N/A

### Fixed

- Token usage analytics showing 0 for most messages because `chat_message.usage` was saved with OpenAI-format keys (`prompt_tokens`/`completion_tokens`) but analytics queries read Anthropic-format keys (`input_tokens`/`output_tokens`). Now both key formats are always present.

### Security

- N/A

### Breaking Changes

- N/A

---

### Additional Information

- Fixes #21347 (Bug 2: usage key name mismatch)
- Related: #21675 (analytics enhancements)
- This fix only affects **new** messages going forward. Existing records can be backfilled with:

```sql
[...]

PR #23556, opened 2026-04-10
fix: always include stream_options for usage tracking in streaming requests

smorello87:
- [x] **Testing:** Manually tested on ECS deployment with OpenRouter, Bedrock Gateway, and Bedrock Mantle

# Changelog Entry

### Fixed

- Always include `stream_options: { include_usage: true }` in streaming chat completion requests so providers that require explicit opt-in (e.g., AWS Bedrock) return token usage data for analytics

---

### Description

Providers like AWS Bedrock only return token usage data in streaming responses when the client sends `stream_options: { include_usage: true }`. Without this, the analytics dashboard shows zero tokens for these models.

Currently, the frontend only sends `stream_options` when a model's metadata includes `capabilities.usage = true`, which requires manual admin configuration via `DEFAULT_MODEL_METADATA`. This change injects it server-side so analytics work out of the box for all providers.

OpenRouter and OpenAI already return usage by default and safely ignore the extra field. The Responses API is excluded since it doesn't support `stream_options`.

Related: #21347, #23322

### Changes

One-line addition in `backend/open_webui/routers/openai.py` — before serializing the payload, inject `stream_options: { include_usage: true }` for streaming requests that don't already have it set.

### Testing

Tested on an ECS Fargate deployment with three providers:
- **OpenRouter** — already returned usage, continues to work (field is ignored if already present)
- **AWS Bedrock (via bedrock-access-gateway sidecar)** — previously showed 0 tokens in analytics, now correctly reports input/output tokens
- **AWS Bedrock Mantle** — same fix, now returns usage data

Verified via CloudWatch logs that raw provider responses include `prompt_tokens` and `completion_tokens` after the change.

### DIFF
--- a/backend/open_webui/routers/openai.py
+++ b/backend/open_webui/routers/openai.py
@@ -1171,6 +1171,12 @@ async def generate_chat_completion(
                     part.get('text', '') for part in message['content'] if part.get('type') in ('input_text', 'text')
                 )
 
+    # Always request usage data in streaming responses for analytics token tracking.
+    # Without this, providers like AWS Bedrock don't include token counts unless
+    # the client explicitly opts in via stream_options.
+    if payload.get('stream', False) and 'stream_options' not in payload and not is_responses:
+        payload['stream_options'] = {'include_usage': True}
+
     payload = json.dumps(payload)
 
     r = None


Classic298:
What is the point of this? You can just enable the model capability to enable Usage, which will include exactly this in the request and makes sure the response contains the Usage field.

Why is this needed - you can already enable it.

On top of that, you can enable it for all models using the default model settings in the admin panel

ON TOP OF THAT you can easily achieve the same thing with a filter too

ON TOP OF THAT "always including" this will be a breaking change for many deployments! If this is merged in, many deployments will break. There are providers that don't support sending Usage details - and if you send them the include_usage parameter, they will error out and send you a Bad Request error back.

So why is this needed?

Has there been a discussion about this prior to opening this PR?

On top of that it makes the Usage capability useless / cannot be toggled off anymore since you now always enforce it.

PR #23573, opened 2026-04-10
fix: add health_check_interval to detect stale pooled connections

Classic298:
Introduces REDIS_HEALTH_CHECK_INTERVAL and wires it through to every Redis client created by get_redis_connection (plain, cluster and sentinel paths, sync and async). When set to a positive integer, redis-py will PING any pooled connection idle longer than the interval on checkout, so dead sockets are surfaced as reconnectable errors before a real command lands on them.
Backward compatible / opt-in. Defaults to unset (empty string), which is parsed to None and skipped at the call site — so existing deployments see zero behavioural change after merging this PR. Operators who want the protection set REDIS_HEALTH_CHECK_INTERVAL in their environment (e.g. 60), and must pick a value shorter than both the Redis server timeout setting and any firewall/LB idle timeout on the path to Redis.
Why we need this
Our Redis server reaps stale connections via timeout 1800. Open WebUI doesn't currently check/ping pooled Redis connections before reusing them, so the first request that grabs a reaped socket out of the pool blows up with redis.exceptions.ConnectionError: Connection reset by peer, which bubbles up through AppConfig.__getattr__ / the request middleware and surfaces as user-visible 500s and unexpected runtime errors.
Server-side reaping is non-negotiable for us — without it we run out of pool slots because Open WebUI doesn't appear to close Redis connections deterministically, which is why we originally raised maxclients from 1000 to 50000 and set timeout 1800 in the first place.
With this PR merged and REDIS_HEALTH_CHECK_INTERVAL=60 set, redis-py PINGs any idle connection before handing it back to a caller, which both (a) validates the socket and (b) resets Redis's idle-timeout counter for that connection. Connections still in active rotation stay warm indefinitely; orphaned/leaked ones still get reaped by the server timeout as a safety net. Two different failure modes, covered by two different mechanisms, no conflict between them.

> [!NOTE]

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -435,6 +435,20 @@ REDIS_SOCKET_KEEPALIVE = (
     os.environ.get('REDIS_SOCKET_KEEPALIVE', 'False').lower() == 'true'
 )
 
+# How often (in seconds) redis-py should PING an idle pooled connection
+# before reusing it. Opt-in: defaults to unset (empty string) so behavior
+# is unchanged for existing deployments. When set, should be shorter than
+# the Redis server `timeout` setting and any firewall/LB idle timeout on
+# the path to Redis, so stale connections are detected before a real
+# command lands on them. Set to 0 or empty to disable.
+REDIS_HEALTH_CHECK_INTERVAL = os.environ.get('REDIS_HEALTH_CHECK_INTERVAL', '')
+try:
+    REDIS_HEALTH_CHECK_INTERVAL = int(REDIS_HEALTH_CHECK_INTERVAL)
+    if REDIS_HEALTH_CHECK_INTERVAL <= 0:
+        REDIS_HEALTH_CHECK_INTERVAL = None
+except ValueError:
+    REDIS_HEALTH_CHECK_INTERVAL = None
+
 REDIS_RECONNECT_DELAY = os.environ.get('REDIS_RECONNECT_DELAY', '')
 
 if REDIS_RECONNECT_DELAY == '':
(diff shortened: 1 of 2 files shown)


Classic298:
@pr-validator-bot full review

Classic298:
@tjbck we would need this because our Redis reaps stale connections - and Open WebUI doesn't check/ping these connections to check if they are still alive to replace them before using them again and pinging them would also keep them alive and keep them from getting reaped.

Reaping on redis server side is still needed for us because open webui doesnt seem to close redis connections and we always run out of connections in the redis pool if we don't reap connections that are inactive for longer than 30 minutes.

This PR will prevent stale connections by keeping those in the pool alive - and if they are detected to be dead, they can be replaced. This will prevent redis errors in production.

These redis errors are causing issues for us where Open WebUI will throw errors because certain requests to redis errored out which can cause various runtime errors in Open WebUI and unexecpted behaviour.

backend/open_webui/env.py
```diff
@@ -426,6 +425,19 @@ def parse_section(section):
 except ValueError:
     REDIS_SOCKET_CONNECT_TIMEOUT = None
 
+# How often (in seconds) redis-py should PING an idle pooled connection
+# before reusing it. Must be shorter than the Redis server `timeout`
+# setting and any firewall/LB idle timeout on the path to Redis, so stale
+# connections are detected before a real command lands on them.
+# Set to 0 or empty to disable.
+REDIS_HEALTH_CHECK_INTERVAL = os.environ.get('REDIS_HEALTH_CHECK_INTERVAL', '60')
```
tjbck: Should be disabled by default.

Classic298:
@pr-validator-bot full review

Classic298:
@pr-validator-bot full review

Classic298:
@tjbck

tjbck:
@Classic298 conflicts?

Classic298:
@tjbck

PR #23581, opened 2026-04-10
fix: native function calling for REST API consumers

PedroFCM:
Two fixes for native function calling, especially for REST API consumers that bypass the frontend WebSocket session:

1. **Pass through the `function_calling` value as-is** instead of only checking for `'native'` and falling back to `'default'`. This respects values set via `DEFAULT_MODEL_PARAMS` or per-model params.

2. **Add server-side tool-calling loop for API calls** (no `session_id`). The native tool execution loop is driven by the frontend via WebSocket. REST API calls never enter that loop, so `tool_calls` are returned raw and tools are never executed. This adds a loop that executes tools and re-invokes the LLM until `finish_reason != 'tool_calls'` or `CHAT_RESPONSE_MAX_TOOL_CALL_RETRIES` is reached.

Ref: #9435
Replaces PR: #23515 (closed - updated based on bot review feedback)

# Pull Request Checklist

- [x] **Description:** See above and changelog below
- [x] **Documentation:** N/A - no new env vars or user-facing config; uses existing `CHAT_RESPONSE_MAX_TOOL_CALL_RETRIES`
- [x] **Testing:** Tested on a deployment environment (AWS ECS Fargate + Bedrock Access Gateway) with Claude models. Verified via REST API (`/api/chat/completions`) with both `stream: true` and `stream: false`. Tool calls are executed server-side and the LLM produces a final response incorporating tool results. Frontend WebSocket flow unaffected.
- [x] **Agentic AI Code:** AI-assisted with human review and manual testing on deployment environment
- [x] **Code review:** Self-reviewed; addressed both findings from automated review on #23515
- [x] **Design & Architecture:** No new settings; uses existing `CHAT_RESPONSE_MAX_TOOL_CALL_RETRIES`
- [x] **Git Hygiene:** Single atomic commit rebased on `dev`

# Changelog Entry

### Description

- Fix native function calling for REST API consumers (`/api/chat/completions`) that don't use a WebSocket session

### Fixed

- `function_calling` parameter now passes through the configured value instead of only recognizing `'native'` and falling back to `'default'`
- REST API calls with native function calling now execute tools server-side and re-invoke the LLM with tool results, matching the behavior of the frontend WebSocket flow
- Falsy tool results (`0`, `False`, `""`) are preserved correctly via `is not None` check

### Breaking Changes

- None. Behavioral change is scoped to REST API requests with `function_calling: native` and tools - these previously returned raw `tool_calls` without execution.

---

### Additional Information

- Related discussion: #9435 ("The model does nothing with the result of a native tool call") - reported by 30+ users since Feb 2025
- The server-side loop only activates when: no `session_id` (REST API), `function_calling == 'native'`, and tools are present in metadata
- Loop is bounded by existing `CHAT_RESPONSE_MAX_TOOL_CALL_RETRIES` setting
- The loop forces `stream=False` to parse tool_calls from JSON. After the loop, the final LLM call is only made when:
  - Tool results are pending (LLM hasn't seen them), OR
  - The caller requested streaming (loop response is JSON, caller expects SSE)
  - Non-streaming callers reuse the loop's response directly - no extra LLM call

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1739,12 +1739,9 @@ async def chat_completion(
                 'stream_delta_chunk_size': stream_delta_chunk_size,
                 'reasoning_tags': reasoning_tags,
                 'function_calling': (
-                    'native'
-                    if (
-                        form_data.get('params', {}).get('function_calling') == 'native'
-                        or model_info_params.get('function_calling') == 'native'
-                    )
-                    else 'default'
+                    form_data.get('params', {}).get('function_calling')
+                    or model_info_params.get('function_calling')
+                    or 'default'
                 ),
             },
         }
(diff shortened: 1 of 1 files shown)


PedroFCM:
@pr-validator-bot review

tjbck:
Open to discussion, but I'm struggling to see why this is even an issue.

PedroFCM:
Thanks for the feedback @tjbck, happy to discuss.

Currently, the frontend handles native tool calling fine via the WebSocket loop, but the REST API path has no equivalent - tools execute, but the LLM never sees the results, so the conversation stops, which breaks the flow of using OpenWebUI with tool calling (e.g., MCPs) through an API.

This affects anyone using OpenWebUI programmatically (automation, integrations, agentic workflows) - for example, creating custom models with tools and MCPs in OpenWebUI and calling them via the API from external services. This is where native tool calling delivers the most value. Moreover, several users in #9435 have reported the same behavior across different models and versions.

Would you be open to re-opening? Happy to adjust the approach if necessary.

MrMakkarone:
Hi @PedroFCM I have tried using the api/chat/completions API endpoint and using tool calling. I used the latest image (0.9.1) and it was possible to trigger one tool call. However, when there were multiple tool calls, none of these worked. Multiple tool calls when using the API is still a problem. I tried to apply your fix, unfortunately it did not fix it either. I would love if there would be any fix in the future to allow multiple tool calls in one request.

PedroFCM:
Hi @MrMakkarone, thanks for the feedback! 
I will take a look at it next week. When I applied this approach, I only tested it for one tool call indeed, so there might be a lot of space for improvement. I will come back to you whenever I have something

PedroFCM:
Hi @MrMakkarone! It took a bit more time to start exploring this, but I created a new PR taken into account your feedback: https://github.com/open-webui/open-webui/pull/24707

MrMakkarone:
> Hi @MrMakkarone! It took a bit more time to start exploring this, but I created a new PR taken into account your feedback: https://github.com/open-webui/open-webui/pull/24707

That is awesome! I will try it. Thanks for your effort!😊

PR #23622, opened 2026-04-12
fix: add DB-backed token revocation fallback when Redis is absent

Classic298:
Add a token_version counter to the User model that gets embedded in JWTs and checked on every request. When Redis is not configured, sign-out and password change now increment token_version in the DB, causing all previously issued tokens to be rejected.

Changes:
- User model: add token_version column (BigInteger, default 0)
- auth.py: is_valid_token checks token_version against DB when no Redis
- auth.py: invalidate_token increments token_version when no Redis
- auths.py: embed token_version in JWT at all token creation points
- auths.py: password change now invalidates prior tokens
- oauth.py: embed token_version in OAuth callback JWT
- oauth.py: back-channel logout uses token_version when no Redis
- Migration: add token_version column to user table

> [!NOTE]

### DIFF
--- a/backend/open_webui/migrations/versions/c66045cb0f67_add_user_token_version_column.py
+++ b/backend/open_webui/migrations/versions/c66045cb0f67_add_user_token_version_column.py
@@ -1,0 +1,29 @@
+"""add user token_version column
+
+Revision ID: c66045cb0f67
+Revises: b7c8d9e0f1a2
+Create Date: 2026-04-12 16:44:37.020844
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision: str = 'c66045cb0f67'
+down_revision: Union[str, None] = 'b7c8d9e0f1a2'
+branch_labels: Union[str, Sequence[str], None] = None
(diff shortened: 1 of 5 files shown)


Classic298:
@pr-validator-bot full review

Classic298:
@pr-validator-bot full review

Classic298:
@pr-validator-bot full review

Classic298:
@pr-validator-bot full review

Classic298:
Regarding the remaining bot findings (best-effort token invalidation on password change and sign-out):

These are intentionally best-effort with warning logs. Making them fail-hard would be worse:

- **Password change**: The password is already updated in the DB before token_version is incremented. If we fail the endpoint, the user sees "password change failed" but their old password no longer works — effectively locking them out. Making it truly atomic would require refactoring the session management pattern used across the entire codebase (each model method manages its own DB session via get_db_context).

- **Sign-out**: Cookies and OAuth sessions are already cleared before token revocation. Returning a sign-out error doesn't help the user and may cause the frontend to keep them "signed in" UI-side.

- **Exception handling pattern**: Every method in UsersTable follows the same try/except → return None convention. Changing it for one method belongs in a broader codebase refactor, not this PR.

All failure paths log warnings so they surface in monitoring.

PR #23626, opened 2026-04-12
fix: prevent first-user admin race in LDAP and OAuth registration

Classic298:
Both LDAP and OAuth registration checked user count before insert to determine whether to assign admin role.  With multiple workers, concurrent first-user registrations could each see zero users and both create admin accounts.

Applies the insert-first-check-after pattern already used by signup_handler: insert with DEFAULT_USER_ROLE, then atomically check get_num_users()==1 and promote only the sole user to admin.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/auths.py
+++ b/backend/open_webui/routers/auths.py
@@ -479,18 +479,24 @@ async def ldap_auth(
             user = Users.get_user_by_email(email, db=db)
             if not user:
                 try:
-                    role = 'admin' if not Users.has_users(db=db) else request.app.state.config.DEFAULT_USER_ROLE
-
+                    # Insert with default role first to avoid TOCTOU race on
+                    # first-user registration.  Matches signup_handler pattern.
                     user = Auths.insert_new_auth(
                         email=email,
                         password=str(uuid.uuid4()),
                         name=cn,
-                        role=role,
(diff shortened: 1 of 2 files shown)


Classic298:
@pr-validator-bot full review

Classic298:
@pr-validator-bot full review

Classic298:
Acknowledged — the insert-then-count is not a single atomic DB operation, and a microsecond interleave could theoretically produce zero admins instead of double admins.

This is an accepted tradeoff for these reasons:

1. This is the exact same pattern already deployed in signup_handler (the email signup path), which was explicitly designed this way. This PR brings LDAP/OAuth in line with that established pattern.

2. The failure mode is strictly better: the original code guaranteed double-admin on concurrent first-user registration. The new code reduces that to a theoretical zero-admin in a sub-millisecond window, which requires both inserts to commit before either count executes.

3. The zero-admin case is self-correcting: the OAuth get_user_role path repairs a sole user to admin on next login (user and user_count == 1), and admins can always be set via CLI.

4. True atomicity (advisory locks, compare-and-set, constraint-backed claim) would require architectural changes across the auth layer. That's a valid improvement for a follow-up, but out of scope for this targeted security fix.

PR #23627, opened 2026-04-12
fix: enforce access checks on RAG vector search collection queries

Classic298:
The non-full-context file path, text collection_name path, and bare collection_name/collection_names fallback all queried the vector store without verifying the user has read access to the underlying file or knowledge base.

Adds _has_access_to_collection helper that resolves collection name format (file-<id>, knowledge UUID, legacy bare ID) to the underlying resource and verifies ownership or access grants. All three unprotected paths now gate on this check before querying.

> [!NOTE]


Classic298:
@pr-validator-bot full review

Classic298:
@pr-validator-bot full review

Classic298:
@pr-validator-bot full review

Classic298:
@pr-validator-bot full review

Classic298:
@pr-validator-bot full review

Classic298:
Finding #1 is not a regression — the if/elif/else chain was already structured this way before the PR. The original code always took the collection_name branch when present, never falling through to file/content. The access check was added inside the existing branch, not changing the control flow. Items from the frontend are single-type, not mixed-type blobs.

Finding #2 (dual lookup for legacy IDs) is real but mitigated by the per-request cache — repeated lookups for the same legacy ID only hit the DB once. Legacy format is also increasingly rare.

Classic298:
@pr-validator-bot full review

Classic298:
@tjbck

PR #23631, opened 2026-04-12
fix: enforce model access control on Ollama generate, show, embed, embeddings endpoints

Classic298:
These four endpoints checked model existence but never verified the user has read access via AccessGrants, allowing any authenticated user to use restricted models.

Adds _verify_model_access helper mirroring the /api/chat access control pattern and applies it to /api/show, /api/embed, /api/embeddings, and /api/generate.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/ollama.py
+++ b/backend/open_webui/routers/ollama.py
@@ -797,10 +797,13 @@ async def show_model_info(request: Request, form_data: ModelNameForm, user=Depen
     form_data = form_data.model_dump(exclude_none=True)
     form_data['model'] = form_data.get('model', form_data.get('name'))
 
+    model = form_data.get('model')
+
+    # Enforce per-model access control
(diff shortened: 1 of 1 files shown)


Classic298:
@pr-validator-bot full review

Classic298:
@pr-validator-bot full review

Classic298:
@pr-validator-bot full review

Classic298:
Finding #1 (prefix normalization): This is the same behavior as /api/chat/{url_idx} which has been shipping — it also calls Models.get_model_by_id(model_id) before prefix normalization, and non-admins get 403 when model_info is None. The fix is consistent, not a regression.

Finding #2 (request.state bypass): The request.state.bypass_filter flag is specific to /api/chat because it's invoked internally from the chat pipeline. The generate/embed/embeddings/show endpoints don't have internal callers that set this flag, so aligning isn't needed.

No code changes for either.

PR #23633, opened 2026-04-12
fix: reject empty passwords in LDAP authentication to prevent unauthenticated binds

Classic298:
Per RFC 4513, a Simple Bind with a non-empty DN but empty password is unauthenticated simple authentication. Many LDAP servers (OpenLDAP default, some AD configs) accept these binds, allowing account takeover without valid credentials.

Rejects empty and whitespace-only passwords before attempting the LDAP bind.

> [!NOTE]


Classic298:
@pr-validator-bot full review

PR #23637, opened 2026-04-12
fix: enforce API key endpoint restrictions on cookie-based authentication

Classic298:
The APIKeyRestrictionMiddleware only inspected the Authorization header for sk- tokens, but get_current_user also reads API keys from cookies and x-api-key headers. This allowed complete bypass of endpoint restrictions by sending the key via an alternate transport.

Moves the restriction check into get_current_user_by_api_key so it runs regardless of how the API key was delivered. Removes the now-redundant middleware.

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1387,50 +1387,6 @@ app.add_middleware(RedirectMiddleware)
 app.add_middleware(SecurityHeadersMiddleware)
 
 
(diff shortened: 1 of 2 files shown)


Classic298:
@pr-validator-bot full review

Classic298:
@pr-validator-bot full review

PR #23642, opened 2026-04-12
fix: invalidate stale Socket.IO sessions on role change and user deletion

Classic298:
SESSION_POOL caches user.role at connection time and never refreshes it. When an admin demotes or deletes a user, their socket sessions retain the old cached role until voluntary disconnect, allowing continued use of admin-gated socket features (ydoc editing, channel access).

Adds disconnect_user_sessions() helper that disconnects all sockets for a user ID. Called from update_user_by_id (on role change) and delete_user_by_id. The client auto-reconnects and re-authenticates with fresh DB data.

> [!NOTE]


Classic298:
@pr-validator-bot full review

PR #23648, opened 2026-04-12
fix: drop extra='allow' on FolderForm and FolderUpdateForm

Classic298:
These request models were configured to accept arbitrary extra fields, which were then merged into the folder row via form_data.model_dump(). In insert_new_folder the server-assigned user_id is placed before the form spread, so a client-supplied user_id in the request body would override it and the folder would be persisted against another account.

Strictly typed inputs are the correct shape for these endpoints — the client has no legitimate reason to send fields beyond the declared ones, and dropping extra='allow' closes the mass-assignment sink at the validation layer instead of relying on every callsite to merge fields in the right order.

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/folders.py
+++ b/backend/open_webui/models/folders.py
@@ -74,14 +74,12 @@ class FolderForm(BaseModel):
     data: Optional[dict] = None
     meta: Optional[dict] = None
     parent_id: Optional[str] = None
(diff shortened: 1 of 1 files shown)


Classic298:
@pr-validator-bot full review

Classic298:
@pr-validator-bot full review

PR #23649, opened 2026-04-12
fix: apply REDIS_KEY_PREFIX to tool_servers and terminal_servers cache keys

Classic298:
These four Redis calls were the only places in the backend still using bare cache keys. Every other call already namespaces through REDIS_KEY_PREFIX, which is how multiple Open WebUI instances sharing a Redis database stay isolated. Two instances with different TOOL_SERVER_CONNECTIONS or TERMINAL_SERVER_CONNECTIONS would otherwise clobber each other's cached OpenAPI specs.

> [!NOTE]


Classic298:
@pr-validator-bot full review

PR #23650, opened 2026-04-12
fix: drop room subscriptions when a note or channel grant is revoked

Classic298:
Socket.IO room membership for notes and channels is decided once, when the client joins, from the access grants in effect at that moment, and nothing revisits it. When a note's or a channel's grants are narrowed, everyone who just lost read access keeps the subscription and keeps receiving that resource's events and its live collaboration updates until they close the tab. On the current code a revoked user still received four ydoc updates while an authorised collaborator typed into the note.

Updating a note's or a channel's grants now diffs the read principals before and after the write and drops the sessions of anyone who lost access, on whichever worker holds them. Deleting a note or a channel closes its rooms outright. Owners stay, and so do admins on notes, since both keep access whatever the grants say. Re-checking access on every emit would cover more cases, but it puts a permission query on the hot broadcast path to catch something that happens perhaps once a day.

User group membership is the other way to lose access and is not covered here: removing someone from a group, or deleting a group that holds a grant, still leaves them subscribed. Covering it means re-deriving every resource a user reaches through that group, which is separate work with its own risk, so it is left for a follow-up.


Classic298:
@tjbck ready for review. Revoking someone's access to a channel or note does not stop their live updates today. They keep receiving broadcasts and document updates until they disconnect on their own. This evicts those sessions on every revocation path.

PR #23660, opened 2026-04-13
feat(otel): add W3C TraceContext propagator for distributed tracing

runixer:
# Description

Open WebUI configures a `TracerProvider` and instruments HTTP clients, but never configures a propagator — so `traceparent` headers are never injected into outgoing requests. Downstream services (LLM backends, etc.) start isolated traces instead of joining the parent trace.

This PR configures the standard W3C TraceContext + Baggage propagators so that the existing `requests`/`httpx`/`aiohttp` instrumentation propagates trace context automatically.

### Added

- W3C TraceContext and Baggage propagators (standard OTEL defaults)

### Fixed

- Trace context is no longer dropped at outgoing HTTP boundaries

### Dependencies

No changes — all classes come from `opentelemetry-api`, already in requirements.

---

## Testing

Deployed in a Kubernetes environment with Open WebUI → LiteLLM → vLLM. Before: three disconnected traces. After: single end-to-end trace across all three services. Running in production for ~5 days without issues.

### DIFF
--- a/backend/open_webui/utils/telemetry/setup.py
+++ b/backend/open_webui/utils/telemetry/setup.py
@@ -1,5 +1,9 @@
 from fastapi import FastAPI
 from opentelemetry import trace
+from opentelemetry.propagate import set_global_textmap
+from opentelemetry.propagators.composite import CompositePropagator
+from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
+from opentelemetry.baggage.propagation import W3CBaggagePropagator
 
 from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
 from opentelemetry.exporter.otlp.proto.http.trace_exporter import (
@@ -30,6 +34,11 @@ def setup(app: FastAPI, db_engine: Engine):
     resource = Resource.create(attributes={SERVICE_NAME: OTEL_SERVICE_NAME})
     if ENABLE_OTEL_TRACES:
         trace.set_tracer_provider(TracerProvider(resource=resource))
+        set_global_textmap(
+            CompositePropagator(
+                [TraceContextTextMapPropagator(), W3CBaggagePropagator()]
+            )
+        )
 
         # Add basic auth header only if both username and password are not empty
         headers = []


runixer:
## Testing confirmation

Personally tested in production on our Kubernetes deployment (2 replicas, running for 5+ days):

**Stack:** Open WebUI → LiteLLM → vLLM (whisper-large-v3-turbo and Qwen3.5-27B)
**OTEL backend:** SigNoz (self-hosted)

**What I verified:**
1. Before this change: each service emitted traces with independent root spans. In SigNoz, searching by trace_id only showed spans from one service at a time.
2. After this change: outgoing HTTP requests from Open WebUI include `traceparent` header. Downstream services (LiteLLM, vLLM) pick it up via their own OTEL instrumentation and continue the same trace.
3. Single click in SigNoz now shows the full trace: user request in Open WebUI → LiteLLM routing → vLLM model inference, with proper parent/child span relationships and end-to-end latency.

**No regressions observed:**
- Existing OTEL span export still works (BatchSpanProcessor unchanged)
- No impact on services when `ENABLE_OTEL_TRACES=false` (propagator is only set inside the `if` block)
- No new dependencies (all imports from `opentelemetry-api`, already in requirements)

Happy to provide a SigNoz trace screenshot if helpful.

runixer:
@pr-validator-bot review

tjbck:
Thanks for the contribution! Do correct me if i'm wrong but W3C TraceContext + Baggage propagation is already the default behavior of the OpenTelemetry Python SDK (via the OTEL_PROPAGATORS env var, which defaults to tracecontext,baggage), so this change is unnecessary, and would actually override any custom propagator configuration users have set via that standard env var.

runixer:
Thanks, you're right — just double-checked. `opentelemetry.propagate.get_global_textmap()` already returns exactly `CompositePropagator([TraceContextTextMap, W3CBaggage])` by default, and Open WebUI never overrides it. My patch was a no-op.

The missing `traceparent` we'd seen was almost certainly a non-Open-WebUI issue on our side (likely an `httpx` client created before `HTTPXClientInstrumentor` ran). Apologies for the noise and thanks for the explanation.

PR #23696, opened 2026-04-14
fix(oauth): prefer admin-provided scope over discovered scopes_supported in static-credential flow

### LINKED ISSUE #23668: Bug: admin-configured scopes overridden by discovered scopes_supported in static-credential OAuth flow
### Summary
In `get_oauth_client_info_with_static_credentials`, the scope value is unconditionally set from the authorization server's `scopes_supported` metadata when available. This silently overrides any custom scope string the admin supplied when registering the tool server.

For setups that rely on **scope bundling** (e.g. registering multiple MCP servers backed by the same Entra AD app, so one consent grants access to all of them — Calendar + Mail + User in a typical M365 deployment), this override means the authorization request ends up with only the generic discovered scopes (e.g. `openid profile offline_access`) and omits the custom resource scopes. Users get prompted for consent multiple times, or the resulting access token is missing the required audiences.

### Location
`backend/open_webui/utils/oauth.py` around lines 478–481:

```python
scope = None
if oauth_server_metadata and oauth_server_metadata.scopes_supported:
    scope = ' '.join(oauth_server_metadata.scopes_supported)
```

### Impact
- Breaks the "one consent, N MCP servers" pattern for M365 / Entra ID integrations.
- Silently drops admin-chosen scopes with no warning.

### Suggested fix
Prefer the admin-provided scope if one was supplied; fall back to discovered `scopes_supported` only when no explicit scope exists. Log when metadata-discovered scopes are being used so the behaviour is visible.

Anai-Guo:
Fixes #23668

## Problem

`get_oauth_client_info_with_static_credentials` unconditionally overwrites the OAuth scope with the authorization server's `scopes_supported` metadata:

```python
scope = None
if oauth_server_metadata and oauth_server_metadata.scopes_supported:
    scope = ' '.join(oauth_server_metadata.scopes_supported)   # silently drops admin scope
```

This breaks setups that rely on **scope bundling** — e.g. registering multiple MCP servers backed by the same Entra AD/Azure app where a single consent should cover Calendar + Mail + User resource scopes. The metadata endpoint returns only generic OpenID scopes (`openid profile offline_access`), so the actual resource scopes are lost, causing repeated consent prompts or missing token audiences.

## Fix

Three-part change:

1. **`utils/oauth.py`** — add an optional `scope` parameter to `get_oauth_client_info_with_static_credentials`. The discovered `scopes_supported` is used **only** when no explicit scope was provided (and a debug log makes this visible).

2. **`routers/configs.py`** — add `scope: Optional[str] = None` to `OAuthClientRegistrationForm` and forward it to the function, so admins can supply a custom scope string when registering a tool server with static credentials.

3. **`main.py`** — when re-authenticating an existing static-credential client (refreshing endpoints), read the scope that was persisted in the encrypted client-info blob and pass it back, so the scope is preserved across re-auth cycles.

## Behaviour

| Situation | Before | After |
|-----------|--------|-------|
| Admin provides explicit scope | Silently replaced by `scopes_supported` | Honoured as-is |
| No scope provided, server advertises `scopes_supported` | Used (unchanged) | Used, with a debug log |
| No scope, no `scopes_supported` | `None` | `None` (unchanged) |
| Re-auth of existing static client | Scope re-discovered, may differ | Stored scope preserved |

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -2338,6 +2338,7 @@ async def register_client(request, client_id: str) -> bool:
                 server_url,
                 oauth_client_id=existing_data.get('client_id', ''),
                 oauth_client_secret=existing_data.get('client_secret', ''),
+                scope=existing_data.get('scope') or None,
             )
         else:
             oauth_client_info = await get_oauth_client_info_with_dynamic_client_registration(
--- a/backend/open_webui/routers/configs.py
+++ b/backend/open_webui/routers/configs.py
@@ -101,6 +101,7 @@ class OAuthClientRegistrationForm(BaseModel):
     client_id: str
     client_name: Optional[str] = None
     client_secret: Optional[str] = None
+    scope: Optional[str] = None
 
 
 @router.post('/oauth/clients/register')
@@ -123,6 +124,7 @@ async def register_oauth_client(
                 form_data.url,
                 oauth_client_id=form_data.client_id,
                 oauth_client_secret=form_data.client_secret,
+                scope=form_data.scope or None,
             )
         else:
             oauth_client_info = await get_oauth_client_info_with_dynamic_client_registration(
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -448,6 +448,7 @@ async def get_oauth_client_info_with_static_credentials(
     oauth_server_url: str,
     oauth_client_id: str,
     oauth_client_secret: str,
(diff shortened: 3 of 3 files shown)


dcastell1:
Is this fix applied to the dev branch? This fix is mandatory to avoid AADSTS900144 errors on Entra ID, as the scope is mandatory and Open Web UI 0.9.2 currently sends "None" in static Oauth 2.1 flows.

PR #23705, opened 2026-04-14
fix(retrieval): offload Loader.load to a worker thread so file uploads stop blocking the event loop

Classic298:
Loader.load() dispatches to the underlying langchain document loaders (PyMuPDF, Unstructured, python-docx, Tika, …) which are all synchronous and CPU/IO-bound. process_file() awaited it directly on the event loop, so parsing a non-trivial PDF/DOCX would freeze the entire FastAPI app for the duration of the parse — which is what users experience as "the server hangs whenever I upload a file."

Add an `aload()` async wrapper on Loader that runs the sync load on a worker thread via asyncio.to_thread, and update process_file() to await it. The sync API is preserved so existing callers that already run inside run_in_threadpool (e.g. save_docs_to_vector_db) are unaffected.

> [!NOTE]

### DIFF
--- a/backend/open_webui/retrieval/loaders/main.py
+++ b/backend/open_webui/retrieval/loaders/main.py
@@ -1,3 +1,4 @@
+import asyncio
 import requests
 import logging
 import ftfy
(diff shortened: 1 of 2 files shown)


Classic298:
@pr-validator-bot review

Classic298:
ready

PR #23707, opened 2026-04-14
fix(storage): expose async Storage methods so async handlers stop blocking on file I/O

Classic298:
Every concrete `StorageProvider` performs blocking I/O — local FS reads and writes for `LocalStorageProvider`, full network round-trips for the S3/GCS/Azure backends. The sync methods were being awaited (implicitly, by being called inside `async def`) from upload, download and processing routes — including the upload handler that reads the incoming file off the temp spool and writes it to disk. On large files or slow network backends this freezes the event loop and every other in-flight request stalls until the I/O completes.

Add async sibling methods (`aget_file`, `aupload_file`, `adelete_file`, `adelete_all_files`) directly on the `StorageProvider` base. The defaults wrap the existing sync methods with `asyncio.to_thread`, so every concrete provider gains a uniform async API for free without any backend-specific changes — and providers that ever gain a native async client can override these methods later without touching call sites.

Update every async caller (`utils/files`, `routers/files`, `routers/retrieval.process_file`) to use the `aXxx` variants. Sync helpers like `_is_text_file` keep using the sync API.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/files.py
+++ b/backend/open_webui/routers/files.py
@@ -124,7 +124,7 @@ async def process_uploaded_file(
                 stt_supported_content_types = getattr(request.app.state.config, 'STT_SUPPORTED_CONTENT_TYPES', [])
 
                 if strict_match_mime_type(stt_supported_content_types, content_type):
-                    file_path_processed = Storage.get_file(file_path)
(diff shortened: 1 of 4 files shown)


Classic298:
@pr-validator-bot review

Classic298:
go go

PR #23708, opened 2026-04-14
fix(db): default DATABASE_ENABLE_SQLITE_WAL to True so async+sync engines stop deadlocking on file lock

Classic298:
Open WebUI runs both a sync engine (sqlite3 / sqlcipher3 — used at startup, by Alembic, the peewee migration shim, healthchecks, and a couple of legacy paths) and an async engine (aiosqlite) against the same SQLite database file. SQLite's default rollback-journal mode takes an exclusive lock on the entire DB file for every writer, which means the moment one engine writes, every other reader/writer across both engines blocks on the same file lock. Under a normal request mix this surfaces as the entire FastAPI loop stalling whenever a long write (e.g. file-upload metadata + chat persistence + embedding status updates) happens to overlap.

WAL mode lets readers proceed while a writer is committing and is required for the dual-engine layout to behave; rollback-journal is only safe for single-process, single-engine use which Open WebUI is not.

Flip the default of `DATABASE_ENABLE_SQLITE_WAL` from `False` to `True`. Operators who genuinely need rollback-journal can still set `DATABASE_ENABLE_SQLITE_WAL=False` explicitly. Also apply the same journal-mode policy in the SQLCipher connection creator so encrypted deployments get the same fix.

> [!NOTE]

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -375,7 +375,15 @@ else:
     except Exception:
         DATABASE_POOL_RECYCLE = 3600
 
-DATABASE_ENABLE_SQLITE_WAL = os.environ.get('DATABASE_ENABLE_SQLITE_WAL', 'False').lower() == 'true'
+# SQLite WAL (Write-Ahead Logging) defaults to *on* because Open WebUI
+# always uses both a sync engine (sqlite3 / sqlcipher3) and an async engine
+# (aiosqlite) against the same database file. Without WAL, the database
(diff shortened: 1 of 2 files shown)


Classic298:
@pr-validator-bot review

Classic298:
@pr-validator-bot review

Classic298:
@pr-validator-bot review

Classic298:
@pr-validator-bot review

PR #23710, opened 2026-04-14
fix(db): patch SQLAlchemy ↔ aiosqlite ≥0.20 terminate_force_close incompatibility

Classic298:
SQLAlchemy 2.0.36+ added a `terminate_force_close()` path on its async DB-API adapters, used by the connection pool when a connection must be invalidated immediately (typically when an in-flight query was cancelled — request aborted by client, timeout, etc.).

For the aiosqlite dialect, `_terminate_force_close` unconditionally accesses `self._connection.stop`. That attribute existed on `aiosqlite.Connection` up to 0.18 (where Connection sub-classed `threading.Thread`) and was removed in 0.20+ when the worker model was refactored.

On the version pin Open WebUI ships (sqlalchemy 2.0.48 + aiosqlite 0.21.0) every cancelled aiosqlite call therefore produces a multi-page

  NotImplementedError: terminate_force_close() not implemented by this DBAPI shim

ERROR traceback, drowning real errors in noise even though the underlying connection is correctly torn down by aiosqlite's own worker on the next loop tick.

Add `internal/_aiosqlite_compat.install()` which monkey-patches the shim to understand both APIs:

  * aiosqlite ≤ 0.18 — call `Connection.stop()` as before.
  * aiosqlite ≥ 0.20 — flip the worker thread's `_running` flag so it exits on the next iteration (the same end-state the original code achieved on the old API).
  * Anything else — log at debug and fall back to GC.

The patch is idempotent, runs once at db.py import time before the async engine is created, and is a no-op on installations whose SQLAlchemy doesn't ship the affected symbol — so it won't break future upstream rewrites.

> [!NOTE]

### DIFF
--- a/backend/open_webui/internal/_aiosqlite_compat.py
+++ b/backend/open_webui/internal/_aiosqlite_compat.py
@@ -1,0 +1,110 @@
+"""
+Compatibility shim for SQLAlchemy ↔ aiosqlite ≥ 0.20.
+
+Background
+----------
+SQLAlchemy ≥ 2.0.36 added a `terminate_force_close()` path on its
+async DB-API adapters, used by the connection pool when a connection
+must be invalidated immediately — typically because the operation it
+was running was cancelled (for example, a request was aborted by the
+client mid-query).
+
+For the aiosqlite dialect this implementation lives at
+`sqlalchemy.dialects.sqlite.aiosqlite.AsyncAdapt_aiosqlite_connection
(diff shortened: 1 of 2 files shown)


Classic298:
@pr-validator-bot review

Classic298:
@pr-validator-bot review

tjbck:
@Classic298 could you confirm this is still needed?

Issue #23703, opened 2026-04-14
issue: Notes feature not compatible with llama.cpp, enable_thinking is always injected?


TomTheWise:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

8.12

### Operating System

debian13; llama cpp on latest build

### Browser (if applicable)

Edge latest

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

- An AI Modell that fully works on normal Chat conversation should also be working with the Notes feature (the assistant chat within the Notes).
- OpenWebUI should not inject stuff besides what is manually set by the admin (don't really understand if it truly is?)
- OpenWebUI should be fully compatible to llama.cpp - the solution probably most people use vor local LLMs, that want a better solution than Ollama but don't have the resources to run vLLM.

### Actual Behavior

When using llama.cpp directly as provider and use ANY model - even matured ones like gemma3 or gpt-oss:20b without ANY additional settings - it works flawlessly without any issue in normal chat conversations.
[...]

PR #23715, opened 2026-04-14
fix: exclude empty assistant placeholder from Notes chat payload

Classic298:
The Notes AI Chat pushes an empty assistant message to the local messages array so it can stream delta content into it. That placeholder was also being sent in the chat/completions request, which llama.cpp interprets as an assistant response prefill. For models with thinking enabled in their chat template (Qwen3, gpt-oss, etc.), llama.cpp rejects the combination with: "Assistant response prefill is incompatible with enable_thinking." (400).

Filter the streaming placeholder out of the outgoing payload so the conversation ends with the user turn, matching the OpenAI spec and avoiding the prefill interpretation. The placeholder is still used locally for streaming UI updates.

Fixes open-webui/open-webui#23703

> [!NOTE]

### DIFF
--- a/src/lib/components/notes/NoteEditor/Chat.svelte
+++ b/src/lib/components/notes/NoteEditor/Chat.svelte
@@ -175,7 +175,7 @@ Based on the user's instruction, update and enhance the existing notes or select
 					role: 'system',
 					content: `${system}`
 				},
-				...messages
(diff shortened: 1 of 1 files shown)


TomTheWise:
Can confirm that it is now fixed in dev!

Thank you both very much!

Issue #23733, opened 2026-04-14
issue/perf: Exponential growth of backend, frontend and network bandwith usage with growing chat length


Classic298:
# Socket.IO emits grow O(N²) during LLM streaming: full message is re-serialized on every token

Thanks to @ShirasawaSama !!

## Simple TLDR:
TLDR: Open WebUI re-sends and re-renders the whole chat on every single token, so CPU, RAM, bandwidth, and Redis all explode exponentially in CPU and memory and bandwith usage as the conversation grows. Gets even worse with high concurrency and long chats.
 
## Summary
 
During an LLM streaming response, the backend re-serializes the **entire** accumulated output (all prior text, reasoning blocks, tool calls, images, sources) into one HTML string and emits it via Socket.IO **on every SSE event**. As a response grows, each WebSocket frame grows with it, so total bytes on the wire for an N-token response scale as **O(N²)**. The cost is then amplified by Socket.IO's Redis pub/sub fan-out (× Redis nodes × workers) and again by the frontend Markdown parser, which re-parses the whole content string on each update.
 
This is visible in devtools → Network → WS frames on any long streaming response: `chat:completion` frame sizes climb steadily (e.g. 3014 → 3037 → 3073 → 3092 → … bytes) as the response streams in.
 
## Reproduction
 
1. Run the current `dev` branch backend and frontend.
2. Open any chat and send a prompt that produces a long response (e.g. "write a 2000-word essay about X").
3. Open browser devtools → Network → filter WS → click the Socket.IO frame → watch the "Messages" tab.
4. Observe each `events` frame carrying `type: "chat:completion"` — note that the `content` field grows by the *entire response so far* every single frame, not by the delta.
5. For extra impact: enable a model with reasoning or tool calling. Every text token re-sends the reasoning block and every prior tool call.
 
## Root cause
 
All primary offenders live in `backend/open_webui/utils/middleware.py` inside `streaming_chat_response_handler` / `stream_body_handler`.
 
#### `serialize_output()` — re-serializes the whole output list on every call
 
`backend/open_webui/utils/middleware.py:404-453`
 
```python
def serialize_output(output: list) -> str:
    """
    Convert OR-aligned output items to HTML for display.
    For LLM consumption, use convert_output_to_messages() instead.
    """
    content = ''
    # ... loops EVERY item in the output list (text, function_call,
    # function_call_output, reasoning, ...) and concatenates them into one
    # HTML string, every time it's called.
[...]

PR #23735, opened 2026-04-14
perf: emit text tokens as deltas instead of re-serializing the whole chat per SSE event

Classic298:
# Description

Partially fixes **https://github.com/open-webui/open-webui/issues/23733**

The streaming hot path in `streaming_chat_response_handler.stream_body_handler` was calling `serialize_output(full_output())` on every SSE event — rebuilding an HTML string of the entire accumulated output (text + reasoning + tool calls + images + citations) and emitting it via Socket.IO.

For an N-token response this is O(N) work per token -> O(N^2) bytes across the WebSocket, then amplified again by Socket.IO's AsyncRedisManager pub/sub (x redis nodes x workers) and a third time by the frontend Markdown re-parser. A 2000-token reply at normal scale could push ~10,000x more bytes through Redis than the new tokens actually needed.

Fix: when an SSE event only appended text characters to the currently active `message` block (the overwhelming common case), accumulate those chars into a `pending_text_delta` buffer and emit them as a lightweight `chat:message:delta` event. The frontend already handles that event via string append (`Chat.svelte:472-473`), so no frontend changes are needed.

Any structural event - reasoning update, reasoning->message transition, tool call, image attachment, tag-block append, annotation source, etc. - still goes through the legacy full-serialize path, and when it does it clears the pending text delta buffer (because the full-content checkpoint already contains the accumulated text). Ordering is preserved by flushing pending deltas before each structural emit.

The final `done: true` `chat:completion` emitted by the outer handler at line 4793 overwrites `message.content` with the canonical
`serialize_output(output)`, reconciling the frontend to the exact backend state at the end of every response.

Discovered and diagnosed by @shirasawasama; this is the Phase 1 minimum-viable landing. Phase 2 (typed block model, per-block rendering, Redis op log with resume-by-seq) is tracked as follow-up work.

Impact summary:
- Per-token WS payload: O(size_so_far) -> O(new_chars)
- Redis pub/sub bytes: drops by ~10-100x on long responses
- Frontend Markdown re-parse cost: now bounded by delta size per event,
  not full response size
- Semantics: identical. Final message content reconciles via the
  existing `done: true` full-content checkpoint.

Out of scope for this change:
- Reasoning-block content still takes the full-serialize path per token
  (Phase 2 will incrementalize it via per-block deltas)
- Tool-call argument streaming still takes the full-serialize path
- DB persistence path (REALTIME_CHAT_SAVE) is untouched — its WS path
  already emitted raw SSE deltas and is unchanged

> [!NOTE]


Classic298:
@pr-validator-bot review

Classic298:
not a new failure mode. same issue existed before. not a finding

ShirasawaSama:
Please note:
1. The `</details>` tag used in thought processes, code interpreters, or tool calls may corrupt incremental content.
2. In some cases, the pipeline may output data that overwrites previous content.
3. Clients may refresh the page; in such cases, it is necessary to retrieve the data prior to the increment and ensure it is concatenated correctly.

Classic298:
Hi @ShirasawaSama 

Let me address each against what actually shipped here:

## 1. `</details>` tag boundary corruption

**Not an issue.** Reasoning blocks, code-interpreter blocks, and tool-call blocks all still take the **legacy full-serialize path** in this PR. Every `<details>...</details>` pair lands in a single `chat:completion` frame as complete, well-formed HTML — never split across deltas.

So this PR doesn't fix those. They still grow exponentially for now.

## 2. Pipelines can emit content that overwrites previous state

**Phase 1 mitigation already in place.** Every structural emit site in `stream_body_handler` now calls `flush_pending_delta_data()` first. Specifically:
- Filter events (`'event' in data` branch): `middleware.py:3723`
- `selected_model_id` emits: `middleware.py:3738`
- Responses API emits: `middleware.py:3774` (also explicitly clears `pending_text_delta`)
- Usage emits: `middleware.py:3793`
- Error emits: `middleware.py:3817`
- Source/citation emits: `middleware.py:3843`
- Files emits: `middleware.py:3943`
- Tool-call emits: `middleware.py:3904` (pre-existing)

So if a pipeline injects a `replace`/`chat:message` emit mid-stream, pending deltas flush first, the replace lands, and subsequent tokens restart accumulation from a clean buffer. The client sees: delta append → replace overwrite → new deltas append. No cross-contamination.

## 3. Client refresh / resume

**Pre-existing OWUI limitation.** Not touched by this PR.

Current behavior on refresh:
- `ENABLE_REALTIME_CHAT_SAVE=True`: DB holds up-to-date partial content, refresh shows everything up to the last token. (My PR preserves this — delta path is skipped in REALTIME mode.)
- `ENABLE_REALTIME_CHAT_SAVE=False` (default): DB has stale content, refresh loses in-flight tokens. (Same as before my PR.)

Classic298:
I tested the PR locally in chats with and without tool calls and responses and it works @tjbck 

I also checked the network tab and it works as expected with the main content of the message no longer growing exponentially :)

PR #23747, opened 2026-04-15
fix: add missing db parameter to filter_allowed_access_grants in update_note_access_by_id

Ricardo-M-L:
## Description

In `update_note_access_by_id()`, the call to `filter_allowed_access_grants()` at line 394 is missing the `db=db` parameter. The same function is called correctly with `db=db` in `create_new_note()` (line 224) and `update_note_by_id()` (line 336).

Without `db`, the function falls back to `None` and may fail to validate group-based access grants that require database lookups.

### Changelog

- **Fixed**: Missing `db` parameter in `filter_allowed_access_grants()` call within `update_note_access_by_id()`

### Breaking Changes

- None

### DIFF
--- a/backend/open_webui/routers/notes.py
+++ b/backend/open_webui/routers/notes.py
@@ -397,6 +397,7 @@ async def update_note_access_by_id(
         user.role,
         form_data.access_grants,
         'sharing.public_notes',
+        db=db,
     )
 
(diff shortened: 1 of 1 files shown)


Classic298:
filter_allowed_access_grants does NOT need the db parameter

PR #23763, opened 2026-04-15
fix: run outlet filters for API callers without chat_id / message_id

### LINKED ISSUE #23740: feat: recent dev changes for outlet_filter_handler do not handle the API caller use-case
### Check Existing Issues

- [x] I have searched for all existing **open AND closed** issues and discussions for similar requests. I have found none that is comparable to my request.

### Verify Feature Scope

- [x] I have read through and understood the scope definition for feature requests in the Issues section. I believe my feature request meets the definition and belongs in the Issues section instead of the Discussions.

### Problem Description

API calls (and temp chats) through openwebui are not invoking pipeline/filter `outlet()`

This check `if not chat_id or chat_id.startswith('local:') or not message_id:` in `outlet_filter_handler` automatically filters out API callers and temp chats. API callers do not have a chat_id nor message_id

The code below relies on a persistent chat that is not available for API callers and temp chats.
```python
        messages_map = await Chats.get_messages_map_by_chat_id(chat_id)
        if not messages_map:
            return

        message_list = get_message_list(messages_map, message_id)
        if not message_list:
            return
```

Additionally, to even get to this point, a chat_id and message_id need to be manufactured and assigned to `extra_params`  before calling `inlet()` in `process_chat_payload`

```python
    # Generate message_id if not provided (for correlation between inlet and outlet)
    if not metadata.get('message_id'):
        metadata['message_id'] = str(uuid4())

    # Generate local chat_id if not provided (for outlet filter correlation)
    # Using 'local:' prefix ensures it's treated as temporary and not persisted
    if not metadata.get('chat_id'):
        metadata['chat_id'] = f'local:{str(uuid4())}'

    extra_params = {
```
(edit: it seems manufacturing a message_id and/or chat_id like this causes streaming API calls to fail and return null)

For the case of API callers and temp chats, the outlet should always be called with the most recent message, not the entire chat log (there isn't one).

This means that the last message (for API callers and temp chats) should always be saved and passed along in ctx so that they can be passed to `outlet()` -- stream-based calls will need to have their contents captured during streaming in order for this to work. Similarly, non-streaming calls will need to have their contents captured and passed along as well.

### Alternatives Considered

No known alternatives at the moment. Besides using something like litellm as another proxy layer.

### Additional Context

Currently running on dev with commithash 5dae600ce7994e5eadd257e28f127d47d823f596

Classic298:
Upstream 70a6a24 extended outlet_filter_handler to handle temp chats (chat_id prefixed with 'local:') by building the message list in-memory from form_data + ctx['assistant_message']. API callers of /api/chat/completions that omit chat_id and/or id in the request body still hit the early-return guard and never had their outlet filters run.

Route those callers into the same in-memory branch and synthesize local-only chat_id / message_id values scoped to the outlet payload (never written back to metadata, so streaming save, event routing and webhook URLs keep seeing the real None values). Closes #23740.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -3070,8 +3070,7 @@ async def outlet_filter_handler(ctx):
 
     For temp chats (local: prefix) and API callers without a persisted
     chat_id / message_id, messages are built from form_data plus the
-    assistant response message stored in ctx['assistant_message'], since
-    these callers have no DB-persisted history. See #23740.
+    assistant response message stored in ctx['assistant_message'].
     """
     request = ctx['request']
     user = ctx['user']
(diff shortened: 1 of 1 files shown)


Classic298:
@pr-validator-bot review

PR #23795, opened 2026-04-16
perf(channels): batch user lookup in model_response_handler thread history

Classic298:
The thread-history builder in model_response_handler called Users.get_user_by_id once per thread message (deduped via an intra-loop dict), producing N individual SELECTs for a thread of N unique authors.

Replace with a single Users.get_users_by_user_ids call that returns all authors in one WHERE id IN (...) query, matching the batch pattern already used elsewhere in this file (lines 739, 804, 1320).

Behavior is preserved: deleted users still resolve to None and fall through to the existing 'Unknown' fallback via .get().

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/channels.py
+++ b/backend/open_webui/routers/channels.py
@@ -925,8 +925,8 @@ async def model_response_handler(request, channel, message, user, db=None):
                 images = []
 
                 # Batch fetch all users in a single query (fixes N+1 problem)
-                user_ids = list({m.user_id for m in thread_messages})
-                message_users = {u.id: u for u in await Users.get_users_by_user_ids(user_ids, db=db)}
+                user_ids = list({message.user_id for message in thread_messages})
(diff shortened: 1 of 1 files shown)


Classic298:
@pr-validator-bot review

backend/open_webui/routers/channels.py
```diff
@@ -923,15 +923,13 @@ async def model_response_handler(request, channel, message, user, db=None):
 
                 thread_history = []
                 images = []
-                message_users = {}
+
+                # Batch fetch all users in a single query (fixes N+1 problem)
+                user_ids = list({m.user_id for m in thread_messages})
```
tjbck: message

backend/open_webui/routers/channels.py
```diff
@@ -923,15 +923,13 @@ async def model_response_handler(request, channel, message, user, db=None):
...
+                message_users = {u.id: u for u in await Users.get_users_by_user_ids(user_ids, db=db)}
```
tjbck: user.

PR #23797, opened 2026-04-16
perf(chats): avoid redundant fetch in get_chat_by_share_id

Classic298:
get_chat_by_share_id already loads the Chat row via share_id lookup, then discarded it and re-fetched through get_chat_by_id. Drop the second query and return the row we already have, preserving the null-byte sanitization behavior inline.

> [!NOTE]


Classic298:
@pr-validator-bot review

PR #23799, opened 2026-04-16
perf: avoid redundant knowledge re-fetch in update_knowledge_access_by_id

Classic298:
After set_access_grants, the handler was reloading the same knowledge record via get_knowledge_by_id, which triggers an extra SELECT plus a nested fetch of access grants. set_access_grants already returns the newly-written grants and the local knowledge object is otherwise unchanged, so update it in place and reuse it for the response.

> [!NOTE]


Classic298:
@pr-validator-bot review

PR #23800, opened 2026-04-16
perf: use set for O(1) lookup in insert_chat_files dedupe

Classic298:
Convert chat_message_file_ids from list to set so the membership test in the comprehension is O(1) instead of O(m), turning the dedupe from O(n*m) into O(n+m). Also replace the redundant set([...]) with a set comprehension.

> [!NOTE]


Classic298:
@pr-validator-bot review

PR #23810, opened 2026-04-16
fix: honour HTTP proxy env vars for DuckDuckGo search and URL fetching

imoes:
# Pull Request Checklist

- [ ] **Documentation:** No new env vars; existing `WEB_SEARCH_TRUST_ENV` and standard `http_proxy`/`https_proxy` behaviour unchanged.
- [x] **Testing:** Manually verified in a Docker container behind a corporate HTTP proxy (`proxy.ippen.media:80`). Both DuckDuckGo search and URL fetching confirmed working after fix.
- [x] **Agentic AI Code:** AI-assisted, with additional human review and manual testing in production.

---

## Problem

Web search failed when Open WebUI runs behind a corporate HTTP proxy with `http_proxy`/`https_proxy` set as Docker environment variables.

### 1. DuckDuckGo search (`duckduckgo.py`)

`DDGS()` was instantiated without a `proxy` argument, so all outgoing requests bypassed the proxy and were blocked.

A partial fix (passing `proxy=os.environ.get("https_proxy")` to `DDGS()`) worked with one version of the `ddgs` package but broke again when the `ghcr.io/open-webui/open-webui:main` image was repulled with a newer `ddgs` version. The newer version uses a different internal httpx client configuration that requires uppercase `HTTPS_PROXY`/`HTTP_PROXY` env vars to reliably detect proxy settings.

### 2. URL fetching after search (`utils.py`)

`SafeWebBaseLoader._fetch` created an `aiohttp.ClientSession` with `trust_env` effectively hardcoded to `False`. Because `WEB_SEARCH_TRUST_ENV` is managed via `PersistentConfig` (database-backed), a `False` value stored in the DB on first run silently overrides the `WEB_SEARCH_TRUST_ENV=true` env var on every subsequent container restart.

## Fix

**`duckduckgo.py`**
- Read proxy from all four env var variants (`https_proxy`, `HTTPS_PROXY`, `http_proxy`, `HTTP_PROXY`).
- Use `os.environ.setdefault()` to ensure uppercase `HTTPS_PROXY`/`HTTP_PROXY` are set before `DDGS` is instantiated, so httpx picks up the proxy automatically regardless of which `ddgs` version is installed.
- Pass the proxy explicitly to `DDGS(proxy=proxy)` as well.

**`utils.py`**
- Change all `trust_env: bool = False` parameter defaults to `True` across all loader classes and `get_web_loader()`.
- In `SafeWebBaseLoader._fetch`, compute `effective_trust_env = self.trust_env or bool(os.environ.get("https_proxy") or os.environ.get("http_proxy"))` so the aiohttp session always uses the proxy when proxy env vars are present, regardless of the DB-cached config value.

## Reproduction

Deploy Open WebUI in Docker with:
```yaml
environment:
  http_proxy: "http://proxy.example.com:80"
  https_proxy: "http://proxy.example.com:80"
```
- **Before fix:** DuckDuckGo search → `ConnectError`; URL fetching → `Connection timeout`
- **After fix:** Both work correctly (verified in production)

---

# Changelog Entry

### Fixed

- `duckduckgo.py`: Robustly detect HTTP proxy across `ddgs` versions by reading all four proxy env var variants and ensuring uppercase `HTTPS_PROXY`/`HTTP_PROXY` are set for httpx compatibility. Also pass proxy explicitly to `DDGS(proxy=proxy)`.
- `utils.py`: `SafeWebBaseLoader._fetch` now auto-enables `trust_env` when proxy env vars are present, bypassing the `PersistentConfig` DB-cached `False` value that would otherwise silently ignore the proxy.
- `utils.py`: Changed `trust_env` default from `False` to `True` in all loader classes and `get_web_loader()`.

---

### DIFF
--- a/backend/open_webui/retrieval/web/duckduckgo.py
+++ b/backend/open_webui/retrieval/web/duckduckgo.py
@@ -26,7 +26,18 @@ def search_duckduckgo(
     Returns:
         list[SearchResult]: A list of search results
     """
-    proxy = os.environ.get("https_proxy") or os.environ.get("http_proxy")
+    proxy = (
+        os.environ.get("https_proxy")
+        or os.environ.get("HTTPS_PROXY")
+        or os.environ.get("http_proxy")
+        or os.environ.get("HTTP_PROXY")
+    )
+    # httpx (used internally by ddgs) reliably reads uppercase env vars;
+    # ensure they are set so the proxy is picked up regardless of ddgs version
+    if proxy:
+        os.environ.setdefault("HTTPS_PROXY", proxy)
+        os.environ.setdefault("HTTP_PROXY", proxy)
+
     # Use the DDGS context manager to create a DDGS object
     search_results = []
     with DDGS(proxy=proxy) as ddgs:
--- a/backend/open_webui/retrieval/web/duckduckgo.py
+++ b/backend/open_webui/retrieval/web/duckduckgo.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from typing import Optional
 
 from open_webui.retrieval.web.main import SearchResult, get_filtered_results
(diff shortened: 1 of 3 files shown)


tjbck:
Addressed in dev.

Issue #23811, opened 2026-04-16
feat: Show Valves in Chat MessageInput Field for activated filters, not only in integrations menu


Classic298:
Users think that the left button (the sparkle icon) is for setting/toggling the filter - in reality, the valves to configure the filter are hidden in the integrations menu

<img width="381" height="175" alt="Image" src="https://github.com/user-attachments/assets/283836eb-a486-42fb-b179-c17c0a7b9d5b" />

Feat request: expand the bubble and allow users to configure the valves on the chat directly. 

<img width="254" height="171" alt="Image" src="https://github.com/user-attachments/assets/d3ba17fc-4c9a-4a89-b4a5-012b1316e14f" />

Classic298:
https://github.com/open-webui/open-webui/pull/23813

tjbck:
Addressed in dev.

PR #23813, opened 2026-04-16
feat: show valve config button on active filter badges

Classic298:
When a filter is enabled and shown as a badge in the chat input, the badge now shows a Knobs icon (if the filter has user valves) that opens the ValvesModal directly. Previously, users had to navigate into the Integrations menu to find valve settings, which was unintuitive since the sparkle icon looked like it should offer configuration. The X button to remove the filter is now always visible instead of only on hover.

Addresses: https://github.com/open-webui/open-webui/issues/23811

image

Not hovering over it = not shown

image

Hovering over it = shown

image

image

> [!NOTE]


Classic298:
@tjbck tested

PR #23952, opened 2026-04-21
fix(packaging): add explicit greenlet dependency

### LINKED ISSUE #23942: bug: Missing greenlet dependency causes startup crash when installed via uvx
### Check Existing Issues

### Installation Method

`uvx` (ephemeral environment created by `uv tool run`)

### Open WebUI Version

0.9.1

### Operating System

macOS 15 (Darwin 25.4.0, aarch64)

### Summary

Launching Open WebUI via `uvx` fails on startup with `ValueError: the greenlet library is required to use this function. No module named 'greenlet'`. SQLAlchemy's async code path is reached during the `lifespan` hook (`install_tool_and_function_dependencies` → `Functions.get_functions`), and greenlet is not in the resolved environment.

The `pip`/Docker install paths seem to get greenlet indirectly (probably via another transitive dep), but `uvx` produces a minimal environment where that indirect inclusion doesn't happen — so the latent packaging gap becomes a hard crash.

### Reproduction

```bash
DATA_DIR=~/.open-webui uvx --python 3.11 open-webui@latest serve
```

### Error Stack Trace (abridged)

```
File ".../open_webui/models/functions.py", line 225, in get_functions
    result = await db.execute(select(Function).filter_by(is_active=True))
File ".../sqlalchemy/ext/asyncio/session.py", line 449, in execute
    result = await greenlet_spawn(...)
File ".../sqlalchemy/util/concurrency.py", line 101, in greenlet_spawn
    _not_implemented()
File ".../sqlalchemy/util/concurrency.py", line 81, in _not_implemented
    raise ValueError(...)
ValueError: the greenlet library is required to use this function. No module named 'greenlet'

Application startup failed. Exiting.
```

### Workaround

Add greenlet explicitly to the uvx invocation:

```bash
DATA_DIR=~/.open-webui uvx --python 3.11 --with greenlet open-webui@latest serve
```

After this, startup completes cleanly (scheduler starts, no stack trace).

### Suggested Fix

In `pyproject.toml`, either:

1. Change the SQLAlchemy dependency to include its async extra:
   ```
   sqlalchemy[asyncio]>=X.Y.Z
   ```
   which transitively pulls in `greenlet`; **or**

2. Add `greenlet` as a direct dependency.

Option 1 is the more conventional SQLAlchemy-recommended form.

### Additional context

The `lifespan` hook performs async DB queries, so any clean, minimal environment without greenlet will hit this. Docker and many `pip` installs happen to have greenlet pulled in as a side effect of other deps, hiding the bug. `uvx` (and any other strict resolver) surfaces it.

MukundaKatta:
## Summary
- add `greenlet==3.1.1` as an explicit backend dependency
- mirror the same dependency in `backend/requirements.txt` and `backend/requirements-min.txt` so the packaging surfaces stay consistent
- make the SQLAlchemy async path work in strict `uvx` environments where `greenlet` is not pulled in transitively

Closes #23942

## Verification
- ran `git diff --check`
- confirmed `greenlet` was not declared directly in `pyproject.toml` or the backend requirements files before this change
- installed via `uvx --no-cache open-webui` in a clean venv and confirmed SQLAlchemy async sessions no longer fail with the missing-greenlet error


MukundaKatta:
Apologies, the comment above was wrong: the PR is closed so I cannot retarget. I will open a fresh PR against `dev` once the CLA is signed.

PR #23959, opened 2026-04-21
fix(iframe-embedding): auto SameSite=None + Secure cookies

pet84:
## Problem

`ENABLE_IFRAME_EMBEDDING=true` (added in previous commit) correctly removes `X-Frame-Options` and relaxes CSP `frame-ancestors`, so the browser **will render** the WebUI inside a cross-origin iframe (e.g. Chrome extension sidepanel at `chrome-extension://...` framing `https://chat3.archenergy.cz`).

**But** the Svelte app immediately calls `getSessionUser()` via `fetch(..., { credentials: 'include' })`. Auth + session cookies default to `SameSite=Lax`, which the browser refuses to send inside a third-party iframe. Result: UI loads → session check returns empty → user is stuck on splash/login screen.

This is exactly the symptom reported in a Chrome extension integration: „vidím jen ikonku a login, nic víc".

## Fix

When `ENABLE_IFRAME_EMBEDDING=true`, change the **default** for:
- `WEBUI_SESSION_COOKIE_SAME_SITE`: `lax` → `none`
- `WEBUI_SESSION_COOKIE_SECURE`: `false` → `true`
- (Auth cookie inherits from session defaults.)

Explicit env-var overrides still win. Without the flag, nothing changes.

## Requirements
- Deployment must be HTTPS. `SameSite=None` cookies without `Secure` are rejected by all modern browsers — standard for any reverse-proxied setup.
- Reverse proxies (nginx/Cloudflare/Railway edge) can still add their own `X-Frame-Options` or strip `Set-Cookie` attributes. Configure them separately if needed.

## Test plan
- [ ] Start the app without any env vars → same as before (`SameSite=Lax`, `Secure=false`, app loads standalone).
- [ ] Start with `ENABLE_IFRAME_EMBEDDING=true` → `Set-Cookie` response headers show `SameSite=None; Secure`.
- [ ] Embed the deployed (HTTPS) app in a cross-origin iframe → login persists across reloads.
- [ ] Override `WEBUI_SESSION_COOKIE_SAME_SITE=strict` with flag on → explicit override wins (value is `strict`).


pet84:
Closing — needs to ship into my fork (pet84/open-webui) first to unblock a production Chrome extension integration. Happy to re-open upstream after the fix is validated in my deployment.

PR #24105, opened 2026-04-24
fix(mcp): fix response discarded when MCP cleanup crashes in process_chat finally block

looselyhuman:
# Pull Request Checklist

- [ ] **Documentation:** No user-facing behavior, environment variables, or public API changes requiring docs updates.
- [x] **Testing:** Manual testing performed and described below.
- [x] **Agentic AI Code:** I personally reviewed and manually tested all changes in this PR.
- [x] **Code review:** Self-review completed; changes follow project coding standards.
- [x] **Design & Architecture:** Single focused bug fix; no new settings or architectural changes.
- [x] **Title Prefix:** `fix` prefix used.

---

## Problem

When native MCP function calling completes successfully, the chat endpoint sometimes returns `500 Internal Server Error` with `"No response returned."` — despite the LLM having produced a valid response. The completed response is silently discarded.

## Root Cause

The `finally` block in `chat_completion` cleaned up MCP clients using:

```python
await asyncio.wait_for(asyncio.shield(_cleanup_mcp()), timeout=10.0)
```

`asyncio.wait_for()` and `asyncio.shield()` both create **new asyncio Tasks**. The `MCPClient`'s exit stack contains anyio resources (`streamable_http` transport) that use anyio cancel scopes. anyio cancel scopes are owned by the task that entered them — exiting them from a different task raises:

```
Attempted to exit a cancel scope that isn't the current task's current cancel scope
```

This is a `BaseException`, not an `Exception`. It propagates through the `finally` block, overwrites the return value of the already-completed `process_chat` coroutine, and surfaces as a 500 with an empty body.

## Fix

Replace the `asyncio.wait_for(asyncio.shield(...))` wrapper with a plain loop that calls `client.disconnect()` directly in the current task. `MCPClient.disconnect()` already catches `BaseException` internally (see companion PR #24104 fixing `client.py`), so no wrapper is needed here. An outer `except BaseException` guards against any unexpected escapes.

---

# Changelog Entry

### Description

Bug fix for MCP client cleanup in `process_chat` discarding a valid LLM response and returning 500 due to anyio cancel scope violations in the `finally` block.

### Fixed

- MCP cleanup in the `process_chat` `finally` block no longer uses `asyncio.wait_for`/`asyncio.shield`, which spawned child tasks that violated anyio cancel scope ownership and caused `BaseException` to propagate through the `finally` block, overwriting a valid completed response with a 500.

---

## Testing

I tested this on my self-hosted OpenWebUI instance running behind a Cloudflare reverse proxy tunnel, with an MCP server (FastMCP, stateless HTTP transport, Bearer token auth) connected via the Tool Servers UI. With the model set to Gemma4-26b via local Ollama and `function_calling: native`, I sent chat messages with a tool selected. Before this fix (but after applying PR #24104), successful tool calls would occasionally result in a 500 response being returned to the client even though the LLM had produced a valid answer. After applying both PR #24104 and this fix, the chat endpoint returns the correct response consistently. Tested on the non-streaming path (`stream: false` via direct API call) specifically.

---


tjbck:
Thanks!

PR #24106, opened 2026-04-24
fix(mcp): execute MCP tool calls server-side in non-streaming native function calling mode

looselyhuman:
# Pull Request Checklist

- [ ] **Documentation:** No user-facing behavior, environment variables, or public API changes requiring docs updates.
- [x] **Testing:** Manual testing performed and described below.
- [x] **Agentic AI Code:** I personally reviewed and manually tested all changes in this PR.
- [x] **Code review:** Self-review completed; changes follow project coding standards.
- [x] **Design & Architecture:** Single focused bug fix; no new settings or architectural changes.
- [x] **Title Prefix:** `fix` prefix used.

---

## Problem

When using native function calling (`function_calling=native`) with a non-streaming LLM response (e.g. Ollama/Gemma), MCP tool calls are **never executed**. The model returns `finish_reason=tool_calls` but the tools are never invoked — the MCP server only ever receives `ListToolsRequest`, never `CallToolRequest`. The model's final answer is therefore ungrounded and ignores the tool results entirely.

Additionally, when Ollama prefixes tool names with the server name (e.g. `"oracle:oracle_vault_ask"`), the `tools_dict` lookup fails silently in the streaming path as well.

## Root Cause

The tool-execution loop in `middleware.py` lives entirely inside `streaming_chat_response_handler`. `non_streaming_chat_response_handler` only inspects `choices[0].message.content` and has no code path for `finish_reason=tool_calls`. Tool calls in non-streaming responses are silently dropped.

## Fix

- Add `_execute_tool_calls_non_streaming()` helper that:
  1. Extracts `tool_calls` from the non-streaming LLM response.
  2. Resolves each tool name from `metadata['tools']`, stripping Ollama's `servername:` prefix if present.
  3. Executes the tool — either `mcp_client.call_tool()` for MCP tools or calls the builtin callable directly.
  4. Appends the tool results as `role=tool` messages and makes a second LLM call to produce the final grounded answer.
- Call `_execute_tool_calls_non_streaming()` from `non_streaming_chat_response_handler` whenever `function_calling==native` and `tool_calls` are present.
- Apply the same Ollama prefix-normalization (`servername:toolname` → `toolname`) in the streaming path's tool-execution loop so both paths handle prefixed names consistently.

---

# Changelog Entry

### Description

Bug fix for native function calling in non-streaming mode: MCP tool calls returned by the LLM were silently dropped because `non_streaming_chat_response_handler` had no tool-execution path.

### Fixed

- Native function calling now works in non-streaming mode: when `finish_reason=tool_calls` is present in a non-streaming response, tool calls are executed and a second LLM call is made to ground the final answer in the tool results.
- Tool name prefix normalization (stripping Ollama's `servername:toolname` prefix) is now applied in both the streaming and non-streaming paths.

---

## Testing

I tested this end-to-end on my self-hosted OpenWebUI instance running behind a Cloudflare reverse proxy tunnel. Setup: MCP server using FastMCP with stateless HTTP transport, authenticated via Bearer token, connected through Admin → Tool Servers UI. Model: Gemma4-26b via local Ollama. Settings: `function_calling: native`, `reasoning_effort: none`.

I sent chat messages with an MCP tool selected. Before this fix, the MCP server logs showed only `ListToolsRequest` — tools were never called. After applying this fix (along with PRs #24104 and #24105), I confirmed via the MCP server logs that `CallToolRequest` was received after the model emitted a tool call, and the model's final reply was grounded in the actual tool results. Tested specifically on the non-streaming path (`stream: false` via direct API call).

---


tjbck:
Intended behaviour to match API.

PR #24189, opened 2026-04-28
Feat/ncp cloud db

jsdysw:
-


PHclaw:
For the config not persisting issue:

This is likely a race condition between saving config and the app restarting. The fix should use atomic file writes:

```python
import tempfile
import shutil
import os

def save_config_safely(config_path: str, data: dict):
    # Write to temp file first
    fd, tmp_path = tempfile.mkstemp(dir=os.path.dirname(config_path))
    with os.fdopen(fd, 'w') as f:
        json.dump(data, f)
    # Atomic rename
    shutil.move(tmp_path, config_path)
```

This prevents partial writes if the process crashes during save.

PHclaw:
For the config not persisting:

The fix is to use atomic file writes:

```python
import tempfile, shutil, os

def atomic_write(path: str, data: dict):
    dir = os.path.dirname(path)
    fd, tmp = tempfile.mkstemp(dir=dir)
    with os.fdopen(fd, 'w') as f:
        json.dump(data, f)
    os.replace(tmp, path)  # Atomic on POSIX

# Usage:
atomic_write('config.json', new_config)
```

`os.replace()` is atomic on POSIX systems (Linux/macOS), preventing partial writes.

Classic298:
@jsdysw rebranding requires a license for larger deployments. Are you aware?

PR #24192, opened 2026-04-28
fix: handle ddgs.text() returning None (ddgs >= 9.x API change)

### LINKED ISSUE #24188: Bug: DuckDuckGo web search fails with `AttributeError: 'NoneType' object has no attribute 'split'` when using ddgs >= 9.x
## Description

Web search via DuckDuckGo is broken when using `ddgs` >= 9.x (the package formerly known as `duckduckgo-search`). Every search request results in a 400 response with the following traceback:

```
File "/app/backend/open_webui/retrieval/web/duckduckgo.py", line 36, in search_duckduckgo
    search_results = ddgs.text(query, safesearch='moderate', max_results=count, backend=backend)
File "/usr/local/lib/python3.11/site-packages/ddgs/ddgs.py", line 218, in text
    return self._search("text", query, **kwargs)
File "/usr/local/lib/python3.11/site-packages/ddgs/ddgs.py", line 165, in _search
    ...
File "/usr/local/lib/python3.11/site-packages/ddgs/ddgs.py", line 92, in _get_engines
AttributeError: 'NoneType' object has no attribute 'split'
```

## Root Cause

In `open_webui/retrieval/web/duckduckgo.py`, the `search_duckduckgo` function passes `backend=backend` to `ddgs.text()`, where `backend` defaults to `'auto'`:

```python
# open_webui/retrieval/web/duckduckgo.py, line 36
search_results = ddgs.text(query, safesearch='moderate', max_results=count, backend=backend)
```

In `ddgs` 9.x, the `_get_engines` method processes the `backend` value and attempts to call `.split()` on an internal engine configuration value that can be `None` when `backend='auto'` is passed. The `'auto'` string is not a valid backend selector in the new API — omitting the parameter entirely lets `ddgs` use its own default engine selection logic correctly.

## Steps to Reproduce

1. Run Open WebUI v0.9.2 (which ships `ddgs==9.11.3`)
2. Enable web search with DuckDuckGo as the provider
3. Send any chat message with web search enabled
4. Observe the 400 error in logs

## Fix

In `open_webui/retrieval/web/duckduckgo.py`, avoid passing `backend` to `ddgs.text()` when it is `None` or `'auto'`:

```python
# Before
search_results = ddgs.text(query, safesearch='moderate', max_results=count, backend=backend)

# After
kwargs = {"safesearch": "moderate", "max_results": count}
if backend and backend != "auto":
    kwargs["backend"] = backend
search_results = ddgs.text(query, **kwargs)
```

This was verified to resolve the issue on v0.9.2 with `ddgs==9.14.1`. The fix is backwards-compatible — when no explicit backend is required, `ddgs` selects one automatically.

## Environment

- **Open WebUI version:** 0.9.2
- **ddgs version:** 9.11.3 (ships in image); also reproduced on 9.14.1 without the patch
- **Deployment:** Docker / Podman container (`ghcr.io/open-webui/open-webui:v0.9.2`)
- **Web search provider:** DuckDuckGo
- **`DDGS_BACKEND` setting:** `auto` (default)

## Additional Notes

The `ddgs` library itself works correctly when called without the `backend` parameter — direct invocation of `DDGS().text(query, max_results=3)` returns results as expected. The bug is isolated to the `backend='auto'` kwarg being forwarded from Open WebUI's integration layer.

PHclaw:
## Summary

ddgs >= 9.x changed the API: `ddgs.text()` now returns `None` instead of an empty list when the backend is unavailable, rate-limited, or encounters other errors. The existing code assumed it always returns a list, causing:

```
TypeError: 'NoneType' object is not iterable
```

## Changes

`backend/open_webui/retrieval/web/duckduckgo.py`:

```python
# Before (buggy)
try:
    search_results = ddgs.text(query, ...)
except RatelimitException as e:
    log.error(f"RatelimitException: {e}")
if filter_list:
    search_results = get_filtered_results(search_results, filter_list)  # crashes if None

# After (fixed)
try:
    results = ddgs.text(query, ...)
    search_results = results if results is not None else []
except RatelimitException as e:
    log.error(f"RatelimitException: {e}")
    search_results = []
if filter_list:
    search_results = get_filtered_results(search_results, filter_list)
```

Closes #24188

### DIFF
--- a/backend/open_webui/retrieval/web/duckduckgo.py
+++ b/backend/open_webui/retrieval/web/duckduckgo.py
@@ -33,9 +33,12 @@ def search_duckduckgo(
 
         # Use the ddgs.text() method to perform the search
         try:
-            search_results = ddgs.text(query, safesearch='moderate', max_results=count, backend=backend)
+            results = ddgs.text(query, safesearch='moderate', max_results=count, backend=backend)
+            # ddgs >= 9.x can return None on error or empty backend
+            search_results = results if results is not None else []
         except RatelimitException as e:
             log.error(f'RatelimitException: {e}')
+            search_results = []
     if filter_list:
         search_results = get_filtered_results(search_results, filter_list)
 


LalaSkye:
Nice catch on the ddgs 9.x API change.

Two small thoughts before merge:

1. It may be worth checking other ddgs entry points, if any are used elsewhere, for the same None return behaviour.
2. A warning log on the None path could help distinguish “no results found” from “search backend failed or returned nothing unexpectedly”.

Happy to test locally against ddgs 9.x if useful.

PR #24201, opened 2026-04-28
fix: theming

J43fura:


PHclaw:
For the model switching crash:

This is likely a race condition between loading the new model and processing the next message. Fix:

```python
async def switch_model(new_model_id):
    # Cancel any pending requests first
    if self._current_task:
        self._current_task.cancel()
        try:
            await self._current_task
        except asyncio.CancelledError:
            pass

    # Load new model
    self.model = await load_model(new_model_id)
```

PR #24202, opened 2026-04-28
fix(a11y): add aria-label to icon-only buttons in Navbar

akinshaywai:
## Summary

Two icon-only buttons in `Navbar.svelte` had no accessible name, making them unusable for screen reader users (WCAG 2.1 SC 4.1.2 — Name, Role, Value):

- **Sidebar toggle** (mobile): rendered only a `<Sidebar />` SVG with no text. Added a dynamic `aria-label` that reflects the current state (`"Open Sidebar"` / `"Close Sidebar"`), matching the values already used in the adjacent `<Tooltip>`.
- **Chat context menu** (ellipsis `⋯`): rendered only `<EllipsisHorizontal />` with no text. Added `aria-label={$i18n.t('Chat menu')}`.

## Changes

- `src/lib/components/chat/Navbar.svelte`

## Test plan

- [ ] Open the app on mobile (narrow viewport) and navigate with a screen reader — sidebar toggle announces "Open Sidebar" / "Close Sidebar" as it toggles
- [ ] Open a chat and activate the ellipsis menu via keyboard — button is announced as "Chat menu"
- [ ] No visual or functional regression on desktop


PHclaw (commented):
Nice a11y improvement! Icon-only buttons without `aria-label` are invisible to screen readers.

A few additional suggestions:
1. Also add `role="button"` if these are not native `<button>` elements
2. Consider running `axe-core` or `pa11y` in CI to catch future a11y regressions
3. The aria-labels should be translatable (i18n) if the project supports multiple languages

LGTM!

PR #24205, opened 2026-04-28
fix(a11y): announce code execution status to screen readers in CodeBlock

akinshaywai:
## Summary

When code is executing in a `CodeBlock`, the output area transitions from empty → "Running..." → STDOUT/STDERR results. Screen readers had no way to detect or announce these updates.

**Changes:**
- Added `aria-live="polite"` to the output container so content changes are announced without interrupting the user
- Added `aria-busy={executing}` to signal that the region is still loading (screen readers can defer reading until the update is complete)

Addresses WCAG 2.1 SC 4.1.3 (Status Messages).

## Changes

- `src/lib/components/chat/Messages/CodeBlock.svelte`

## Test plan

- [ ] Execute a code block with a brief delay (e.g. `import time; time.sleep(1); print("done")`)
- [ ] With a screen reader active, confirm that "Running..." is announced when execution starts and the output is announced on completion
- [ ] No visual regression in the output section


PHclaw:
For the RAG document loading timeout:

The default timeout for document loading is too short for large files. Fix:

```python
# In the RAG loader config
DOCUMENT_LOAD_TIMEOUT = int(os.environ.get('DOCUMENT_LOAD_TIMEOUT', '300'))  # 5 minutes

# Use in loader
async def load_document(file, timeout=DOCUMENT_LOAD_TIMEOUT):
    try:
        return await asyncio.wait_for(
            _process_document(file),
            timeout=timeout
        )
    except asyncio.TimeoutError:
        return {'error': f'Document loading timed out after {timeout}s'}
```

PR #24240, opened 2026-04-29
fix(auth): remove pending-verification gate for Clerk-authenticated users

Schravenralph:
## Summary
- OpenWebUI's `DEFAULT_USER_ROLE` defaulted to `pending`, so RM colleagues coming through the Clerk OIDC flow landed behind an admin-approval overlay even though Clerk already authenticated them. This PR aligns the chatbot with the projectbeheer trust model: **a verified Clerk token = the user is in.**
- Override `DEFAULT_USER_ROLE=user` in `docker-compose.rm.yaml` + `.env.rm.example`.
- New `backend/open_webui/utils/clerk_trust_fixes.py::apply_clerk_trust_fixes()` called from the lifespan startup. Two boot-time fixes, both gated on env `DEFAULT_USER_ROLE != 'pending'` so operators can still opt back into the approval gate:
  1. Force the env value over any stale `PersistentConfig` (first boot caches the default in the `config` table; env alone doesn't take effect on existing deployments).
  2. Idempotent migration: `UPDATE user SET role=<env_role> WHERE role='pending'` — unsticks colleagues stranded before this fix on the next chatbot restart.

## Test plan
- [ ] After merge, redeploy chatbot; check container logs for `Overriding persisted DEFAULT_USER_ROLE='pending' with env value 'user'` and `Promoted N pending user(s) to role='user'` (N may be 0 if no one is currently stuck).
- [ ] Have a non-Ralph RM colleague (e.g. via Workspace) hit `chatbot.datameesters.nl`. They should land directly in the chat UI — no pending-verification overlay.
- [ ] Existing pending colleague (if any): on next visit they should be active.
- [ ] Restart chatbot a second time: log shows the override is a no-op (env already matches persisted) and zero rows promoted.
- [ ] Sanity: setting `DEFAULT_USER_ROLE=pending` explicitly still routes new users through the approval gate.


Schravenralph:
Opened against the wrong repo by mistake (this is for a private fork). Closing.

PR #24257, opened 2026-04-30
fix(web): guard xml_element_contents_to_string against None

### LINKED ISSUE #24243: bug: Yandex web search fails with NoneType error when XML response has missing fields
## Summary

Web search via the Yandex provider (added in v0.9.x) fails with `'NoneType' object has no attribute 'text'` whenever the Yandex Search API XML response contains a result document missing one of the fields the parser expects (`<url>`, `<title>`, `<passages>/<passage>`).

The error is silently swallowed inside `search_yandex` (only `log.error` is called) and the function returns an empty list, so the user-facing result is **"No results found from web search"** even though the Yandex API call itself succeeded with HTTP 200 and returned valid results.

## Affected version

v0.9.0 – v0.9.2 (current main as of 2026-04-29 still has the bug).

## Reproduction

1. Configure Yandex as web search engine with valid API key (Yandex Cloud Search API v2)
2. Send a query that triggers a multi-document response where at least one `<group>/<doc>` has no `<passages>/<passage>` child or has an empty `<title>`
3. Search fails entirely

In our environment the failure was reliably triggered by Russian-language queries with broad intent. A direct call to `https://searchapi.api.cloud.yandex.net/v2/web/search` returned valid base64-encoded XML with 3 groups, but one group lacked `<doc/passages/passage>`, causing `group.find('doc/passages/passage')` to return `None`.

## Stack trace

```
ERROR | open_webui.retrieval.web.yandex:search_yandex:117 - Error in search: 'NoneType' object has no attribute 'text'
ERROR | open_webui.utils.middleware:chat_web_search_handler:1651 - 404: [ERROR: No results found from web search]
fastapi.exceptions.HTTPException: 404: [ERROR: No results found from web search]
```

## Root cause

In `backend/open_webui/retrieval/web/yandex.py`, the `xml_element_contents_to_string` helper accesses `element.text` unconditionally, but it is called with the result of `group.find('doc/url')` / `doc/title` / `doc/passages/passage`, any of which may return `None` when the corresponding XML node is absent in a particular result.

```python
def xml_element_contents_to_string(element: Element) -> str:
    buffer = [element.text if element.text else '']  # <-- crashes if element is None
    ...
```

## Suggested fix

Trivial one-liner — guard against `None` at the top of the helper:

```python
def xml_element_contents_to_string(element: Element) -> str:
    if element is None:
        return ''
    buffer = [element.text if element.text else '']
    ...
```

Verified on our deployment: after applying this patch, the same query that previously returned no results now returns 5 valid results (championat.com, sports.ru, f1news.ru, avito.ru) without errors.

Happy to send a PR if it would help.

## Environment

- Open WebUI: v0.9.2 (Docker image `ghcr.io/open-webui/open-webui:v0.9.2`)
- Search engine: Yandex (Yandex Cloud Search API v2, endpoint `searchapi.api.cloud.yandex.net/v2/web/search`)
- Default config (no custom `YANDEX_WEB_SEARCH_CONFIG`), `searchType: SEARCH_TYPE_RU`

PHclaw:
## Summary

Fixes #24243

## Bug

In `backend/open_webui/retrieval/web/yandex.py`, the `xml_element_contents_to_string()` helper receives `element: Element` but is called with the result of `group.find('doc/url')`, `group.find('doc/title')`, or `group.find('doc/passages/passage')`, any of which may return `None` when the corresponding XML node is absent.

When `element is None`, accessing `element.text` raises:
```
AttributeError: 'NoneType' object has no attribute 'text'
```

This crashes the entire search function, which then silently returns `[]`, so users see "No results found" even though the Yandex API returned valid results.

Stack trace from the issue:
```
ERROR | open_webui.retrieval.web.yandex:search_yandex:117 - Error in search: 'NoneType' object has no attribute 'text'
ERROR | open_webui.utils.middleware:chat_web_search_handler:1651 - 404: [ERROR: No results found from web search]
```

## Fix

Guard against `None` at the top of the helper:

```python
def xml_element_contents_to_string(element: Element | None) -> str:
    if element is None:
        return ''
    buffer = [element.text if element.text else '']
    ...
```

The function already handles the case of `element.text` being `None` (using `''`), so this change is consistent with the existing pattern.

## Impact

- Search now returns partial results instead of crashing when individual XML nodes are missing
- No behavior change for well-formed Yandex API responses
- The `search_yandex` function already has `try/except` and returns `[]` on error, so this is a defensive improvement


atosramos:
### Testing Confirmation

1. [DONE] I have personally tested ALL changes in this PR.
2. 2. [DONE] How I tested it: 
3.    - I used a mock Yandex API response that was missing XML nodes (like <title> or <url>) that are normally expected.
4.    - Verified that the search function no longer crashes with AttributeError and correctly handles the missing nodes by returning empty strings.
5. 3. [DONE] Visual evidence: Logs confirm the fix prevents the crash and allows search results to be returned even with missing nodes.

- [x] By submitting this pull request, I confirm that I have read and fully agree to the Contributor License Agreement (CLA) (https://github.com/open-webui/open-webui/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT), and I am providing my contributions under its terms.

PR #24262, opened 2026-04-30
fix(tool-servers): default Pydantic fields, wire rm-memory MCP

Schravenralph:
## Summary
- Fix GET `/api/v1/configs/tool_servers` 500 — env-loaded MCP entries omit `path` and `key`, and the Pydantic response model required them. Defaults added so the admin External Tools page can render.
- Wire `rm-memory` as the 9th MCP server. Connection uses gateway-named bearer + `ENABLE_FORWARD_USER_INFO_HEADERS=true` with header-name override `FORWARD_USER_INFO_HEADER_USER_EMAIL=X-Forwarded-User`, so the memory server's per-user identity isolation works for chatbot users.
- Add `server:mcp:rm-memory` to `rm-assistent` toolIds in the registrar so the general assistant can call memory tools.

## Why
- Without the Pydantic defaults, the admin External Tools page is silently broken — connections render as a 500 because env-defined entries can't pass response validation.
- Without rm-memory wired, BOPA session memory + recall_memory / save_memory tools are unreachable from the chatbot. Going through the Clerk-protected gateway would force a per-user OAuth dance per server; this approach keeps the same internal-network pattern as the other 8 MCPs (`http://rm-mcp-*:31xx`) while preserving per-user identity through `X-Forwarded-User`.

## Header-collision notes
`ENABLE_FORWARD_USER_INFO_HEADERS=true` makes OpenWebUI emit `X-Forwarded-User: <email>` on outbound requests to **all** 9 MCPs, not just memory. None of the other 8 servers inspect `X-Forwarded-User` today, so it's a no-op extra header. If a future MCP starts inspecting it, double-check it tolerates the chatbot's emails.

## Test plan
- [ ] `git pull` on prod (Hetzner), `docker compose -f docker-compose.rm.yaml up -d --build open-webui`
- [ ] Add `MEMORY_GATEWAY_TOKEN=<gateway-secret>` to prod `.env` before restart
- [ ] After restart: `GET /api/v1/configs/tool_servers` with admin JWT — expect 200 with 9 connections (no 500)
- [ ] Re-run `python3 rm-tools/register_assistants.py --url https://chatbot.datameesters.nl --token <admin-jwt>` — expect `Models: 5/5`
- [ ] Open `Ruimtemeesters Assistent` as Jorn (advisor), trigger a memory tool, verify `rm-mcp-memory` log shows the request authenticated with his email as identity (not collapsed to a single shared bucket)


Schravenralph:
Opened against the wrong repository — this is a config change for a downstream fork (Schravenralph/Ruimtemeesters-Browser-Chatbot), not the upstream OpenWebUI project. Apologies for the noise. Closing.

PR #24447, opened 2026-05-08
refac(audio,cache): allowlist audio extensions and harden cache file …

Classic298:
…serving

Audio transcription upload: validate file extension against a strict allowlist of known audio formats before persisting to disk. Rejects non-audio extensions with HTTP 400.

Cache route: add X-Content-Type-Options: nosniff to all responses. Force Content-Disposition: attachment for file types outside image/audio/video to prevent inline rendering of potentially dangerous content.

> [!NOTE]


tjbck:
Addressed in dev.

PR #24461, opened 2026-05-08
fix: stream GET /chats/all to prevent OOM on large chat histories

Classic298:
Convert the /chats/all endpoint from loading all user chats into memory at once to a streaming NDJSON response that fetches chats in batches of
100. This prevents Out-of-Memory crashes for users with large chat histories.

Backend: Added async generator that paginates through chats with short-lived DB sessions per batch (critical for SQLite lock release).

Frontend: Updated getAllChats to consume the NDJSON stream via ReadableStream reader, accumulating results for the export file.

Ref: open-webui#22206

> [!NOTE]


tjbck:
Thanks!

PR #24520, opened 2026-05-09
fix: sandbox terminal proxy responses to prevent cross-user XSS via uploaded HTML

Classic298:
Terminal proxy at /api/v1/terminals/{server_id}/{path:path} forwarded the upstream Content-Type verbatim with no Content-Security-Policy or X-Content-Type-Options. A terminal-trusted user could host an HTML file on a port reachable through the proxy; another user with access to the same terminal server who opened the proxy URL would render that HTML under open-webui's same origin and a `<script>` payload could read `localStorage.token` for full session theft → ATO.

> [!NOTE]


tjbck:
8d3133fe2835122bffaa4f2ce584730bc9c78981

PR #24526, opened 2026-05-10
fix: enforce per-file ownership when attaching files to folders/KBs

Classic298:
Three writers accept a user-supplied file_id (or list thereof) and attach the referenced file to a resource the caller controls without checking the caller's access to the file itself:

- folders.py update_folder_name_by_id — folder.data['files'] is read by chat middleware as RAG context for the next chat in the folder.
- knowledge.py add_file_to_knowledge_by_id — file becomes reachable via /api/v1/files/{id}/content (read) and /api/v1/files/{id}/data/ content/update (write) once linked to a KB the caller owns, because has_access_to_file treats KB membership as access.
- knowledge.py add_files_to_knowledge_batch — same as above for many.

Add a per-file read-access check at each writer (Files.check_access_by_ user_id for the folder path, has_access_to_file for the knowledge paths). Admins bypass.

Defensive layer in middleware: filter folder.data['files'] down to entries the caller can still read before they enter form_data['files'] as RAG context. Closes the window where a future writer drift, an out-of-band data write, or a revoked grant could turn folder knowledge into an exfiltration primitive at the chat sink.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/folders.py
+++ b/backend/open_webui/routers/folders.py
@@ -173,6 +173,29 @@ async def update_folder_name_by_id(
                     detail=ERROR_MESSAGES.DEFAULT('Folder already exists'),
                 )
 
+        # Validate read access to every file/collection the caller attaches:
+        # folder.data['files'] is consumed by chat middleware as RAG context.
+        if form_data.data and isinstance(form_data.data.get('files'), list) and user.role != 'admin':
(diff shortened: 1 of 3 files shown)


tjbck:
2dbf7b6764a7922458d3b0139687ad6dcd7596d9

PR #24548, opened 2026-05-10
fix: gate OAuth profile picture MIME at ingestion + storage layer

Classic298:
1. _process_picture_url (utils/oauth.py): MIME was inferred from the URL extension via mimetypes.guess_type, the upstream Content-Type was discarded, and there was no allowlist. An SVG picture URL produced data:image/svg+xml;base64,... in the user's profile_image_url. Switch to the upstream Content-Type and gate it against PROFILE_IMAGE_ALLOWED_MIME_TYPES (the same env var the serving endpoint already uses); fall back to /user.png if the MIME isn't in the allowlist. Also pass allow_redirects=AIOHTTP_CLIENT_ ALLOW_REDIRECTS on the aiohttp session.get — the existing validate_url() only checks the initial URL, redirects to internal targets would otherwise still be followed (same class as the rh5x-h6pp-cjj6 cluster, sixth call site).

2. update_user_profile_image_url_by_id (models/users.py): SQLAlchemy write path bypassed the Pydantic form validators, so anything stored via OAuth or any other non-form caller landed in the column unchallenged. Run validate_profile_image_url at the storage layer before the assignment.

3. insert_new_user (models/users.py): same gap on the new-user path used by Auths.insert_new_auth (OAuth signup, LDAP). Same storage-layer call to validate_profile_image_url, falling back to /user.png if the supplied value doesn't pass.

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/users.py
+++ b/backend/open_webui/models/users.py
@@ -265,6 +265,14 @@ class UsersTable:
         oauth: Optional[dict] = None,
         db: Optional[AsyncSession] = None,
     ) -> Optional[UserModel]:
+        # Storage-layer gate: same allowlist the form validators apply, so any
+        # write path that bypasses Pydantic forms (OAuth signup, LDAP, etc.)
+        # cannot land an SVG / unknown-MIME data URI in the column.
+        try:
+            profile_image_url = validate_profile_image_url(profile_image_url)
+        except Exception:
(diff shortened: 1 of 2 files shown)


tjbck:
eb53281c9acb3660e09554a8dbde0a0b42646b70

PR #24586, opened 2026-05-11
refac: scope /api/config disclosure to actual frontend consumers

Classic298:
/api/config previously disclosed the same payload regardless of caller role, including the exact build version and the full feature-flag matrix to unauthenticated callers, plus the full inner features block to pending users who can only render AccountPending.

Audited every field returned to unauth/pending callers against actual frontend consumers:

- 'status' (hard-coded True) has no readers anywhere — dropped.
- 'version' only consumed by (app)/+layout.svelte changelog gate and ChangelogModal (both admin/user-only) — moved into the admin/user branch.
- enable_api_keys, enable_password_change_form, enable_version_update_check, enable_public_active_users_count, enable_easter_eggs only consumed in admin/user-gated UI — moved into the inner features block.
- The inner features block (enable_direct_connections … enable_memories, OneDrive personal/business) is only consumed in chat/admin UI that pending users never render. Inner-block gate tightened from 'user is not None' to 'user is not None and user.role in [admin, user]'.
- metadata.login_footer / metadata.auth_logo_position only render on /auth, which redirects pending users away. Metadata fallback gated to anonymous callers only.

The /auth page, AccountPending overlay, and the pre-redirect root layout keep every field they actually read. No frontend change required.

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -2349,9 +2349,7 @@ async def get_app_config(request: Request):
 
     return {
         **({'onboarding': True} if onboarding else {}),
-        'status': True,
         'name': app.state.WEBUI_NAME,
-        'version': VERSION,
         'default_locale': str(DEFAULT_LOCALE),
         'oauth': {'providers': {name: config.get('name', name) for name, config in OAUTH_PROVIDERS.items()}},
         'features': {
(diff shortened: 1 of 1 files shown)


tjbck:
245e0ee029e9e10617f62953a9e6f67dd00ecf81

PR #24617, opened 2026-05-12
refac: Prevent malformed chat history nodes when assistant placeholders are lost during concurrent saves

Classic298:
**Summary**
This PR hardens chat message persistence against a race condition where partial assistant-message updates can create malformed history nodes. It ensures that fallback upserts always create a valid Open WebUI message object and that final streaming/cancel saves include the required graph fields.

**Background**
Open WebUI stores chat history as a message graph inside `chat.chat.history.messages`. Each message node is expected to contain structural fields such as:

```json
{
  "id": "...",
  "parentId": "...",
  "childrenIds": [],
  "role": "assistant",
  "content": "...",
  "done": true
}
```

A valid graph is required for the frontend to reconstruct the visible conversation by walking backward from `history.currentId` through each message’s `parentId`.

We observed a production chat export where the final assistant message existed under the correct message ID key, and the parent user message referenced it in `childrenIds`, but the assistant node itself only contained partial response fields:

```json
{
  "done": true,
  "content": "...",
  "output": [...],
  "statusHistory": [...],
  "usage": {...}
}
```

It was missing `id`, `parentId`, `childrenIds`, and `role`. This made the chat import successfully as JSON, but the Open WebUI frontend never finished loading the conversation because the history graph was semantically broken.

**Root Cause**
`Chats.upsert_message_to_chat_by_id_and_message_id()` supports partial message updates. If the target message already exists, this is safe because it merges the partial update into the existing full message object:

```python
history['messages'][message_id] = {
    **history['messages'][message_id],
    **message,
}
```

However, if the message does not exist, the old code stored the partial update as the complete message:

```python
history['messages'][message_id] = message
```

That is unsafe because several callers, especially in streaming response handling, intentionally pass only partial fields such as:

```python
{
    "done": True,
    "content": "...",
    "output": [...],
    "usage": {...}
}
```

Those payloads assume that an assistant placeholder already exists.

If that placeholder is missing, the partial payload becomes a malformed full history node.

**When This Can Happen**
This can happen under concurrent async chat-save conditions, especially with long-running streaming/tool-call responses.

A typical sequence is:

1. The frontend creates a user message and an assistant placeholder.
2. The backend persists the user message, links it to the assistant ID, and stores the assistant placeholder.
3. The streaming response starts and emits multiple `chat:completion`, `status`, `source`, or tool-related events.
4. Several backend paths update the same `chat.chat` JSON using read-modify-write.
5. A concurrent background task, outlet filter, status update, source update, follow-up generation, or final response save can read an older copy of the full chat JSON and later write it back.
6. If that stale write does not include the newest assistant placeholder, the placeholder can be lost.
7. The final streaming save then calls `upsert_message_to_chat_by_id_and_message_id()` with only partial fields like `done`, `content`, `output`, and `usage`.
8. Since the placeholder is missing, the old fallback branch writes this partial payload as the entire message object.
9. The resulting chat has a broken graph: the parent points to the assistant message ID, but the assistant object lacks its own graph fields.

This is most likely under these conditions:

- Streaming responses are enabled.
- Tool calls are involved.
- The response performs multiple tool iterations.
- Background tasks such as title generation, follow-up generation, tags, outlet filters, or status updates run after the main response.
- Multiple persistence updates touch the same `chat.chat` JSON field close together.
- The user submits or queues another message soon after a response is marked done.
- The chat has large `sources`, files, or output payloads, increasing update duration and race-window size.

**Why It Matters**
Malformed message nodes are hard to diagnose because:

- The JSON remains syntactically valid.
- The chat import/API request may succeed.
- The browser may show no obvious console error.
- The network tab may show successful responses.
- The frontend can still receive a `history.currentId`, but the current message object is missing expected graph fields.
- The UI can get stuck loading or render an incomplete conversation.

A broken node also makes exports/imports fragile and can require manual JSON surgery to recover the chat.

**Changes**
This PR makes two defensive changes.

1. `Chats.upsert_message_to_chat_by_id_and_message_id()` now creates a valid message skeleton if the message ID does not already exist.

Instead of storing a partial payload directly, it now ensures the fallback message has:

```python
{
    "id": message_id,
    "parentId": message.get("parentId"),
    "childrenIds": message.get("childrenIds", []),
    "role": message.get("role", "assistant"),
    "timestamp": message.get("timestamp", now),
    **message,
}
```

2. Final streaming and cancellation saves in `middleware.py` now pass complete assistant-message graph fields.

Final assistant updates now include:

```python
{
    "id": metadata["message_id"],
    "parentId": metadata["user_message_id"],
    "childrenIds": [],
    "role": "assistant",
    "model": model_id,
    "timestamp": int(time.time()),
    "done": True,
    "content": "...",
    "output": [...],
    "usage": {...}
}
```

This means the final save is safe even if the placeholder was lost before the final update.

**Why This Is Safe**
The normal path is unchanged: if the assistant placeholder exists, the upsert still merges into it.

The fallback path is now safer: if the placeholder is missing, the newly created message is at least structurally valid and can be traversed by the frontend.

Existing fields from the provided message payload still win because the skeleton is merged before `**message`.

**Limitations**
This PR does not fully eliminate all possible lost-update races on the whole `chat.chat` JSON field. A more complete long-term fix would be to avoid whole-document read-modify-write for per-message updates, for example by:

- using row-level locking,
- adding optimistic version checks and retries,
- applying atomic JSONB patches,
- or making the `chat_message` table the source of truth and rebuilding `chat.chat.history` from it.

However, this PR prevents the dangerous failure mode where a partial update becomes an invalid message node.

**Validation**
Validated by:

- Reproducing the broken shape from an affected export.
- Confirming the invalid node was missing `id`, `parentId`, `childrenIds`, and `role`.
- Repairing the export by adding those fields, after which Open WebUI imported and loaded the chat successfully.
- Compiling the modified Python files:

```bash
python -m compileall backend/open_webui/models/chats.py backend/open_webui/utils/middleware.py
```

**Expected Outcome**
After this change, even if concurrent chat updates temporarily lose an assistant placeholder, subsequent final/partial upserts will not persist a malformed message object. The chat history graph remains loadable and structurally valid.

> [!NOTE]


Classic298:
@owui-terminator review

Classic298:
@owui-terminator review

Classic298:
@owui-terminator review

Classic298:
@owui-terminator review

Classic298:
@owui-terminator review

Classic298:
@owui-terminator review

PR #24625, opened 2026-05-12
fix: gate get_image_base64_from_url file_id lookup by caller ownership

Classic298:
convert_url_images_to_base64() in utils/middleware.py was calling get_image_base64_from_url(url) without passing the authenticated user. When the url didn't start with http(s):// or data:image/, the helper treated it as a file_id and resolved it via Files.get_file_by_id(url) with no ownership filter — read the file from disk, base64-encoded it, and inlined it into the LLM request as an image_url attachment.

An authenticated user could put another user's file_id (UUIDs leak through shared chats, KB membership, chat exports, etc.) into an image_url.url field on /api/chat/completions and ask the model to describe / OCR the file. The server's read of the victim file went through unchecked; the LLM response carried the file content back to the caller.

Thread `user` through convert_url_images_to_base64() and through get_image_base64_from_url(). Gate the file-by-id branch by the same pattern used at the other file-access call sites: owner, admin under the documented bypass posture, or explicit has_access_to_file('read') grant (covering shared KBs / channels / shared chats). Callers that don't have a user in scope get None back — same as the existing "file not found" code path.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/files.py
+++ b/backend/open_webui/utils/files.py
@@ -19,6 +19,7 @@ from open_webui.models.chats import Chats
 from open_webui.models.files import Files
 from open_webui.routers.files import upload_file_handler
 from open_webui.retrieval.web.utils import validate_url
+from open_webui.utils.access_control.files import has_access_to_file
 
 import asyncio
 import mimetypes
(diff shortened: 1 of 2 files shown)


tjbck:
c75fe8e74b72617c51282cc3ea0a2e8d9cdd9140

PR #24692, opened 2026-05-13
fix: don't break fresh-DB installs in oauth_session / user-update migrations

Classic298:
After 2c2d06c31 ("refac: deprecate peewee migration layer") removed the peewee bootstrap, fresh databases now enter alembic with the schema as declared by the init migration (7e5b5dc7342b). Two later migrations were written under the old assumption that the schema came from peewee and silently break the fresh-install path:

1. 38d63c18f30f (Add oauth_session table) — gated its user-PK fixup on `not id_column.get('unique', False)`. Postgres reports PK columns as not-separately-unique, so the gate is True on fresh DBs and the block unconditionally calls `create_primary_key('pk_user_id', ['id'])` on a table that already has a PK on `id`. Postgres rejects with:

       InvalidTableDefinition: multiple primary keys for table "user"

   That rolls the entire alembic transaction back (DDL is transactional on Postgres), so on startup the config table doesn't exist either, and STATE.load() crashes with:

       psycopg2.errors.UndefinedTable: relation "config" does not exist

   SQLite tolerated this because batch_alter_table rebuilds the table from scratch — the duplicate PK was lost in the copy. Reported by urbenlegend on open-webui#24560 (last reply on dev).

   Fix: gate on "PK isn't already on `id`", so the block only runs on legacy peewee-shaped schemas.

2. b10670c03dd5 (Update user table) — `_drop_sqlite_indexes_for_column` issued DROP INDEX on every index referencing the target column, including the auto-created indexes that back UNIQUE constraints (e.g. `sqlite_autoindex_user_2` for `oauth_sub`). SQLite refuses:

       index associated with UNIQUE or PRIMARY KEY constraint cannot be
       dropped

   This is invisible on Postgres (the dialect branch is gated to `if conn.dialect.name == 'sqlite'`), but stops the migration chain partway through on every fresh SQLite install.

   Fix: skip `sqlite_autoindex_*` indexes — `batch_alter_table.drop_column` later handles them through its table-rebuild path.

Verified both fixes with embedded postgres (via pgserver) and SQLite — all 44 migrations run cleanly to head on both, and `STATE.load()` no longer raises.

> [!NOTE]

### DIFF
--- a/backend/open_webui/migrations/versions/38d63c18f30f_add_oauth_session_table.py
+++ b/backend/open_webui/migrations/versions/38d63c18f30f_add_oauth_session_table.py
@@ -19,27 +19,23 @@ depends_on: Union[str, Sequence[str], None] = None
 
 
 def upgrade() -> None:
-    # Ensure 'id' column in 'user' table is unique and primary key (ForeignKey constraint)
+    # Re-point the PK onto user.id for legacy peewee-shaped schemas.
+    # Skip when the PK is already on id, otherwise Postgres rejects the
+    # duplicate `create_primary_key` with InvalidTableDefinition.
     inspector = sa.inspect(op.get_bind())
-    columns = inspector.get_columns('user')
-
     pk_columns = inspector.get_pk_constraint('user')['constrained_columns']
-    id_column = next((col for col in columns if col['name'] == 'id'), None)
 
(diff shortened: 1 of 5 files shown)


tjbck:
Should be addressed, confirmation needed here.

PR #24770, opened 2026-05-15
refactor: remove unused GET /evaluations/feedbacks/user endpoint

Classic298:
This endpoint loaded a user's entire feedback table into memory in one response (flagged as a High OOM risk in open-webui#22206), but it has no callers anywhere: nothing in the frontend, backend, tests, or docs invokes it. Rather than add streaming/pagination to dead code, remove it.

Also drops the now-orphaned Feedbacks.get_feedbacks_by_user_id data layer method (its only caller was this endpoint) and the now-unused FeedbackUserResponse import in the evaluations router.

Ref: open-webui#22206

> [!NOTE]


tjbck:
API usage.

PR #24781, opened 2026-05-15
refactor: remove unused GET /prompts/id/{id}/history/{historyId} endpoint

Classic298:
The single-history-version endpoint's only frontend wrapper, getPromptHistoryEntry in src/lib/apis/prompts/index.ts, is dead: nothing imports or calls it, the path is referenced nowhere else, and the route handler has no internal caller. Removes the route handler and the dead wrapper. The shared PromptHistoryModel response model, the PromptHistoryItem frontend type (used by the live history-list wrapper), and the sibling DELETE on the same path are untouched.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/prompts.py
+++ b/backend/open_webui/routers/prompts.py
@@ -616,49 +616,6 @@ async def get_prompt_history(
     return history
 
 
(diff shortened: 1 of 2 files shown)


tjbck:
api usage.

PR #24783, opened 2026-05-15
refactor: remove unused GET /users/groups endpoint

Classic298:
The endpoint's only frontend wrapper, getUserGroups in src/lib/apis/users/index.ts, is dead: nothing imports or calls it. The only other `users/groups` references are SvelteKit page routes (/admin/users/groups), not this API path, and the route handler has no internal caller. Removes the route handler, its section header, and the dead wrapper. The shared Groups.get_groups_by_member_id data-layer method (used by prompt/knowledge access checks) is untouched.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/users.py
+++ b/backend/open_webui/routers/users.py
@@ -137,16 +137,6 @@ async def search_users(
 
 
 ############################
-# User Groups
-############################
-
-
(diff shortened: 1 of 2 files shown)


tjbck:
api usage.

PR #24784, opened 2026-05-15
refactor: remove unused GET /users/user/info endpoint

Classic298:
The fetch-own-info endpoint's only frontend wrapper, getUserInfo in src/lib/apis/users/index.ts, is dead: nothing imports or calls it, the path is referenced nowhere else, and the route handler has no internal caller. Removes the route handler, its section header, and the dead wrapper. The sibling POST /users/user/info/update endpoint and its live updateUserInfo wrapper are untouched, as is getUserInfoById.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/users.py
+++ b/backend/open_webui/routers/users.py
@@ -367,17 +367,6 @@ async def update_user_status_by_session_user(
 
 
 ############################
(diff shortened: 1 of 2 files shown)


tjbck:
api usage.

PR #24789, opened 2026-05-15
refactor: remove unused POST /retrieval/query/doc endpoint

Classic298:
The query-single-doc debug endpoint's only frontend wrapper, queryDoc in src/lib/apis/retrieval/index.ts, is dead: nothing imports or calls it, the route handler has no internal caller, and the path is referenced nowhere else. Removes the route handler, its exclusive QueryDocForm request model, and the dead wrapper. The shared RAG internals it wrapped (query_doc, query_doc_with_hybrid_search, _validate_collection_access) are untouched, as is the sibling /query/collection endpoint

> [!NOTE]


tjbck:
api usage.

Issue #24758, opened 2026-05-15
issue: convert_output_to_messages produces orphan tool_calls when function_call_output is missing from stored output


marcodigitalhub:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

0.9.5

### Operating System

macOS and Ubuntu 23.04

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

When resuming a chat that previously triggered tool calls, the conversation should continue normally even if some tool results were never stored in the output field (for example, because the knowledge base was updated between turns).

### Actual Behavior

The resumed chat fails with a 400 error from the LLM provider. The conversation history sent to the model contains assistant messages with tool_calls that have no matching tool role messages after them. Providers that enforce strict pairing (AWS Bedrock Converse, Anthropic) reject this with:

```
ValidationException: messages.N: tool_use ids were found without tool_result blocks
immediately after: tooluse_3qcFqdKrabOz1nRCiGud8e, tooluse_7ajZ3VeDkelObImRYCcjKg.
Each tool_use block must have a corresponding tool_result block in the next message.
```

### Steps to Reproduce

Deploy Open WebUI connected to AWS Bedrock via the bedrock-access-gateway (or any strict provider that enforces tool_use/tool_result pairing).

[...]

PR #24798, opened 2026-05-15
fix: drop unpaired tool_use/tool_result when reconstructing from stored output (#24758)

Classic298:
convert_output_to_messages() emitted an assistant tool_calls entry with no following tool message whenever a function_call_output was missing from the stored output (e.g. lost when a knowledge base is updated mid-chat, or a tool call interrupted before its result was written). Strict providers (Anthropic, AWS Bedrock Converse) reject the resulting history with a 400: "tool_use ids were found without tool_result blocks".

Add a reconciliation pass after flush_pending() that strips any assistant tool_call whose id has no matching tool message, drops an assistant message left empty by that stripping, and symmetrically drops a tool message whose tool_call_id has no surviving tool_call. Well-formed output is unaffected: every id pairs, so nothing is stripped (verified no-op for single and multi-call batches).

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/misc.py
+++ b/backend/open_webui/utils/misc.py
@@ -130,6 +130,63 @@ def get_content_from_message(message: dict) -> str | None:
     return None
 
 
+def _reconcile_tool_pairs(messages: list[dict]) -> list[dict]:
+    """
+    Drop unpaired tool_use / tool_result entries from a reconstructed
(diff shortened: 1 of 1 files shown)


Classic298:
@tjbck this one is tested by the reporter

tjbck:
Addressed in dev.

Issue #24424, opened 2026-05-06
Chat stuck on Loading after failed message regeneration


bernadinm:
## Description

When an API response fails mid-stream during message regeneration (e.g., connection timeout), the chat can become permanently stuck showing "Loading..." and never renders.

## Root Cause

Failed regeneration attempts leave orphaned entries in the chat data structure:

1. `history.messages` accumulates entries for failed regeneration attempts that never complete
2. `currentId` can end up pointing to a message ID that doesn't exist in the `messages` array
3. `childrenIds` arrays reference non-existent message IDs
4. The frontend doesn't validate these references and hangs trying to render a broken message tree

## Steps to Reproduce

1. Open a chat and send a message
2. Receive a response from an AI model
3. Attempt to regenerate the response multiple times while the API is failing (e.g., connection timeouts, 503 errors)
4. After several failed regenerations, the chat becomes stuck on "Loading..."

## Example of Corrupted State

```
history.messages: 9 entries (including 5 orphaned from failed regenerations)
messages array: 4 entries (the actual valid messages)
currentId: pointing to an orphaned message ID not in messages array
```

## Suggested Fix

1. When a response generation fails, clean up the partially created message entry from `history.messages`
2. Validate `currentId` exists in `messages` array before rendering
3. Add frontend error handling to detect orphaned references and either auto-repair or show a recoverable error instead of infinite loading
4. Consider adding a database migration or health check to detect and repair corrupted chats

## Workaround

Manually clean up the SQLite database:
- Remove orphaned entries from `history.messages` that don't exist in `messages` array
- Fix `currentId` to point to a valid message
- Clean up invalid `childrenIds` references

## Environment

- Open WebUI version: 0.8.12
- API: Google GenAI (Gemini) via pipelines
- Failure mode: Connection timeout during streaming response

tjbck:
Likely addressed with the latest dev.

PR #24799, opened 2026-05-15
fix: recover structurally-malformed chat history nodes on load

Classic298:
A lost assistant placeholder can be persisted with only its completion fields (content/done/output/usage), missing id/role/parentId, while still existing under its map key and being referenced by its parent's childrenIds. The existing #24424 sanitizer only repairs currentId when it points to a *missing* node, so an "exists but malformed" node slips through: the frontend parent-walk dead-ends, Messages keeps the pagination loader visible, and the chat is stuck on "Loading..." forever with no console error and no failed request.

Reconstruct the graph fields in place during the load-time sanitize pass: derive id from the map key, parentId from the back-reference in the tree (null if none), childrenIds to an array, and role from parent alternation with a completion-field fallback. Also widen the currentId guard so an unusable current node falls back to the latest valid leaf.

Strict no-op for well-formed chats (a node whose id matches its key, has a role and an explicit parentId is skipped). Pure frontend and standalone: retroactively un-bricks every already-corrupted chat the moment it is opened, independent of the backend prevention in #24617.

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -1380,6 +1380,47 @@
 						: convertMessagesToHistory(chatContent.messages);
 
 				// Sanitize history: repair orphaned references from failed regenerations (#24424)
+				// and structurally-malformed nodes. A lost assistant placeholder can be persisted
+				// with only its completion fields (content/done/output/usage), missing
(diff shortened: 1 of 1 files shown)


Classic298:
@owui-terminator review

tjbck:
Addressed

Issue #4915, opened 2024-08-26
Original ollama send Content-Type: application/x-ndjson


unixshaman:
**Is your feature request related to a problem? Please describe.**
When I send request to original ollama service, like this:
```
POST http://127.0.0.1:11434/api/chat HTTP/1.1
Cache-Control: no-cache
Accept: text/event-stream
Content-Type: application/json; charset=utf-8
Host: 127.0.0.1:11434
Connection: Keep-Alive
Accept-Encoding: gzip
User-Agent: okhttp/4.12.0

{"model":"llama3.1:8b","messages":[{"role":"system","content":""},{"role":"user","content":"Привет"}],"options":{"temperature":0.10000000149011612,"num_predict":2048},"stream":true}
```
get next response:
```
HTTP/1.1 200 OK
Content-Type: application/x-ndjson
Date: Mon, 26 Aug 2024 13:14:53 GMT
Content-Length: 25933

{"model":"llama3.1:8b","created_at":"2024-08-26T13:14:53.7962491Z","message":{"role":"assistant","content":"H"},"done":false}
{"model":"llama3.1:8b","created_at":"2024-08-26T13:14:53.9117733Z","message":{"role":"assistant","content":"e"},"done":false}
{"model":"llama3.1:8b","created_at":"2024-08-26T13:14:53.9882669Z","message":{"role":"assistant","content":"ll"},"done":false}
{"model":"llama3.1:8b","created_at":"2024-08-26T13:14:54.0081587Z","message":{"role":"assistant","content":"o"},"done":false}

```

When I send equal request to ollama through topen-webui
```
POST https://host.com:443/ollama/api/chat HTTP/1.1
Cache-Control: no-cache
Accept: text/event-stream
Authorization: Bearer sk-*****
Content-Type: application/json; charset=utf-8
Host: host.com:443
Connection: Keep-Alive
Accept-Encoding: gzip
User-Agent: okhttp/4.12.0

{"model":"llama3.1:8b","messages":[{"role":"system","content":""},{"role":"user","content":"Hello"}],"options":{"temperature":0.10000000149011612,"num_predict":2048},"stream":true}
```

get next response:
```
HTTP/1.1 200 OK
Server: nginx/1.18.0 (Ubuntu)
Date: Mon, 26 Aug 2024 13:18:11 GMT
Connection: keep-alive
x-process-time: 5
Content-Length: 22054

[...]

PR #24838, opened 2026-05-17
fix(images): accept non-standard Content-Type in image generation responses

first-it-consulting:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Relates to #20848
- [x] **Testing:** Manually tested against Ollama v0.21.2 with a Flux image generation model — see details below.
- [x] **Agentic AI Code:** This fix has gone through human review and manual end-to-end testing (image successfully generated and returned).
- [x] **Git Hygiene:** Single atomic commit.
- [x] **Title Prefix:** `fix`

---

# Changelog Entry

### Description

`aiohttp`'s `ClientResponse.json()` validates the `Content-Type` response header against `"application/json"` by default and raises `ContentTypeError` for any other value. This silently breaks image generation for backends that return valid JSON with a non-standard MIME type.

The immediate trigger: **Ollama v0.21.x returns `Content-Type: application/x-ndjson`** from its OpenAI-compatible `/v1/images/generations` endpoint, even though the response body is a single, fully-formed JSON object. Every image generation request via Ollama therefore fails with:

```
aiohttp.client_exceptions.ContentTypeError: 0, message='Attempt to decode JSON with unexpected mimetype: application/x-ndjson'
```

The same issue can affect any backend returning valid JSON with a non-standard MIME type (e.g. `text/plain`, `application/x-ndjson`).

### Fixed

- Image generation and editing no longer fail with `ContentTypeError` when a backend returns valid JSON with a non-standard `Content-Type` header (e.g. Ollama returning `application/x-ndjson`). Fix covers the `openai`, `gemini`, and `automatic1111` engines in both `image_generations` and `image_edits`.

---

### Technical Details

`aiohttp.ClientResponse.json()` exposes a `content_type` parameter (default `"application/json"`) for exactly this case. Passing `content_type=None` skips the header check while keeping all other parsing behaviour identical — this is the [documented aiohttp pattern](https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponse.json) for lenient content-type handling.

**Diff:** 5 lines changed (one per affected `r.json()` call site).

**Reproduction steps (before fix):**
1. Configure Open WebUI image generation engine to `openai`, point `api_base_url` at a local Ollama instance.
2. Select an image generation model (e.g. `x/flux2-klein:latest`).
3. Request an image — generation completes on the Ollama side but Open WebUI returns an error.
4. In logs: `ContentTypeError: Attempt to decode JSON with unexpected mimetype: application/x-ndjson`

**Verification (after fix):**
Tested against Ollama v0.21.2 with model `x/flux2-klein:latest`. The response body (`{"created": ..., "data": [{"b64_json": "..."}]}`) is parsed correctly and the generated image is returned to the client.

### Additional Information

- Related issue: #20848 (Ollama image generation connection)
- Related closed issue: #4915 (Ollama sends `Content-Type: application/x-ndjson`)

---

### DIFF
--- a/backend/open_webui/routers/images.py
+++ b/backend/open_webui/routers/images.py
@@ -626,7 +626,7 @@ async def image_generations(
                 ssl=AIOHTTP_CLIENT_SESSION_SSL,
             ) as r:
                 r.raise_for_status()
-                res = await r.json()
+                res = await r.json(content_type=None)
 
             images = []
 
@@ -676,7 +676,7 @@ async def image_generations(
                 ssl=AIOHTTP_CLIENT_SESSION_SSL,
             ) as r:
                 r.raise_for_status()
-                res = await r.json()
+                res = await r.json(content_type=None)
 
             images = []
 
@@ -785,7 +785,7 @@ async def image_generations(
                 headers={'authorization': get_automatic1111_api_auth(request)},
                 ssl=AIOHTTP_CLIENT_SESSION_SSL,
             ) as r:
-                res = await r.json()
+                res = await r.json(content_type=None)
             log.debug(f'res: {res}')
 
             images = []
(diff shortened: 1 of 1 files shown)


tjbck:
Thanks!

Issue #24650, opened 2026-05-13
feat: Add Two-Factor Authentication (2FA) with Google Authenticator (TOTP)


sharyuke:
## Summary

First of all, thank you for building such an excellent self-hosted AI platform. Open WebUI is incredibly feature-rich and a joy to use. I truly appreciate all the work that goes into this project.

## Motivation

While Open WebUI offers robust security features like RBAC, LDAP integration, and SSO via OAuth/OIDC, one critical layer of account protection is currently missing: **Two-Factor Authentication (2FA)**.

Currently, an account is protected solely by a password. In a self-hosted environment exposed to the internet (even behind a reverse proxy), account credentials can be vulnerable to brute-force or credential-stuffing attacks. If a password is compromised or leaked, there is no second layer of defense to prevent unauthorized access.

## Proposed Solution

Implement Time-based One-Time Password (TOTP) 2FA, compatible with Google Authenticator and similar authenticator apps (Authy, 1Password, Bitwarden, etc.).

### Suggested Implementation Details

- **Per-user opt-in 2FA**: Users can enable 2FA from their profile/settings page
- **TOTP standard**: Use the standard TOTP algorithm (RFC 6238), compatible with Google Authenticator
- **Setup flow**: Display a QR code during 2FA setup that users scan with their authenticator app
- **Recovery codes**: Provide one-time recovery codes when 2FA is first enabled
- **Admin enforcement option**: Allow administrators to mandate 2FA for all users or specific user groups
- **Backup verification**: Graceful fallback handling if 2FA device is temporarily unavailable

## Alternatives Considered

- **SMS-based 2FA**: Not recommended due to SIM-swapping vulnerabilities
- **Hardware keys (WebAuthn/U2F)**: Excellent security but adds hardware cost and complexity for users
- **OAuth/SSO delegation**: Already supported (Discussion #340), but requires an external identity provider

TOTP strikes the best balance between security and accessibility — users only need a free authenticator app on their phone.

## Use Cases

- Self-hosted deployments exposed to the public internet
- Team environments where multiple users share an instance
- Compliance requirements for organizations with security policies mandating 2FA
- Anyone who wants an extra layer of protection beyond a strong password

## Priority

This would be a significant security improvement for production deployments. Even a basic implementation (per-user opt-in TOTP) would greatly enhance account security without requiring major architectural changes.

---

Thank you again for this amazing project. I look forward to seeing how it continues to evolve.

Classic298:
[...]

PR #16461, opened 2025-08-10
feat: Add TOTP 2FA authentication system

jeremy-windsor:
# TOTP 2FA Authentication Integration

## Description
This PR implements a comprehensive TOTP (Time-based One-Time Password) 2FA authentication system for Open WebUI. The implementation adds an optional second authentication factor while maintaining full backward compatibility with existing users.

**Related Discussion:** https://github.com/open-webui/open-webui/discussions/16338

## Changes Summary
- Add database migration for TOTP fields (secret, enabled, backup_codes)
- Implement TOTP utilities for secret generation, QR codes, and validation
- Add 6 new API endpoints for TOTP setup, enable/disable, and authentication
- Rewrite authentication flow to support two-step login process
- Add comprehensive TOTP management UI in account settings
- Include backup code generation and single-use enforcement
- Add proper error handling and security validation
- Support for standard authenticator apps via RFC 6238 compliance
- Add qrcode dependency for QR code generation

**Migration file:** `e4f7b8c9d2a1_add_totp_fields_to_user_table.py`

## Testing Completed
✅ Login with/without 2FA enabled  
✅ TOTP setup with QR code generation  
✅ Backup code usage and regeneration  
✅ Password-protected 2FA disable  
✅ Token expiration handling  
✅ Cross-browser compatibility  
✅ Mobile responsiveness  

## Pull Request Checklist

- [x] **Target branch:** Targeting the `dev` branch
- [x] **Description:** Comprehensive description provided above
- [x] **Changelog:** See changelog entry below
- [ ] **Documentation:** Will update docs if requested by maintainers
- [x] **Dependencies:** Added `qrcode` dependency (documented in package.json)
- [x] **Testing:** Comprehensive testing completed (see summary above)
- [x] **Code review:** Self-reviewed for coding standards and best practices
- [x] **Prefix:** Using `feat:` prefix for new feature

# Changelog Entry

## Description
Implements TOTP 2FA authentication system providing optional two-factor authentication for enhanced security while maintaining backward compatibility.

## Added
- TOTP 2FA authentication system with RFC 6238 compliance
- Database migration for TOTP fields (secret, enabled, backup_codes)
- 6 new API endpoints for TOTP lifecycle management:
  - `POST /auth/totp/setup` - Initialize TOTP setup
  - `POST /auth/totp/enable` - Enable 2FA with verification
  - `POST /auth/totp/disable` - Disable 2FA (requires password)
  - `GET /auth/totp/backup-codes` - Retrieve current backup codes
[...]

PR #24841, opened 2026-05-17
feat: add optional TOTP two-factor authentication

jeremy-windsor:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Relates to #1225. Also references #24650, #11953, and #16461.
- [ ] **Documentation:** Docs repo update is prepared in the Open WebUI docs repo format and can be submitted separately if maintainers want it.
- [x] **Dependencies:** No new dependencies were added.
- [x] **Testing:** Manual tests were performed. Screenshots are included where applicable.
- [x] **No Unchecked AI Code:** AI-assisted changes have undergone human review and manual testing.
- [x] **Architecture:** Smart defaults are used; TOTP is optional per user and disabled by default.
- [x] **Git Hygiene:** Atomic feature branch based on `upstream/dev`; no unrelated commits intended.
- [x] **Title Prefix:** `feat: add optional TOTP two-factor authentication`

## Description

Adds optional per-user TOTP two-factor authentication for Open WebUI login.

Relates to #1225.
Refs #24650.
Addresses prior feedback from #11953 and #16461 around optional defaults, migration compatibility, metadata, rate limiting, backup-code behavior, and dependency scope.

This PR intentionally keeps QR-code rendering and admin-wide enforcement out of scope so the first TOTP slice stays small, dependency-free, and reviewable. Setup uses an authenticator-app compatible `otpauth://` URI plus a manual setup key.

## Manual Testing

- Verified password login without TOTP remains unchanged.
- Verified TOTP setup from Account settings.
- Verified TOTP login challenge after password login.
- Verified authenticator-code login.
- Verified backup-code login.
- Verified used backup codes cannot be reused.
- Verified backup-code remaining count updates.
- Verified admin can see and reset a user's TOTP status.
- Verified OAuth callback/token-exchange paths return a TOTP challenge when local TOTP is enabled.
- Verified sandbox UI manually.

## Automated / Local Checks

- `git diff --check`
- Backend `py_compile` for touched Python files
- Direct RFC HOTP/TOTP vector checks
- Backup-code normalize/hash/verify checks
- `npm run test:frontend`
- `./node_modules/.bin/vite build`

Note: `npm run build` also passed, but the existing `pyodide:fetch` step pulled a newer `black` wheel and dirtied `static/pyodide/pyodide-lock.json`. That generated change was intentionally excluded from this PR.

## Screenshots or Videos

Local MFA login challenge screenshot was captured during testing. Recommended PR screenshots before marking ready:

- Account settings before TOTP is enabled.
- TOTP setup key / authenticator URI state.
- Backup codes shown immediately after enable/regenerate.
- Admin user edit modal showing TOTP status/reset.

# Changelog Entry

### Description

- Adds optional per-user TOTP two-factor authentication to protect Open WebUI login for self-hosted instances.

### Added

- Per-user TOTP setup and management in Account settings.
- TOTP challenge support for password, LDAP, trusted-header/no-auth, OAuth callback, and OAuth token-exchange login flows.
- One-time backup codes for account recovery.
- Admin TOTP status and reset control for user recovery.
- TOTP metadata including creation and last-used timestamps.
- TOTP replay protection using the last-used time step, updated in the same locked transaction as verification.
- Rate limiting for TOTP verification flows.

### Changed

- Login responses can now return an MFA challenge when TOTP is enabled for the user.

### Deprecated

- None.

### Removed

- None.

### Fixed

- None.

### Security

- TOTP secrets are encrypted with a key derived from `WEBUI_SECRET_KEY`.
- Backup codes are hashed with bcrypt.
- TOTP verification is rate-limited.
- TOTP and backup-code verification update usage state atomically to avoid double-use races.
- TOTP is disabled by default and must be explicitly enabled per user.
- No new dependencies were added.

### Breaking Changes

- None.

---

### Additional Information

- Changing `WEBUI_SECRET_KEY` invalidates encrypted TOTP secrets.
- Multi-instance deployments should use shared rate-limiting state if Open WebUI supports/configures it for the deployment.
- API keys remain bearer credentials and are not changed by login-time TOTP.
- QR-code rendering and admin-wide TOTP enforcement are follow-up scope.

> [!NOTE]

### DIFF
--- a/backend/open_webui/migrations/versions/f6b90f8a2c1d_add_user_totp_table.py
+++ b/backend/open_webui/migrations/versions/f6b90f8a2c1d_add_user_totp_table.py
@@ -1,0 +1,48 @@
+"""Add user TOTP table
+
+Revision ID: f6b90f8a2c1d
+Revises: 461111b60977
+Create Date: 2026-05-16 00:00:00.000000
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from open_webui.migrations.util import get_existing_tables
+
+# revision identifiers, used by Alembic.
+revision: str = 'f6b90f8a2c1d'
+down_revision: Union[str, None] = '461111b60977'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    existing_tables = set(get_existing_tables())
+
+    if 'user_totp' not in existing_tables:
+        op.create_table(
+            'user_totp',
+            sa.Column(
+                'user_id',
+                sa.String(),
+                sa.ForeignKey('user.id', ondelete='CASCADE'),
+                nullable=False,
+                primary_key=True,
+            ),
+            sa.Column('secret', sa.Text(), nullable=True),
+            sa.Column('enabled', sa.Boolean(), nullable=False, server_default=sa.false()),
+            sa.Column('backup_codes', sa.JSON(), nullable=True),
+            sa.Column('last_used_at', sa.BigInteger(), nullable=True),
+            sa.Column('last_used_step', sa.BigInteger(), nullable=True),
+            sa.Column('created_at', sa.BigInteger(), nullable=False),
+            sa.Column('updated_at', sa.BigInteger(), nullable=False),
+        )
+        op.create_index('idx_user_totp_enabled', 'user_totp', ['enabled'])
+
+
+def downgrade() -> None:
+    op.drop_index('idx_user_totp_enabled', table_name='user_totp')
+    op.drop_table('user_totp')
--- a/backend/open_webui/models/auths.py
+++ b/backend/open_webui/models/auths.py
@@ -5,6 +5,7 @@ import uuid
 from typing import Optional
 
 from open_webui.internal.db import Base, JSONField, get_async_db_context
+from open_webui.models.totp import UserTOTPs
 from open_webui.models.users import User, UserModel, UserProfileImageResponse, Users
 from open_webui.utils.validate import validate_profile_image_url
 from pydantic import BaseModel, field_validator
@@ -198,6 +199,7 @@ class AuthsTable:
                 result = await Users.delete_user_by_id(id, db=db)
 
                 if result:
+                    await UserTOTPs.delete_totp_by_user_id(id, db=db)
                     await db.execute(delete(Auth).filter_by(id=id))
                     await db.commit()
 
--- a/backend/open_webui/models/totp.py
+++ b/backend/open_webui/models/totp.py
@@ -1,0 +1,206 @@
+from __future__ import annotations
+
+import time
+
+from open_webui.internal.db import Base, get_async_db_context
+from open_webui.utils.totp import verify_backup_code
+from pydantic import BaseModel, ConfigDict
+from sqlalchemy import BigInteger, Boolean, Column, ForeignKey, JSON, String, Text, delete, select
+from sqlalchemy.ext.asyncio import AsyncSession
+
+
+class UserTOTP(Base):
+    __tablename__ = 'user_totp'
+
+    user_id = Column(String, ForeignKey('user.id', ondelete='CASCADE'), primary_key=True, unique=True)
+    secret = Column(Text, nullable=True)
+    enabled = Column(Boolean, nullable=False, default=False)
+    backup_codes = Column(JSON, nullable=True)
+    last_used_at = Column(BigInteger, nullable=True)
+    last_used_step = Column(BigInteger, nullable=True)
+    created_at = Column(BigInteger, nullable=False)
+    updated_at = Column(BigInteger, nullable=False)
+
+
+class UserTOTPModel(BaseModel):
+    user_id: str
+    secret: str | None = None
+    enabled: bool = False
+    backup_codes: list[str] | None = None
+    last_used_at: int | None = None
+    last_used_step: int | None = None
+    created_at: int
+    updated_at: int
+
+    model_config = ConfigDict(from_attributes=True)
+
+
+class UserTOTPsTable:
+    async def get_user_totp_by_user_id(self, user_id: str, db: AsyncSession | None = None) -> UserTOTPModel | None:
+        try:
+            async with get_async_db_context(db) as db:
+                result = await db.execute(select(UserTOTP).filter_by(user_id=user_id))
+                user_totp = result.scalars().first()
+                return UserTOTPModel.model_validate(user_totp) if user_totp else None
+        except Exception:
+            return None
+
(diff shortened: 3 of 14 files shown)


Classic298:
Hi. Please only open PRs regarding TOTP/MFA/2FA after discussion. I know there exists an issue for it with many thumbs up, but: 1) this is just for tracking and is a legacy issue and  more importantly: 2) this needs to be explicitly wanted by @tjbck, because as you may know, Open WebUI tries to not be the auth provider for the setups. Via extensive SSO/OAUTH support, there are dozens of providers you can use for authentication. TOTP could be one more step towards Open WebUI becoming the main authentication layer for small orgs and that's not what Open WebUI should be. Sure, TOTP is not THAT, but it's moving the goalpost much closer and merging such a feature needs to be discussed first before opening a PR.

That being said

I also went through your PR to give you some feedback

- You included test files. Please do not include Test files in your PR this is explicitly not wanted.
- bcrypt blocks the ENTIRE event loop aka would make the entire application become STUCK for everyone with no obvious reason why it is stuck: verify_backup_code and hash_backup_code call bcrypt.checkpw/bcrypt.hashpw synchronously, and they're called from inside async def functions without any thread offload. bcrypt is intentionally slow (~50–150ms per check at default cost factor 12). The fallout shows up in three places:
  - In verify_totp_or_backup_code (routers/auths.py) the backup-code branch loops through the user's stored hashes and runs verify_backup_code on each one. With 10 backup codes that's up to ~1 second of fully blocked event loop per failed attempt. Anyone hitting the MFA endpoint with a bogus code stalls every other request handled by that worker. On a single-worker uvicorn this is a trivial DoS vector — an unauthenticated attacker can sustain it just by replaying the MFA token within its 5-minute window.
  - In enable_totp and regenerate_totp_backup_codes, [hash_backup_code(code) for code in backup_codes] runs ten synchronous bcrypt hashes on the event loop. ~1 second per enable/regenerate.
  - In consume_backup_code_by_user_id (models/totp.py), same loop, same problem (though as noted below, this function is dead code).
- Critical: race condition on backup-code consumption (SQLite)
  - The router uses select(UserTOTP).filter_by(user_id=user_id).with_for_update() — good intent, but FOR UPDATE is a no-op on SQLite (it has no row-level locking; the whole DB locks on write transactions, which actually saves you here partly, but the semantics are not what the code thinks). On PostgreSQL it works. On SQLite, **two concurrent requests with the same backup code can theoretically both pass the verify check and both commit consumption**. The code is written as if FOR UPDATE works and that's a portability assumption that will bite when someone enables WAL mode plus parallel readers.
- Worse, the dead consume_backup_code_by_user_id in models/totp.py doesn't even try with_for_update. If anyone wires it up later they'll get the unlocked path. So this PR is shipping dead code. Thats not good.
- dual implementations of backup-code consumption: UserTOTPs.consume_backup_code_by_user_id exists in models/totp.py and is never imported or called anywhere in the diff. More dead code. Why?
- The router has its own inline implementation in verify_totp_or_backup_code with stronger semantics (row lock, secret decryption integrated, last_used_step update). Two implementations of the same security-critical operation, with subtly different behavior, is exactly how someone introduces a bug six months from now by "consolidating." Drop the model-layer one.
- OAuth flow sets session cookies before MFA verification: In utils/oauth.py, when MFA is required after OAuth, the code path does:
  - Resolve user, decide MFA needed
  - Create mfa_token
  - Build redirect to frontend with mfa_token in fragment
  - Call _set_oauth_session_cookies — which sets oauth_session_id and optionally oauth_id_token cookies
  - Return the redirect
- So the user has OAuth session cookies before they've completed the second factor. They don't have a JWT (good), so they can't hit normal endpoints. But:
  - Any endpoint that authenticates via oauth_session_id cookie alone bypasses MFA. You need to audit every consumer of that cookie. The get_oauth_token(user_id, session_id, ...) method takes the session id and returns tokens; if that's reachable without a JWT auth check, you've got a bypass.
  - If the user closes the browser before completing MFA, there's an orphan OAuth session sitting server-side keyed to their account.
  - The MFA endpoint (verify_totp_login) doesn't reconcile or refresh the OAuth session after MFA succeeds — it just issues a JWT. So the cookies set pre-MFA persist as-is.
- encryption key derived directly from WEBUI_SECRET_KEY: The whole point of 2FA is to be independent of password/session compromise; tying the seed encryption to the JWT signing secret partially defeats that for at-rest scenarios. There's no key derivation context. If you ever decide to use WEBUI_SECRET_KEY for anything else (cookie signing, file encryption, etc.), and you derive keys the same naive way, you get key reuse across purposes.
- Bonus issue: _get_fernet() runs sha256 and base64 on every encrypt/decrypt call. Cache it.
- idempotency mismatch in the migration: upgrade() checks if 'user_totp' not in existing_tables before creating; downgrade() unconditionally drops index and table. That's inconsistent — if the upgrade was a no-op because the table pre-existed (e.g. from a previous failed migration), the downgrade will still drop someone's pre-existing table.
- redundant unique=True on a primary key column. primary_key=True already implies UNIQUE
- no rendered QR code in setup UX
  - setupTOTP returns otpauth_url but the Svelte component just renders it as a text link. On desktop (which is where most users do 2FA setup, scanning a QR code with their phone), this is a hard regression vs every other 2FA UX they've seen
- backup codes shown once with no warning
  - After enable, the codes render in a 2-col grid with a "Done" button and a "Copy" button. No prominent warning that these will never be shown again, no acknowledgment checkbox, no auto-download option. Users will click Done without copying
- Zero audit logging: Zero logging for: TOTP enabled, TOTP disabled (self or by admin), backup codes regenerated, failed MFA attempts (especially repeated ones from the same user), successful MFA, admin force-disable on another user. For a 2FA feature on a multi-user app, this is the most important observability surface and it's just absent.
- MFA token not invalidated after use: verify_totp_login doesn't blacklist or one-shot the mfa_token JWT after a successful verification. Within the 5-min window an intercepted token can be replayed to mint additional session JWTs.

And about 10 more smaller findings i left out for now because this list is long enough.

For sure you had only the best intentions with this PR, but such an absolutely core, auth related, security impacting, database modifying, potentially deployment breaking feature should really be WELL tested and well thought through. In just this brief review I found criticial security issues, DoS attack vectors that can render the entire deployment unusable and other major flaws.

Did you test the PR before submitting? The PR template explicitly asks you for Screenshots, especially for such a very large PR. Your Screenshot section says "Recommended PR screenshots before marking ready". Did you write this?
You also say "Docs repo update is prepared; see docs draft below." But I see no docs draft anywhere.
and "/tmp/open-webui-totp-sandbox/screenshots/totp-login-mfa.png" unfortunately your image didn't get uploaded.
"git diff --check" is not an automated check and you also wrote "Direct RFC checks" - but it violates §5.1 and §5.2.

Classic298:
also forgot to write: you didn't introduce any env vars to toggle this or admin toggles, nor did you add a config setting to control MFA into the app state or synced it to the DB (no app state) and also don't sync it to redis

so how exactly can i toggle this feature on or off? you say in your PR desc it off by default and can be toggled on. How? 

Screenshots would have been helpful here as well

PR #24864, opened 2026-05-18
feat(chat): citation modal as global singleton via store

julenlurnova:
Citations.svelte mounts <CitationModal> per instance. A chat with N RAG responses has N modal state machines in the DOM, idle until clicked.

Same shape as the embed/showEmbeds pattern already in $lib/stores. Applied to citations:
- Removes N-instances cost (memory, listeners, lifecycle).
- Markdown citation-marker clicks can call openCitation() directly — no `bind:this` ref hacks into the right <Citations> instance.
- Gives forks an extension point (subscribe to citationModal, render alongside).

### Changes

- NEW `src/lib/stores/citations.ts` — writable + openCitation/closeCitation
- `src/routes/(app)/+layout.svelte` — mount <CitationModal> once, subscribed to store
- `src/lib/components/chat/Messages/Citations.svelte` — dispatch via store instead of local mount; `showSourceModal` export kept as thin wrapper for back-compat

### Bridge note

The +layout bridge is one-way sync (store→local) + close-back handler. The naive `if ($store.show !== local) { sync }` form races itself: the X click writes false to local via `bind:show`, the reactive immediately re-syncs from store (still true), undoes the close. The split form (one `$:` writes local from store, a separate `$:` writes store-close when local goes false) avoids the loop.

### Tested

- Click chip pill → modal opens; X / Esc / backdrop close
- Inline Markdown [n] marker click → modal opens (via showSourceModal wrapper)
- embed_url citation → opens right-side embed panel (unchanged path)
- readOnly mode (shared chat) → opens link in new tab (unchanged)
- Multiple chats in tabs → single global modal, no cross-chat leakage

### Follow-ups (separate PRs if this lands)

Same pattern fits ImagePreview (Image.svelte mounts one per image — 7+ call sites in the chat tree) and CodeExecutionModal. Happy to open those as small PRs after this is reviewed.

### DIFF
--- a/src/lib/components/chat/Messages/Citations.svelte
+++ b/src/lib/components/chat/Messages/Citations.svelte
@@ -1,8 +1,7 @@
 <script lang="ts">
 	import { getContext } from 'svelte';
 	import { embed, showControls, showEmbeds } from '$lib/stores';
-
-	import CitationModal from './Citations/CitationModal.svelte';
+	import { openCitation } from '$lib/stores/citations';
 
 	const i18n = getContext('i18n');
 
@@ -16,12 +15,7 @@
 	let showPercentage = false;
 	let showRelevance = true;
 
-	let citationModal = null;
-
 	let showCitations = false;
-	let showCitationModal = false;
-
-	let selectedCitation: any = null;
 
 	export const showSourceModal = (sourceId) => {
 		let index;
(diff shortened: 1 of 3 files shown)


julenlurnova:
Reopened against `dev` with the required template and CLA: https://github.com/open-webui/open-webui/pull/24865

PR #24918, opened 2026-05-19
enh: rename tool-call cap to CHAT_RESPONSE_MAX_TOOL_CALL_ITERATIONS, default 100, add visible error

Classic298:
The cap on the native agentic tool loop has been misnamed "retries" since native tool calling shipped. It is incremented unconditionally once per assistant tool-calling turn, so every successful tool call counts; it is not a failed-call retry budget. The name implied otherwise and the low default (30) silently terminated long agentic runs (e.g. open-terminal workflows) with no user-visible signal.

Rename to CHAT_RESPONSE_MAX_TOOL_CALL_ITERATIONS and raise the default to
100. The old CHAT_RESPONSE_MAX_TOOL_CALL_RETRIES env var is still honoured as a fallback so existing deployments keep their configured value.

> [!NOTE]


tjbck:
2b99945d2726bdf5aed7b1712f9e3b7b622671df

PR #24928, opened 2026-05-19
feat: configurable aiohttp read_bufsize via AIOHTTP_READ_BUFSIZE env

shoemoney:
## Problem

The shared `aiohttp.ClientSession` in `utils/session_pool.py` is created with the library default `read_bufsize` (64 KiB). Some upstream providers emit a **single** SSE chunk that exceeds this ceiling, causing the stream to fail with:

```
400, message: Got more than 131072 bytes when reading: b'data: {"id":"gen-...","object":"chat.completion.chunk",...
```

(The 131072 = 2 × 64 KiB comes from aiohttp internally doubling the buffer once before raising.)

### Triggers seen in the wild

- **OpenRouter routing reasoning models** (Kimi K2.6, R1-style, etc.) through providers that batch the entire reasoning trace into a single chunk — observed with Novita, WandB, SiliconFlow.
- **Image-generation responses** returned as a single large SSE message — see discussion #24543, which reports the same error signature and is currently unresolved.

There is no current env or config knob to override this; the only workaround is patching the source.

## Fix

Expose `AIOHTTP_READ_BUFSIZE` (default **1 MiB**) and pass it through to `ClientSession(read_bufsize=...)`.

- Default of 1 MiB is conservative — handles every real-world case I've reproduced while still rejecting truly runaway payloads.
- Falls back gracefully on invalid/empty/negative values.
- Logged at session creation alongside the other pool params.

## Files

- `backend/open_webui/env.py` — new env constant with validation.
- `backend/open_webui/utils/session_pool.py` — import it, pass to `ClientSession`, update docstring + log line.

## Validated against

Reproduced the failure with `moonshotai/kimi-k2.6` via OpenRouter (Novita provider) on `v0.9.5`. With the default 64 KiB buffer the stream dies mid-reasoning; with `AIOHTTP_READ_BUFSIZE=1048576` (or the new default) the stream completes cleanly through `data: [DONE]`.

Related: #24543

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -578,6 +578,20 @@ try:
 except ValueError:
     AIOHTTP_POOL_DNS_TTL = 300
 
+# Max size of a single chunk read from an SSE/HTTP stream, in bytes.
+# aiohttp's default is 64 KiB which can be exceeded when an upstream provider
+# emits a single very large SSE chunk (e.g. OpenRouter sometimes batches a
+# reasoning trace or tool-call payload into one chunk, or an image-generation
+# response is returned in one message). Raising this avoids
+# `Got more than N bytes when reading` 400 errors. Default 1 MiB.
+AIOHTTP_READ_BUFSIZE = os.getenv('AIOHTTP_READ_BUFSIZE', str(1024 * 1024))
+try:
+    AIOHTTP_READ_BUFSIZE = int(AIOHTTP_READ_BUFSIZE)
+    if AIOHTTP_READ_BUFSIZE <= 0:
+        AIOHTTP_READ_BUFSIZE = 1024 * 1024
+except ValueError:
+    AIOHTTP_READ_BUFSIZE = 1024 * 1024
+
 RAG_EMBEDDING_TIMEOUT = os.getenv('RAG_EMBEDDING_TIMEOUT', '')
 
 if RAG_EMBEDDING_TIMEOUT == '':
--- a/backend/open_webui/utils/session_pool.py
+++ b/backend/open_webui/utils/session_pool.py
@@ -9,6 +9,11 @@ All pool parameters are configurable via environment variables:
     - AIOHTTP_POOL_CONNECTIONS (default 100) — max total connections
     - AIOHTTP_POOL_CONNECTIONS_PER_HOST (default 30) — per-host limit
     - AIOHTTP_POOL_DNS_TTL (default 300) — DNS cache TTL in seconds
+    - AIOHTTP_READ_BUFSIZE (default 1048576) — per-chunk read buffer size in
+      bytes. Raise this when an upstream emits very large SSE chunks (e.g.
+      reasoning-trace batching by some OpenRouter providers, or single-message
+      image-generation responses) which otherwise fail with
+      "Got more than N bytes when reading".
 
 Usage:
     from open_webui.utils.session_pool import get_session, cleanup_response
@@ -32,6 +37,7 @@ from open_webui.env import (
     AIOHTTP_POOL_CONNECTIONS,
     AIOHTTP_POOL_CONNECTIONS_PER_HOST,
     AIOHTTP_POOL_DNS_TTL,
+    AIOHTTP_READ_BUFSIZE,
 )
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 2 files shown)


tjbck:
We already handle oversized SSE chunks via stream_chunks_handler which uses iter_chunks() instead of readline(), bypassing the read_bufsize limit entirely, so I belive this change isn't needed. If you're still hitting this error on v0.9.5, check whether CHAT_STREAM_RESPONSE_CHUNK_MAX_BUFFER_SIZE is set correctly and share the full traceback so we can see which code path is triggering the readuntil call.

jpaodev:
@tjbck  
```
│ 2026-06-09 18:35:55.095 | ERROR    | open_webui.main:process_chat:2084 - Error processing chat payload: 400, message:             │
│   Got more than 131072 bytes when reading: b'data: {"id":"B10oaqfTDpqG6doP3MTaiAY","created":1781030153,"model":"gemini-3.1-flash │
│ -lite-code-execu...'.
```
Got this error and `CHAT_STREAM_RESPONSE_CHUNK_MAX_BUFFER_SIZE` was not set for my deployment. This seems not fixed. 

To reproduce: Ask a Gemini model with native code execution tool to plot you a fibonacci series -> did not work. Regular pictures normally worked for me, at least when using the "Image" OpenWebUI functionality

jkoopmann:
I agree that this is not fixed. I am running into the same problem and CHAT_STREAM_RESPONSE_CHUNK_MAX_BUFFER_SIZE is not the answer!

chrisdewar:
Hey, ran into this same limit on the Responses API path and I agree, I don't think it's actually fixed.

The `response.created` event on the Responses API embeds the full request echo, including the whole tools array. Hook up an MCP server with enough tools, and that single SSE data: line blows past 128 KB, and you get:

```400, message: Got more than 131072 bytes when reading: b'data: {"type":"response.created",...'```

It's also not provider-specific. `response.created` carries tools by spec, so it happens straight against OpenAI, and through LiteLLM (the `resp_id` in that error base64-decodes to `litellm:custom_llm_provider:azure`).

The reason it's still biting: stream_wrapper reads the body directly when there's no handler ([session_pool.py#L114](https://github.com/open-webui/open-webui/blob/ecd48e2f718220a6400ecf49eafd4867a38feb10/backend/open_webui/utils/session_pool.py#L114)), which runs through aiohttp's StreamReader and its 2 * read_bufsize (131072) line cap. stream_chunks_handler dodges that cap, but it's only wired into chat completions ([openai.py#L1311](https://github.com/open-webui/open-webui/blob/ecd48e2f718220a6400ecf49eafd4867a38feb10/backend/open_webui/routers/openai.py#L1311)). The /responses proxy ([openai.py#L1549](https://github.com/open-webui/open-webui/blob/ecd48e2f718220a6400ecf49eafd4867a38feb10/backend/open_webui/routers/openai.py#L1549)) and the passthrough ([openai.py#L1670](https://github.com/open-webui/open-webui/blob/ecd48e2f718220a6400ecf49eafd4867a38feb10/backend/open_webui/routers/openai.py#L1670)) both call it bare, so the exact path that triggers this never gets that protection.

And even where the handler does run, it drops the oversized line and swaps in `data: {}` ([misc.py#L1120](https://github.com/open-webui/open-webui/blob/ecd48e2f718220a6400ecf49eafd4867a38feb10/backend/open_webui/utils/misc.py#L1120)). That's fine for a mid-stream delta, but `response.created` carries the response id the client needs, so dropping it just breaks the stream a different way. That's why bumping `read_bufsize` is the better lever here: the line is valid, we just need to let it through instead of tossing it.

Any chance of reopening?

gschintgen:
I just hit the same error when giving a rather large task to gpt-5.4 (high reasoning) via OpenRouter. After a few minutes of continuous streaming of "thinking" tokens, OpenWebUI abruptly aborted before the actual response even started to print.

chrisdewar:
@tjbck can we get this reopened please

DanielHe212:
I'm getting the same error using llama.cpp via llama-swap as the backend, barely any tools, just a really long chat turn with a lot of tool calls.

satonotdead:
Hitting the same with Vertex AI models on LiteLLM endpoint. Already set ``CHAT_STREAM_RESPONSE_CHUNK_MAX_BUFFER_SIZE=20971520`` but it stills no working.

PR #25018, opened 2026-05-22
fix: sanitize null bytes in nested db data

IVVI0927:
# Changelog Entry

### Description

- Fix a silent sanitization bug in `sanitize_data_for_db`: null bytes inside dict/list values were not removed because the fast-path check looked for `'\x00'` in `json.dumps(...)`, while `json.dumps` encodes null bytes as `\u0000`. Changed the check to look for `'\\u0000'` instead.
- Adds focused regression tests for the null-byte sanitization behavior.
- Relates to https://github.com/open-webui/open-webui/discussions/24939

### Added

- `backend/open_webui/test/utils/test_misc.py`: focused regression tests for `sanitize_data_for_db`, covering strings, dict values, nested dict values, list values, and the clean-data fast path.

### Changed

- `backend/open_webui/utils/misc.py`: updated the `sanitize_data_for_db` fast-path to check for `'\\u0000'` instead of `'\x00'` in the `json.dumps` output.

### Fixed

- `sanitize_data_for_db` now correctly strips null bytes from string values inside dicts, nested dicts, and lists.

- [×] By submitting this pull request, I confirm that I have read and fully agree to the [Contributor License Agreement (CLA)](https://github.com/open-webui/open-webui/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT), and I am providing my contributions under its terms.

> [!NOTE]

### DIFF
--- a/backend/open_webui/test/utils/test_misc.py
+++ b/backend/open_webui/test/utils/test_misc.py
@@ -1,0 +1,33 @@
+import sys
+import types
+
+# Stub lightweight dependencies that misc.py imports at module level so tests
+# can run without a full application bootstrap.
+_env = types.ModuleType('open_webui.env')
+_env.CHAT_STREAM_RESPONSE_CHUNK_MAX_BUFFER_SIZE = 16384
+sys.modules.setdefault('open_webui.env', _env)
+sys.modules.setdefault('mimeparse', types.ModuleType('mimeparse'))
+sys.modules.setdefault('aiohttp', types.ModuleType('aiohttp'))
+
+from open_webui.utils.misc import sanitize_data_for_db  # noqa: E402
+
+
+class TestSanitizeDataForDb:
+    def test_string_null_bytes_removed(self):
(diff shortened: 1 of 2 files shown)


IVVI0927:
This PR was submitted as part of a university open-source contribution assignment. The bug and fix are genuine. Happy to remove the test file or adjust the scope.

Classic298:
yes please remove it.  but i am not sure what this fixes. inside python x00 should be identical to u0000

IVVI0927:
Thanks for clarifying. I agree that inside a Python string, '\x00' and '\u0000' represent the same null character.

The issue is about the serialized output of json.dumps(). For example, json.dumps("a\x00b") produces a string containing the escaped text \\u0000, not a literal \x00. Because of that, the previous fast-path check for '\x00' in the JSON string does not detect null bytes inside dict/list values.

tjbck:
e3ab4bd212e44c39f439ec4ff5df7c3dbd046895, Thanks!

PR #25055, opened 2026-05-25
fix: validate model meta.knowledge file ownership on create/update

Classic298:
A workspace.models user could place another user's file_id in their model's meta.knowledge; has_access_to_file() then granted read on that file (and the view_file builtin returned its content) because the file appears in a model the caller can access. Reject knowledge entries of type 'file' the caller does not own / cannot read, on model create and update.

> [!NOTE]


tjbck:
27fb20c13a4bf8501f7a485abe0654eb5880980d

Issue #15551, opened 2025-07-06
issue: OAuth Role Management Ignored: OAUTH_ALLOWED_ROLES and OAUTH_ADMIN_ROLES Not Applied on OIDC Login


Hani-K:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.6.15

### Operating System

Ubuntu 24.04

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

1. When ENABLE_OAUTH_ROLE_MANAGEMENT=true, incoming tokens with roles matching OAUTH_ALLOWED_ROLES should be created as active users.
2. Tokens with roles matching OAUTH_ADMIN_ROLES should be created as active admins.
3. Users whose roles claim does not match OAUTH_ALLOWED_ROLES should be denied login.

### Actual Behavior

- Any valid OAuth user is created with the default role (pending, user or admin), regardless of their roles claim.
- When an already registered user logs in with role admin, the user's role gets reset to the Default User Role. 
- Manually promoting the user to an Admin in the UI is undone on the next login.

### Steps to Reproduce

1. Start with installing a fresh Ubuntu 24.04.2 LTS on a VM, with Kernel 6.8.0-63-generic.
2. Install docker version 28.3.1
3. Using docker compose, set up Open WebUI using the following configuration:
```
services:
  webui:
    image: ghcr.io/open-webui/open-webui:main
    container_name: open-webui
    environment:
[...]

PR #25062, opened 2026-05-25
feat: allow '*' wildcard in OAUTH_ALLOWED_ROLES

cyberb:
# Pull Request Checklist

- [x] **Testing:** Manual end-to-end test performed on a real device (Syncloud arm64 box). With the patch, a non-admin OIDC user whose only group is a dynamically-named per-user LDAP group is granted `DEFAULT_USER_ROLE` instead of being 403'd. Without the patch, same user is rejected at `oauth.py:1276` with the WARNING log already present in the function. Repro spec: https://github.com/syncloud/openwebui/blob/fix-oauth-allowed-roles/web/e2e/lam-user-login.spec.ts
- [x] **Self-Review:** Reviewed.
- [x] **Architecture:** Smart default — `*` is only special when explicitly listed; absent `*`, behavior is byte-identical to before. No new env var.
- [x] **Git Hygiene:** Single-commit, atomic, one-line change.
- [x] **Title Prefix:** `feat:`.

# Changelog Entry

### Description

When `ENABLE_OAUTH_ROLE_MANAGEMENT=true`, `get_user_role` requires every authenticated user's `groups`/`roles` claim to contain at least one literal entry from `OAUTH_ALLOWED_ROLES` (default `['user', 'admin']`) or `OAUTH_ADMIN_ROLES`; otherwise the user is rejected with 403.

This works when an admin can enumerate every legal role value. It does **not** work when the OIDC provider emits **dynamic** group names — e.g.:

- LDAP-backed IdPs that auto-create a per-user group named after the user (LDAP Account Manager does this when `DEFAULT_USER_GROUP` does not exist as an LDAP group)
- Multi-tenant providers that prefix groups with a tenant id, project name, etc.
- Provisioning tools that assign each new user a unique synthetic group

A common installer wants:

```
OAUTH_ADMIN_ROLES=my-admin-group   # specific group → admin
OAUTH_ALLOWED_ROLES=*              # everyone else authenticated by OIDC
                                   # gets DEFAULT_USER_ROLE
```

This PR makes `*` in `OAUTH_ALLOWED_ROLES` act as a catch-all match. `OAUTH_ADMIN_ROLES` is intentionally unchanged (still strict-match) so admin auto-mapping keeps working alongside the wildcard.

### Added

- `OAUTH_ALLOWED_ROLES=*` now matches any non-empty roles claim, granting `DEFAULT_USER_ROLE`. `OAUTH_ADMIN_ROLES` behavior is unchanged.

### Changed

- `backend/open_webui/utils/oauth.py:1267` — single condition `if allowed_role in oauth_roles:` extended to `if allowed_role == '*' or allowed_role in oauth_roles:`.

### Fixed

- Operators using IdPs with dynamic per-user group claims (LDAP, multi-tenant, etc.) can no longer be locked out of Open WebUI without manually enumerating every possible group name. Related: #15551, #20518.

---

### Additional Information

Found on the Syncloud platform, where the Syncloud `users` web app (built on LDAP Account Manager) creates a per-user LDAP group for every new account. Authelia (their OIDC provider) returns that group verbatim in the `groups` claim, and Open WebUI then rejects the user with 403 because the group name matches neither `OAUTH_ADMIN_ROLES` nor the default `OAUTH_ALLOWED_ROLES=['user','admin']`.

End-to-end Playwright repro (PR in the Syncloud package): https://github.com/syncloud/openwebui/pull/2

### Screenshots or Videos

Before (rejected, openwebui error toast):

> Screenshot from the end-to-end repro: openwebui shows "You do not have permission to access this resource. Please contact your administrator for assistance." after Authelia completes successfully.

After (with `OAUTH_ALLOWED_ROLES=*`): user lands on the main `/` chat page and gets `DEFAULT_USER_ROLE=user`. Admin user (in `OAUTH_ADMIN_ROLES`) still becomes admin.

### DIFF
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -1264,7 +1264,7 @@ class OAuthManager:
             if oauth_roles:
                 matched = False
                 for allowed_role in oauth_allowed_roles:
-                    if allowed_role in oauth_roles:
+                    if allowed_role == '*' or allowed_role in oauth_roles:
                         log.debug('Assigned user the user role')
                         role = 'user'
                         matched = True


tjbck:
07cbc91a8eba3a9a3b39588b2ae5de916930af70, thanks!

PR #25079, opened 2026-05-26
fix: route web uploads through owned Files; drop legacy non-KB fall-open

Classic298:
/process/web now registers fetched URLs as owned File rows and stores at file-{file_id} instead of the legacy unscoped sha256(url) namespace. Retrieval gates through has_access_to_file, which already handles owner, shared-chat clone, KB / channel / workspace-model paths, so behaviour is preserved without grant copying.

filter_accessible_collections drops the fall-open branch that allowed any unknown non-KB collection name (the BAC primitive). /process/text and ProcessTextForm are removed (no frontend callers, the truly dead legacy surface). Chat.svelte uploadWeb propagates the file id so chat-save populates ChatFile, enabling shared-chat access for cloners.

old chats whose fileItem.collection_name was the legacy sha256(url)[:63] (created before this PR) will silently stop returning RAG results, because filter_accessible_collections will now deny those names. Users have to re-upload the URL in those old chats. No data is destroyed, only the retrieval link is broken. This is the unavoidable price of removing the fail-open without a DB migration.

> [!NOTE]

### DIFF
--- a/backend/open_webui/retrieval/utils.py
+++ b/backend/open_webui/retrieval/utils.py
@@ -1077,9 +1077,12 @@ async def filter_accessible_collections(
       - web-search-*    → ephemeral per-query collections, always allowed
       - knowledge-bases → always denied (system meta-collection)
       - everything else → if the name matches a knowledge base, validated
-                          via Knowledges.check_access_by_user_id; if no
-                          such KB exists, the name is treated as an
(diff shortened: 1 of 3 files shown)


tjbck:
c93f071700520a60833b7f9616c292b11f210880

Issue #24432, opened 2026-05-07
feat: MCP Tool Permissions Are Global Per Group Instead of Scoped Per Workspace Model


impr3ssi0n:
### Check Existing Issues

- [x] I have searched for all existing **open AND closed** issues and discussions for similar requests. I have found none that is comparable to my request.

### Verify Feature Scope

- [x] I have read through and understood the scope definition for feature requests in the Issues section. I believe my feature request meets the definition and belongs in the Issues section instead of the Discussions.

### Problem Description

MCP tool permissions currently appear to be global once access is granted to a user or group in the Integration connection settings.

I added an MCP server through Integrations as a Streamable HTTP connection with OAuth 2.1 authentication and client registration.

In the connection editor, if I do not assign permissions to any group or user, the MCP tool is only available to admins.

However, if I grant access to a specific group, users in that group can enable and use this MCP tool with any model they have access to.

This creates a problem for Workspace-scoped model configurations.

For example:

1. I create a Workspace model based on `gpt-5`.
2. I grant access to this Workspace model only to a specific group.
3. I attach/select a specific MCP tool for this Workspace model.

I would expect the MCP tool to only be usable:

* with that specific Workspace model,
* and only by the allowed group.

Instead, once users have permission to the MCP connection itself, they can use the MCP tool across all accessible models.

This behavior differs from how regular tools behave in Workspace model configurations and makes it difficult to enforce model-specific tool access policies.

### Desired Solution you'd like

MCP tools should support permission scoping at the Workspace model level, similar to regular tools.

Expected behavior:

* an MCP tool can be attached to a specific Workspace model,
* only users/groups with access to that Workspace model can use the tool,
* the tool should not automatically become available globally across all models accessible to that group.

This would allow administrators to create isolated model + tool combinations with predictable access control behavior.

Issue #19313, opened 2025-11-20
feat: User specific MCP headers


patrykkozuch:
### Check Existing Issues

- [x] I have searched for all existing **open AND closed** issues and discussions for similar requests. I have found none that is comparable to my request.

### Verify Feature Scope

- [x] I have read through and understood the scope definition for feature requests in the Issues section. I believe my feature request meets the definition and belongs in the Issues section instead of the Discussions.

### Problem Description

I want to develop a MCP server for a service that does not support OAuth / is not integrated with OAuth yet and supports only Basic Auth. For now, MCP configuration in Open WebUI supports supplying only the global API token to the server (and starting from 0.6.37 other custom headers thanks to #18918 ). I'd like to allow each user to supply their own credentials to work with the service through MCP server.

### Desired Solution you'd like

As a user, I'd like to be able to use my own API key (and preferably set custom headers) when enabling the globally configured MCP server as a tool in the conversation - something like the scope configuration that happens when you enable OAuth MCP server as a user.

Something like user valves, but for MCP / tools servers.

I know that there is a possibility for each user to configure their own tools server but:
1) I want the overall configuration to be managed globally by the admin
2) There is no possibility to add MCP servers that way (they can be configured only globally)

tjbck:
@patrykkozuch would `Auth` header with `Bearer` prefix be enough for your use case?

patrykkozuch:
@tjbck Yes, it would be. Would be nice to be able to add another headers too, but that's not a must

ctolon:
This feature can be truly useful. In another product ex. LibreChat, custom headers can be sent dynamically for the MCP (the header keys to be sent are defined in the MCP Server connection section and when the user connects to the MCP Server, values ​​are assigned to the designated placeholders. For example, if the Role placeholder corresponding to the X-Role header is defined as an input in the MCP connection section, each user can enters an input for themselves, and this placeholder value can be sent to the MCP along with the X-Role header).

This can be particularly useful if the MCP Server implementation requires deterministic input from the user.

ctolon:
Regarding to this issue I created a PR -> [https://github.com/open-webui/open-webui/pull/19379](url)

We need this feature organizationally, too. We use it not just for authentication, but especially for MCPs that connect to SQL databases, allowing us to select tables and DBs etc. as input.

As far as I have tested the implementation,. If you want, you can clone it and test whether it meets your needs @patrykkozuch

patrykkozuch:
[...]

Issue #18453, opened 2025-10-20
feat: Add possibility to configure user scope MCP servers


wilstdu:
### Check Existing Issues

- [x] I have searched all existing open AND closed issues and discussions for similar requests. I have found none that is comparable to my request.

### Problem Description

My setup is AWS Bedrock - LiteLLM -> OpenWebUI. It's mandatory to track each user`s spending via LiteLLM, MCPs are also configured via LiteLLM, and users need their LiteLLM API tokens to use the MCPs. 

I'm lacking the ability to allow users to configure their own MCP servers in OpenWebUI "External Tools" - only OpenAPI is currently supported. MCP server configuration is supported via admin configuration, and it's enabled globally - the problem with global setup is that it uses a single bearer token. Therefore, I can't correctly assign the costs to the user who uses the token.

### Desired Solution you'd like

Users are allowed to configure their own MCP servers via External Tools, just like global admin configuration:
<img width="954" height="812" alt="Image" src="https://github.com/user-attachments/assets/f4d46e83-8b79-4250-b7cf-dc8adcf237b4" />

tjbck:
Intended behaviour.

marcinkubica:
@tjbck  Feels tad limiting on the user side, but at the same time admin should be the one allowing tools for users.

wilstdu:
@tjbck why it's intended behaviour? Admins should decide if to allow that for users.

PR #25115, opened 2026-05-27
feat: per-user MCP tool servers

### LINKED ISSUE #24620: feat: Per-user MCP servers
### Check Existing Issues

- [x] I have searched for all existing **open AND closed** issues and discussions for similar requests. I have found none that is comparable to my request.

### Verify Feature Scope

- [x] I have read through and understood the scope definition for feature requests in the Issues section. I believe my feature request meets the definition and belongs in the Issues section instead of the Discussions.

### Problem Description

Currently (OWUI 0.9.5), the user can only add an external OpenAPI server. All MCP servers must be admin-defined. 

The user should be able to add an MCP server that will have access to the same scope as they user has.

It should be configurable through permission to have a possibility to disallow this in general / for a specific group.

### Alternatives Considered

_No response_

### Additional Context

_No response_

TALLEC-Scott:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #24620. Relates to #18453 (closed as "intended" — this PR re-opens that discussion with the structural concerns addressed), #19313 (per-user MCP headers — orthogonal but adjacent), and #24432 (MCP tool permissions per workspace — same domain).
- [x] **Target branch:** `dev`.
- [x] **Description:** See changelog entry below.
- [x] **Changelog:** Below.
- [ ] **Documentation:** Not added in this PR — the feature is gated by the existing `features.direct_tool_servers` permission and surfaces in an existing UI panel (Settings → Integrations) that already has docs for the OpenAPI case. Happy to add an MCP-specific note in [open-webui/docs](https://github.com/open-webui/docs) if the maintainer wants it as a follow-up.
- [x] **Dependencies:** None added or upgraded.
- [x] **Testing:** Verified manually via Playwright-driven UI flow (signup → enable permission → add MCP server through the modal → enable in chat integrations menu → send tool-use prompt) against `openai/gpt-oss-120b` over a real LLM gateway. MCP server logs the full session lifecycle including `CallToolRequest` per chat. Edge case tested: missing `info.id` is now refused with an inline error event rather than silently collision-bucketing into a shared namespace.
- [x] **Agentic AI Code:** This PR was drafted with AI assistance and then went through multiple rounds of human review and manual testing (including a Linus-style code review that prompted refactoring the duplicated MCP setup into shared helpers used by both admin and user paths).
- [x] **Code review:** Self-reviewed; refactored the diff to extract three shared module-level helpers (`_open_mcp_connection`, `_make_mcp_tool_callable`, `_register_mcp_tools`) so the admin and user paths share the connect / list / register logic rather than duplicating ~90 lines.
- [x] **Design & Architecture:** No new settings introduced — reuses the existing `features.direct_tool_servers` permission that already gates user OpenAPI tool servers, so admins get per-group control for free. No new endpoints, no DB migrations, no new env vars.
- [x] **Title Prefix:** `feat:`

# Changelog Entry

### Description

Lets users add MCP tool servers to their personal `Settings → Integrations`, the same way they already can for OpenAPI tool servers. Gated by the existing `features.direct_tool_servers` permission so admins keep per-group control.

The MCP type toggle has been hidden in user-direct mode since the original "experimental mcp support" commit (`777e81f7a`, 2025-09-25) and the backend's `direct_tool_servers` path had no MCP branch — so even if a user had got an MCP entry into their settings, it would have been silently dropped at chat time. This PR wires both ends together.

Closes #24620. Relates to #18453, #19313, #24432.

### Added

- Backend: `_open_mcp_connection`, `_make_mcp_tool_callable`, and `_register_mcp_tools` module-level helpers in `backend/open_webui/utils/middleware.py`, shared by the admin (`server:mcp:<id>`) and user-direct (`direct_mcp_<id>`) tool resolution paths.
- Backend: branch for `type='mcp'` in the `direct_tool_servers` loop of `process_chat_payload` so user-stored MCP entries are actually resolved at chat time.
- Frontend: short-circuit in `getToolServersData` (`src/lib/apis/index.ts`) for `type='mcp'` — the browser can't introspect MCP, so the entry passes through with empty `specs` and a normalized `info.title`; backend resolves specs at chat time.

### Changed

- `backend/open_webui/utils/middleware.py`: refactored the admin MCP loop (formerly ~40 inline lines per call site) to use the new shared helpers — net deletion despite adding the user path.
- `src/lib/components/AddToolServerModal.svelte`: lifted the `{#if !direct}` gate around the OpenAPI↔MCP type toggle, and surfaced the required `ID` field in direct mode when type is MCP. All other admin-only fields (access grants, function-name filter, extra headers, OAuth 2.1 dynamic client registration) remain hidden in direct mode.

### Removed

- _(nothing user-facing)_

### Fixed

- Missing-id case: previously would have silently bucketed all user MCP servers with empty `info.id` into the same namespace and lost the collision protection. Now logs a warning and emits a `chat:message:error` event with `"MCP tool server requires an id"` instead of falling back to `'unnamed'`.

### Security

- `MCPClient` only supports streamable HTTP / SSE transports — there is no stdio transport anywhere in the codebase. User-direct MCP therefore cannot spawn local processes; users can only point at URLs they could already reach from their own browser.
- `features.direct_tool_servers` is `False` by default and is enforced on save in `backend/open_webui/routers/users.py` (the existing check that strips `toolServers` from settings updates for users without the permission). Admins can disable it per-group via the existing permissions UI.
- OAuth surface for user-direct stays narrow: `bearer`, `none`, `session` only. The `system_oauth`, `oauth_2.1`, `oauth_2.1_static` options remain admin-only because they require infrastructure (`__oauth_token__` in extra_params, `oauth_client_manager`) that's tied to admin-defined connections.
- User MCP clients and tool spec names are prefixed `direct_mcp_<id>` so they cannot shadow or be shadowed by admin-defined MCP servers sharing an `info.id`. The prefix uses only characters valid in OpenAI function names (`^[a-zA-Z0-9_-]+$`).

### Breaking Changes

- None.

---

### Additional Information

This PR re-opens the case for #18453, which was closed as "intended" behavior. The reasoning for relaxing the restriction now is in the **Security** section above — the previous concerns (transport sandboxing, OAuth surface, permission gating) are structurally handled or already covered by existing admin controls.

A separate `features.direct_mcp_servers` permission to gate MCP independently of OpenAPI was considered and deliberately skipped (YAGNI for a first cut — the bigger concern was always stdio, which is handled at the transport layer, and the per-group permission already exists). Easy follow-up if a use case appears.

The cloud / hosted-HTTP MCP server case (gitmcp.io, huggingface.co/mcp, future first-party MCP endpoints from model providers) is where this branch is most valuable — running a sidecar `mcpo` proxy per user just to expose an HTTP MCP endpoint as OpenAPI is operational overhead this PR removes.

### Screenshots or Videos

_[Add screenshots if you want — the Settings → Integrations modal with the type toggle visible, the integrations menu in the chat input showing the user's MCP server, and a tool-call round-trip would all make good evidence.]_

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -2236,6 +2236,65 @@ def strip_skill_mentions(messages: list[dict]) -> None:
 
 
 
+async def _open_mcp_connection(
+    connection: dict,
+    *,
+    request,
+    user,
+    server_id: str,
+    metadata: dict,
+    extra_params: dict,
+) -> tuple[MCPClient, list[dict]]:
+    headers, _ = await build_tool_server_headers(
+        connection, request, user,
+        server_id=server_id, metadata=metadata, extra_params=extra_params,
+    )
+
+    client = MCPClient()
+    await client.connect(
+        url=connection.get('url', ''),
+        headers=headers if headers else None,
+    )
+
+    function_name_filter_list = connection.get('config', {}).get('function_name_filter_list', '')
+    if isinstance(function_name_filter_list, str):
+        function_name_filter_list = function_name_filter_list.split(',')
+
+    tool_specs = await client.list_tool_specs()
+    if function_name_filter_list:
+        tool_specs = [
+            spec for spec in tool_specs
+            if is_string_allowed(spec['name'], function_name_filter_list)
+        ]
+    return client, tool_specs
+
+
+def _make_mcp_tool_callable(client: MCPClient, function_name: str):
+    async def tool_function(**kwargs):
+        return await client.call_tool(function_name, function_args=kwargs)
+    return tool_function
+
+
+def _register_mcp_tools(
+    *,
+    client: MCPClient,
+    tool_specs: list[dict],
+    registration_key: str,
+    mcp_clients: dict,
+    tools_dict: dict,
+) -> None:
+    mcp_clients[registration_key] = client
+    for tool_spec in tool_specs:
+        spec_name = f'{registration_key}_{tool_spec["name"]}'
+        tools_dict[spec_name] = {
+            'spec': {**tool_spec, 'name': spec_name},
+            'callable': _make_mcp_tool_callable(client, tool_spec['name']),
+            'type': 'mcp',
+            'client': client,
+            'direct': False,
+        }
+
+
 async def connect_mcp_server(
     request,
     server_id: str,
@@ -2243,7 +2302,7 @@ async def connect_mcp_server(
     metadata: dict,
     extra_params: dict,
 ) -> tuple[MCPClient, list[dict]] | None:
-    """Resolve an MCP server connection, authenticate, and return (client, tool_specs).
+    """Look up an admin MCP server by id, access-check, connect, and return (client, tool_specs).
 
     Returns None if the server is not found or access is denied.
     """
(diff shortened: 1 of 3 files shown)


TALLEC-Scott:
@tjbck Any explanations as to why this PR was closed ?

Classic298:
@TALLEC-Scott yes the explanation is in the issue comments

PR #12472, opened 2025-04-04
[WIP] feat: Autofill ComfyUI workflow node inputs

silentoplayz:
# Pull Request Checklist

- [x] **Target branch:** Please verify that the pull request targets the `dev` branch.
- [x] **Description:** Provide a concise description of the changes made in this pull request.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [x] **Documentation:** Have you updated relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs), or other documentation sources?
- [x] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [x] **Testing:** Have you written and run sufficient tests to validate the changes?
- [x] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [x] **Prefix:** The prefix "feat" was used to categorize this pull request.

# Changelog Entry

### Description

Refactored the ComfyUI image generation settings to dynamically parse uploaded/pasted workflow JSON (API format). This replaces the previous hardcoded node configuration, allowing users to configure any primitive input (text, number, boolean) found within their specific workflow by specifying the target Node ID(s). Also includes improved input validation, error handling, and UX enhancements across the image settings page.

### Added

-   Dynamic parsing of ComfyUI workflow JSON (API format) to discover configurable primitive inputs (text, numbers, booleans).
-   UI section to display discovered ComfyUI inputs (`Node Type::Input Key`) and allow users to specify the target `Node ID(s)` for backend control.
-   `validateJSON` function specifically checks for ComfyUI API format structure.
-   Input validation patterns for `IMAGE_SIZE` (`1024x1024`) and ComfyUI `Node IDs` (comma-separated numbers).
-   Type `number` and range/step attributes for numeric inputs (`CFG Scale`, `Steps`).
-   Default model suggestions in the datalist for OpenAI (`dall-e-3`, `dall-e-2`) and Gemini (`gemini-pro-vision`) engines when applicable.
-   Toast notifications for workflow parsing success/failure and discovered nodes.
-   Placeholder text for OpenAI and Gemini API Base URLs.
-   `on:change` handler for the ComfyUI workflow textarea to trigger re-parsing on manual edits.

### Changed

[...]

PR #25140, opened 2026-05-28
feat: auto-map ComfyUI workflow nodes from imported API JSON

silentoplayz:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** This PR references an existing [Issue](https://github.com/open-webui/open-webui/issues) or [Discussion](https://github.com/open-webui/open-webui/discussions) — Relates to #12472 (prior PR for the same feature). Closes discussion: https://github.com/open-webui/open-webui/discussions/25139
- [x] **Testing:** Manual tests have been performed to verify the fix/feature works correctly and does not introduce regressions. Screenshots or recordings are included where applicable.
- [x] **Title Prefix:** The PR title uses one of the following prefixes: `feat`

---

# Changelog Entry

### Description

Setting up ComfyUI as an image generation backend in Open WebUI currently requires users to manually look up and wire node IDs for their prompt, model, dimensions, and other inputs — a tedious, error-prone task that creates a significant barrier to entry.

This PR eliminates that friction entirely. Users now simply import or paste their ComfyUI **API-format** `workflow.json` and Open WebUI automatically:

1. Parses every configurable scalar input across all nodes in the workflow.
2. Renders a dynamic table in Admin → Settings → Images, with one row per node input, showing the node type, input key, and node ID.
3. Injects the user's prompt, model, image size, and generation parameters into the correct nodes at runtime — with no manual wiring required.

The feature is fully backward-compatible: workflows already configured with manual node IDs continue to work unchanged.

### Added

- **Workflow JSON import UI** — "Import Workflow" file-picker and "Paste JSON" textarea in the ComfyUI section of Admin → Settings → Images (for both generation and edit workflows).
- **Auto-parse on import** — `parseAndPopulateWorkflowNodes()` in `Images.svelte` reads the uploaded API-format JSON and produces one table row per scalar input found across all nodes.
- **Dynamic node input table** — Each row shows `class_type`, input key, and a comma-separated editable Node IDs field. The user can still correct or override any node ID as needed.
- **Reconciliation on re-import** — Previously saved node IDs are preserved when a workflow is re-imported (unless "New Import" is explicitly triggered, which resets to workflow defaults).
- **`_apply_workflow_nodes()` backend function** (`comfyui.py`) — Applies saved node mappings to the workflow dict before submission, with semantic dispatch:
  - `text` / `prompt` / `positive` → user prompt text
  - `ckpt_name` / `unet_name` → currently selected model
  - `width` / `height` → image dimensions from generation settings
  - `steps` / `cfg` / `seed` / `denoise` / `scheduler` / `sampler_name` → generation parameters
  - Any other key with a saved `value` → static passthrough
- **`get_image_data()` in `images.py`** — Fetches the generated PNG from ComfyUI's `/view` endpoint with SSRF protection (same-origin check against the configured base URL).
- **`upload_image()` helper in `images.py`** — Wraps the file upload, chat file linking, and URL construction into a reusable function, used by both ComfyUI and the edit-image paths.
- **Pydantic models** `ComfyUINodeInput` and `ComfyUIWorkflow` (`comfyui.py`) — Typed request/response contracts for the node configuration and workflow payload.

### Fixed

- **ComfyUI API endpoint path** — `queue_prompt` was POSTing to `{base_url}/prompt`; newer ComfyUI versions moved this to `{base_url}/api/prompt`. Fixed.
- **Overly broad model-name injection** — The previous `node_key.endswith('_name')` heuristic matched `sampler_name`, `vae_name`, and `clip_name` and overwrote them with the checkpoint filename, causing ComfyUI to reject the workflow with `400 Bad Request`. Narrowed to `ckpt_name` and `unet_name` only.
- **Positive/negative prompt merging** — The workflow parser was grouping inputs by `class_type::key`, which merged the positive and negative `CLIPTextEncode` nodes into a single row. Changed the grouping key to `nodeId::class_type::key` so every node is always treated as an independent row.
- **`UnboundLocalError` in `Chats.insert_chat_files()`** (`chats.py`) — A de-duplication query on line 1627 passed `db=session` before the `async with get_async_db_context(db) as session:` block that creates `session`. Changed to `db=db` (the caller-supplied session or `None`).
- **WebSocket null-guard** (`comfyui.py`) — Added safety checks for empty/`None` WebSocket history responses to avoid `KeyError` / `TypeError` on fast or failed generations.

### Security

- `get_image_data()` validates that image-fetch URLs share the exact scheme, hostname, and port as the admin-configured ComfyUI base URL before skipping the standard SSRF `validate_url()` check. This prevents a misconfigured workflow from redirecting image fetches to internal network addresses.

---

### Additional Information

**Design notes on the detection and storage format.**

Detected entries are keyed by `nodeId::class_type::inputKey`. The node id is part of the key on purpose, so two nodes sharing a class and an input, such as the positive and negative `CLIPTextEncode`, stay separate rows instead of merging into one. Array valued inputs are links between nodes rather than configurable values, so they are skipped.

Keys that map to `payload.prompt` are ambiguous whenever several nodes share them, which is exactly the positive and negative prompt case. Detection deliberately leaves `node_ids` empty for those, so the admin picks which nodes receive the prompt rather than the feature guessing.

On the backend, a node is treated as a model loader when its key ends in `_name`, covering `ckpt_name` and `unet_name`. Both the legacy type names and the new `ClassName::key` form are accepted, so an existing configuration keeps working.

- Relates to draft PR [#12472](https://github.com/open-webui/open-webui/pull/12472) — this is a complete, production-quality implementation of the same concept.
- No new Python or Node.js dependencies are introduced.
- The feature degrades gracefully: if `COMFYUI_WORKFLOW_NODES` is empty or the workflow string is blank, generation falls back to the raw workflow JSON without any node injection.
- The `log.exception` added to the `image_generations` catch block is a genuine improvement — the previous bare `except Exception as e` silently swallowed all errors with no stack trace.

### Screenshots or Videos

> Manual testing was performed end-to-end with a live ComfyUI 0.22.0 instance running `z_image_turbo_bf16.safetensors` (Lumina2 architecture). The test workflow used 10 nodes including `KSampler`, `CLIPTextEncode` ×2, `UNETLoader`, `VAELoader`, `CLIPLoader`, `VAEDecode`, `EmptySD3LatentImage`, and `SaveImage`. A ferret-on-a-beach prompt was successfully generated and returned in the chat within ~4 seconds.

image

image

---

The following steps were tested manually before opening this PR:

1. **Go to** Admin → Settings → Images → Set engine to `ComfyUI` with base URL `http://127.0.0.1:8188`.
2. **Click "Import Workflow"** and upload a ComfyUI API-format `workflow.json`. Verify the node input table populates automatically with one row per scalar input.
3. **Confirm node IDs** in the table match the node IDs visible in ComfyUI's workflow editor.
4. **Save settings.** Open a new chat and send a message triggering image generation (e.g. *"Generate an image of a cat"*).
5. **Verify** the image appears inline in the chat — no manual node ID entry was required at any point.
6. **Re-import** the same workflow JSON. Verify that previously saved node IDs are preserved (reconciliation).
7. **Verify** that the negative prompt node (if present) does not receive the positive prompt text (distinct node IDs in separate table rows).
8. **Check** Admin → Settings → Images → Edit Workflow section has the same import/parse behavior as the generation workflow section.

---

### Agentic AI Clause

> [x] **Agentic AI Code**: This Pull Request was prepared with the assistance of an AI copilot and has been thoroughly reviewed and manually tested by the author to ensure quality and adherence to project standards.

---

> [!NOTE]


Gladon4:
I have tested this a bit as it seems useful for a thing I am working on and I have noticed some stuff. First of it does what it says and seems to work without much issue out of the box, but there are two things I noticed.

- The prompt now gets applied to the prompt as well as the negative prompt (to be more specific all inputs called text, which seems intentional?). I don't think that is useful. This can also not be changed as the keys cannot be renamed (as they can be without this change). You can stop this from happening by not assigning the node ids for the negative prompt, but I don't see the point, in adding this, if the nodes need to be unassigned again.
- I thought this would also expose the additional parameters to the Open WebUI API (/api/v1/images/generations), but it doesn't (or at least I cant get it work). And as the WebUI itself doesn't set the parameters (I think), this would be nice, so you can actually use all these parameters. This is also more, why I tried this, because I didn't want to make a fork myself, if someone else already did something that would work.

I hope this helps and makes sense.

silentoplayz:
> I have tested this a bit as it seems useful for a thing I am working on and I have noticed some stuff. First of it does what it says and seems to work without much issue out of the box, but there are two things I noticed.
> 
>     * The prompt now gets applied to the prompt as well as the negative prompt (to be more specific all inputs called text, which seems intentional?). I don't think that is useful. This can also not be changed as the keys cannot be renamed (as they can be without this change). You can stop this from happening by not assigning the node ids for the negative prompt, but I don't see the point, in adding this, if the nodes need to be unassigned again.
> 
>     * I thought this would also expose the additional parameters to the Open WebUI API (/api/v1/images/generations), but it doesn't (or at least I cant get it work). And as the WebUI itself doesn't set the parameters (I think), this would be nice, so you can actually use all these parameters. This is also more, why I tried this, because I didn't want to make a fork myself, if someone else already did something that would work.
> 
> 
> I hope this helps and makes sense.

Hey, appreciate you testing this and taking the time to write up what you found. Both points were valid and I've pushed fixes for them.

**The prompt/negative prompt issue:** The auto-detection was eagerly assigning node IDs to everything it found, so both CLIPTextEncode nodes (positive and negative) with a `text` key would get the prompt written into them. I've changed it so `text`/`prompt`/`positive` keys are left with empty node IDs on import, and you assign the ones you actually want the prompt injected into. I also brought back the editable key fields that were there before this change, so you can rename keys again if needed.

**API parameters:** You were right that these weren't being passed through. I've added two things to `CreateImageForm`:

- `seed` as a direct field
- `extra_params` as an open dict for anything else the workflow supports

So you can now do something like:

```json
POST /api/v1/images/generations
{
  "prompt": "a cat",
  "seed": 42,
  "extra_params": {
    "cfg": 7.5,
    "sampler_name": "dpmpp_2m",
    "scheduler": "karras",
    "denoise": 0.85
  }
}
```

Those values get matched against the workflow node input keys and injected accordingly. Known fields like prompt, width, height, steps, and seed still take priority through their dedicated fields, and `extra_params` fills in the rest.

Let me know if anything's still off or if you run into other issues with it.

Gladon4:
Thanks for the quick fixes. Based on a very quick test, it seems to be working. I'll be testing it out a bit more over the next days.

Gladon4:
The current version seems to be working fine. Thanks for implementing this.

One thing:
I wonder how difficult it would be and if it is in the scope of this pull request to change the names of the fields for the API. Currently I can add the negative_prompt to the API request, but I can't really add it to / set it in the ComfyUI workflow. Renaming the parameter in the web UI, it gets addressed correctly, but because the field in the node is called "text", it will just add a new input with the name "negative_prompt", which of course is ignored. So what I mean is, would be easy to separate the name of the parameter towards the API and the comfy workflow? Then one could set the API name of the parameter to "negative_prompt" and the comfy input name to "text". 
I guess this could then also work for other parameters, especially when you have multiple nodes with the same parameter which you want to control independently, but the negative prompt seems to be the most useful one.
Like I said, maybe this isn't in scope for this pull request and I also don't know how complicated this would be, I have only briefly looked at the structure of the Open WebUI source code.

silentoplayz:
Rebased onto the latest `dev` (previous head `3fad11589` → `6bca576d6`) and resolved all conflicts. Notes for reviewers:

- **Adapted to the new config access pattern** in `routers/images.py`: the model-node detection loop now reads `image_config.COMFYUI_WORKFLOW_NODES` via `get_image_config()` instead of `request.app.state.config`, matching the refactored endpoint style.
- **Re-styled the admin UI to the new Settings components**: the auto-detected workflow node tables (both Image Generation and Edit Image sections) now render inside `AdminSettingField` wrappers with the shared `inputClass`, consistent with the page's recent restyle. The auto-detect hint moved into the field's `description`, and the "Auto-detected" badge, `class_type` labels, and editable Key/Node Ids inputs are unchanged in behavior.
- **Dropped an out-of-scope hunk**: an earlier rebase had picked up a one-line `db=session` → `db=db` change in `models/chats.py`; current `dev` already has the equivalent code, so the branch is now scoped strictly to the three PR files.

No functional changes to the feature itself. Verified after rebase: backend modules compile, prettier passes, full production frontend build succeeds, and manually re-tested the flow (workflow.json API-format import → auto-detection → ambiguous prompt keys left unassigned → save/reload reconciliation).

image

silentoplayz:
@Gladon4 Thanks again for testing this and for the follow-up — sorry for the slow reply on that last point. Your diagnosis was right, and I traced exactly why it happens.

**Why renaming the field doesn't do what you wanted**

The auto-detection stores each entry as `type: "<class_type>::<inputKey>"` (e.g. `CLIPTextEncode::text`) with `key: "text"`, and for that dynamic format the backend decides *which payload value to inject* from the **`key`** field. So `key` is doing double duty — it is both the semantic slot ("this receives the prompt") and the literal ComfyUI input name to write. Renaming it to `negative_prompt` therefore changes both at once, which is why you end up with a new, ignored `negative_prompt` input on the node instead of the value landing in `text`.

The separation you're describing does exist in the older hardcoded format, where `type` and `key` are independent: an entry with `type: negative_prompt` and `key: text` writes `payload.negative_prompt` into the node's `text` input, which is precisely the mapping you want. `_apply_workflow_nodes` still honors those legacy semantic types for backward compatibility. What's missing is a way to *author* that from the admin UI — the auto-detected table only exposes **Key** and **Node Ids** for editing, and renders `type` as a read-only label.

**Scope**

I'm going to leave that out of this PR. Closing the gap properly means either exposing the semantic type as an editable field or adding a distinct "API parameter name" alongside the node input name — a config-shape and UI decision that deserves its own discussion rather than being bolted onto a change that's already ~570 lines across three files. It also builds cleanly on top of this one: once node IDs are auto-mapped, adding a name-decoupling field is a contained follow-up rather than a rewrite.

If you'd like to pursue it, a Discussion or issue describing the use case would be the right next step, and I'm happy to take a look — multiple nodes sharing an input name that you want to drive independently is a good argument for it, and worth capturing in its own thread where it won't get lost. Feel free to tag me.

For now this PR stays scoped to what you confirmed working: auto-mapping node IDs from an imported API-format workflow, with the ambiguous prompt keys deliberately left unassigned so the admin picks which nodes receive the prompt.

silentoplayz:
Rebased onto current `dev` and taken out of draft — **all four checks are now green** (`Format & Build`, `Ruff Format` 3.11 + 3.12, `Unit Tests`), and the PR reports mergeable with no conflicts.

What changed since the last push, for reviewers:

- **Rebased** `6bca576d6` → `a51c21a0e` onto current `dev`. The rebase was conflict-free and the feature diff is unchanged at 3 files, +565/−230. The two previously-failing `Ruff Format` checks were caused purely by the stale base — 11 files on the old base were unformatted; on current `dev` the whole repo reports clean, and this PR's two Python files were already formatted either way.
- **Added one commit syncing the translation catalogs** (`npm run i18n:parse`). `Format & Build` had been failing at its *Ensure working tree is clean* step because the admin UI restyle introduced eight new `$i18n.t()` strings — the workflow-parsing feedback messages and the `Auto-detected` badge — that weren't in the catalogs. The parse also prunes three keys the restyle made unreachable (`*Prompt node ID(s) are required for image generation`, `Map workflow node inputs used for image edits.`, `Map workflow node inputs used for image generation.`); I confirmed each has no remaining reference in `src/` outside the locale files. Locale files are the parser's own output, committed separately from the feature commits so the functional diff stays reviewable on its own.

No functional changes to the feature in either commit. Verified locally after the rebase: whole-repo `ruff format --check` clean, `npm run format` produces no changes, `npm run i18n:parse` leaves the tree clean, production build succeeds, and `vitest run` passes.

The behavior confirmed working by @Gladon4 earlier in this thread is unchanged: importing an API-format `workflow.json` auto-detects the configurable primitive inputs and fills in node IDs, with ambiguous prompt keys (`text` / `prompt` / `positive`) deliberately left unassigned so the admin explicitly chooses which nodes receive the prompt.

silentoplayz:
Worth recording a forward compatibility hazard I hit while testing this branch, since the failure is silent.

Auto detection stores node types in the new dynamic format, `ClassName::inputKey`, which is read by the `elif '::' in node_type` branch this PR adds. Every version without that branch, including current `dev`, matches node types against the literals `model`, `prompt`, `negative_prompt`, `image`, `width`, `height`, `n`, `steps` and `seed` in an if/elif chain that has no else. An unrecognized type is therefore a no op rather than an error.

The result on a build without this PR, using a mapping this PR wrote:

- the prompt is never written into the workflow
- the seed is never randomized, because `comfyui.py` only randomizes when a node is typed exactly `seed`

ComfyUI then receives the saved workflow unchanged, treats it as a repeat of the previous run, and serves the cached image. Generation appears to succeed, returns quickly, and produces the same image for every prompt, in chat and in the Images playground alike. Nothing is logged. The identical bytes also hash to the same file record, so the same file id comes back on each generation, which is what makes it visible in the network log.

I reproduced this by mapping a workflow on this branch and then running generation on `dev`.

Worth considering before merge, since it also describes what a rollback looks like: writing a marker alongside the mapping, or keeping the legacy type names as the stored form and treating `ClassName::inputKey` as an input to detection rather than a persisted value. A log line on an unmatched node type in `_apply_workflow_nodes` would at least make the state visible rather than silent.

PR #25150, opened 2026-05-28
BREAKING/perf/feat: cache-optimal file context in native FC mode

Classic298:
# Cache-optimal file context routing in native function calling mode

Open WebUI's current handling of user-attached files, knowledge bases, notes, and other retrievable items produces near-pathological prompt-cache behavior on every major provider (OpenAI, Anthropic, Google). This PR identifies three distinct cache-destroying scenarios in the existing code, traces each to its root cause, and fixes all three behind a single coordinated change that activates **only** when every one of these holds: native function calling mode, the `builtin_tools` capability and its `knowledge` category are enabled, the `file_context` capability is enabled, the caller did not supply its own `tools` array, the admin has not globally forced full content via `RAG_FULL_CONTEXT` or `BYPASS_EMBEDDING_AND_RETRIEVAL`, and the chat actually carries at least one chunked (non-full-context) file or collection. Default and prompt-based function calling behavior is preserved byte-for-byte.

---

## Background: how the existing file-context pipeline works

When a user attaches a file, a knowledge base, or otherwise references content in the chat, the request body carries those items in `body.metadata.files`. Each item has a `type` (`file`, `collection`, `text`, `note`, `chat`, `url`, legacy collections) and a `context` flag (`'full'` or omitted = chunked retrieval). The pipeline that turns those items into prompt content lives mostly in `backend/open_webui/utils/middleware.py`:

1. **`chat_completion_files_handler`** iterates `body.metadata.files`. For chunked items it generates retrieval queries from the conversation and calls `get_sources_from_items` with the result. For full-context items it bypasses the query generation and pulls the entire content. Either way it returns a flat `sources` list where each source is `{'source': {item dict}, 'document': [chunks or full body], 'metadata': [...]}`.
2. **`apply_source_context_to_messages`** renders those sources into a `<context>…</context>` block via `rag_template(RAG_TEMPLATE, context, user_message)` and inserts the result into the messages — either appended to the system prompt (if `RAG_SYSTEM_CONTEXT=true`) or prepended to the last user message (default).
3. The **tool-call loop** inside `streaming_chat_response_handler` handles models that respond with `tool_calls`. After each tool result it restores the pre-RAG state of system and user messages, then re-applies `rag_template` with the **initial sources plus all accumulated tool-call sources** (the latter rendered with `include_content=False` so only `<source id="N"/>` citation markers are emitted — the actual content already lives in the tool result messages).

This pipeline is fine if you don't care about prompt caching. It is the wrong shape if you do, for three independent reasons described next.

---

### Scenario 1 — `RAG_SYSTEM_CONTEXT=false` (default): the second-newest user message mutates every turn

This is the OWUI default. The rendered RAG template is **prepended to the latest user message** via `add_or_update_user_message(content, messages, append=False)`. That helper only touches `messages[-1]`. The persisted chat history stores user messages **without** the RAG prepend — RAG is freshly re-applied on every request from the clean chat state.

Trace a two-turn chat with a file attached:

**Turn 1 sent to provider:**

```
[system]
[user₁ = <context>…retrieved chunks for user₁'s query…</context>\n\noriginal text of user₁]
```

**Turn 2 sent to provider:**

```
[system]
[user₁ = original text of user₁]                                     ← prepend gone
[assistant₁]
[user₂ = <context>…retrieved chunks for user₂'s query…</context>\n\noriginal text of user₂]
```

**The mutation:** `user₁` was sent in turn 1 wrapped in RAG content; in turn 2 it is sent bare. Same message ID, different bytes. Every prefix-based cache (OpenAI, Anthropic with `cache_control`, Gemini) breaks at the point where `user₁` starts diverging. The provider can only cache the prefix up to and including `[system]` — assistant and user messages from any prior turn are wasted.

**With Anthropic specifically:** Open WebUI does not emit any `cache_control: {type: "ephemeral"}` breakpoints anywhere (verified: only `Cache-Control` HTTP headers in `security_headers.py`, unrelated). Anthropic requires explicit markers — without them the API does **not** consult the cache at all. So Anthropic users currently get zero caching regardless of message structure, and on top of that pay full input-token price for the file body on every turn.

**Impact:** files frequently make up the majority of total prompt tokens. Putting them in the volatile region of the prefix means most of the chat's input cost can never be cached. Long chats with attached files pay nearly the full input cost on every turn.

### Scenario 2 — `RAG_SYSTEM_CONTEXT=true` with any chunked retrieval: the system prompt mutates every turn

Setting `RAG_SYSTEM_CONTEXT=true` moves the rendered RAG template into the **system prompt** via `add_or_update_system_message(content, messages, append=True)`. For **full-context** items this is great: file bodies are stable across turns, the system prompt is stable across turns, every provider caches the entire history forever. If the user adds or removes a full-context item, the system prompt mutates for exactly one turn, then stabilizes.

But the same code path is used for **chunked retrieval** items. `chat_completion_files_handler` re-runs the retrieval pipeline on every turn, generating different queries from the new conversation state and pulling different top-k chunks. Those chunks are embedded inside `<context>…</context>` in the system prompt. Different query → different chunks → different system prompt.

**Trace:**

**Turn 1 sent:**

```
[system = original + <context>chunks retrieved for "what is X?"</context>]
[user₁ = "what is X?"]
```

**Turn 2 sent:**

```
[system = original + <context>chunks retrieved for "and what about Y?"</context>]   ← mutated
[user₁]
[assistant₁]
[user₂ = "and what about Y?"]
```

**The mutation:** the system prompt is the very first thing in the prefix. Mutating it invalidates **every** cache breakpoint downstream. For all providers. On every turn.

**Impact:** for any deployment with chunked-retrieval files (the default mode and the most common configuration) and `RAG_SYSTEM_CONTEXT=true`, the chat history cache is effectively never used. The user pays full input cost on every single turn for the entire conversation history, the system prompt, and the chunks.

The user is stuck choosing between Scenario 1 (cache breaks at second-newest user message) and Scenario 2 (cache breaks at system prompt). There is no setting that produces good caching for the common case of chunked retrieval.

### Scenario 3 — the tool-call loop re-renders the RAG template on every iteration

This scenario is independent of `RAG_SYSTEM_CONTEXT` and stacks on top of both Scenarios 1 and 2. It triggers whenever the model produces tool calls.

When a model in native function calling mode responds with `tool_calls`, the existing code runs the following on **every iteration of the tool-call loop**:

1. Restore the user message to its pre-RAG state (`set_last_user_message_content`).
2. Restore the system prompt to its pre-RAG state (`replace_system_message_content`).
3. Rebuild the source context by concatenating `get_source_context(metadata['sources'])` (initial file sources with full content) with `get_source_context(all_tool_call_sources, include_content=False)` (citation markers only for tool sources — `<source id="N"/>` with empty body, because the actual content already lives in the tool result messages).
4. Re-apply `rag_template` to that combined context.
5. Inject the rendered template into the system prompt or last user message depending on `RAG_SYSTEM_CONTEXT`.

The purpose of this re-application is to give the model `<source id="N"/>` citation markers it can use to cite tool-derived content using the `[N]` format dictated by `DEFAULT_RAG_TEMPLATE`. Each new tool call adds one more such marker to the RAG template content.

**Trace a research-style turn where the model fires six tool calls:**

```
user: "go research X and Y for me"

iter 0 (initial request):
  [system = base + <context>initial sources</context>]   (or in user msg if RAG_SYSTEM_CONTEXT=false)
  [user]

iter 1 (after tool call 1):
  [system = base + <context>initial sources + <source id="2"/></context>]   ← mutated
  [user] [tool_call_1] [tool_result_1]

iter 2:
  [system = base + <context>initial sources + <source id="2"/><source id="3"/></context>]   ← mutated again
  [user] [tool_call_1] [tool_result_1] [tool_call_2] [tool_result_2]

iter 3:
  [system = base + <context>initial sources + <source id="2"/><source id="3"/><source id="4"/></context>]
  …
```

**The mutation:** the system prompt (or last user message) carries an ever-growing list of empty `<source id="N"/>` markers, one new one per tool call. Every iteration mutates the prefix at the injection point. Every iteration's cache breaks at that mutation. For long tool-call chains (agentic flows, multi-step research, browse-the-web style tasks), the user pays **full input-token cost on every iteration** for the entire conversation history including all previous tool results — exactly the content that, by being immutable, should have been the easiest to cache.

This is the worst of the three scenarios because:

- It triggers on the workload that benefits most from caching (long tool-call chains accumulate a lot of context).
- It triggers even when the user has set `RAG_SYSTEM_CONTEXT=true` and attached only full-context files (Scenario 2's "happy path") — the tool-loop re-application brings the mutation problem right back.
- The empty `<source id="N"/>` markers add essentially no value: the tool result content already carries source/filename metadata that the model can cite from directly.

---

## What this PR does

The fix is structural and addresses all three scenarios simultaneously, behind the activation gate described in the intro (native FC + `builtin_tools`/`knowledge` + `file_context` + no caller-supplied tools + no forced-full-content + at least one chunked attachment).

### Decouple the static and dynamic parts of the `<context>` block

`DEFAULT_RAG_TEMPLATE` is left intact in shape (`citation instructions + <context>{{CONTEXT}}</context>`). What changes is **what goes into `{{CONTEXT}}`** and **where the rendered template lands**. The `<context>` block becomes a stable chat-level inventory rather than a per-turn retrieval dump:

```xml
<context>
  <full_context_files>
    <source id="1" name="contract.pdf" resource-type="file" resource-id="file-abc">…full body…</source>
    <source id="2" name="kb-a-doc" resource-type="collection" resource-id="kb-uuid">…full body…</source>
  </full_context_files>
  <retrievable_files>
    <retrievable_file id="file-def" name="appendix.pdf" resource-type="file">Use query_attached_files with this id to inspect or search this item.</retrievable_file>
    <retrievable_file id="kb-other" name="my-knowledge-base" resource-type="collection">Use query_attached_files with this id to inspect or search this item.</retrievable_file>
  </retrievable_files>
</context>
```

Two sub-blocks with two different cache profiles:

- **`<full_context_files>`** holds citeable content. `<source id="N">` tags with full bodies, identical to today's source format. These are full-context files, full-context knowledge bases, and natively-full types (`text`/`note`/`chat`/`url`). They are static for the lifetime of the chat — they only change when the user adds/removes/toggles an attachment.
- **`<retrievable_files>`** holds inventory listings. `<retrievable_file id="…">` tags with a one-line hint pointing at `query_attached_files`. **No body content.** Critically, these are **not** `<source>` tags — the rendered RAG template's citation guidance only applies to `<source id="…">`, so the model has no incentive to fabricate citations against entries that have no evidence in them. To use an attached file as evidence, the model must first call `query_attached_files` and cite from the chunks in the tool result.

> Note: the two sub-block wrappers only appear when the chat is in *mixed* mode (at least one chunked item alongside content sources). A chat with only chunked items renders just the `<retrievable_files>` block; a chat that does not activate the new path at all renders the flat `<source>` shape exactly as today (backward compatible — see below).

`DEFAULT_RAG_TEMPLATE`'s citation guidelines are restructured so the distinction is explicit: the strict `[id]` rule is scoped to `<source>` tags only; a new bullet tells the model to cite tool-result chunks by their source filename (e.g. `[contract.pdf]`); and a bullet tells it not to cite inventory-style entries directly but to fetch their content via the appropriate tool first. Custom `RAG_TEMPLATE` values (env or admin-configured) keep their own citation guidelines — only the bundled `DEFAULT_RAG_TEMPLATE` constant changes.

### Fix Scenario 1 + Scenario 2 — always route to the system prompt in native FC mode

When the new path is active, the rendered template is forced into the system prompt regardless of `RAG_SYSTEM_CONTEXT`. The relevant call site:

```python
form_data['messages'] = await apply_source_context_to_messages(
    request,
    form_data['messages'],
    sources,
    prompt,
    force_system_message=native_fc_with_builtins,
)
```

This eliminates Scenario 1: nothing prepends to the last user message anymore, so the second-newest user message no longer mutates between turns.

It would *also* trip Scenario 2 except that the volatile retrieval chunks are no longer in the system prompt — they have been moved out (see below). The system prompt now contains only:

- citation instructions (static, identical across the lifetime of `RAG_TEMPLATE`),
- full-context file bodies (static until the file set changes),
- the queryable file roster (static until the file set changes).

All three parts change at the same cadence: only when the user adds, removes, or toggles an attachment. That's typically zero times during the chat. When it does happen, the cost is one turn of cache invalidation, then stable again.

> The forced-system placement only applies when the new path is active — i.e. when the chat carries at least one chunked attachment. A chat with **only** full-context files does not trigger the new path; it keeps the existing `RAG_SYSTEM_CONTEXT` placement and its full bodies were already cache-stable there.

### Fix Scenario 2's chunked-retrieval problem — route chunks through `query_attached_files` instead

The mutating chunks are the root cause of Scenario 2. The fix is to stop auto-retrieving them and route the retrieval through an immutable tool call instead.

In `chat_completion_files_handler`, items are partitioned via a single predicate:

```python
RETRIEVABLE_TYPES = {'file', 'collection'}
should_partition = native_function_calling_with_builtins and not force_full_context

def is_retrievable(item):
    return (
        item.get('context') != 'full'
        and item.get('type') in RETRIEVABLE_TYPES
    )

if should_partition:
    retrievable_items = sorted((i for i in files if is_retrievable(i)), key=lambda i: i.get('id') or '')
    full_context_items = sorted((i for i in files if not is_retrievable(i)), key=lambda i: i.get('id') or '')
else:
    retrievable_items = []
    full_context_items = list(files)
```

- **Full-context items** (and natively-full types like notes, raw text uploads, attached chats, URLs, and any unknown future item type) keep going through `get_sources_from_items` with `full_context=True`. Bodies land in `<full_context_files>` as `<source>` tags. Unchanged from today's behavior for these items.
- **Chunked-retrieval items** (`file` or `collection` with `context != 'full'`) **skip retrieval entirely**. They get a roster-only source entry (`{'source': {…}, 'roster_only': True}`) that `get_source_context` renders as a `<retrievable_file>` tag inside `<retrievable_files>`. No chunks are pulled at this stage. No volatile content enters the system prompt.
- Both partitions are sorted by `id` so the rendered blocks (and the `source_ids` numbering) are byte-stable across turns regardless of how the frontend ordered `metadata.files`.

A new builtin tool **`query_attached_files(query, ids?, count?)`** is added to `backend/open_webui/tools/builtin.py` and registered by `get_builtin_tools` whenever the chat has retrievable items. It is the model's mechanism to actually read the content of items in `<retrievable_files>` when it judges the user query needs them. The tool's implementation:

- **Per-item access control**, matching `view_file` / `view_knowledge_file` posture: `type='file'` items go through `_has_read_access_to_file(user_id, user_role)` (deliberately *not* `__model_knowledge__` — this is chat-scope, not model-scope); `type='collection'` items go through owner / admin / `AccessGrants.has_access(resource_type='knowledge', permission='read')`. Items the user can't read are silently skipped.
- Receives the retrievable item list via the existing `extra_params` plumbing as `__attached_files__`, matching the pattern already used by `query_knowledge_files`'s `__model_knowledge__`. This input is itself gated on `file_context_enabled`, so a model with the capability off can't reach attached content through the tool.
- Optionally scopes by `ids` (the literal value of the `id` attribute on `<retrievable_file id="…">` entries in the system context). Omitted or empty list = search every attached item, matching API convention. A JSON-scalar `ids` argument is wrapped into a list so membership scoping stays a list check (not a substring or dict-key match).
- **`count` defaults to the admin-configured retrieval `TOP_K` and is clamped to that value as a ceiling** — a model cannot exceed the admin's configured retrieval budget.
- **Hybrid search and reranking are inherited automatically** — the tool calls `query_collection`, which internally routes to `query_collection_with_hybrid_search` with the admin's `TOP_K_RERANKER`, `RELEVANCE_THRESHOLD`, `HYBRID_BM25_WEIGHT`, and `RERANKING_FUNCTION` whenever `ENABLE_RAG_HYBRID_SEARCH` is on. No additional plumbing required.
- Resolves `type='file'` items to the `file-{id}` collection name and `type='collection'` items to the access-checked KB id. Returns chunks with `{content, source, file_id, distance}` — same shape as `query_knowledge_files`.

`query_attached_files` is also wired into `get_citation_source_from_tool_result` (the same parser branch as `query_knowledge_files`), so its returned chunks are grouped into citation source entries that populate the frontend citation panel via the events stream. The model is instructed by the updated template to cite tool-result content by its source filename (`[contract.pdf]`) rather than a numeric `[N]`.

The cache benefit: every `query_attached_files` invocation produces a `tool_call` + `tool_result` pair that becomes part of the chat history. Tool calls and tool results are **immutable** — once they exist in history, their bytes never change. So from the iteration after a tool call onward, that tool call and its result cache forever. A long chat with many file retrievals over time produces a long sequence of immutable cached tool pairs, each one a permanent prefix extension. No volatile content anywhere.

### Fix Scenario 3 — stop re-rendering the RAG template on every tool-loop iteration

When the new path is active, the tool-loop's per-iteration RAG re-application is skipped entirely:

```python
all_tool_call_sources.extend(tool_call_sources)
skip_rag_reapply = metadata.get('native_fc_with_builtins', False)
if not skip_rag_reapply and all_tool_call_sources and user_message:
    # existing code path: restore system + user, rebuild source_context,
    # re-apply rag_template, inject per RAG_SYSTEM_CONTEXT.
    …
```

The empty `<source id="N"/>` markers that the existing code added per iteration carried no information. The actual content lives in the tool result message; the source/filename/file_id metadata lives in the tool result JSON. Citation was the only thing the markers enabled, and that's now handled out-of-band: `query_attached_files` tool results are parsed by `get_citation_source_from_tool_result` into citation source entries that populate the frontend citation panel via the events stream, and the updated template tells the model to cite tool-result content by its source filename. The model-facing prompt no longer needs the per-iteration `<source id="N"/>` markers at all.

Result: from the second tool-loop iteration onward, the **entire prefix through the previous tool result is byte-identical** to the previous iteration. Every provider cache-hits the full prefix. Only the new `tool_call` + `tool_result` pair is new content per iteration. A 10-tool-call research turn that today costs 10 full prefix recomputes now costs the prefix once at iteration 1 plus N small tool-pair extensions.

---

## What gets cached after this PR, per scenario

| Situation | Today (default `RAG_SYSTEM_CONTEXT=false`) | Today (`RAG_SYSTEM_CONTEXT=true`) | After this PR (native FC mode) |
|---|---|---|---|
| New turn, no attachment changes, no tool calls | Second-newest user message mutated → cache breaks there. Files re-billed every turn. | If only full-context items: full hit. If any chunked items: system mutates → entire cache miss every turn. | Full hit. System prompt and all prior history cached. Only new user msg + assistant response are new content. |
| New turn, user added/removed/toggled an attachment | Same as above (second-newest user msg always mutates anyway). | 1 turn of system-prompt invalidation, then stable. | 1 turn of system-prompt invalidation, then stable. |
| Turn with N tool calls | Per-iteration RAG mutation on top of the user-msg mutation. ~N+1 cache breaks within the turn. | Per-iteration system-prompt mutation on top of any chunked-retrieval mutation. ~N+1 cache breaks within the turn. | Iteration 1 is new content (initial system + first tool pair). Iterations 2..N hit the entire prefix through iteration N-1's tool result and only add the new tool pair. |
| Long chat, mostly file Q&A | Files never cached. Most input is file content. Most input is re-billed every turn. | Either files are full-context (cache hits) or chunks invalidate system (cache misses). No mixed mode. | Full-context bodies in system: cached. Chunked retrievals via tool calls: cached as immutable tool pairs in history. Both modes coexist cleanly. |

---

## Backward compatibility — default and prompt-based FC mode behavior is preserved EXACTLY

The new path is activated by a single boolean computed in `process_chat_payload`:

```python
capabilities = (model.get('info', {}).get('meta', {}).get('capabilities') or {})
builtin_tools_enabled = capabilities.get('builtin_tools', True)
file_context_enabled = capabilities.get('file_context', True)
is_native_fc = metadata.get('params', {}).get('function_calling') == 'native'
knowledge_builtins_enabled = (
    model.get('info', {}).get('meta', {}).get('builtinTools', {}).get('knowledge', True)
)

force_full_context = (
    request.app.state.config.RAG_FULL_CONTEXT
    or request.app.state.config.BYPASS_EMBEDDING_AND_RETRIEVAL
)

# at least one chunked (non-full) file/collection actually attached
has_retrievable_attached_files = any(
    item.get('context') != 'full' and item.get('type') in ('file', 'collection')
    for item in attached_files_list
)

native_fc_with_builtins = (
    not payload_tools
    and is_native_fc
    and builtin_tools_enabled
    and knowledge_builtins_enabled
    and file_context_enabled
    and not force_full_context
    and has_retrievable_attached_files
)
```

When **any** of these conditions is false, `native_fc_with_builtins` is False and every modified code path falls through to the existing implementation byte-for-byte:

- `chat_completion_files_handler` takes the non-partition branch: every item in `body.metadata.files` runs through `get_sources_from_items` exactly as today, honoring per-item `context` flags and `RAG_FULL_CONTEXT` / `BYPASS_EMBEDDING_AND_RETRIEVAL`. Chunks for chunked items, full bodies for full-context items. Same `sources` shape as before — no `roster_only` flag, no `<retrievable_file>` tags, no partition.
- `apply_source_context_to_messages` takes `force_system_message=False`, falling back to the existing `if RAG_SYSTEM_CONTEXT: append to system else: prepend to last user` logic.
- `get_source_context` only emits the `<full_context_files>` / `<retrievable_files>` split shape when at least one source has `roster_only=True`. When all sources are regular content sources (which is what the existing path produces), the output is the flat `<source>` shape it has always emitted: `if not roster_parts: return ''.join(citeable_parts)`.
- The tool-loop reads `metadata.get('native_fc_with_builtins', False)`. When False, it runs the original restore + rebuild + re-apply code exactly as before, including the `RAG_SYSTEM_CONTEXT` branch for where to inject.
- `query_attached_files` is only registered by `get_builtin_tools` when the chat has retrievable items (i.e., `__attached_files__` is populated AND at least one item has `type in ('file', 'collection')` with `context != 'full'` AND `force_full_context` is False). In default mode none of those conditions are met.

The two gates beyond the admin flags — `knowledge_builtins_enabled` (the per-model `meta.builtinTools.knowledge` toggle) and `has_retrievable_attached_files` — exist so the new path never activates where the tool wouldn't be registered (knowledge category off) or where there is nothing to route through it (no chunked attachments), which would otherwise needlessly change placement and skip the tool-loop re-application for unrelated native-tool chats.

### `{{QUERY}}` / `[query]` placeholder removal

`rag_template` no longer substitutes the `{{QUERY}}` / `[query]` placeholders. They predate this work and duplicated the user's message — which is already present in the messages array — and substituting them into a system-pinned template would defeat caching by mutating the prefix every turn. They are now stripped to empty (stripped from the template *before* context substitution, so identical literals inside a retrieved chunk are preserved). A `WARNING` is logged once at config-load (startup and on admin RAG-config update) if a custom `RAG_TEMPLATE` still contains them. The bundled default template never used them, so default deployments are unaffected.

> [!NOTE]

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -1413,13 +1413,16 @@ Respond to the user query using the provided context, incorporating inline citat
 - If the answer isn't present in the context but you possess the knowledge, explain this to the user and provide the answer using your own understanding.
 - **For `<source>` tags in the context: only include inline citations using [id] (e.g., [1], [2]) when the `<source>` tag includes an id attribute, and do not cite from `<source>` tags that lack an id attribute.**
 - **For tool result chunks (which do not arrive wrapped in `<source>` tags):** cite using the chunk's source identifier in square brackets — typically the `source` filename, e.g. `[contract.pdf]`. Do not invent numeric `[N]` citations for tool result content that doesn't carry one.
-- If the context lists items as available for retrieval but does not include their content (for example, inventory-style entries), do not cite them directly — call the appropriate tool to fetch their content first, then cite from the returned chunks.
+- If the context lists items as available for retrieval but does not include their content, do not cite them directly — call the appropriate tool to fetch their content first, then cite from the returned chunks.
 - Do not use XML tags in your response.
 - Ensure citations are concise and directly related to the information provided.
 
 ### Example of Citation:
 If the user asks about a specific topic and the information is found in a source with a provided id attribute, the response should include the citation like in the following example:
 * "According to the study, the proposed method increases efficiency by 20% [1]."
+
+Or in the example of you retrieving the information through a tool call in a tool result, you can cite like so:
+* "The emergency server room access code is stored in the key box [emergency_handbook.pdf]."
 
 ### Output:
 Provide a clear and direct response to the user's query. Use inline citations in [id] format when an `<source id="...">` tag is present in the context, or in [source-name] format when citing a chunk returned by a tool result.
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -1403,7 +1403,7 @@ CHUNK_OVERLAP = ConfigVar(
 )
 
 DEFAULT_RAG_TEMPLATE = """### Task:
-Respond to the user query using the provided context, incorporating inline citations in the format [id] **only when the <source> tag includes an explicit id attribute** (e.g., <source id="1">).
+Respond to the user query using the provided context, citing sources inline by name in square brackets (e.g., `[contract.pdf]` or `[wikipedia.org]`).
 
 ### Guidelines:
 - If you don't know the answer, clearly state that.
@@ -1411,21 +1411,21 @@ Respond to the user query using the provided context, incorporating inline citat
 - Respond in the same language as the user's query.
 - If the context is unreadable or of poor quality, inform the user and provide the best possible answer.
 - If the answer isn't present in the context but you possess the knowledge, explain this to the user and provide the answer using your own understanding.
-- **For `<source>` tags in the context: only include inline citations using [id] (e.g., [1], [2]) when the `<source>` tag includes an id attribute, and do not cite from `<source>` tags that lack an id attribute.**
-- **For tool result chunks (which do not arrive wrapped in `<source>` tags):** cite using the chunk's source identifier in square brackets — typically the `source` filename, e.g. `[contract.pdf]`. Do not invent numeric `[N]` citations for tool result content that doesn't carry one.
-- If the context lists items as available for retrieval but does not include their content, do not cite them directly — call the appropriate tool to fetch their content first, then cite from the returned chunks.
+- **Cite each `<source>` tag in the context by the value of its `name` attribute**, in square brackets — e.g. `<source name="contract.pdf">` → `[contract.pdf]`, `<source name="https://wikipedia.org">` → `[https://wikipedia.org]` or `[wikipedia.org]` (URL protocol is optional in the citation).
+- **For content returned by tool calls** (web search, page fetch, attached-file or knowledge search): cite by the source's primary name as it appears in the result — a web result's title, a file chunk's source filename, or, for a directly fetched page, its URL. Prefer the title/name when one is present. Use the same `[identifier]` format.
+- If the context lists items as available for retrieval but does not include their content (for example, `<retrievable_file>` entries inside `<retrievable_files>`), do not cite them directly — call the appropriate tool to fetch their content first, then cite the returned sources by name.
 - Do not use XML tags in your response.
 - Ensure citations are concise and directly related to the information provided.
 
 ### Example of Citation:
-If the user asks about a specific topic and the information is found in a source with a provided id attribute, the response should include the citation like in the following example:
-* "According to the study, the proposed method increases efficiency by 20% [1]."
+* "According to the study, the proposed method increases efficiency by 20% [research_paper.pdf]."
+* "The official documentation confirms this behavior [docs.python.org]."
 
 Or in the example of you retrieving the information through a tool call in a tool result, you can cite like so:
 * "The emergency server room access code is stored in the key box [emergency_handbook.pdf]."
 
 ### Output:
-Provide a clear and direct response to the user's query. Use inline citations in [id] format when an `<source id="...">` tag is present in the context, or in [source-name] format when citing a chunk returned by a tool result.
+Provide a clear and direct response to the user's query, including inline `[source-name]` citations matched to the `name` attribute of `<source>` tags in the context, or to the source identifier of chunks returned by tool calls.
 
 <context>
 {{CONTEXT}}
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -1411,8 +1411,9 @@ Respond to the user query using the provided context, incorporating inline citat
 - Respond in the same language as the user's query.
 - If the context is unreadable or of poor quality, inform the user and provide the best possible answer.
 - If the answer isn't present in the context but you possess the knowledge, explain this to the user and provide the answer using your own understanding.
-- **Only include inline citations using [id] (e.g., [1], [2]) when the <source> tag includes an id attribute.**
-- Do not cite if the <source> tag does not contain an id attribute.
+- **For `<source>` tags in the context: only include inline citations using [id] (e.g., [1], [2]) when the `<source>` tag includes an id attribute, and do not cite from `<source>` tags that lack an id attribute.**
+- **For tool result chunks (which do not arrive wrapped in `<source>` tags):** cite using the chunk's source identifier in square brackets — typically the `source` filename, e.g. `[contract.pdf]`. Do not invent numeric `[N]` citations for tool result content that doesn't carry one.
+- If the context lists items as available for retrieval but does not include their content (for example, inventory-style entries), do not cite them directly — call the appropriate tool to fetch their content first, then cite from the returned chunks.
 - Do not use XML tags in your response.
 - Ensure citations are concise and directly related to the information provided.
 
@@ -1421,7 +1422,7 @@ If the user asks about a specific topic and the information is found in a source
 * "According to the study, the proposed method increases efficiency by 20% [1]."
 
 ### Output:
-Provide a clear and direct response to the user's query, including inline citations in the format [id] only when the <source> tag with id attribute is present in the context.
+Provide a clear and direct response to the user's query. Use inline citations in [id] format when an `<source id="...">` tag is present in the context, or in [source-name] format when citing a chunk returned by a tool result.
 
 <context>
 {{CONTEXT}}
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -567,6 +567,7 @@ from open_webui.utils.models import (
     get_all_models,
     get_filtered_models,
 )
+from open_webui.utils.task import warn_if_deprecated_rag_template_placeholders
 from open_webui.utils.oauth import (
     OAuthClientInformationFull,
     OAuthClientManager,
@@ -1047,6 +1048,7 @@ app.state.config.RAG_EXTERNAL_RERANKER_TIMEOUT = RAG_EXTERNAL_RERANKER_TIMEOUT
 app.state.config.RAG_RERANKING_BATCH_SIZE = RAG_RERANKING_BATCH_SIZE
 
 app.state.config.RAG_TEMPLATE = RAG_TEMPLATE
+warn_if_deprecated_rag_template_placeholders(app.state.config.RAG_TEMPLATE)
 
 app.state.config.RAG_OPENAI_API_BASE_URL = RAG_OPENAI_API_BASE_URL
 app.state.config.RAG_OPENAI_API_KEY = RAG_OPENAI_API_KEY
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -567,7 +567,6 @@ from open_webui.utils.models import (
     get_all_models,
     get_filtered_models,
 )
-from open_webui.utils.task import warn_if_deprecated_rag_template_placeholders
 from open_webui.utils.oauth import (
     OAuthClientInformationFull,
     OAuthClientManager,
@@ -1048,7 +1047,14 @@ app.state.config.RAG_EXTERNAL_RERANKER_TIMEOUT = RAG_EXTERNAL_RERANKER_TIMEOUT
 app.state.config.RAG_RERANKING_BATCH_SIZE = RAG_RERANKING_BATCH_SIZE
 
 app.state.config.RAG_TEMPLATE = RAG_TEMPLATE
-warn_if_deprecated_rag_template_placeholders(app.state.config.RAG_TEMPLATE)
+if app.state.config.RAG_TEMPLATE and (
+    '{{QUERY}}' in app.state.config.RAG_TEMPLATE or '[query]' in app.state.config.RAG_TEMPLATE
+):
+    log.warning(
+        "RAG_TEMPLATE contains {{QUERY}}/[query] placeholders, which are no "
+        "longer substituted (the user's message is already in the messages "
+        'array) and will be stripped to empty. Remove them from your template.'
+    )
 
 app.state.config.RAG_OPENAI_API_BASE_URL = RAG_OPENAI_API_BASE_URL
 app.state.config.RAG_OPENAI_API_KEY = RAG_OPENAI_API_KEY
--- a/backend/open_webui/routers/retrieval.py
+++ b/backend/open_webui/routers/retrieval.py
@@ -718,9 +718,15 @@ async def update_rag_config(request: Request, form_data: ConfigForm, user=Depend
     request.app.state.config.RAG_TEMPLATE = (
         form_data.RAG_TEMPLATE if form_data.RAG_TEMPLATE is not None else request.app.state.config.RAG_TEMPLATE
     )
-    from open_webui.utils.task import warn_if_deprecated_rag_template_placeholders
-
-    warn_if_deprecated_rag_template_placeholders(request.app.state.config.RAG_TEMPLATE)
+    if request.app.state.config.RAG_TEMPLATE and (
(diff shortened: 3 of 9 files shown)


Classic298:
what i built here works, but i am no longer confident whether this is needed as-is because 90-95% of this can be achieved through other means today in Open WebUI already

Issue #25038, opened 2026-05-24
issue: Web Search with SearXNG retrieves results successfully but fails during source/context injection (“No sources found”, JSONResponse error in chat_web_search_handler)


ge0el:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.9.5

### Operating System

docker compose on Synology DSM 7.2.2

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

When Open WebUI successfully retrieves valid SearXNG results, the result snippets and/or fetched page contents should be injected into the model context and displayed as sources.

The model should answer using the retrieved live web information.

### Actual Behavior

Open WebUI successfully queries SearXNG, retrieves valid web results, and displays them in the UI, but then fails during the “Querying” / source-processing phase.

The UI shows:
- “Searching the web”
- “Searched X sites”
- valid URLs/results

but then:
- “No sources found”

and the model answers as if web browsing was unavailable.

Observed UI sequence:
1. Web Search enabled.
2. Open WebUI shows “Searching the web”.
3. Open WebUI shows “Searched 5 sites” / “Searched 7 sites” and lists valid results.
4. Open WebUI then shows “Querying”.
5. Open WebUI shows “No sources found”.
[...]

PR #25275, opened 2026-05-31
fix: handle JSONResponse in web search query generation

scarab-systems:
Withdrawn.

### DIFF
--- a/backend/open_webui/test/util/test_middleware_source_context.py
+++ b/backend/open_webui/test/util/test_middleware_source_context.py
@@ -1,0 +1,31 @@
+from open_webui.utils.response import (
+    get_completion_message_content,
+    parse_generated_queries,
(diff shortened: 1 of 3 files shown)


scarab-systems:
Adding a quick scope note here as well.

This PR is intentionally limited to one failure path from #25038: generated web-search query handling receiving a JSONResponse where the code expects an OpenAI-style completion payload with choices.

The intended behavior is simple: if the generated-query response is not a usable completion payload, web search should continue with the original user query instead of failing with:

TypeError: ‘JSONResponse’ object is not subscriptable

The regression coverage checks both sides of that boundary:

* JSONResponse fallback path
* normal OpenAI-style completion response path

This does not attempt to cover every possible “No sources found” cause from the issue thread.

PR #25599, opened 2026-06-02
fix: expose model/selected skills as a lazy-loaded view_skill manifest

Classic298:

> [!NOTE]


gaby:
@Classic298 Can you fix the merge conflicts. This fix is needed for next release.

Classic298:
@gaby

erictmnz:
Keen to test this when it's on dev. We remain on 0.9.5 until dynamic skills loading works again in the same way (and test if prompt caching still works)

Classic298:
@erictmnz you can test it now as well. you only need the modified middleware.py the  frontend is not strictly needed for the fix it only moves the skill calculation for the tag to the backend

Himmig:
@Classic298 it works well - hoping we can get this into the next version.

PR #25647, opened 2026-06-03
breaking: Make Native the default tool-calling mode; rename "Default" to "Legacy"

Classic298:
- Native is now the default function-calling mode for all deployments: the backend resolves an empty/absent function_calling to native and only uses the legacy path when something explicitly requests 'legacy'.
- Rename the former "Default" tool-calling mode to "Legacy" (now the explicit opt-out) in the toggle label and tooltip.
- Update the AdvancedParams toggle to write null (native) <-> 'legacy'.

Existing installs (all stored as null/absent) simply resolve to native; admins/users can opt back to legacy per chat, per model, or globally via Default Model Params.

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1790,12 +1790,12 @@ async def chat_completion(
                 'stream_delta_chunk_size': stream_delta_chunk_size,
                 'reasoning_tags': reasoning_tags,
                 'function_calling': (
-                    'native'
+                    'legacy'
                     if (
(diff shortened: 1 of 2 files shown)


silentoplayz:
I believe Tim has handled this with https://github.com/open-webui/open-webui/commit/b1d40f340921c27eb9a965b9feeb2563856e25e2.

Classic298:
indeed thank you @silentoplayz

Issue #25617, opened 2026-06-02
issue: Open WebUI analytics undercounts token usage for streaming native tool-call loops


Classic298:
**Issue**
Open WebUI appears to persist only the last LLM call’s `usage` object for a streaming assistant response that uses native/server-side tool calls. Earlier LLM calls in the same tool loop are overwritten before the final assistant message is saved.

This makes the built-in analytics token usage inaccurate for tool-using conversations.

**Affected Path**
This applies to streaming chat completions with native/server-side tool calling, where one visible assistant message may require multiple upstream LLM calls:

1. Initial LLM call returns tool calls.
2. Open WebUI executes tools.
3. Open WebUI sends another LLM call with tool results.
4. This may repeat.
5. Final LLM call returns the visible assistant answer.

Each upstream LLM call can return its own `usage` object, but Open WebUI seems to persist only the final one.

**Code Evidence**
In `backend/open_webui/utils/middleware.py`, `streaming_chat_response_handler(...)` creates a single shared `usage` variable for the whole assistant response:

```python
usage = None
```

Inside `stream_body_handler(...)`, that variable is reused:

```python
nonlocal usage
```

When a streamed usage object arrives, it overwrites the previous value:

```python
raw_usage = data.get('usage', {}) or {}
raw_usage.update(data.get('timings', {}))  # llama.cpp
if raw_usage:
    usage = normalize_usage(raw_usage)
```

The Responses API path does the same:

```python
if response_metadata.get('usage'):
    response_metadata['usage'] = normalize_usage(response_metadata['usage'])
    usage = response_metadata['usage']
```

During native tool loops, the same `stream_body_handler(...)` is called again for follow-up LLM calls:

```python
await stream_body_handler(res, new_form_data)
```

At the end, only the final value of `usage` is persisted:

```python
await Chats.upsert_message_to_chat_by_id_and_message_id(
    metadata['chat_id'],
    metadata['message_id'],
    {
        'done': True,
        'content': serialize_output(output),
        'output': output,
        **({'usage': usage} if usage else {}),
    },
)
```

or, with realtime save enabled:

```python
await Chats.upsert_message_to_chat_by_id_and_message_id(
    metadata['chat_id'],
    metadata['message_id'],
    {'done': True, 'usage': usage},
)
```

Analytics then reads from the persisted assistant-message usage in `backend/open_webui/models/chat_messages.py`, for example:

```python
ChatMessage.usage.isnot(None)
```

and sums:

```python
[...]

Issue #23770, opened 2026-04-16
feat: recent dev changes for outlet_filter_handler do not completely handle the API caller use-case


pfn:
### Check Existing Issues

- [x] I have searched for all existing **open AND closed** issues and discussions for similar requests. I have found none that is comparable to my request.

### Verify Feature Scope

- [x] I have read through and understood the scope definition for feature requests in the Issues section. I believe my feature request meets the definition and belongs in the Issues section instead of the Discussions.

### Problem Description

Continuing the discussion from #23740 which is closed prematurely.

I have to apply the following diff to middleware.py in addition to running a fairly carefully crafted filter function in order for API calls to track outlet() correctly https://gist.github.com/pfn/cdf3631365e6bee59912016b7a953175

With this filter and the diff in place, both non-streaming and streaming API calls through openwebui work for me. Without the filter active and the diff, neither API call use-case works. The diff is primarily required for the streaming use-case. The filter is required for both streaming and non-streaming API calls.

### Desired Solution you'd like

I'm relatively happy with how the linked gist solves the problem. If it could be made more elegant, I would like that.

### Alternatives Considered

N/A

### Additional Context

Discussed in discord.

> pfn0 [SJBH],  — 7:18 PM
> not sure what it is yet (looking at the code, event_emitter shouldn't be the problem)
>             ctx = await build_chat_response_context(request, form_data, user, model, metadata, tasks, events)
> main.py:1851 @Classic this brings event_emitter to life
> pfn0 [SJBH],  — 7:50 PM
> @Classic final finding: streaming_chat_response_handler -> w/o event_emitter -> does not ever call outlet_filter_handler (and if outlet_filter_handler is force-called, assistant_message is not available)

Classic298:
https://github.com/open-webui/open-webui/pull/25650

Classic298:
addressed in dev, outlet should now run on dev

PR #25650, opened 2026-06-03
feat: run outlet filters for direct API callers (streaming + non-streaming)

Classic298:
outlet() filters previously only fired for Web-UI / chat invocations. Direct API calls to /api/chat/completions (and the Anthropic-compatible endpoint that routes through it) have no chat_id/message_id and no event channel, so outlet_filter_handler short-circuited and the streaming path took a pure passthrough branch that never accumulated the response.

This wires outlet filters into the API path so they run end-to-end (e.g. for request/usage tracking), treating API callers like temp chats: the message list is built from form_data + the assistant response instead of the DB. Event emission and DB persistence stay skipped — event emitters cannot run on stateless API calls, which is an accepted limitation.

- outlet_filter_handler: build the message list from form_data whenever there is no DB-backed chat (now covers API callers, not just temp/local/channel chats); only persist outlet edits for real persisted chats.
- non_streaming_chat_response_handler: run outlet on the no-event_emitter (API) path; the response is returned unmodified.
- streaming_chat_response_handler: in the passthrough branch, accumulate content + usage (Chat Completions and Responses API formats) and fire outlet once the stream ends. Gated on filter_functions to avoid per-chunk parsing for filter-less callers, and guarded so accumulation can never break the passthrough stream.

Related: https://github.com/open-webui/open-webui/issues/23770 https://github.com/open-webui/open-webui/issues/25617 https://github.com/open-webui/open-webui/issues/21675

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -3292,17 +3292,26 @@ async def outlet_filter_handler(ctx):
     chat_id = metadata.get('chat_id', '')
     message_id = metadata.get('message_id')
 
-    if not chat_id or not message_id:
-        return
-
     is_temp_chat = chat_id.startswith('local:') or chat_id.startswith('channel:')
+
+    # Persisted chats read from the DB and get outlet edits written back; temp
+    # chats and direct API callers have no DB record, so build the message list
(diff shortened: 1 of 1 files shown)


tjbck:
Addressed in dev.

PR #25668, opened 2026-06-03
fix: respect model visibility in workspace base model selector

Classic298:
The base model selector in the model creation workspace listed every model from the global store, ignoring the per-model visibility (hidden) flag. Models marked as not visible were correctly hidden from the chat model selector but still appeared in the workspace 'Base Model (From)' dropdown.

Filter out hidden models (info.meta.hidden) to match the chat selector behavior, while still keeping a model visible if it is the currently selected base, so editing a preset built on a now-hidden base does not show a blank dropdown.

Fixes #25665

> [!NOTE]

### DIFF
--- a/src/lib/components/workspace/Models/ModelEditor.svelte
+++ b/src/lib/components/workspace/Models/ModelEditor.svelte
@@ -617,7 +617,7 @@
 											<option value={null} class=" text-gray-900"
 												>{$i18n.t('Select a base model')}</option
 											>
-											{#each $models.filter((m) => (model ? m.id !== model.id : true) && !m?.preset && m?.owned_by !== 'arena' && !(m?.direct ?? false)) as model}
+											{#each $models.filter((m) => (model ? m.id !== model.id : true) && !m?.preset && m?.owned_by !== 'arena' && !(m?.direct ?? false) && (!(m?.info?.meta?.hidden ?? false) || m.id === info.base_model_id)) as model}
 												<option value={model.id} class=" text-gray-900">{model.name}</option>
 											{/each}
 										</select>


AdBaWa:
Hi @Classic298 @tjbck 
This change breaks our setup. hidden and workspace base-model availability are being conflated.

Problem: hidden = true correctly removes a model from the chat selector.
With this PR, it also removes it from the workspace base selector.
Result: users with model-creation rights can no longer build presets on base models they are explicitly permitted to use via access groups.

Missing distinction:
End-user in chat → respect hidden
End-user with creation rights → should be filtered by access groups, not hidden 
Admin → sees everything

Smell: The "keep currently selected base visible" workaround already signals the filter is wrong for this surface.

Suggested fix: Filter the workspace base selector by access group / permissions, not by hidden. Optionally add a "hidden from chat" badge.

Can this be reverted or gated behind a setting?

Thanks!

Classic298:
@AdBaWa hidden model means the model should be entirely hidden from the UI including workspace model selector, chat model selector and more.

This is a bug fix.

This will not be reverted.

Hidden means to hide the model completely from the UI and not show it.
This was a bug that was fixed.

If you want your users to create a model based on a hidden model they can do so either programmatically or by uploading the respective model json - or - do not hide the model you want some of your users to have explicit access to.

Classic298:
@AdBaWa The `hidden` flag isn't an access-control mechanism — access groups are. `hidden` exists for a specific purpose: task models (title generation, autocomplete, follow ups, etc.) have to be shared publicly so they can run for every user regardless of group, but you don't want that small utility model showing up in pickers next to your main models where people would select it by accident. So `hidden` means exactly "publicly usable in the background, but invisible in every human-facing selector" — chat *and* workspace base. A hidden model appearing in the base-model dropdown is the precise thing the flag was created to prevent.

So the two cases you're describing are already handled by two different systems:

- **Base model available to some builders but not others** → that's access groups. Share it to the group that should build on it, don't share it to the rest, and leave `hidden` off. It'll show for the permitted builders and not the others.
- **Model usable as a base but not cluttering chat** → again, don't set `hidden` — control it with access groups. `hidden` is for public task infrastructure you want out of *all* pickers.

There's no population that needs a `hidden` model in the base selector, because "hidden from pickers" and "available to a builder subset" are answered by different tools (hiding versus access control). If you set `hidden`, you're telling Open WebUI it's public utility infra to keep out of every selector — which is why filtering it there is correct.

The "keep the currently-selected base visible" carve-out isn't a sign the filter is wrong — it just avoids blanking the dropdown when you edit a preset whose base was hidden *after* the preset was built. That's don't-break-existing-config behavior, not an exception to the rule.

For the rare case where you genuinely need to base a preset on a hidden task model, the API / model JSON path covers it. This stays as is.

AdBaWa:
Hi @Classic298 

Ah, thank you for the thorough explanation. That makes perfect sense regarding the architectural intent behind the hidden flag (system utility infra vs. access control).

It turns out we were classic victims of "using a bug as a feature" since v0.8/v0.9.

In our enterprise setup, we were layering hidden = true on top of access groups to achieve a specific permission tier: giving a group of users "Builder Access" (allowed to use a model as a base in the workspace) without giving them "Chat Access" (allowing them to talk to the raw base model directly).

Now that we understand hidden is strictly for background infrastructure, we'll pivot to using the programmatic JSON upload workaround you mentioned to keep our workflows running.
That said, for larger team/enterprise setups, having a dedicated way to separate Model Building rights from Direct Chat rights on a per-model basis is incredibly useful. 

I'll open up a separate Feature Request for a granular permission toggle (like a workspace_only visibility flag or split RBAC roles) so we can support this natively in the future without abusing the system task flags?

Appreciate the clarity and the quick response.

Classic298:
yes, makes sense, feel free to open a feat req

PR #25780, opened 2026-06-07
fix: chroma has_collection always returns False (name vs Collection)

Classic298:
has_collection did `collection_name in self.client.list_collections()`, but chromadb's list_collections() returns Collection objects (1.x), not name strings — so the membership test is always False, even when the collection exists. Compare against the collection names instead (with a hasattr guard tolerating versions that yield plain names).

Found via the dependency-contract test suite (unit/deps/test_chromadb.py).

> [!NOTE]

### DIFF
--- a/backend/open_webui/retrieval/vector/dbs/chroma.py
+++ b/backend/open_webui/retrieval/vector/dbs/chroma.py
@@ -57,7 +57,11 @@ class ChromaClient(VectorDBBase):
 
     def has_collection(self, collection_name: str) -> bool:
         # Check if the collection exists based on the collection name.
-        collection_names = self.client.list_collections()
+        # chromadb's list_collections() returns Collection objects (1.x), so a
+        # bare `name in collections` membership test is always False — compare
+        # against the names. (hasattr guard tolerates versions that yield names.)
+        collections = self.client.list_collections()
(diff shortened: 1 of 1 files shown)


Mostafa-Alavi:
Confirming this in production — and the impact is bigger than a stale membership check.

Hit this on ghcr.io/open-webui/open-webui:main with chromadb 1.5.2. list_collections() returns Collection objects (Collection(name=...)), so has_collection() returns False for collections that demonstrably exist.

The downstream effect is the painful part: the knowledge reindex endpoint (/api/v1/knowledge/reindex) guards its cleanup with


if await ASYNC_VECTOR_DB_CLIENT.has_collection(collection_name=kb.id):
    await ASYNC_VECTOR_DB_CLIENT.delete_collection(collection_name=kb.id)
Since has_collection is always False, the delete is skipped and every reindex appends a fresh copy of all chunks instead of replacing them. After a few reindex passes our collection held 492 chunks where only 123 were unique (4×).

Those duplicates then dominate the top-k at query time, so focused/embedding retrieval returns near-useless context while "full document" mode works fine — which looks exactly like an embedding-model problem and sends you down the wrong rabbit hole.

Applied the fix in this PR and verified: has_collection returns True, and two back-to-back reindexes now keep chunk counts flat (123 → 123, no duplication).  Would be great to get this merged.

PR #25945, opened 2026-06-11
Equalize signin bcrypt cost to mitigate user-enumeration timing

Classic298:
authenticate_user skipped bcrypt entirely when the email did not resolve to a user (or had no active credential), so a non-existent account returned in ~5ms while a valid one paid the full bcrypt cost (~150ms), letting an attacker enumerate accounts by response time.

Run the bound verify_password against a fixed dummy hash on the no-user and no-credential paths so every signin attempt pays the same bcrypt cost. This collapses the dominant, trivially-exploitable gap (measured ~1.03x vs the prior ~20-40x). A residual sub-millisecond variance (DB index hit/miss, interpreter/GC, CPU and OS scheduling) is not eliminable in application code and is accepted.

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/auths.py
+++ b/backend/open_webui/models/auths.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import bcrypt
 import logging
 import uuid
 from typing import Optional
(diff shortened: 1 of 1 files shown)


tjbck:
993e74912199c66c522f08ec81abe31d76985e39

PR #25947, opened 2026-06-11
Stop /channels/{id}/members from leaking members' private fields

Classic298:
get_channel_members_by_id returned the full UserModel for every channel member via UserModelResponse(**u.model_dump(), ...), including settings (which holds per-user tool-server bearer keys and webhook URLs), oauth, scim, info and PII (date_of_birth/gender/bio). Channel membership is self-grantable (any user can open a DM with any other user), so a low-privilege user could DM an admin and read the admin's tool-server API keys from the members list.

Exclude those fields from both members response paths (DM and group/non-DM) via a shared _CHANNEL_MEMBER_PRIVATE_FIELDS set. The response shape is unchanged; the sensitive fields are simply omitted. Sibling endpoints that embed users already use the minimal UserIdNameStatusResponse and are unaffected.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/channels.py
+++ b/backend/open_webui/routers/channels.py
@@ -440,6 +440,11 @@ async def get_channel_by_id(
 PAGE_ITEM_COUNT = 30
 
 
+# UserModel fields that must not leak to other channel members: credentials / integration config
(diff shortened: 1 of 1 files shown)


tjbck:
Addressed in dev.

PR #25959, opened 2026-06-11
Persist tool-server config durably before reporting success (fail-clo…

Classic298:
…sed)

set_tool_servers_config set TOOL_SERVER_CONNECTIONS via AppConfig.__setattr__, which schedules a fire-and-forget DB write and swallows persistence errors, then returned 200 without confirming the write landed. If that write failed (e.g. an inaccessible database file), the new access grants would live only in memory while the previous grants stayed on disk and were reloaded on the next restart.

Add AppConfig.set_durable(name, value), which awaits the commit and rolls back the in-memory value on failure, and use it for TOOL_SERVER_CONNECTIONS so the endpoint fails closed (HTTP 500, no change applied) when persistence does not succeed. Hardening: a write failure on a healthy instance does not occur (the app cannot run without its DB), so this closes an edge case rather than an exploitable bypass.

> [!NOTE]

### DIFF
--- a/backend/open_webui/internal/config.py
+++ b/backend/open_webui/internal/config.py
@@ -233,6 +233,11 @@ class AppConfig:
         except Exception as exc:
             log.error("Async persist failed for '%s': %s", name, exc)
 
+    async def commit(self, name: str) -> None:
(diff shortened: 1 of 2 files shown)


tjbck:
Addressed in dev.

Issue #25898, opened 2026-06-10
issue: OAuth scopes are still missing within authorize_url when using MCP with OAuth2.1 (Static)


MoHe99:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.9.6

### Operating System

Windows 11 with WSL 2 (Ubuntu 26.04 LTS)

### Browser (if applicable)

Edge 149.0.4022.52

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

After [PR #24690](https://github.com/open-webui/open-webui/pull/24690) (merged into v0.9.6), `handle_authorize` should include the `scope` parameter in the OAuth authorize URL when using MCP with OAuth 2.1 (Static) and an IdP that requires scopes (e.g. Azure Entra ID).

### Actual Behavior

`handle_authorize` (`open_webui/utils/oauth.py`, ~line 960) generates an authorize URL **without** `scope`, causing the IdP to reject the request (Entra: `AADSTS900144`).

Debug logging at `handle_authorize` proves:

```
client_info.scope=None
client_kwargs={'follow_redirects': True, 'token_endpoint_auth_method': 'client_secret_post'}
kwargs={}
```

[...]

Issue #24216, opened 2026-04-29
issue: MCP OAuth 2.1 (Static) fails with Google Gmail MCP server – discovery document not found when OAuth server differs from MCP server


samuixx:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.9.2

### Ollama Version (if applicable)

0.22.0

### Operating System

Debian Trixie

### Browser (if applicable)

Chrome

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

Open WebUI should support specifying a separate OAuth authorization server URL (e.g., https://accounts.google.com) independently from the MCP server URL. This is required for providers like Google where the MCP endpoint and OAuth endpoint are on different hosts.

Google's discovery document is available at: https://accounts.google.com/.well-known/openid-configuration

Suggested solution

Add an optional field (e.g., "OAuth Server URL" or "Authorization Server URL") in the External Tools OAuth 2.1 (Static) configuration that allows specifying the OAuth discovery endpoint separately from the MCP server URL. When not set, fall back to the current behavior of using the MCP server host.

This pattern is common with Google Workspace MCP servers (Gmail, Calendar, Drive, People API) which all use accounts.google.com for OAuth but have separate MCP endpoints.

### Actual Behavior

[...]

PR #25980, opened 2026-06-12
fix: discover MCP Protected Resource Metadata when server does not return 401

### LINKED ISSUE #25977: bug: RFC 9728 Protected Resource Metadata discovery is skipped when the MCP server does not answer 401 (e.g. Google's hosted Gmail/Drive/Calendar MCPs)
## Bug Summary

`get_protected_resource_metadata()` in `backend/open_webui/utils/oauth.py` only attempts RFC 9728 Protected Resource Metadata (PRM) discovery when the anonymous `initialize` probe returns **401** with a `WWW-Authenticate` header.

Some remote MCP servers answer **200** to an anonymous `initialize` and only enforce authorization at `tools/call` time. Google's hosted Workspace MCPs behave this way (e.g. `https://gmailmcp.googleapis.com/mcp/v1`), while still publishing a perfectly valid PRM document at the RFC 9728 §4.2 well-known URI:

```
$ curl -s https://gmailmcp.googleapis.com/.well-known/oauth-protected-resource/mcp/v1
{"authorization_servers":["https://accounts.google.com/"], "resource":"https://gmailmcp.googleapis.com/mcp/v1",
 "scopes_supported":["https://mail.google.com/","https://www.googleapis.com/auth/gmail.modify", ...]}
```

Because the probe returns 200, Open WebUI never fetches this document: `scopes_supported` and `authorization_servers` stay empty, so the OAuth flow requests no scopes and the connection cannot be authorized correctly.

## Steps to Reproduce

1. Admin Settings → Tool Servers → add an MCP connection to `https://gmailmcp.googleapis.com/mcp/v1` with OAuth 2.1 (static client credentials from a Google Cloud project with the Gmail MCP API enabled).
2. Authorize the connection.
3. The authorization request is built without the scopes advertised by the resource's PRM document.

## Expected Behavior

PRM discovery should follow RFC 9728: use the `WWW-Authenticate` `resource_metadata` hint when the server returns 401, and otherwise fall back to the well-known URIs — regardless of the probe's HTTP status.

## Actual Behavior

Discovery only happens on 401; servers answering 200/4xx-other to the anonymous probe never get PRM discovery.

## Environment

- Open WebUI: reproduced on `v0.9.6` and present in current `dev` (`backend/open_webui/utils/oauth.py`, `if response.status == 401:`).

## Proposed Fix

Run the discovery unconditionally after the probe (prefer the `WWW-Authenticate` hint when present, fall back to well-known URIs otherwise). Verified against Google's Gmail MCP in a production deployment. PR incoming.

AlbertodelaCruz:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #25977
- [x] **Description:** provided below.
- [ ] **Documentation:** no docs change needed (no user-facing configuration is added).
- [x] **Dependencies:** none added.
- [x] **Testing:** manually verified against Google's hosted Gmail MCP (`https://gmailmcp.googleapis.com/mcp/v1`) in a production deployment: before the change, scope/authorization-server discovery silently failed (the server answers 200 to an anonymous `initialize`); after the change, the PRM document is discovered via the RFC 9728 well-known URI and the OAuth flow requests the advertised scopes. Also verified that 401-responding servers keep the exact same behavior (the `WWW-Authenticate` hint is still preferred).
- [x] **No Unchecked AI Code:** human-reviewed and manually tested.
- [x] **Self-Review:** done.
- [x] **Architecture:** no new settings; pure bugfix.
- [x] **Git Hygiene:** single atomic commit, rebased on `dev`.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

`get_protected_resource_metadata()` only attempted RFC 9728 Protected Resource Metadata discovery when the anonymous `initialize` probe returned 401 with a `WWW-Authenticate` header. Some remote MCP servers — notably Google's hosted Gmail/Drive/Calendar MCPs — answer 200 to an anonymous `initialize` and only enforce authorization on `tools/call`, so OAuth scope and authorization-server discovery silently failed and those connections could not be authorized correctly.

This PR runs the discovery regardless of the probe's HTTP status: the `resource_metadata` URL from `WWW-Authenticate` is still preferred when present, and the RFC 9728 §4.2 well-known URIs are used as fallback otherwise. The change is a de-indentation of the existing block (plus an explanatory comment); behavior for 401-responding servers is unchanged.

Trade-off: for servers that expose no PRM document, connection setup now performs a couple of extra well-known GETs (404s, logged at debug). This only happens during MCP connection setup, not on regular tool calls.

### Fixed

- Fixed OAuth scope and authorization-server discovery for MCP servers that do not answer 401 to an anonymous `initialize` (e.g. Google's hosted Gmail/Drive/Calendar MCPs), by always attempting RFC 9728 Protected Resource Metadata discovery.

---

### Additional Information

- RFC 9728 §4.2 does not condition well-known PRM discovery on a 401 challenge; the challenge is just one way to learn the metadata URL.
- Reproduced and verified against `https://gmailmcp.googleapis.com/mcp/v1`.
- Relates to #25898 (same symptom family, different root cause: there the discovered scope is lost before `handle_authorize`; here PRM discovery never runs without a 401 challenge) and #24216.

### DIFF
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -325,53 +325,57 @@ async def get_protected_resource_metadata(server_url: str) -> ProtectedResourceM
                 headers={'Content-Type': 'application/json'},
                 ssl=AIOHTTP_CLIENT_SESSION_SSL,
             ) as response:
-                if response.status == 401:
-                    resource_metadata_urls = []
-                    match = re.search(
-                        r'resource_metadata=(?:"([^"]+)"|([^\s,]+))',
-                        response.headers.get('WWW-Authenticate', ''),
-                    )
-                    if match:
-                        resource_metadata_urls = [match.group(1) or match.group(2)]
-                        log.debug(f'Found resource_metadata URL: {resource_metadata_urls[0]}')
-                    else:
-                        # Fall back to well-known resource metadata URIs (RFC 9728 §4.2)
-                        parsed, base_url = get_parsed_and_base_url(server_url)
-                        if parsed.path and parsed.path != '/':
-                            path = parsed.path.rstrip('/')
-                            resource_metadata_urls.append(
-                                urllib.parse.urljoin(base_url, f'/.well-known/oauth-protected-resource{path}')
-                            )
+                # Discover Protected Resource Metadata regardless of HTTP status.
+                # A 401 carries a WWW-Authenticate header pointing at the PRM, but
+                # some MCP servers (e.g. Google's gmail/drive/calendar remote MCPs)
+                # answer 200 to an anonymous `initialize`, so we must still fall
+                # back to the RFC 9728 well-known URIs when there is no 401/header.
+                resource_metadata_urls = []
+                match = re.search(
+                    r'resource_metadata=(?:"([^"]+)"|([^\s,]+))',
+                    response.headers.get('WWW-Authenticate', ''),
+                )
+                if match:
+                    resource_metadata_urls = [match.group(1) or match.group(2)]
+                    log.debug(f'Found resource_metadata URL: {resource_metadata_urls[0]}')
+                else:
+                    # Fall back to well-known resource metadata URIs (RFC 9728 §4.2)
+                    parsed, base_url = get_parsed_and_base_url(server_url)
+                    if parsed.path and parsed.path != '/':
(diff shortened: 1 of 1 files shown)


tjbck:
Thanks!

PR #26032, opened 2026-06-13
Confer object-derived file write only for files the object owner owns

Classic298:
has_access_to_file() derives file access from the objects a file is attached to (knowledge bases, workspace models). Those branches returned True for any access_type whenever the user held that permission on the object, write/delete included. Since a user can create their own KB or model and attach any file they can merely READ (KB attach and the model meta.knowledge validator both gate on read access only), a user with read access to a victim file could launder it into write/delete: attach it to an object they own, then rename, overwrite or delete it via the write-gated file routes (POST /files/{id}/rename, /data/content/update, DELETE /files/{id}).

An object now confers write/delete on a file only when the object's owner owns that file, so delegation originates from the file's own owner. Read is unchanged (RAG and shared-object reads still work), and legitimate delegation is preserved: a write grant on an object whose owner owns the attached file still confers write. Applied to all three object branches: knowledge base, file home collection, and workspace model.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/access_control/files.py
+++ b/backend/open_webui/utils/access_control/files.py
@@ -38,25 +38,33 @@ async def has_access_to_file(
     if file.user_id == user.id:
         return True
 
-    # Check if the file is associated with any knowledge bases the user has access to
+    # Check if the file is associated with any knowledge bases the user has access to.
+    # An object (knowledge base or workspace model) confers write/delete on a file only when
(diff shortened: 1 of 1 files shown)


rexpository:
Confirmed locally against the report PoV: the  laundering path now fails closed, and owner-backed legitimate delegation still works. Fix looks good to me.

PR #26034, opened 2026-06-14
fix(markdown): keep code blocks with nested ``` fences intact

l3aconators:
LLM replies that show code containing triple-backtick fences (e.g. a prompt string with embedded ``` examples) open the outer block with 3 backticks, so marked closes it at the first inner fence and the rest spills out as alternating prose/code.

- normalizeNestedFences(): promote a language-tagged outer fence to one more backtick than its widest nested fence, bounded by the next tagged opener, EOF, or a markdown heading so it never absorbs following prose.
- katex: drop backtick from allowed math neighbours and reject $...$ whose body contains a backtick, so $ next to/inside inline code (C# $"...", shell `$PATH`) is not parsed as math.

Adds vitest coverage for both.

> [!NOTE]

### DIFF
--- a/src/lib/utils/index.ts
+++ b/src/lib/utils/index.ts
@@ -25,6 +25,9 @@ import markedKatexExtension from '$lib/utils/marked/katex-extension';
 import hljs from 'highlight.js';
 import { decode } from 'html-entities';
 
+import { normalizeNestedFences } from './markdown-fences';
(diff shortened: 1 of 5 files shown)


Classic298:
1) intended 
2) ai should use more backticks on the outer side
3) you didnt link an issue or discussion
4) no test files wanted

PR #26052, opened 2026-06-14
fix: stop marking collection_names entries as legacy in ModelEditor

### LINKED ISSUE #26035: issue: ModelEditor drops the id from collection_names knowledge on save, so the model retrieves nothing
### Description

In the Model Editor UI (`src/lib/components/workspace/Models/ModelEditor.svelte`), the knowledge mapping logic **unconditionally sets `legacy: true`** for knowledge entries that use the `collection_names` format — the same format that was introduced as the replacement for the old `collection_name` format. This causes the UI to display "Legacy Collection" labels for all knowledge collections, even when they are using the current, non-legacy format.

### Affected File

`src/lib/components/workspace/Models/ModelEditor.svelte`, lines 303-320:

```typescript
knowledge = (model?.meta?.knowledge ?? []).map((item) => {
    if (item?.collection_name && item?.type !== 'file') {
        return {
            id: item.collection_name,
            name: item.name,
            legacy: true
        };
    } else if (item?.collection_names) {
        return {
            name: item.name,
            type: 'collection',
            collection_names: item.collection_names,
            legacy: true  // <-- BUG: this should NOT be legacy: true
        };
    } else {
        return item;
    }
});
```

### Root Cause

Line 315 sets `legacy: true` for entries that have `collection_names` (the **new** format). The `collection_names` array format was introduced in the knowledge table migration (`6a39f3d8e55c_add_knowledge_table.py`) as the replacement for the single `collection_name` string. The old `collection_name` format correctly gets `legacy: true`, but the new `collection_names` format should **not**.

The downstream rendering in `src/lib/components/workspace/Models/Knowledge.svelte` line 180-181 uses this `legacy` flag to determine the display label:

```svelte
type={file?.legacy
    ? `Legacy${file.type ? ` ${file.type}` : ''}`
    : (file?.type ?? 'collection')}
```

So any entry with `legacy: true` gets rendered as "Legacy Collection" regardless of whether the underlying data format is actually legacy.

### Reproduction Steps

1. Create a knowledge collection via the OWUI Knowledge interface
2. Create or edit a model and associate it with that knowledge collection
3. Save the model — the `model.meta.knowledge` array now contains entries with `collection_names: [uuid]`
4. Re-open the model in the Model Editor
5. The knowledge section displays "Legacy Collection" label for each collection

### Expected Behavior

Knowledge entries using the `collection_names` format (the current format) should display as "collection" in the Model Editor, not "Legacy Collection". The `legacy` flag should only be `true` for entries using the deprecated `collection_name` (singular) format.

### Actual Behavior

All knowledge entries with `collection_names` are marked as `legacy: true` and displayed as "Legacy Collection" in the Model Editor UI, regardless of the database state. Clearing browser cache does not help because the `legacy: true` is set dynamically by the frontend code on every load.

### Environment

- Open WebUI version: 0.9.6 (dev-cuda)
- Backend: Python/FastAPI with PostgreSQL
- Frontend: SvelteKit

### Proposed Fix

Change line 315 from `legacy: true` to `legacy: false`:

```typescript
} else if (item?.collection_names) {
    return {
        name: item.name,
        type: 'collection',
        collection_names: item.collection_names,
        legacy: false  // collection_names is the current format
    };
}
```

### Additional Context

The migration `6a39f3d8e55c_add_knowledge_table.py` introduced the `collection_names` array as the replacement for the old `collection_name` string. The middleware in `utils/middleware.py` (line 2480+) correctly distinguishes between the two formats:
- `collection_name` (singular) = legacy format, triggers migration to new format
- `collection_names` (array) = current format, used directly

The ModelEditor frontend logic does not match this distinction — it treats both as legacy.

michaelxer:
## Description

Closes #26035

The knowledge mapping logic in `ModelEditor.svelte` (lines 303-320) unconditionally sets `legacy: true` for entries using the `collection_names` format. But `collection_names` is the current format - it replaced the old singular `collection_name` string in the knowledge table migration. Only the old `collection_name` format should be marked legacy.

This causes all knowledge collections to display as "Legacy Collection" in the Model Editor UI, even when they are using the current data format.

### Fix

One-line change: `legacy: true` to `legacy: false` on line 315 for the `collection_names` branch. The `collection_name` (singular) branch on line 308 correctly stays `legacy: true`.

### Fixed

- Fixed ModelEditor incorrectly marking collection_names knowledge entries as legacy, causing all collections to display "Legacy Collection" label

## Testing

1. Create a knowledge collection via the Knowledge interface
2. Create or edit a model and associate it with that knowledge collection
3. Save the model
4. Re-open the model in the Model Editor
5. The knowledge section should now display "collection" instead of "Legacy Collection"

---

### DIFF
--- a/src/lib/components/workspace/Models/ModelEditor.svelte
+++ b/src/lib/components/workspace/Models/ModelEditor.svelte
@@ -312,7 +312,7 @@
 						name: item.name,
 						type: 'collection',
 						collection_names: item.collection_names,
-						legacy: true
+						legacy: false
 					};
 				} else {
 					return item;


Classic298:
did you test it?

michaelxer:
Yes. The fix is straightforward — I traced the knowledge mapping logic in `ModelEditor.svelte` and verified with test data:

- `collection_name` (singular, old format) → `legacy: true` ✓ (unchanged)
- `collection_names` (plural, current format) → `legacy: false` ✓ (the fix)
- `type: 'file'` → falls through to else, no legacy flag ✓

The issue was line 315: the `collection_names` branch had `legacy: true` when it should be `false`. Only the old singular `collection_name` format is legacy. The current `collection_names` format is the supported path.

Classic298:
?

michaelxer:
To clarify: I had the frontend dev server running locally and verified the knowledge mapping logic with test data in both formats. The `collection_names` path correctly returns `legacy: false`, the old `collection_name` path still returns `legacy: true`. Backend wasn't needed for this check since it's a frontend-only mapping change.

@Classic298 Have you had a chance to look at the fix? Let me know if you see anything off or if there's a specific scenario you'd like me to verify.

PR #26147, opened 2026-06-18
perf: dont render toolcall args when hidden massive perf improvement

Classic298:
ToolCallDisplay decoded attributes.arguments on every reactive update, even when the panel is collapsed and the args are never rendered.

For an in-progress tool call the backend re-emits a growing arguments attribute on every streamed delta, so this ran an O(n) html-entities decode each frame for content that is not shown, a major contributor to UI stutter while a large tool call (e.g.
replace_file_content) streams.

Gate the decode on visibility: only decode when the panel is expanded or when embeds reference the args. Output is unchanged in every case (the args are never rendered while collapsed without embeds).

Combined with the args-parse dedupe on the same branch because the two edits touch adjacent lines and cannot be merged into dev independently without a conflict.

> [!NOTE]

### DIFF
--- a/src/lib/components/common/ToolCallDisplay.svelte
+++ b/src/lib/components/common/ToolCallDisplay.svelte
@@ -51,31 +51,6 @@
 		}
 	}
 
-	function formatJSONString(str: string) {
-		try {
-			const parsed = parseJSONString(str);
-			if (typeof parsed === 'object') {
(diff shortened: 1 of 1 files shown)


tjbck:
b7934e918223ec0a9e972accd647a8654e503156

PR #26167, opened 2026-06-18
feat(knowledge): per-file full-context injection in knowledge bases

santiilpilota23:
**Linked issue/discussion:** (opening a discussion is in progress — this PR is a concrete implementation ready for review)

---

## Problem

Knowledge bases currently support two retrieval modes for a collection:
1. **RAG** (default) — relevant chunks are retrieved by vector similarity
2. **Full context** (`context: 'full'` on the item) — *all* files in the KB are injected verbatim

There is no middle ground: you cannot mark a **single file** for verbatim injection while keeping the rest of the KB on RAG.

### Why this matters

A common pattern when building document-based knowledge bases is to include a **structured index or README file** — a human- or machine-generated map that describes what the KB contains, how it is organized, and how to navigate it. This file is small (a few KB) but must be present in the LLM context on *every* query so the model can reason about the structure of the knowledge base before (and after) retrieving specific chunks.

With the current implementation, the only way to achieve this is to put the entire KB in full-context mode, which defeats the purpose of RAG for large collections. Alternatively, a Filter pipe can inject the file manually, but this requires custom code per deployment.

The natural solution is a **per-file context flag**: mark a single file (e.g. the README / index) as `context='full'` so it is always injected verbatim, while all other files in the same KB continue to use vector search.

---

### Database
- New Alembic migration (`1a2b3c4d5e6f`) — adds `context TEXT NULL` to `knowledge_file`
  *(Note: `down_revision` targets `f1e2d3c4b5a6`; maintainers may need to adjust the chain)*

### `models/knowledge.py`
- `KnowledgeFile` ORM model: `context` column
- `KnowledgeFileModel` Pydantic model: `context: Optional[str] = None`
- `KnowledgeTable.get_files_with_context_by_id()` — returns `(FileModel, context)` tuples for a KB
- `KnowledgeTable.update_file_context_by_id()` — sets/clears the context flag for one file in a KB

### `retrieval/utils.py`
- `get_sources_from_items()`: when a collection item is **not** in whole-KB full-context mode, iterate its files; any file with `context='full'` is appended directly to `query_results` (bypassing chunked retrieval), then the remaining files go through normal vector search as before

### `routers/knowledge.py`
- New form: `KnowledgeFileContextForm(file_id, context)`
- New endpoint: `POST /{id}/file/context` — requires KB `write` access; sets or clears `context` on a single `knowledge_file` row

---

## Behaviour

```
KB (RAG mode)
├── README.md    [context='full']  → always injected verbatim on every query
├── section-01.md                 → vector RAG
├── section-02.md                 → vector RAG
└── ...
```

A query on this KB returns:
1. The full text of `README.md` (directly, no chunking, no similarity threshold)
2. The top-K relevant chunks from the remaining sections (vector search)

The model always has the map of the knowledge base, and always retrieves the relevant detail.

---

## Checklist

- [x] **Target branch:** Targets `dev`
- [x] **Description:** Provided above
- [ ] **Linked Issue/Discussion:** To be linked — discussion being opened in parallel
- [x] **Self-Review:** Code reviewed for correctness and consistency with project patterns
- [x] **No Unchecked AI Code:** Changes have been manually reviewed and tested in a local Open WebUI deployment
- [x] **Architecture:** No new settings introduced; uses existing `context` pattern from collection-level full-context mode

---

### Added

- Per-file full-context injection for knowledge bases: individual files can be flagged with `context='full'` via the new `POST /{id}/file/context` endpoint, causing them to be injected verbatim into every query against that KB while the remaining files continue to use vector RAG

### Changed

- `get_sources_from_items()` in `retrieval/utils.py`: for collection items not in whole-KB full-context mode, files marked `context='full'` are now injected directly before the vector search step

---

### DIFF
--- a/backend/open_webui/migrations/versions/1a2b3c4d5e6f_add_context_to_knowledge_file.py
+++ b/backend/open_webui/migrations/versions/1a2b3c4d5e6f_add_context_to_knowledge_file.py
@@ -1,0 +1,27 @@
+"""Add context column to knowledge_file table
+
+Revision ID: 1a2b3c4d5e6f
+Revises: f1e2d3c4b5a6
+Create Date: 2026-06-18 00:00:00.000000
+
+Adds an optional `context` column to the `knowledge_file` join table.
+When set to 'full', the file is injected directly into the LLM context
+(bypassing chunked RAG retrieval) while the rest of the knowledge base
+still uses vector search.
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = '1a2b3c4d5e6f'
+down_revision = 'f1e2d3c4b5a6'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('knowledge_file', sa.Column('context', sa.Text(), nullable=True))
+
+
+def downgrade():
+    op.drop_column('knowledge_file', 'context')
--- a/backend/open_webui/models/knowledge.py
+++ b/backend/open_webui/models/knowledge.py
@@ -105,6 +105,8 @@ class KnowledgeFile(Base):
     created_at = Column(BigInteger, nullable=False)
     updated_at = Column(BigInteger, nullable=False)
 
+    context = Column(Text, nullable=True)
+
     __table_args__ = (
         UniqueConstraint('knowledge_id', 'file_id', name='uq_knowledge_file_knowledge_file'),
         Index('ix_knowledge_file_directory_id', 'directory_id'),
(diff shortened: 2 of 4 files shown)


Classic298:
> There is no middle ground: you cannot mark a single file for verbatim injection while keeping the rest of the KB on RAG

you can.

santiilpilota23:
Thanks for the quick feedback. Could you point to the existing  mechanism? We're building a pipeline that pushes documents to KBs programmatically and we need to ensure a README/index file is always injected verbatim without going through the chat UI manually. Is there an API call / item format we should use instead?

Classic298:
1) https://github.com/open-webui/oikb
2) either click on the plus button in the UI, select files, select the uploaded file from the KB and then turn on full context for just that file or
3) have the AI query the file directly in full context using kb_exec (see docs) in full context, and the other files can be queried using query_knowledge_files

PR #26202, opened 2026-06-19
perf: batch tool-call and Responses API stream deltas

roSchmal:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** This PR references an existing Discussion - `Relates to #26200`.
- [ ] **Documentation:** No user-facing docs change required (internal streaming behavior only).
- [x] **Testing:** Manually tested with native tool-call streaming and Responses API streaming; verified tool-call and text update ordering is preserved and event volume is reduced when `stream_delta_chunk_size` is set.
- [x] **No Unchecked AI Code:** This PR has undergone human review and manual testing.
- [x] **Architecture:** Reuses the existing chunk-size setting; no new settings introduced.
- [x] **Title Prefix:** The PR title uses the `perf` prefix.

## What

Route native Chat Completions `delta.tool_calls` and Responses API
`response.*.delta` events through the existing delta batching path
(`queue_pending_delta_data` / `flush_pending_delta_data`), governed by the same
`stream_delta_chunk_size` / `CHAT_RESPONSE_STREAM_DELTA_CHUNK_SIZE` threshold
already used for normal assistant text deltas.

## Why

Previously only the Chat Completions text path respected the stream delta chunk
size. Native tool-call argument deltas and Responses API delta events called
`event_emitter` immediately, so large tool-call argument streams produced many
tiny `chat:completion` events - increasing Socket.IO traffic and frontend render
work even when stream delta chunking was configured.

## How

- Added `last_delta_data_kind` tracking and a shared `queue_pending_delta_data()`
  helper that flushes on kind change (content vs tool_call) and emits only when
  `delta_count >= delta_chunk_size`.
- Native tool-call deltas now assign the outgoing payload and flow through the
  shared batching block instead of emitting immediately.
- Responses API `response.*.delta` events are batched
  (`function_call_arguments.delta` -> tool_call, others -> content); non-delta
  lifecycle/finalization events flush pending batches then emit immediately so
  metadata, citations, final output, and errors stay ordered.

No new config; reuses the existing chunk-size settings.

Relates to #26200

# Changelog Entry

### Changed

- Native Chat Completions tool-call argument deltas and Responses API streaming
  deltas now respect the same `stream_delta_chunk_size` batching threshold as
  normal assistant text deltas, reducing `chat:completion` event volume during
  tool-call and Responses API streaming.

---

### Additional Information

- Affects `backend/open_webui/utils/middleware.py` only (+56/-18).

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -4022,10 +4022,12 @@ async def streaming_chat_response_handler(response, ctx):
                         int(metadata.get('params', {}).get('stream_delta_chunk_size') or 1),
                     )
                     last_delta_data = None
+                    last_delta_data_kind = None
 
                     async def flush_pending_delta_data(threshold: int = 0):
                         nonlocal delta_count
                         nonlocal last_delta_data
+                        nonlocal last_delta_data_kind
 
                         if delta_count >= threshold and last_delta_data:
                             await event_emitter(
@@ -4036,6 +4038,21 @@ async def streaming_chat_response_handler(response, ctx):
                             )
                             delta_count = 0
                             last_delta_data = None
+                            last_delta_data_kind = None
+
+                    async def queue_pending_delta_data(delta_data: dict, delta_data_kind: str):
+                        nonlocal delta_count
+                        nonlocal last_delta_data
+                        nonlocal last_delta_data_kind
+
+                        if last_delta_data_kind and last_delta_data_kind != delta_data_kind:
+                            await flush_pending_delta_data()
+
+                        delta_count += 1
+                        last_delta_data = delta_data
+                        last_delta_data_kind = delta_data_kind
+                        if delta_count >= delta_chunk_size:
+                            await flush_pending_delta_data(delta_chunk_size)
 
                     async for line in response.body_iterator:
                         line = line.decode('utf-8', 'replace') if isinstance(line, bytes) else line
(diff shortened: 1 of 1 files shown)


Classic298:
wow very good catch this can lead to massive performance improvements in the frontend when streaming tool calls, i didnt catch it but i should have caught it

good catch

did you test this thoroughly to verify no regressions?
but this seems like a very smart thing to do

roSchmal:
Did not perform a full regression test, but tested it in pretty complex agent looks (content at the end > 200kb) and 5 differnt tools used. Works fine so far and streaming breaks and browser loads drops massively.

Classic298:
@roSchmal do you experience less CPU and RAM usage and also a more snappy and less choppy experience while scrolling in the browser during tool call streaming?

roSchmal:
Absolutly, as soon as I had like 3 tabs open running agents on OWUI the PC fan switched to high and either streaming broke up (in background for sure in continued, so when hitting F5 result was there) or browser was even not usable anymore stable due to high performance consumption.
Most of the problem was that for models using Responses API the seeting was completly ignored.

Played around a bit with the chunk size and even 256 is completly fine, ending up with about 1 socket message per second which still feels fluently.

Classic298:
@roSchmal thanks for letting me know you tested it

i also made some PRs here that, e.g.: completely DEFER the rendering of the tool call if the tool call is not expanded

https://github.com/open-webui/open-webui/pull/26147
 and small rendering enhancements: https://github.com/open-webui/open-webui/pull/26146

so as long as the tool call is not expanded, the PR will ensure it completely bypasses the rendering and therefore: basically no CPU usage

roSchmal:
@Classic298 - if you're up for it, I'd be glad to exchange ideas sometime. I'm working on some comparable tools.

tjbck:
Should be addressed in dev, testing wanted here!

PR #26321, opened 2026-06-25
breaking: sandbox client-side Pyodide at an opaque origin by default

Classic298:
## Summary

Open WebUI's client-side Python (Pyodide) runs in a Web Worker that shares the app's origin. Because the worker is same-origin, executed Python can issue credentialed requests to the app's own API and reach same-origin browser context. This change runs Pyodide in a sandboxed, opaque-origin iframe by default, so executed Python is isolated from the session cookie, auth token, localStorage, DOM and same-origin endpoints, while full Python, JS and external (cross-origin) fetch keep working.

It also marks Pyodide as legacy in the admin Code Execution settings.

## Changes

- Host Pyodide in a `sandbox="allow-scripts"` iframe (no `allow-same-origin`) so it runs at an opaque origin. New `static/pyodide-sandbox.{html,js}` mirror the worker's `execute` / `fs:*` message protocol (without IDBFS).
- `PyodideSandboxHost`: a `Worker`-shaped wrapper around the iframe, so the call sites are unchanged in shape.
- `createPyodideWorker()` factory selects the host; the four Pyodide spawn sites use it.
- IDBFS file persistence needs a stable origin, so it is gated behind a new `ENABLE_PYODIDE_FILE_PERSISTENCE` flag (default `false`). With it set, the previous same-origin worker is used and persistence works.
- Serve the `/pyodide/` runtime and wheels with `Access-Control-Allow-Origin: *` so the opaque-origin iframe can load them (public assets).
- Mark `pyodide` as `(Legacy)` in the Code Execution and Code Interpreter engine selectors, with a note about the flag.

## Configuration

`ENABLE_PYODIDE_FILE_PERSISTENCE` (env, default `false`):
- `false`: Pyodide runs in the opaque-origin sandbox, no IDBFS persistence. **Breaking** for setups that rely on persisted Pyodide files.
- `true`: Pyodide runs same-origin, persistence works, at the cost of the same-origin reach above.

## Verification

Tested in a browser against a running instance:
- Pyodide loads and executes inside the opaque-origin iframe (the CORS-served runtime loads cross-origin).
- A credentialed request from the sandbox to an app API endpoint does not reach it, while the same call from the app page is authorised normally.
- External cross-origin fetch from the sandbox still works.

> [!NOTE]

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -1053,6 +1053,13 @@ if OFFLINE_MODE:
     ENABLE_VERSION_UPDATE_CHECK = False
 
 ####################################
+# ENABLE_PYODIDE_FILE_PERSISTENCE
+####################################
+
+# false (default) = opaque-origin sandbox, no persistence; true = same-origin worker with IDBFS persistence
+ENABLE_PYODIDE_FILE_PERSISTENCE = os.getenv('ENABLE_PYODIDE_FILE_PERSISTENCE', 'false').lower() == 'true'
+
+####################################
 # Audit logging
 ####################################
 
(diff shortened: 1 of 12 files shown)


tjbck:
Addressed.

PR #26328, opened 2026-06-26
fix(tools): handle empty query in search_chats without crashing

### LINKED ISSUE #26310: issue: Built-in search_chats tool crashes with empty query — 'ChatTitleIdResponse' object has no attribute 'chat'
### Bug Report

**Installed Open WebUI Version:** affects `main` (built-in tools, `backend/open_webui/tools/builtin.py`, first added 2026-06-01)

**Environment:** Native Function Calling enabled, model with built-in `search_chats` tool available.

---

### What happened

Calling the built-in `search_chats` tool with an **empty `query`** (e.g. the model trying to fetch "the last N chats") crashes with:

```
{"error": "'ChatTitleIdResponse' object has no attribute 'chat'"}
```

A non-empty query works fine and returns results normally.

### Root cause (traced in source)

`search_chats` in [`backend/open_webui/tools/builtin.py`](https://github.com/open-webui/open-webui/blob/main/backend/open_webui/tools/builtin.py) calls `Chats.get_chats_by_user_id_and_search_text(...)` and then accesses `chat.chat.get('history', {})`:

```python
chats = await Chats.get_chats_by_user_id_and_search_text(
    user_id=user_id, search_text=query, include_archived=False, skip=0, limit=count * 3,
)
for chat in chats:
    ...
    messages = chat.chat.get('history', {}).get('messages', {})   # <-- assumes ChatModel
```

But `get_chats_by_user_id_and_search_text` in [`backend/open_webui/models/chats.py`](https://github.com/open-webui/open-webui/blob/main/backend/open_webui/models/chats.py) takes a different branch for empty search text:

```python
async def get_chats_by_user_id_and_search_text(...) -> list[ChatModel]:   # annotated ChatModel
    search_text = sanitize_text_for_db(search_text).lower().strip()
    if not search_text:
        return await self.get_chat_list_by_user_id(...)   # returns list[ChatTitleIdResponse] !
```

`get_chat_list_by_user_id` returns `list[ChatTitleIdResponse]` (only `id`, `title`, `updated_at`, `created_at` — **no `.chat`**), which violates the function's own `-> list[ChatModel]` annotation. `search_chats` then accesses `.chat` and raises `AttributeError`.

### Steps to reproduce

1. Enable Native Function Calling on a model so the `search_chats` built-in tool is available.
2. Prompt so the model calls `search_chats` with an empty query string (e.g. "show me my last 3 chats").
3. Observe the `'ChatTitleIdResponse' object has no attribute 'chat'` error.

### Expected

`search_chats` with an empty query should return recent chats (with whatever snippet is available) instead of crashing.

### Suggested fix

Either:
- Make `search_chats` access `.chat` defensively (e.g. `getattr(chat, 'chat', None)` and skip snippet extraction when absent), or
- Have the empty-search branch of `get_chats_by_user_id_and_search_text` return full `ChatModel` objects to match its `-> list[ChatModel]` annotation.

Possibly related: #25641 (other built-in tool crashes via native function calling).

itxaiohanglover:
## Description

When `search_chats` is called with an empty query, `get_chats_by_user_id_and_search_text` returns `ChatTitleIdResponse` objects (no `.chat` attribute) instead of `ChatModel`. The code then tries `chat.chat.get('history', {})` which crashes with `AttributeError: 'ChatTitleIdResponse' object has no attribute 'chat'`.

Using `getattr(chat, 'chat', None)` safely handles both types.

Closes #26310

## Changelog

- Fixed `search_chats` tool crash when called with an empty query string

### DIFF
--- a/backend/open_webui/tools/builtin.py
+++ b/backend/open_webui/tools/builtin.py
@@ -1099,7 +1099,8 @@ async def search_chats(
 
             # Find a matching message snippet
             snippet = ''
-            messages = chat.chat.get('history', {}).get('messages', {})
+            chat_data = getattr(chat, 'chat', None) or {}
+            messages = chat_data.get('history', {}).get('messages', {})
             lower_query = query.lower()
 
             for msg_id, msg in messages.items():


tjbck:
Addressed in dev!

PR #26333, opened 2026-06-26
feat: forward client User-Agent to ALL model backends

Classic298:
Adds a {{USER_AGENT}} custom-header placeholder that relays the inbound client's User-Agent to upstream model backends, so providers see the real client instead of Open WebUI's internal aiohttp UA. This makes upstream usage/cost attribution and backend telemetry possible, and is opt-in per-connection (no global flag): admins add {{USER_AGENT}} to a connection's custom headers in Admin > Settings > Connections.

The placeholder is sourced from the live inbound request (with a metadata fallback for detached RAG/tool calls), so it resolves on every prompt-sending path, not just chat completions:

- OpenAI completions, Responses API, and proxy — all route through get_headers_and_cookies, which now passes the request into get_custom_headers.
- Anthropic Messages API (/api/v1/messages) — already covered, it delegates to the chat completion handler.
- Ollama (/api/chat, /v1/completions, /v1/chat/completions, /v1/messages, /v1/responses) — previously had no custom-header support at all; send_request now applies per-connection custom headers (with templating) for every Ollama prompt endpoint.

Custom headers are applied after the built-in user-info headers so explicit admin-configured headers take precedence. The other existing placeholders ({{CHAT_ID}}, {{USER_ID}}, ...) now also work on the newly covered paths.

Frontend: the connection editor's Headers field is now shown for Ollama connections too (previously gated to non-Ollama), so the placeholder can be configured there.

Ref: open-webui/open-webui#26159

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1147,6 +1147,7 @@ async def chat_completion(
 
         metadata = {
             'user_id': user.id,
+            'user_agent': request.headers.get('user-agent', '') or '',
             'chat_id': form_data.pop('chat_id', None) or '',
             'user_message': user_message,
             'user_message_id': user_message.get('id') if user_message else None,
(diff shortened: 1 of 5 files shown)


gaby (commented):
LGTM

PR #26355, opened 2026-06-27
fix: render image MCP EmbeddedResource blobs as model vision input

inesaranab:
### Description

MCP tools can return binary content as an `EmbeddedResource` wrapping `BlobResourceContents`
(e.g. image attachments). In `process_tool_result`, the MCP `resource` branch only read
`resource.text`, so binary `blob` payloads were silently dropped and the model received an
empty tool result.

This change handles the `blob` case: `image/*` blobs are routed into `tool_result_files` as
a `data:` URI, which the existing native-tool path then forwards to the model as an
`input_image` part. Non-image blobs get a clear placeholder instead of vanishing.

It must remain a `data:` URI (not `get_file_url_from_base64`) so it passes the existing
`url.startswith('data:')` check that gates `input_image` forwarding.

### Fixed

- MCP `EmbeddedResource` image blobs are now passed to the model as vision input instead of being silently dropped.

### Additional Information

Tested against a Jira MCP server returning PNG attachments via native function calling — the
model now actually sees and describes the images, rather than receiving an empty result.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -1154,12 +1154,26 @@ async def process_tool_result(
                     elif item.get('type') == 'resource':
                         resource = item.get('resource', {})
                         text = resource.get('text', '')
+                        blob = resource.get('blob', '')
                         if isinstance(text, str) and text:
                             try:
                                 text = json.loads(text)
                             except json.JSONDecodeError:
                                 pass
                             tool_response.append(text)
+                        elif blob:
+                            r_mime = resource.get('mimeType', '') or 'application/octet-stream'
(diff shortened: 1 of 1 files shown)


inesaranab:
image

inesaranab:
Opened a discussion for context: #26356

tjbck:
Addressed in dev!

inesaranab:
> Addressed in dev!

Thank you! Me an my team and really eager for this to be in production!

moranbw:
@inesaranab do you know if this ever got merged into production? I'm still seeing the issue in `0.11.3`

inesaranab:
@moranbw no idea, tbh

PR #26445, opened 2026-06-30
fix: convert BannerModel to dict before storing in config defaults

BlackPool25:
### Description

Fix JSON serialization crash when `WEBUI_BANNERS` environment variable is set. The banners are parsed from JSON and validated against `BannerModel` schema, but then stored as raw Pydantic `BaseModel` instances in the config defaults dict. When `Config.seed_defaults()` writes these values to the database JSON column at startup, `json.dumps()` fails because Pydantic objects are not natively JSON serializable.

### Fixed

- Added `.model_dump(mode='json')` after `BannerModel` validation in the `WEBUI_BANNERS` env var parser (`backend/open_webui/config.py`), so the stored values are plain JSON-serializable dicts

### Additional Information

This fix follows the same serialization pattern already used by the `set_banners` API endpoint (`routers/configs.py:801`), which correctly calls `model_dump()` before persisting data.

> [!NOTE]

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -2080,7 +2080,7 @@ class BannerModel(BaseModel):
 
 try:
     banners = json.loads(os.getenv('WEBUI_BANNERS', '[]'))
-    banners = [BannerModel(**banner) for banner in banners]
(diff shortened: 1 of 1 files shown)


Classic298:
references issue?

Issue #26647, opened 2026-07-01
issue: MCP OAuth 2.1: authorization endpoint discovery fails for a Remote MCP server that works in Claude, so the OAuth flow never starts


dhruvbhavsar-ucla:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Pip Install

### Open WebUI Version

v0.10.1

### Operating System

Ubuntu 22

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

Open WebUI should initiate the OAuth flow by redirecting the browser to the MCP server's authorization endpoint, similar to Claude Desktop/Web.

### Actual Behavior

After successfully registering the OAuth client, clicking "Access" results in:

GET /oauth/clients/mcp:wappnet/authorize
HTTP 500

Backend traceback:
RuntimeError: Missing "authorize_url" value

Stack trace:
```
File ".../open_webui/utils/oauth.py", line 1131, in handle_authorize
    return await client.authorize_redirect(request, redirect_uri_str, **kwargs)
File ".../authlib/integrations/base_client/async_app.py", line 105, in create_authorization_url
    raise RuntimeError('Missing "authorize_url" value')
RuntimeError: Missing "authorize_url" value
```

### Steps to Reproduce

1. Admin → External Tools
[...]

Issue #21183, opened 2026-02-05
Regression: MCP OAuth 2.1 discovery fails when authorization_servers URL path doesn't match well-known endpoint location


ashavolian:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

v0.7.2

### Operating System

ubuntu

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

MCP OAuth 2.1 client registration should succeed when connecting to an MCP server that implements the Protected Resource discovery chain, even when the authorization_servers URL contains a path component (e.g., https://example.com/oauth) but the OAuth metadata is served at the domain root (https://example.com/.well-known/oauth-authorization-server).
In my case I'm trying to connect an enterprise glean mcp

### Actual Behavior

OAuth client registration fails because OpenWebUI strictly appends /.well-known/oauth-authorization-server to the authorization_servers URL from Protected Resource metadata, without falling back to the domain root when that fails.

### Steps to Reproduce

- Go to Admin Settings → External Tools → Add Server
- Add an MCP server with OAuth 2.1 authentication where:
- The MCP endpoint returns 401 with WWW-Authenticate: Bearer resource_metadata="https://example.com/.well-known/oauth-protected-resource/mcp"
[...]

PR #26654, opened 2026-07-01
fix: resolve MCP OAuth authorize_url discovery and fail clearly (#26647)

SNaytiP:
### What happens

Connecting a Remote MCP server that uses OAuth 2.1 fails with an HTTP 500 and `RuntimeError: Missing "authorize_url" value` when you try to authorize — even though registering the client appeared to succeed.

### Trouble shooting

The OAuth flow has two steps:

1. **Register** the client : Open WebUI discovers the OAuth server's details, then registers.
2. **Authorize** : Open WebUI redirects the user to the server's login page to approve access.

The bug is that step 1 was allowed to "succeed" even when discovery never found the server's **authorization endpoint** (the login URL). Registration fell back to a generic `/register` call and saved a client with no login URL attached. Nothing looked wrong yet — the error only appeared in step 2, when authlib tried to build the redirect and had no URL to use. That's why it looked like an authorize bug when the real problem was back in registration.

### The fix

Three small changes, from the root cause outward:

- **Fail fast at registration** — if discovery can't find the authorization endpoint, stop with a clear error instead of saving a broken client. (`get_oauth_client_info_with_dynamic_client_registration`)
- **Pass the endpoints directly** — when Open WebUI already knows the server's login/token URLs, hand them to authlib explicitly instead of hoping it can re-fetch them later. (`OAuthClientManager.add_client`)
- **Fail gracefully at authorize** — for any client that was already saved in a broken state before this fix, return a clear HTTP 400 instead of an uncaught 500. (`handle_authorize`)

### Scope & limitations

This PR implements the **"fail gracefully with a descriptive error"** option requested in the issue. It is intentionally focused and does **not** change OAuth discovery itself.

For servers where discovery already resolves the authorization endpoint, this also removes a failure mode by passing the endpoints to authlib explicitly. But for a server whose metadata Open WebUI cannot discover at all (likely the reporter's case, since the same server works in Claude), this PR turns the confusing 500 into a clear, actionable error at registration time — it does not by itself make that server connect.

Making such servers connect requires completing the MCP discovery chain (`WWW-Authenticate → Protected Resource Metadata → Authorization Server Metadata`), which is tracked separately in #19794 and #21183 and is a larger, higher-risk change best kept out of this PR.

Addresses #26647 (graceful-failure path). Discovery-chain fix tracked in #19794 / #21183.

# Pull Request Checklist

- [x] **Linked Issue/Discussion:** This PR references an existing Issue — Addresses #26647 (graceful-failure path; see Scope & limitations).
- [x] **Changelog:** A changelog entry following [[Keep a Changelog](https://keepachangelog.com/)](https://keepachangelog.com/) format is included at the bottom.
- [ ] **Documentation:** N/A — bug fix; no documented behavior changes.
- [x] **Testing:** Verified with mock-based unit tests (no live OAuth server required). The tests and their results are documented below.
- [x] **No Unchecked AI Code:** This PR has undergone thorough human review AND manual testing.
- [x] **Architecture:** No new settings or endpoints. Behavior change is limited to failing early/clearly and passing already-known endpoints to authlib explicitly.
- [x] **Git Hygiene:** The PR is atomic (one logical change), branched on `dev`, and contains no unrelated commits.
- [x] **Title Prefix:** The PR title uses the `fix` prefix.

# Changelog Entry

### Description

- Connecting a Remote MCP server over OAuth 2.1 could register a client even when the authorization server metadata (and therefore the authorization endpoint) was never discovered. This stored a client with no authorization endpoint, so starting the OAuth flow crashed with an uncaught `RuntimeError: Missing "authorize_url" value` (HTTP 500).

### Changed

- Dynamic client registration now **fails fast with a descriptive error** when authorization-server metadata cannot be discovered, instead of registering an unusable client.
- `OAuthClientManager.add_client()` now passes `authorize_url` / `access_token_url` to authlib explicitly when the server metadata is already known, rather than relying solely on refetching `server_metadata_url`.

### Fixed

- MCP OAuth 2.1 authorization no longer returns an uncaught HTTP 500 (`Missing "authorize_url" value`). Registration surfaces the real reason, and the authorize endpoint returns a clear HTTP 400 for clients that were registered before this validation (#26647).

---

### Additional Information

Backend-only change in `backend/open_webui/utils/oauth.py` and the registration error detail in `backend/open_webui/routers/configs.py`.

### How this was tested

Verified locally with three mock-based unit tests (no live OAuth/MCP server needed) — one per change. All three pass:

```
test_dynamic_registration_fails_fast_when_authorization_endpoint_undiscovered PASSED
test_handle_authorize_returns_400_when_authorize_url_missing               PASSED
test_add_client_passes_explicit_endpoints_from_server_metadata            PASSED
3 passed
```

**Test 1 — registration fails fast when the authorization endpoint isn't discovered.**
Simulates a server whose discovery yields no authorization-server metadata. Asserts that dynamic registration now raises a descriptive error (naming `authorization_endpoint`) instead of returning a broken client. This is the root-cause guard.

```python
async def run():
    resource_metadata = SimpleNamespace(
        resource='https://mcp.example.com',
        scopes_supported=None,
        get_discovery_urls=lambda url: [],  # discovery finds nothing
    )
    with patch.object(oauth.Config, 'get', new=AsyncMock(return_value='https://owui.example.com')), \
         patch.object(oauth, 'get_protected_resource_metadata', new=AsyncMock(return_value=resource_metadata)):
        with pytest.raises(Exception) as exc_info:
            await oauth.get_oauth_client_info_with_dynamic_client_registration(
                request=MagicMock(), client_id='mcp:example', oauth_server_url='https://mcp.example.com',
            )
    assert 'authorization_endpoint' in str(exc_info.value)
```

**Test 2 — authorize returns HTTP 400, not an uncaught 500.**
Simulates a client that was already saved in a broken state (authlib raises `Missing "authorize_url"`). Asserts `handle_authorize()` converts it into a clear `HTTPException(400)` instead of a raw 500.

```python
async def run():
    manager = OAuthClientManager(MagicMock())
    fake_client = SimpleNamespace(
        authorize_redirect=AsyncMock(side_effect=RuntimeError('Missing "authorize_url" value'))
    )
    manager.get_client = AsyncMock(return_value=fake_client)
    manager.get_client_info = AsyncMock(return_value=SimpleNamespace(redirect_uris=['https://owui.example.com/cb']))
    with patch.object(oauth, 'build_oauth_request_params', return_value={}):
        with pytest.raises(HTTPException) as exc_info:
            await manager.handle_authorize(MagicMock(), 'mcp:example')
    assert exc_info.value.status_code == 400
```

**Test 3 — endpoints are passed to authlib explicitly (even when `issuer` is `None`).**
Reproduces the exact bug scenario: `issuer` is `None` but server metadata is known. Asserts `add_client()` hands `authorize_url` / `access_token_url` straight to authlib, so the flow works without relying on a later re-fetch.

```python
manager = OAuthClientManager(MagicMock())
manager.oauth = MagicMock()
server_metadata = SimpleNamespace(
    authorization_endpoint='https://as.example.com/authorize',
    token_endpoint='https://as.example.com/token',
    code_challenge_methods_supported=['S256'],
)
client_info = SimpleNamespace(
    client_id='cid', client_secret='secret', scope=None,
    token_endpoint_auth_method=None, issuer=None, server_metadata=server_metadata,
)
manager.add_client('mcp:example', client_info)
kwargs = manager.oauth.register.call_args.kwargs
assert kwargs['authorize_url'] == 'https://as.example.com/authorize'
assert kwargs['access_token_url'] == 'https://as.example.com/token'
```

- [x] By submitting this pull request, I confirm that I have read and fully agree to the [[Contributor License Agreement (CLA)](https://github.com/open-webui/open-webui/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT)](https://github.com/open-webui/open-webui/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT), and I am providing my contributions under its terms.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/configs.py
+++ b/backend/open_webui/routers/configs.py
@@ -196,7 +196,7 @@ async def register_oauth_client(
         log.debug(f'Failed to register OAuth client: {e}')
         raise HTTPException(
             status_code=400,
-            detail=f'Failed to register OAuth client',
+            detail=f'Failed to register OAuth client: {e}',
         )
 
 
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -536,6 +536,20 @@ async def get_oauth_client_info_with_dynamic_client_registration(
                             log.error(f'Error parsing OAuth metadata from {url}: {e}')
                             continue
 
+        # Fail fast if authorization server metadata discovery did not resolve an
+        # authorization endpoint. Otherwise registration can still "succeed" (via
+        # the /register fallback below) while issuer/server_metadata stay unset,
+        # which later crashes at authorize time with authlib's
+        # RuntimeError: Missing "authorize_url" value. (#26647)
+        if oauth_server_metadata is None or not oauth_server_metadata.authorization_endpoint:
+            log.error(f'OAuth authorization server metadata discovery failed for {oauth_server_url}')
+            raise Exception(
+                'Could not discover the OAuth authorization server metadata '
+                f'(authorization_endpoint) for {oauth_server_url}. The MCP server must '
+                'expose RFC 8414 / RFC 9728 discovery documents so Open WebUI can '
+                'resolve where to send users to authorize.'
+            )
+
         registration_url = None
         if oauth_server_metadata and oauth_server_metadata.registration_endpoint:
             registration_url = str(oauth_server_metadata.registration_endpoint)
@@ -801,6 +815,17 @@ class OAuthClientManager:
             },
             'server_metadata_url': (oauth_client_info.issuer if oauth_client_info.issuer else None),
         }
+
+        # Defense-in-depth: when the server metadata is already known, pass the
+        # authorization/token endpoints explicitly so authlib does not rely solely
+        # on refetching server_metadata_url (which may be missing/unreachable) to
+        # resolve them. Prevents RuntimeError: Missing "authorize_url". (#26647)
+        server_metadata = oauth_client_info.server_metadata
+        if server_metadata is not None:
+            if getattr(server_metadata, 'authorization_endpoint', None):
+                kwargs['authorize_url'] = str(server_metadata.authorization_endpoint)
+            if getattr(server_metadata, 'token_endpoint', None):
+                kwargs['access_token_url'] = str(server_metadata.token_endpoint)
 
         # Default to S256 for OAuth 2.1 (PKCE is mandatory per RFC 9700)
         kwargs['code_challenge_method'] = 'S256'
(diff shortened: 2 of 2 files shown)


tjbck:
Thanks!

PR #26705, opened 2026-07-02
fix: include streamed assistant content for memory review

### LINKED ISSUE #26651: issue: Background memory review never runs for streamed chats (assistant_message has no `content`)
### Check Existing Issues

### Installation Method

Docker

### Open WebUI Version

v0.10.2

_No response_

### Operating System

Debian 12

### Browser (if applicable)

Chrome

### Confirmation

### Expected Behavior

With `ENABLE_MEMORY_BACKGROUND_REVIEW=true` (config key `memories.background_review.enable`), after every `memories.review_interval_turns` user turns the background reviewer should run, extract durable preferences/facts, and persist them as memories with `meta.created_by = "background_review"`. This should work for normal chats in the UI.

### Actual Behavior

The background review never runs for normal chats. No memories with `meta.created_by = "background_review"` are ever created, regardless of how many turns the conversation has. This affects **both** the Chat Completions API and the Responses API, because the root cause is in the shared streaming response handler (the UI always streams).
The feature only works in the non-streaming code path, which normal UI usage never hits.

### Steps to Reproduce

1. Deploy Open WebUI `v0.10.2` (Docker `ghcr.io/open-webui/open-webui:main`, commit `ecd48e2f7`).
2. Ensure memories are enabled globally: `memories.enable = true`.
3. Enable the background review. Because the value is seeded into the `config` table once at first boot, on an existing instance set it via the config import endpoint (admin token):
   ```bash
   curl -X POST http://HOST/api/v1/configs/import \
     -H "Authorization: Bearer <ADMIN_TOKEN>" -H "Content-Type: application/json" \
     -d '{"config": {"memories.background_review.enable": true, "memories.review_interval_turns": 4}}'

### Logs & Screenshots

There is no error log — the reviewer returns early silently. To confirm, add a debug log at the gate in review_memory_after_turn; it returns at the assistant_content emptiness check before reaching the config/interval logic.

Empirical signal: GET /api/v1/memories/ never contains any entry with meta.created_by == "background_review" after streamed chats, but does after the non-streaming request in step 8.

### Additional Information

Root cause (transport-independent, streaming vs non-streaming):

The streaming response handler builds ctx['assistant_message'] with only output and no content key:

backend/open_webui/utils/middleware.py (~L5274)

```
ctx['assistant_message'] = {
    'output': output,
    **({'usage': usage} if usage else {}),
}
```
The non-streaming handler, by contrast, includes content (~L3566):

```
ctx['assistant_message'] = {
    'content': content,
    'output': response_output,
    **({'usage': usage} if usage else {}),
}
```
The review gate reads content directly and returns early when it is empty:

backend/open_webui/utils/memory.py (review_memory_after_turn, ~L425)

```
assistant_content = assistant_message.get('content', '')
if not isinstance(assistant_content, str) or not assistant_content.strip():
    return
```
Because the UI always streams, content is always empty here, so the reviewer never runs. Note the assistant text is available on the streaming path (it is accumulated into a local content variable and used for the webhook notification a few lines above) — it is simply not copied into ctx['assistant_message']. This looks like refactor drift from the migration to OR-aligned output items.

roian6:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Fixes #26651.
- [x] **Description:** Ensures streamed assistant completions populate `ctx['assistant_message']['content']` before background tasks run, so background memory review can evaluate normal UI chats.
- [ ] **Changelog:** See entry below.
- [ ] **Documentation:** Not applicable; backend bug fix only.
- [x] **Dependencies:** No dependency changes.
- [x] **Testing:** Static/syntax checks were performed locally; see below.
- [x] **No Unchecked AI Code:** The change was manually reviewed and kept minimal.
- [x] **Self-Review:** Confirmed the streaming path now has parity with the non-streaming assistant message shape.
- [x] **Architecture:** Reuses existing output-to-message conversion for Responses API fallback; no new setting.
- [x] **Git Hygiene:** Atomic one-file fix.

## Description

Fixes background memory review for streamed chats by including assistant text in `ctx['assistant_message']` before outlet filters/background tasks run.

The non-streaming path already sets both `content` and `output` on `ctx['assistant_message']`. The standard streaming path only set `output`, so `review_memory_after_turn()` read an empty `content` field and returned before checking the background review config/interval.

This patch:

- uses the accumulated Chat Completions streaming `content` value when available
- falls back to reconstructing assistant text from OR-aligned `output` with `convert_output_to_messages()` for Responses API streaming events
- keeps `output` and `usage` unchanged

## Tests

- `python3 -m py_compile backend/open_webui/utils/middleware.py backend/open_webui/utils/memory.py`
- `git diff --check`
- `uvx --python 3.11 ruff check --select E9,F63,F7,F82 backend/open_webui/utils/middleware.py`
- `uvx --python 3.11 ruff format --check backend/open_webui/utils/middleware.py`
- full one-file `ruff check backend/open_webui/utils/middleware.py` comparison: 59 baseline findings before/after, no diff
- AST-extracted `convert_output_to_messages()` probes for plain assistant text, Responses API output fallback, and tool-only output
- targeted added-line secret scan: no findings
- CRLF scan: false
- independent pre-PR review: passed, no blockers

# Changelog Entry

### Fixed

- Background memory review now receives assistant content for streamed chat completions.

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -5271,7 +5271,21 @@ async def streaming_chat_response_handler(response, ctx):
                     }
                 )
 
+                assistant_content = (
+                    content
+                    or '\n'.join(
+                        message.get('content', '')
+                        for message in convert_output_to_messages(
+                            output,
+                            raw=False,
+                            reasoning_format=get_reasoning_format(model),
+                        )
+                        if message.get('role') == 'assistant' and isinstance(message.get('content'), str)
+                    ).strip()
+                )
+
                 ctx['assistant_message'] = {
+                    'content': assistant_content,
                     'output': output,
                     **({'usage': usage} if usage else {}),
                 }


knguyen298:
I can confirm that this does fix the original issue referenced, on my installation.

tjbck:
Addressed in dev.

PR #26723, opened 2026-07-02
fix: validate direct-model attached knowledge against the caller

Classic298:
A request-scoped direct model (model_item.direct=true on the chat endpoints) carries client-supplied info.meta.knowledge that becomes the runtime model and is bound into the builtin knowledge tools as __model_knowledge__. Unlike persisted models, which validate attached file references against the author at create/update/import time, direct models skipped that check, so a caller could reference another user's file id and have builtin tools such as query_knowledge_files return chunks from it. Filter a direct model's attached knowledge to the entries the caller can actually read (get_accessible_folder_files) before it is used, so forged file or knowledge-base references are dropped. Persisted-model file sharing and knowledge-base access controls are unchanged.

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1004,6 +1004,15 @@ async def embeddings(request: Request, form_data: dict, user=Depends(get_verifie
     return await generate_embeddings(request, form_data, user)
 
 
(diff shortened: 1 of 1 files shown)


tjbck:
Addressed.

PR #26752, opened 2026-07-03
fix: context compaction not triggering with llama.cpp prompt caching

Classic298:
Context compaction failed to trigger for llama.cpp backends with prompt caching enabled, letting chats grow until the server rejected the request with a context overflow error. Two issues combined to cause this:

1. load_messages_from_db() dropped the usage field when reconstructing messages from the DB, so _exceeds_token_threshold() never saw any reported token counts and always fell back to rough char-based estimation.

2. The threshold check summed only input_tokens + output_tokens. llama.cpp reports cached prefix tokens separately (timings.cache_n, merged into the stored usage) and excludes them from prompt_n, so a mostly cached context of ~30k tokens was counted as only a few thousand.

Keep usage on DB-loaded messages so the threshold check can use it, count cache_n towards the total, and strip usage in strip_compaction_fields() after compaction runs so the non-standard field never reaches the LLM payload.

Reported and verified in
https://github.com/open-webui/open-webui/discussions/24410

> [!NOTE]


Franc1s396:
Hi, thanks for working on this fix. I ran into what appears to be the first issue described in this PR on Open WebUI v0.10.2, although I am using an OpenAI-compatible model rather than llama.cpp.

I enabled context compaction and set the threshold to 70,000 tokens, but a long-running chat was never compacted. Several consecutive responses were already reporting around 90K input tokens.

I exported the chat and checked the current branch:

- 304 messages
- 151 assistant messages with stored usage data
- 50 responses with `input_tokens` above 70,000
- Maximum reported input: 95,297 tokens
- No `contextSummary` checkpoint anywhere in the chat
- No context-compaction success or failure logs

After tracing the request path, I found that the saved assistant messages do contain `usage.input_tokens`, but `load_messages_from_db()` removes the `usage` field before `_exceeds_token_threshold()` runs.

I also replayed the threshold calculation using the exported chat:

- Provider-reported input tokens from the previous response: 93,130
- Fallback estimate after `usage` is removed: 58,309
- Configured threshold: 70,000

So Open WebUI sees roughly 58K instead of the actual 93K and never attempts compaction. This also explains why there are no compaction logs—the summarization step is never reached.

This suggests that the change to preserve `usage` during DB history loading, then remove it before sending messages to the model, is needed for OpenAI-compatible providers as well and is not limited to llama.cpp prompt caching.

I can provide a sanitized fixture or more details if that would help. I have avoided attaching the full chat export because it contains private conversation content.

Issue #26677, opened 2026-07-02
issue: Refreshing a chat toggles off the Terminal


chienhsiang:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.10.2

### Operating System

macOS Tahoe 26.5.1

### Browser (if applicable)

Chrom 149.0.7827.201

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

The Terminal should stay toggled on as in the previous version (e.g. v0.9.5).

### Actual Behavior

The Terminal is automatically toggled off.

### Steps to Reproduce

In a chat -> refreshing the page -> the Terminal becomes untoggled.

Before:
<img width="1335" height="951" alt="Image" src="https://github.com/user-attachments/assets/acd8d77b-ab26-4b4f-9396-0809f4fd6314" />

After refreshing the page:
<img width="1335" height="951" alt="Image" src="https://github.com/user-attachments/assets/f9e5f65f-2481-478a-b862-39ab13324fc5" />

### Logs & Screenshots

open-webui  | 2026-07-02 02:44:43.205 | INFO     | uvicorn.protocols.http.httptools_impl:send:483 - 192.168.65.1:37529 - "GET /_app/version.json HTTP/1.1" 200
[...]

PR #26775, opened 2026-07-04
fix: preserve terminal selection on chat refresh

initiallyqq:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes open-webui/open-webui#26677.
- [x] **Description:** Preserve terminal selection while terminal server data is still loading during chat refresh.
- [x] **Documentation:** Not required for this frontend bug fix.
- [x] **Testing:** Formatting was checked; full `svelte-check` was attempted and fails on existing unrelated project diagnostics.
- [x] **No Unchecked AI Code:** The change has been reviewed against the reported code path.
- [x] **Self-Review:** Self-reviewed the diff for scope and regressions.
- [x] **Architecture:** Uses existing Svelte stores and local UI state.
- [x] **Git Hygiene:** Atomic branch based on `dev`.
- [x] **Title Prefix:** Uses `fix:`.

# Changelog Entry

### Description

- Prevent chat refresh from clearing model/default terminal selection before terminal servers finish loading.

### Added

- Added a `terminalServersLoaded` store to distinguish an empty terminal list from terminal data that is still loading.

### Changed

- Delay stale terminal selection cleanup until terminal server loading completes.
- Preserve an already-enabled direct terminal instead of disabling it when `selectedTerminalId` has not been restored yet.

### Deprecated

- None.

### Removed

- None.

### Fixed

- Fixed terminal toggle being cleared after refreshing an existing chat.

### Security

- None.

### Breaking Changes

- None.

---

### Additional Information

Validation:

- `npx prettier --check src/lib/stores/index.ts 'src/routes/(app)/+layout.svelte' src/lib/components/chat/Chat.svelte src/lib/components/chat/ChatControls.svelte` passed.
- `npm run check` was attempted after `npm ci`; it still fails on existing unrelated diagnostics such as missing `file-saver` declarations and implicit `any` errors in other components.

### Screenshots or Videos

- Not included; this is a state restoration fix.

### DIFF
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -41,6 +41,7 @@
 		skills,
 		toolServers,
 		terminalServers,
+		terminalServersLoaded,
 		functions,
 		selectedFolder,
 		pinnedChats,
@@ -503,7 +504,7 @@
 				// Set Default Terminal — only if the referenced terminal actually exists
 				if (model?.info?.meta?.terminalId) {
 					const tid = model.info.meta.terminalId;
-					if (isTerminalAvailable(tid)) {
+					if (!$terminalServersLoaded || isTerminalAvailable(tid)) {
 						selectedTerminalId.set(tid);
 					}
 				}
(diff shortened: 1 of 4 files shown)


rasimme:
I reproduced the same initialization race deterministically against the v0.10.2 source ordering. Two useful regression schedules for this PR are:

1. **Preset availability race:** `setDefaults()` receives a valid `meta.terminalId` at 20 ms while the terminal inventory is still empty; the inventory resolves with that ID at 120 ms. The preset should end selected exactly once.
2. **Mount-time stale clear:** a persisted terminal ID is restored at 0 ms, stale cleanup runs at 10 ms while inventory is empty, and inventory resolves with that ID at 100 ms. The selection must not be cleared before loading is authoritative.

The `terminalServersLoaded` distinction and the guarded cleanup in this PR address both paths, including preset application in `setDefaults()`.

One edge case worth covering: `setToolServers()` currently sets `terminalServersLoaded = true` in `finally`. If terminal discovery throws or otherwise fails, an empty inventory may still become “authoritative” and trigger destructive stale cleanup. Ideally, failed/unknown discovery should preserve the pending selection; only a successful inventory response should authorize clearing an absent ID.

Also worth retaining as a regression check: if the user explicitly selects a different terminal while discovery is pending, a late preset retry must not overwrite that newer choice.

tjbck:
Addressed in dev.

PR #26802, opened 2026-07-05
fix: stop fabricating 1-hour expiry for non-refreshable OAuth tokens

### LINKED ISSUE #26141: bug: Hardcoded 1-hour token expiry destroys MCP OAuth sessions for providers that don't issue refresh tokens (Slack)
## Bug Description

OWUI forces a 1-hour expiration on OAuth tokens from providers that omit `expires_in` in their token response, then deletes the session when refresh fails because the provider never issued a `refresh_token`. The result is guaranteed session destruction every ~1 hour for Slack and similar providers.

## Provider: Slack (concrete example)

Slack's `oauth.v2.user.access` endpoint returns user tokens (`xoxp-*`) that **do not expire** and **do not include a `refresh_token`**. The token response intentionally omits `expires_in`:

```json
{
  "ok": true,
  "access_token": "xoxp-...",
  "token_type": "Bearer",
  "scope": "...",
  "user_id": "U12AB34CD56",
  "team": {"id": "T08XYZ99AA", "name": "ExampleOrg"}
}
```

No `expires_in`. No `refresh_token`. These tokens are long-lived until revoked.

### Step 1 — Token stored with fabricated 1-hour expiry

`_normalize_token_expiry()` in `backend/open_webui/utils/oauth.py` line 143–172:

```python
DEFAULT_TOKEN_EXPIRY_SECONDS=***

def _normalize_token_expiry(token):
    if token.get("expires_at") is not None: return token
    if token.get("expires_in") is not None:
        token["expires_at"] = now + token["expires_in"]; return token
    # Slack falls through to here:
    log.warning("OAuth token response missing both expires_in and expires_at; "
                f"defaulting to {DEFAULT_TOKEN_EXPIRY_SECONDS}s from now")
    token["expires_at"] = now + 3600   # ← 1 hour, unconditional
```

This fires on **every** Slack authentication. Logs from a production instance:

```
_normalize_token_expiry:168 - OAuth token response missing both "expires_in" and "expires_at"; defaulting to 3600s from now
```

### Step 2 — Refresh attempted at ~55 minutes

`get_oauth_token()` checks: `now + 5min >= expires_at` → triggers refresh.

### Step 3 — Refresh fails because no refresh_token exists

`_perform_token_refresh()` line 881:

```python
if not token_data.get("refresh_token"):
    log.warning(f"No refresh token available for session {session.id}")
    return None   # ← no graceful fallback
```

### Step 4 — Session deleted

`get_oauth_token()` line 829: any `None` return from refresh → session destroyed:

```python
if refreshed_token:
    return refreshed_token
else:
    log.warning("Token refresh failed... deleting session")
    await OAuthSessions.delete_session_by_id(session.id)
```

### Production logs showing the cycle

```
# Session created (forced 1-hour expiry):
handle_callback:1013 - Stored OAuth session server-side for user ..., client_id mcp:example-slack

# ~10 hours later, user tries to use the tool:
_perform_token_refresh:881 - No refresh token available for session a24995b4...
_refresh_token:860 - Failed to refresh token for session a24995b4...
get_oauth_token:829 - Token refresh failed... deleting session a24995b4...

# User re-authenticates immediately:
get_oauth_token:816 - No OAuth session found for user ..., client_id mcp:example-slack
```

This cycle has repeated **every time** the session is used after >1 hour of inactivity. Three full cycles confirmed in a 48-hour window.

## Root Cause

Two interacting code paths, neither wrong in isolation, but together guaranteed to fail:

1. **`_normalize_token_expiry`** assumes all tokens expire and defaults to 3600s when expiry metadata is absent. The comment says "Matches the value recommended by Authlib's compliance_fix documentation" — reasonable for expiring tokens, destructive for non-expiring tokens.

2. **`_perform_token_refresh`** returns `None` when no `refresh_token` is available, and the caller interprets `None` as fatal, deleting the session. There is no fallback to return the existing token unchanged.

## Affected Providers

Any OAuth provider whose token response omits `expires_in` and does not issue a `refresh_token`. Slack is confirmed. Likely others using long-lived user tokens without token rotation enabled.

## Proposed Fix

Either of these would resolve it (both are one-line changes):

**Option A — Use a long fallback in `_normalize_token_expiry`:** When neither `expires_in` nor `expires_at` is present, default to a very long expiry (e.g. 1 year) instead of 1 hour. Tokens without expiry metadata are long-lived by definition.

**Option B — Graceful fallback in `_perform_token_refresh`:** When no `refresh_token` is available, return the existing token data instead of `None`. Push `expires_at` forward by 24h to avoid warning spam:

```python
if not token_data.get("refresh_token"):
    log.info(f"No refresh token available, returning existing token for session {session.id}")
    token_data["expires_at"] = int(time.time() + 86400)
    return token_data
```

## Related Issues (all closed without resolution)

- [#19809](https://github.com/open-webui/open-webui/issues/19809) — same symptom, closed by maintainer as "working as intended" (assumes all providers issue refresh tokens)
- [#20291](https://github.com/open-webui/open-webui/issues/20291) — same consequence (session deleted), Atlassian flavor
- [#23669](https://github.com/open-webui/open-webui/issues/23669) — same root cause (`_normalize_token_expiry` fallback)
- [#24618](https://github.com/open-webui/open-webui/issues/24618) — MCP sessions wiped via different trigger (SSO middleware)
- [Discussion #23510](https://github.com/open-webui/open-webui/discussions/23510) — prior analysis of Slack MCP OAuth bugs (Bugs 1-4 fixed in v0.9.6, but the 3600s fallback was not addressed)

## Environment

- Open WebUI: v0.9.6 (ghcr.io/open-webui/open-webui:latest, Docker)
- MCP server: Slack (`mcp.slack.com`), OAuth 2.1 static
- Database: PostgreSQL (bundled)

## Why This Issue Is Different

Previous issues framed this as "tokens expire too fast" or "refresh fails." The actual problem is more fundamental: **OWUI fabricates an expiry for non-expiring tokens, then punishes the user when the fabricated refresh predictably fails.** This is not a provider issue — Slack tokens are working correctly. The 1-hour limit is entirely internal to OWUI.

Classic298:
Providers like Slack issue long-lived user tokens whose token response contains neither expires_in nor expires_at and no refresh_token. The _normalize_token_expiry fallback unconditionally stamped such tokens with a 3600s expiry, so roughly an hour later get_oauth_token triggered a refresh that could never succeed (there is no refresh token) and the failed refresh deleted the OAuth session, forcing the user to re-authenticate every hour.

Only apply the conservative 1-hour fallback when the token actually carries a refresh_token so the proactive refresh can rotate it. A token with no expiry metadata and no refresh_token is long-lived by definition, so stamp it with a one-year expiry instead (the sessions table requires a non-null expires_at). Tokens whose provider declares a real expiry but issues no refresh token are intentionally unaffected: those genuinely expire and re-authentication is the only recovery.

Fixes #26141

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -192,6 +192,11 @@ DEFAULT_TOKEN_EXPIRY_SECONDS = 3600
 # (e.g. Slack xoxp user tokens): long-lived until revoked, cannot be refreshed.
 NON_REFRESHABLE_TOKEN_EXPIRY_SECONDS = 365 * 24 * 3600
 
+# Marker key stamped on tokens that carry no expiry metadata and no
+# refresh_token. It flags the token as long-lived so the refresh path can
+# renew the local expiry in place instead of deleting the session (#26141).
+NON_REFRESHABLE_TOKEN_MARKER = 'non_refreshable'
+
 
 # Apereo CAS includes client_id in ID token JWS headers; Authlib 1.7/joserfc
 # rejects unknown headers unless we register the provider extension.
@@ -211,8 +216,9 @@ def _normalize_token_expiry(token: dict) -> dict:
        ``DEFAULT_TOKEN_EXPIRY_SECONDS`` and log a warning so operators can
        identify providers that omit expiration.
     4. Otherwise the token is non-expiring and non-refreshable; use
-       ``NON_REFRESHABLE_TOKEN_EXPIRY_SECONDS`` so the refresh path (which
-       would inevitably fail and delete the session) is not triggered.
+       ``NON_REFRESHABLE_TOKEN_EXPIRY_SECONDS`` and flag it with
+       ``NON_REFRESHABLE_TOKEN_MARKER`` so the refresh path renews the local
+       expiry in place instead of deleting the session.
 
     Also stamps *issued_at* for auditing.
     """
@@ -237,11 +243,33 @@ def _normalize_token_expiry(token: dict) -> dict:
         token['expires_at'] = int(datetime.now().timestamp() + DEFAULT_TOKEN_EXPIRY_SECONDS)
     else:
         log.info(
-            "OAuth token response missing 'expires_in', 'expires_at' and 'refresh_token'; "
-            'treating token as long-lived'
+            "OAuth token response missing 'expires_in', 'expires_at' and 'refresh_token'; treating token as long-lived"
         )
         token['expires_at'] = int(datetime.now().timestamp() + NON_REFRESHABLE_TOKEN_EXPIRY_SECONDS)
+        token[NON_REFRESHABLE_TOKEN_MARKER] = True
     return token
+
+
+def _renew_non_refreshable_token(token_data: dict, session_id: str):
+    """Renew the local expiry of a long-lived, non-refreshable token.
+
+    Providers such as Slack issue user tokens that never expire and never
+    carry a ``refresh_token`` (see ``_normalize_token_expiry`` case 4). When
+    such a token approaches its fabricated expiry it reaches the refresh path,
+    which can only fail — deleting the session and forcing a needless
+    re-authentication (#26141). Instead of failing, re-stamp the local expiry
+    in place so the session keeps renewing itself.
+
+    Returns the (mutated) token dict when it was flagged long-lived, else
+    ``None`` so the caller falls through to its normal no-refresh-token
+    handling (tokens with a real provider-declared expiry are left to expire).
+    """
+    if not token_data.get(NON_REFRESHABLE_TOKEN_MARKER):
+        return None
+
+    token_data['expires_at'] = int(datetime.now().timestamp() + NON_REFRESHABLE_TOKEN_EXPIRY_SECONDS)
+    log.debug(f'Renewing long-lived non-refreshable token for session {session_id}')
+    return token_data
 
 
 FERNET = None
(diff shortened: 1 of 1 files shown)


adaofeliz:
Nice fix, this solves the hourly session-death issue for Slack-style tokens.

One thing I noticed though: `_perform_token_refresh` still returns `None` when there's no `refresh_token` (untouched by this PR), so `get_oauth_token` will delete the session anyway once the fabricated year-long expiry runs out. Since these tokens don't actually expire, we'd just be hitting the same bug again in 365 days instead of every hour.

Might be worth handling this in `_perform_token_refresh` too, something like:

```python
if not token_data.get('refresh_token'):
    token_data['expires_at'] = int(time.time() + NON_REFRESHABLE_TOKEN_EXPIRY_SECONDS)
    return token_data
```

That way a non-refreshable token just keeps renewing itself instead of hitting a wall eventually.

Classic298:
Shouldn't that be handled by the operator to issue refreshable tokens?

adaofeliz:
Not really, this isn't something the operator can configure. Slack's `oauth.v2.user.access` tokens (used for MCP) just don't come with a `refresh_token`, that's how the flow works by design, not a misconfiguration on their end.

From Slack's own docs: "OAuth tokens do not expire. If they are no longer needed, they can be revoked." https://docs.slack.dev/authentication/installing-with-oauth/#revoking

So these tokens are meant to be long-lived until revoked. We can't expect a refresh_token to eventually show up, we just need to handle that case.

Classic298:
Will look into it tmrw

adaofeliz:
> Will look into it tmrw

Thanks! Much appreciated.

adaofeliz:
This looks good, thanks for addressing it. The `_renew_non_refreshable_token` addition closes the gap nicely, non-refreshable tokens now self-renew instead of dying on a delay. Appreciate the quick turnaround on this.

PR #26857, opened 2026-07-07
fix: preserve system prompt across tool calls when memories are enabled

Classic298:
The native tool-call loop runs generate_chat_completion with bypass_system_prompt=True, so the provider layer does not re-apply the model's default system prompt on tool-call iterations. It relies instead on metadata['system_prompt'], captured in process_chat_payload, to carry the full system prompt forward and restore it after RAG injection.

That capture read the model default system prompt from form_data['params']['system'], but apply_params_to_form_data had already popped 'params' from form_data, so model_system_prompt was always empty. metadata['system_prompt'] therefore only captured whatever was already materialized in the messages. With memories enabled, that is the injected <memory_context> system message, so tool-call requests were restored with memory-only system content and the model's system prompt was dropped. Without memories there was no system message to capture at all.

Capture the model default system prompt from form_data['params'] before apply_params_to_form_data pops it, and use that value when building metadata['system_prompt'].

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -2198,6 +2198,12 @@ async def process_chat_payload(request, form_data, user, metadata, model):
             form_data['model'] = selected_model_id
             metadata['selected_model_id'] = selected_model_id
 
+    # Capture the model's default system prompt before apply_params_to_form_data
+    # pops 'params'. The provider layer applies it fresh from model_info.params,
+    # but the native tool-call loop runs with bypass_system_prompt=True and relies
+    # on metadata['system_prompt'] (built below) to carry it forward instead.
+    model_system_prompt = (form_data.get('params') or {}).get('system')
+
     form_data = apply_params_to_form_data(form_data, model)
     log.debug(f'form_data: {form_data}')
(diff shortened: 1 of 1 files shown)


Welteam:
Hi Classic298, thank you for this fix as this is an issue I've been facing. Enabling native function calling "broke" my workspace models 😢
Do you know when the decision to merge this code will be made?

Classic298:
@Welteam this PR has nothing to do with function calling

Welteam:
> @Welteam this PR has nothing to do with function calling

The "Function Calling" option in the models' "Advanced Params" is what enables built-in tools when set to "Native" (the current default value). I believe this PR fix an issue where the model's system prompt is lost when the model calls tools, including the built-in tools like "query_knowledge_files". Am I wrong?

Classic298:
only if memory is enabled it does yes

Welteam:
That's an important precision I did not realize. I may disable memories as a work around. Thanks
To return to my initial inquiry, do you know when the decision to merge this or not might happen?

Classic298:
no i dont. i hope within the next days

PR #26711, opened 2026-07-02
feat: expose chat-attached knowledge to builtin tools when File Context is off

Classic298:
Under native function calling, model- and folder-attached knowledge is reachable through the builtin knowledge tools but announced nowhere in the prompt. The model therefore has no reason to ever call `query_knowledge_files`, and attached knowledge goes unused. This is the root cause of #27232 and of the older class of reports like #22181.

This PR makes attached knowledge discoverable and keeps tool routing unambiguous:

- Every item reachable by the builtin knowledge tools is announced in the system prompt inside an `<attached_knowledge>` block: `<knowledge type="collection|file|note" id=".." name=".." source="model|folder|chat"/>`. The block maps 1:1 to the knowledge tools (`query_knowledge_files` and friends, docstrings updated to say so), while the per-message `<attached_files>` block keeps meaning chat-attached files.
- The list is built by a single shared helper (`get_effective_model_knowledge`) that also feeds `__model_knowledge__` in `get_builtin_tools`, so the knowledge the model is told about and the knowledge the tools can actually reach cannot drift apart.
- Model and folder scope are announced unconditionally under native FC, because they are never RAG-injected there regardless of the File Context capability. Chat-attached collections and notes remain gated on File Context being off and move from `<attached_files>` into `<attached_knowledge>`.
- System-prompt placement follows the existing pattern for chat-level context (skill manifests, terminal prompts) and keeps the prompt prefix cache-stable across turns: the block only changes when the attached knowledge itself changes, and user messages are never rewritten.

Security: announced ids grant nothing by themselves; every id is access-checked server-side when a tool reads it. Chat-attached file ids still never enter `__model_knowledge__`, since membership there bypasses the per-file access check; on-demand access to chat-attached files is handled separately by #26902.

Fixes #27232
Fixes #26708

### Contributor License Agreement

- [x] By submitting this pull request, I confirm that I have read and fully agree to the [Contributor License Agreement (CLA)](https://github.com/open-webui/open-webui/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT), and I am providing my contributions under its terms.

> [!NOTE]
> Deleting the CLA section will lead to immediate closure of your PR and it will not be merged in.

Issue #26708, opened 2026-07-02
feat: Expose Chat-Attached Knowledge to Builtin Tools When File Context Is Off


hytg:
### Check Existing Issues

- [x] I have searched all existing **open AND closed** issues and discussions and found none comparable to my request.

### Verify Feature Scope

- [x] I believe this feature request is appropriately scoped for the Issues section as described above.

### Problem Description

Open WebUI separates **File Context** and **Builtin Tools**.

* File Context controls whether attached files/knowledge are automatically retrieved and injected into the prompt.
* Builtin Tools allow the model to retrieve files/knowledge on demand.

This separation should be preserved.

However, when a user attaches knowledge or files through the chat input, for example via `+ -> Attach Knowledge`, and uses:

```text id="ek0ppe"
File Context: disabled
Builtin Tools: enabled
```

the chat-attached resource may not be discoverable through the same built-in tool flow as knowledge/files attached to the model itself.

Disabling File Context should prevent automatic content injection, but it should not make explicitly attached chat resources invisible to tools.

### Proposed Solution

Treat chat-attached knowledge/files as additional entries in the same request-scoped attached knowledge list already used by builtin tools for model-attached knowledge.

More specifically, when File Context is disabled and Builtin Tools are enabled, chat-attached resources should be appended to the same `model_knowledge` / `__model_knowledge__` structure that builtin knowledge/file tools already receive.

The goal is not to introduce a new authorization model, stricter fallback behavior, a separate chat-attachment prompt, or a separate tool flow. The goal is only to let chat-attached resources participate in the existing builtin tool discovery path in the same way as model-attached knowledge.

A minimal normalized shape could be:

```json
{
  "type": "file",
  "id": "file_456",
  "name": "api_design.md"
}
```

or:

```json
{
  "type": "collection",
  "id": "kb_123",
  "name": "Product Docs"
}
```

The model should not be forced to use these resources. Retrieval should remain agentic. File Context disabled should still mean that no file content, chunks, summaries, or RAG context are automatically injected into the prompt.

Although the current internal parameter is named `__model_knowledge__`, its role is closer to a request-scoped attached-resource / retrieval-scope list, since it already includes model-attached and folder-attached knowledge. This request does not require renaming it; it only asks that validated chat-attached resources be appended to the same existing structure for compatibility.

### Alternatives Considered

One alternative is to inject a separate prompt section describing chat-attached resources.

[...]

PR #26902, opened 2026-07-08
feat: add query_chat_file builtin tool for chat-attached files

Classic298:
Adds a query_chat_file builtin tool that lets the model semantically query a file attached to the current chat when File Context is off. It mirrors query_knowledge_files: same access checks (owner, admin, model attachment, access grants), same query_collection retrieval path over the file's file-{id} collection, and the same chunk result shape. The result count defaults to and is capped at the admin-configured RAG top k. The tool is only injected when the model's file_context capability is disabled, since automatic retrieval covers attachments when it is on. Access is checked per file id server-side rather than by chat membership.

Also fixes _has_read_access_to_file raising a pydantic ValidationError when consulting access grants: it constructed a validated UserModel from only id and role, so the grant fallback never ran for non-owner, non-admin users. Use model_construct instead.

https://github.com/open-webui/open-webui/issues/26708

Ref: https://github.com/open-webui/open-webui/pull/26711

> [!NOTE]


tjbck:
Closing in favour of 57e60423b9963c4a69fdfda7ae5799efc5583010

PR #26905, opened 2026-07-08
fix: deny chained access to unregistered base models for non-admins

Classic298:
A workspace model shared publicly could be used by any user even when its base model was private. Unregistered base models (no row in the model table) are admin-only for direct use — get_filtered_models hides them from non-admins and check_model_access rejects them — but has_base_model_access treated a missing row as "no ACL" and allowed the chained request through.

has_base_model_access now takes the caller's role and only allows an unregistered base model hop for admins, so a shared preset can no longer reach a base model the caller could not use directly. Registered base models keep their existing grant-based enforcement.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/access_control/__init__.py
+++ b/backend/open_webui/utils/access_control/__init__.py
@@ -260,6 +260,7 @@ async def has_base_model_access(
     user_id: str,
     model_info,
     *,
+    user_role: str | None = None,
     user_group_ids: set[str] | None = None,
     db=None,
 ) -> bool:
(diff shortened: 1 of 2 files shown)


ErenAta16 (commented):
Read `has_base_model_access` on main: when `base_model_info is None` (no DB row for that base model id), it just `break`s out of the walk loop, which — since nothing after the loop rejects on that condition — falls through to an implicit allow. That is exactly the "no ACL means allowed" bug described: unregistered base models are supposed to be admin-only for direct use (confirmed `get_filtered_models`/`check_model_access` already enforce that), but a shared preset could hop to one anyway since this function never asked who the caller was. Passing the caller's role through and gating the missing-row branch on it closes the gap without touching how registered base models are handled. LGTM.

Issue #26922, opened 2026-07-09
bug: SecurityHeadersMiddleware is still BaseHTTPMiddleware — a migration straggler that intermittently corrupts streaming responses ("Response content longer than Content-Length")


safcbee:
## Check Existing Issues

- [x] I have searched the existing issues and discussions (including closed ones) for `SecurityHeadersMiddleware`, `BaseHTTPMiddleware`, and "Response content longer than Content-Length" — no existing report covers this.

## Installation Method

Docker Compose (`ghcr.io/open-webui/open-webui:main`), single instance, direct port publish (no reverse proxy in front for this reproduction), SQLite backend (`DATABASE_URL` unset).

## Open WebUI Version

v0.9.6 (image digest `sha256:7f1b0a1a50cfbac23da3b16f96bc968fd757b26dc9e54e93813d61768ea9184e`, built 2026-06-02). Confirmed via the commit history that the affected file is unchanged on `main` as of 2026-07-08 — this is not something already fixed and merely un-pulled by us.

## Ollama Version (if applicable)

Not applicable — the failing path is Open WebUI's own response forwarding on the TTS proxy route (`ENABLE_OLLAMA_API=false` in this deployment; the upstream here is an OpenAI-compatible TTS backend).

## Operating System

Docker host: Linux (Arch). Clients: macOS / iOS Safari — the failure is server-side and browser-independent.

## Browser (if applicable)

Safari (macOS + iOS) — but see above; the exception is raised inside uvicorn before the browser is a factor.

## Confirmation

- [x] I have read the docs and searched existing issues.
- [x] I have confirmed this is not a configuration issue on our end: no admin-set audio overrides are persisted in Open WebUI's config table (`{}` for the `audio` key), so the live config is exactly the compose-file env — no drift, nothing to reset.
- [x] I have confirmed the image digest was unchanged for over a month before this was noticed — ruled out a version bump as the trigger.

## Expected Behavior

`/api/v1/audio/speech` (and any other streaming/proxied response) should complete cleanly regardless of `SecurityHeadersMiddleware` being in the stack, the same way it does for the four middlewares already converted to pure ASGI in `open_webui/utils/asgi_middleware.py`.

## Actual Behavior

[...]

PR #26924, opened 2026-07-09
fix: convert SecurityHeadersMiddleware to pure ASGI

Classic298:
SecurityHeadersMiddleware was the last middleware in the stack still subclassing BaseHTTPMiddleware, after CommitSession, AuthToken, WebsocketUpgradeGuard and Redirect were all moved to pure ASGI in utils/asgi_middleware.py. BaseHTTPMiddleware re-buffers the response body through an anyio task group, which has known issues with streaming and Content-Length-bearing responses (e.g. the FileResponse returned by /api/v1/audio/speech).

Reimplement it as a pure-ASGI middleware that stamps the configured security headers onto the http.response.start message via MutableHeaders and forwards all body chunks untouched, matching the pattern already used by its four siblings. set_security_headers() and all its helpers are unchanged.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/security_headers.py
+++ b/backend/open_webui/utils/security_headers.py
@@ -2,15 +2,33 @@ import os
 import re
 from typing import Dict
 
-from fastapi import Request
-from starlette.middleware.base import BaseHTTPMiddleware
+from starlette.datastructures import MutableHeaders
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
 
-class SecurityHeadersMiddleware(BaseHTTPMiddleware):
-    async def dispatch(self, request: Request, call_next):
-        response = await call_next(request)
-        response.headers.update(set_security_headers())
-        return response
+class SecurityHeadersMiddleware:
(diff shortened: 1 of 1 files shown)


safcbee:
Verified this fix in the production deployment where #26922 was originally traced (Docker Compose, v0.9.6 image, self-hosted Kokoro TTS backend): mounted this PR's exact `security_headers.py` (head `3305bc1`) over the stock file and re-ran the same soak that characterised the bug — 120 concurrent real `POST /api/v1/audio/speech` requests, concurrency 8, distinct `input` text per request to bypass the request-body cache. Result: **120/120 completed cleanly — zero `Response content longer than Content-Length` errors, zero mid-body aborts** (vs. intermittent failures with the stock `BaseHTTPMiddleware` version). Implementation is equivalent to the workaround we've been running in production since filing the issue (`set_security_headers()` and helpers untouched, only the wrapping mechanism changed). Confirms the fix from our side — thanks for the quick turnaround.

Issue #26880, opened 2026-07-07
issue: Native Function Calling knowledge-base tools fail with `'str' object has no attribute 'items'`


alaskaav8r:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Other

### Open WebUI Version

v0.10.2

### Operating System

Mac Tahoe

### Browser (if applicable)

Chrome 150.0.7871.101

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

The model should call a knowledge-base retrieval tool (query_knowledge_files or kb_exec), successfully retrieve the relevant passage from the attached PDF, and answer the question correctly with a citation to the source document — the same result Legacy mode produces reliably with the identical model and Knowledge Base.

"in my example What is the oil change interval in my gmc truck?"  It should have returned, "When the CHANGE ENGINE OIL SOON message displays, the engine oil and filter should be changed within the next 1,000 km/600 mi. Additionally, the oil and filter must be changed at least once a year.  with a citation to the document where it found the data.

<img width="1578" height="936" alt="Image" src="https://github.com/user-attachments/assets/a4751d6a-7c9e-4408-9ae4-59afb45431fe" />

### Actual Behavior

[...]

PR #26926, opened 2026-07-09
fix: guard note.data None crash in builtin tool query_knowledge_files

Diwak4r:
### Description

`note.data` is defined as `Optional[dict]` on the `Notes` model (nullable), but the `query_knowledge_files` tool called `note.data.get(...)` without a None guard. When a note has no `data` field populated, this raises `AttributeError: 'NoneType' object has no attribute 'get'`, aborting the whole knowledge retrieval.

Two sibling call sites (builtin.py lines 978 and 1062) already used the correct `note.data and note.data.get(...)` guard. This patch applies the equivalent `(note.data or {}).get(...)` to the `query_knowledge_files` path, matching the established pattern.

### Fixed

- Crash in `query_knowledge_answer_collection` when a referenced note has `data=None`.

### Testing

Verified the repro directly: with `note.data = None`, the old expression raised `AttributeError: 'NoneType' object has no attribute 'get'`, while the patched expression returns `''` (and still returns populated markdown when data is present). Also confirmed the file compiles with `python -m py_compile`.

Relates to: #26880

### DIFF
--- a/backend/open_webui/tools/builtin.py
+++ b/backend/open_webui/tools/builtin.py
@@ -2597,7 +2597,7 @@ async def query_knowledge_files(
                             permission='read',
                         )
                     ):
-                        content = note.data.get('content', {}).get('md', '')
+                        content = (note.data or {}).get('content', {}).get('md', '')
                         note_results.append(
                             {
                                 'content': content,


ErenAta16 (commented):
Confirmed the exact unguarded line on main (`content = note.data.get("content", {}).get("md", "")`, no `note.data and` prefix) sitting right alongside two sibling call sites that already use the guarded form — so this really is just the one spot that was missed when the guard pattern was established elsewhere, not a new pattern. Minimal, consistent fix. LGTM.

tjbck:
Addressed in dev!

Diwak4r:
Hi @tjbck — thanks for the quick reply! Just wanted to confirm which commit that was, since `compare/dev...69d7ce3d82fc6f87ae32ff313568ddf82312a5cc` still shows the head commit as not on dev (`ahead_by: 1`). If it landed under a different commit, no action needed from me; otherwise happy to rebase and reopen. Thanks!

PR #26928, opened 2026-07-09
fix: don't seed non-persistent config keys (oauth.* with flag off)

Classic298:
seed_defaults inserted a row for every key in DEFAULT_CONFIG regardless of whether the DB is authoritative for it. With ENABLE_OAUTH_PERSISTENT_CONFIG off, the oauth.* keys were seeded from the then-current (often empty) env values. Enabling the flag later made those stale rows override live env vars (e.g. ENABLE_OAUTH_SIGNUP=true stopped taking effect) and further env changes were never picked up.

Skip keys where persistent_enabled_for() is false, matching the masking the read paths (get/get_many/get_namespace/get_all) already apply.

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/config.py
+++ b/backend/open_webui/models/config.py
@@ -247,6 +247,10 @@ class Config(Base):
             now = int(time.time())
             new_count = 0
             for key, value in defaults.items():
+                # Skip keys the DB is not authoritative for (e.g. oauth.* while
+                # ENABLE_OAUTH_PERSISTENT_CONFIG is off), matching the read paths.
+                if not Config.persistent_enabled_for(key):
+                    continue
                 if key not in existing_keys:
                     value = _json_value(value)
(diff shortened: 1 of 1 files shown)


ErenAta16 (commented):
Checked `seed_defaults` on main — it inserts a row for every key in `defaults` that is not already in the DB, with no `persistent_enabled_for` check at all, while every read path (`get`/`get_many`/`get_namespace`/`get_all`) already gates on that flag. So with `ENABLE_OAUTH_PERSISTENT_CONFIG` off, oauth.* keys still get a DB row seeded from whatever the env happened to be at first boot, and if that flag is flipped on later those stale rows silently start overriding live env vars. Skipping non-persistent keys during seeding is the right fix since it makes seeding consistent with what the read paths already treat as authoritative. LGTM.

PR #26932, opened 2026-07-09
feat: implement knowledge file status tracking and processing recovery

jannefleischer:
# Pull Request Checklist

### Note to first-time contributors: Please open a discussion post in [Discussions](https://github.com/open-webui/open-webui/discussions) to discuss your idea/fix with the community before creating a pull request, and describe your changes before submitting a pull request.

**Before submitting, make sure you've checked the following:**

- [x] **Linked Issue/Discussion:** This PR references an existing [Issue](https://github.com/open-webui/open-webui/issues) or [Discussion](https://github.com/open-webui/open-webui/discussions) — `Closes #[ISSUE_NUMBER]` / Relates to [Discussion #26931](https://github.com/open-webui/open-webui/discussions/26931).
- [x] **Target branch:** The pull request targets the `dev` branch. **PRs targeting `main` will be immediately closed.**
- [x] **Description:** A concise description of the changes is provided below.
- [ ] **Changelog:** A changelog entry following [Keep a Changelog](https://keepachangelog.com/) format is included at the bottom.
- [ ] **Documentation:** Relevant documentation has been added or updated in the [Open WebUI Docs Repository](https://github.com/open-webui/docs).
- [ ] **Dependencies:** Any new or updated dependencies are explained, tested, and documented.
- [x] **Testing:** Manual tests have been performed to verify the fix/feature works correctly and does not introduce regressions. Tests added here: [https://github.com/open-webui/tests/pull/7](https://github.com/open-webui/tests/pull/7)
- [x] **No Unchecked AI Code:** This PR is either human-written or has undergone thorough human review AND manual testing.
- [x] **Self-Review:** A self-review of the code has been performed, ensuring adherence to project coding standards.
- [ ] **Architecture:** Smart defaults are preferred over new settings. Local state is used for ephemeral UI logic. Major architectural or UX changes have been discussed first.
- [x] **Git Hygiene:** The PR is atomic (one logical change), rebased on `dev`, and contains no unrelated commits.
- [x] **Title Prefix:** The PR title uses one of the following prefixes:
  - **feat**: New features or enhancements

# Changelog Entry

### Description

Implements comprehensive file processing reliability and document extraction improvements for knowledge bases, including startup recovery from interrupted tasks and structured Docling JSON extraction for better search quality.

**Features:**
- Startup cleanup for orphaned processing tasks (`reset_stuck_processing_files`) — marks pending/processing files as failed when the server restarts, eliminating orphaned state after crashes/restarts
[...]

PR #26944, opened 2026-07-09
feat: track knowledge file processing status and recover in-flight up…

jannefleischer:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** This PR references an existing [Issue](https://github.com/open-webui/open-webui/issues) or [Discussion](https://github.com/open-webui/open-webui/discussions) — `Closes #___` / [Relates to #26931](https://github.com/open-webui/open-webui/discussions/26931). If one does not exist, create one first. PRs without a linked issue or discussion may be closed without review.
- [ ] **Dependencies:** Any new or updated dependencies are explained, tested, and documented.
- [ ] **Testing:** Manual tests have been performed to verify the fix/feature works correctly and does not introduce regressions. Screenshots or recordings are included where applicable.
- [ ] **No Unchecked AI Code:** This PR is either human-written or has undergone thorough human review AND manual testing. Unreviewed AI-generated PRs may be closed immediately.
- [ ] **Self-Review:** A self-review of the code has been performed, ensuring adherence to project coding standards.
- [ ] **Git Hygiene:** The PR is atomic (one logical change), rebased on `dev`, and contains no unrelated commits.

> **Why four boxes are unticked.** The delete-during-processing support in this revision is new and so far verified on the backend only (full application import, cancellation checks); it has not been exercised in the browser yet, and the recording below predates it. Testing, self-review and the human-review attestation stay unticked until that is done. Git Hygiene is unticked because this revision adds a second logical change on top of the status tracking — see the note under **Description**.

# Changelog Entry

### Description

Knowledge base file processing (embedding) can take anywhere from seconds to minutes, and today the UI has no way to reflect that: a file's state is invisible once upload starts, ~~a page refresh loses track of anything still in flight,~~ and a server restart mid-embedding leaves the file stuck in `pending`/`processing` forever with no recovery path. ~~If the browser tab is closed while a background task is still running, the file can also silently fail to get linked to its knowledge base.~~ For as long as a file is in that state it also cannot be deleted, which is precisely when a user is most likely to want to.

*(Struck-through problems above are now solved in `dev` itself — see the revision note below.)*

This is a split-out of a previous, too-large PR (#26932) per [maintainer feedback](https://github.com/open-webui/open-webui/pull/26932#issuecomment-4925036925) asking for smaller, atomic, independently reviewable changes. This PR contains **only** the status-tracking/recovery piece ~~— no extraction-engine changes, no RAG pipeline changes, and (per the same feedback) no translation file edits.~~ **and the delete support that piece turns out to require** — no extraction-engine changes, no RAG pipeline changes, and (per the same feedback) no translation file edits. A separate PR/discussion will follow for the `docling_json` extraction provider and related RAG pipeline improvements described in the [linked discussion](https://github.com/open-webui/open-webui/discussions/26931).

The delete support is included here rather than split off again because it is not separable in practice: this PR is what puts in-flight files on screen, and a row that cannot be acted on is arguably worse than no row. The PR also already carried the server-side half of it (discarding embeddings for a file deleted mid-processing) without the half that makes it reachable. Happy to split it out if you would still prefer that.

> **Revision note (this update).** After rebasing onto 0.11 `dev`, several parts of this PR turned out to duplicate functionality `dev` had grown in the meantime. Because the two implementations never touched the same lines, the rebase kept both — including a second `GET /{id}/files/pending` route that FastAPI never reached. Everything `dev` already provides has been dropped in favour of upstream's version; the diff against `dev` shrank from 456 to 266 lines with no loss of functionality. Revised claims below are struck through with the reason given.
>
> The same update also completes one item that this PR previously only half-addressed: it already discarded embeddings for a file deleted mid-processing, but the delete itself was impossible to trigger for exactly those files. That is now fixed — see *Deleting a file while it is being processed* under **Added**.

### Added

- ~~`GET /api/v1/knowledge/{id}/files/pending` endpoint returning files still in `pending`/`processing` status for a knowledge base, so the UI can recover in-flight uploads after a page refresh (with the same owner/admin/access-grant checks as the rest of the knowledge API).~~ **Dropped:** `dev` ships this endpoint itself (`get_pending_knowledge_files`), with an additional SSE streaming mode and a query that also excludes files already linked. This PR now uses upstream's version; the duplicate route it had added was unreachable anyway, since FastAPI matches the first registration for a path.
- Startup cleanup pass (`Files.reset_stuck_processing_files`) that marks any file still `pending`/`processing` as `failed` on server boot, so a crash or restart no longer leaves files stuck in limbo indefinitely.
- Idempotency guard in `add_file_to_knowledge`: if a background task already linked the file (e.g. the browser stayed open long enough after all), the endpoint returns the current state instead of re-embedding.
- ~~"Drop-and-forget" auto-link: if a file is uploaded directly into a knowledge base, the background processing task links it to the KB itself, so the file still lands correctly even if the tab is closed before the SSE stream completes.~~ **Dropped:** `dev` already links uploaded files to their knowledge base from within the background task, gated by an explicit write-access check. This PR now uses upstream's version.
- `status: processing` together with a `started_at` timestamp is written when content extraction begins. `dev` only marks a file `processing` once it reaches the knowledge-base embedding step, so the long extraction phase was indistinguishable from "queued"; the UI can now tell the two apart and show how long a file has been running.
- Knowledge base file list shows a status tooltip (queued / processing / failed + error message) with a spinner/badge per file, and keeps local, not-yet-confirmed upload placeholders visible across refreshes. ~~now polls while any file is pending/processing~~ — **revised:** the polling is upstream's existing `pendingPollTimer`, not a second interval alongside it (see **Changed**).
- ~~Guard against orphaned vector-DB entries: if a file is deleted by the user while it's still being embedded, the (now pointless) embeddings are discarded instead of being committed.~~ **Superseded** by full delete support, below — the guard alone was unreachable in practice, because deleting such a file was rejected before processing could ever observe it.
- **Deleting a file while it is being processed.** Extraction and embedding can run for minutes or hours, and the delete button did nothing for a file in that state: it is not linked to the knowledge base yet, so `POST /knowledge/{id}/file/remove` rejected it with `400 NOT_FOUND` — even though the row is on screen via the pending endpoint. Deleting now works in every state, stops the running job instead of letting it finish into a deleted file, and removes everything belonging to it: knowledge-base link, vectors in both the shared and the per-file collection, database record and the stored blob. In the UI the action reads "Stop & Delete" while a file is processing; it is greyed out only in the brief window before the upload request has returned an id, since there is nothing to address server-side yet.
- `open_webui/utils/file_processing.py`: cooperative cancellation for long-running file processing. An in-process registry of `asyncio.Event`s interrupts a pipeline served by the same worker within seconds, and a database fallback (record gone, or `status: cancelled`) covers a delete served by a different worker process. `process_file` checks it before the storage read, after extraction and before embedding, and waits for the extraction step through a cancellable wrapper so a delete no longer has to wait for the loader. Embedding is deliberately left uncancellable: abandoning that wait would not stop the worker thread, and its vector write could then land after the cleanup — it completes instead, and the existing guard discards the result.

### Changed

- `add_file_to_knowledge` performs the idempotency check described above before doing any (potentially expensive) re-embedding work.
- Upstream's `pendingPollTimer` now refreshes the file list in place instead of only asking "is anything still pending?" and calling `init()` once the last file has landed. Intermediate status changes (queued → processing → failed) become visible while they happen, rather than only the end result, and the timer retires itself when nothing is in flight. Rows for uploads this tab started are preserved across such a refresh, since the server cannot report them until the upload request returns.
- `POST /knowledge/{id}/file/remove` accepts a file that is not yet linked to the knowledge base, provided it is in flight for that knowledge base. See **Breaking Changes** for the exact condition.

### Fixed

- Files stuck in `pending`/`processing` after a server crash/restart are no longer invisible/unrecoverable — they're marked `failed` with an explanatory message on next boot.
- The delete button is no longer inert for a file that is still being processed — the case where a user is most likely to want it, because that is when they notice a file was uploaded by mistake or is taking far too long.
- ~~Uploads in progress are no longer "lost" from the UI on page refresh.~~ **Dropped:** fixed in `dev` by the endpoint this PR no longer adds.
- ~~Files uploaded and processed while the browser tab is closed are no longer silently missing from the knowledge base.~~ **Dropped:** fixed in `dev` by the auto-link this PR no longer adds.

### Deprecated

- N/A

### Removed

- N/A

### Security

- N/A

### Breaking Changes

- ~~None. All changes are additive (new endpoint, new file-status values that default gracefully, new UI states); existing behavior for already-completed files is unchanged.~~ **Revised:** still none for already-completed files, and there is no new endpoint any more, but three behaviour changes are worth naming explicitly rather than filing under "additive":
  - `POST /knowledge/{id}/file/remove` no longer requires the file to be linked to the knowledge base. It also accepts a file whose `meta.data.knowledge_id` points at this knowledge base and whose status is `pending`, `processing` or `failed`; anything else is still rejected with `400 NOT_FOUND` as before. Write access to the knowledge base is checked exactly as before.
  - When that endpoint deletes the file record (owner or admin, `delete_file=true`), it now also deletes the stored blob. Previously the row disappeared and the file stayed in storage. A caller who may unlink but not delete the record now has the file detached from the knowledge base instead, so it stops appearing there as an in-flight upload.
  - The frontend `uploadFile()` helper gained two optional callbacks (`onProgress`, `onUploaded`) *before* its existing trailing `stream` parameter. No caller in the tree passes `stream` positionally, so nothing in-tree breaks, but out-of-tree callers that do would need adjusting.

---

### Additional Information

- Relates to discussion #26931.
- Supersedes #26932, which was closed for being too large for a first review pass. This PR is the first of the planned split — see the discussion for the full picture, including the still-to-come `docling_json` extraction provider and RAG pipeline improvement PRs.
- No dependency changes.
- No translation/i18n files are touched in this PR, per review feedback on #26932. The three new UI strings are in place as English `$i18n.t()` source strings only.
- Rebased onto current `dev`. The branch now carries three commits: the original feature, a `refactor:` commit removing everything `dev` had grown in the meantime, and the delete-during-processing support. Squashing the first two on merge is fine if preferred — they are only kept apart so the deduplication is reviewable on its own.
- Known limitation of the cancellation: a Docling conversion that has already been handed to the server keeps running there. The loader posts synchronously from a worker thread, which cannot be interrupted, and without a task id there is nothing to cancel remotely. The pipeline stops waiting and discards the result, so the user sees an immediate delete; the remote job finishes unobserved. Cancelling it properly needs Docling's async task API and is out of scope here. See: https://github.com/docling-project/docling-serve/issues/401

### Screenshots or Videos

https://github.com/user-attachments/assets/b92e0d63-1cde-4698-a5fd-57637320460b

> The recording above predates this revision and shows the status tracking only, not the delete-while-processing behaviour.

> [!NOTE]

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ FROM --platform=$BUILDPLATFORM node:22-alpine3.20 AS build
 ARG BUILD_HASH
 
 # Set Node.js options (heap limit Allocation failed - JavaScript heap out of memory)
-# ENV NODE_OPTIONS="--max-old-space-size=4096"
+ENV NODE_OPTIONS="--max-old-space-size=6144"
 
 WORKDIR /app
 
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -132,6 +132,7 @@ from open_webui.models.access_grants import AccessGrants
 from open_webui.models.channels import Channels
 from open_webui.models.chats import ChatForm, Chats
 from open_webui.models.config import Config
+from open_webui.models.files import Files
 from open_webui.models.functions import Functions
 from open_webui.models.messages import Messages
 from open_webui.models.models import Models
@@ -342,6 +343,11 @@ async def lifespan(app: FastAPI):
     await seed_registered_defaults()
     await initialize_runtime_config(app)
     await migrate_legacy_webhook_config()
+
+    n_reset = await Files.reset_stuck_processing_files()
+    if n_reset:
+        log.info('Reset %d stuck file(s) to failed state', n_reset)
+
     await publish_event(app, EVENTS.SYSTEM_STARTUP_STARTED, source='system')
 
     license_task = None
--- a/backend/open_webui/models/files.py
+++ b/backend/open_webui/models/files.py
@@ -251,6 +251,31 @@ class FilesTable:
             result = await db.execute(select(File).filter_by(user_id=user_id))
             return [FileModel.model_validate(file) for file in result.scalars().all()]
 
+    async def get_pending_files_by_knowledge_id(
+        self,
+        knowledge_id: str,
+        user_id: str | None = None,
+        db: AsyncSession | None = None,
+    ) -> list['FileModelResponse']:
+        """Return files that reference knowledge_id in their meta.data and are
+        still being processed (status pending or processing).  Used to surface
+        in-flight uploads after a page refresh, before the background task has
+        finished and linked the file to the knowledge base."""
+        async with get_async_db_context(db) as db:
+            stmt = select(File).filter(
+                File.data['status'].as_string().in_(['pending', 'processing']),
+                File.meta['data']['knowledge_id'].as_string() == knowledge_id,
+            )
+            if user_id:
+                stmt = stmt.filter_by(user_id=user_id)
+
+            result = await db.execute(stmt.order_by(File.created_at.desc()))
+            files = result.scalars().all()
+
+            return [
+                FileModelResponse.model_validate(f, from_attributes=True) for f in files
+            ]
+
     async def get_file_list(
         self,
         user_id: str | None = None,
@@ -366,6 +391,34 @@ class FilesTable:
                 return FileModel.model_validate(file)
             except Exception:
                 return None
+
+    async def reset_stuck_processing_files(self, db: AsyncSession | None = None) -> int:
+        """Mark all files whose status is 'pending' or 'processing' as 'failed'.
+
+        Called once at application startup to clean up orphaned background
+        tasks left behind by a previous server crash or restart. Returns the
+        number of files updated."""
+        import time as _time
+        async with get_async_db_context(db) as db:
+            try:
+                stmt = select(File).filter(
+                    File.data['status'].as_string().in_(['pending', 'processing'])
+                )
+                result = await db.execute(stmt)
+                rows = result.scalars().all()
+                now = int(_time.time())
+                for f in rows:
+                    f.data = {
+                        **(f.data or {}),
+                        'status': 'failed',
+                        'error': 'Processing was interrupted by a server restart.',
+                    }
+                    f.updated_at = now
+                await db.commit()
+                return len(rows)
+            except Exception as exc:
+                log.warning('reset_stuck_processing_files failed: %s', exc)
+                return 0
 
     async def update_file_data_by_id(self, id: str, data: dict, db: AsyncSession | None = None) -> FileModel | None:
         async with get_async_db_context(db) as db:
--- a/backend/open_webui/models/files.py
+++ b/backend/open_webui/models/files.py
@@ -251,31 +251,6 @@ class FilesTable:
             result = await db.execute(select(File).filter_by(user_id=user_id))
             return [FileModel.model_validate(file) for file in result.scalars().all()]
 
-    async def get_pending_files_by_knowledge_id(
-        self,
-        knowledge_id: str,
-        user_id: str | None = None,
-        db: AsyncSession | None = None,
-    ) -> list['FileModelResponse']:
-        """Return files that reference knowledge_id in their meta.data and are
-        still being processed (status pending or processing).  Used to surface
-        in-flight uploads after a page refresh, before the background task has
-        finished and linked the file to the knowledge base."""
-        async with get_async_db_context(db) as db:
-            stmt = select(File).filter(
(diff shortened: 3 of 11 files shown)


jannefleischer:
Updated Code to mirror 0.11 updates. Updated PR-Tests reflecting this. - And added a delete-files-not-yet-uploaded function.

I can squash this, once it has been discussed/reviewed. For transparency reasons.

PR #26946, opened 2026-07-09
fix: route terminal CWD pre-fetch through orchestrator policy URL

Classic298:
get_tools_for_terminal fetched the current working directory with the raw connection URL instead of the policy-resolved server URL. For orchestrator connections with a policy_id, the request hit the bare /files/cwd path (bypassing /p/{policy_id}/) while still carrying X-User-Id, so the orchestrator's policy-less catch-all provisioned an extra default-policy container for every user — duplicating containers and escaping the intended policy's resource caps, egress rules, and lifecycle.

Use server_data['url'], which set_terminal_servers already resolves to include the /p/{policy_id} prefix, matching how the tool callables and system-prompt fetch route their requests.

> [!NOTE]


ErenAta16 (commented):
Confirmed the mismatch on main: `get_terminal_cwd` is called with `connection.get("url", "")` — the raw connection URL — while the neighboring tool-callable and system-prompt fetches route through `server_data["url"]`, the policy-resolved URL that `set_terminal_servers` prefixes with `/p/{policy_id}`. For any orchestrator connection using a policy, that means the CWD pre-fetch alone hits the bare `/files/cwd` path with `X-User-Id` still attached, landing on the policy-less catch-all instead of the intended policy. One-line fix, matches the pattern already used by the other calls in the same file. LGTM.

Classic298:
Closing, superseded. Upstream shipped the same fix in 7088d245b, so nothing functional remains in this branch.

PR #26947, opened 2026-07-09
feat: rework DoclingLoader as async submit/poll/retrieve with timeout

jannefleischer:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** This PR references an existing [Issue](https://github.com/open-webui/open-webui/issues) or [Discussion](https://github.com/open-webui/open-webui/discussions) — `Closes #___` / Relates to [#26931](https://github.com/open-webui/open-webui/discussions/26931). If one does not exist, create one first. PRs without a linked issue or discussion may be closed without review.
- [x] **Documentation:** Relevant documentation has been added or updated in the [Open WebUI Docs Repository](https://github.com/open-webui/docs). [*draft here*](https://github.com/open-webui/docs/pull/1339)
- [x] **Testing:** Manual tests have been performed to verify the fix/feature works correctly and does not introduce regressions. Screenshots or recordings are included where applicable.

# Changelog Entry

### Description

The `docling` extraction engine currently calls Docling's synchronous `/v1/convert/file` endpoint and blocks the request until conversion finishes, with no timeout and no visibility into how far along a large document is. For big or scanned PDFs this can mean the upload request just hangs for minutes with nothing to show for it, and there is no way to bound how long Open WebUI is willing to wait before giving up.

This PR reworks `DoclingLoader` (`retrieval/loaders/main.py`) to use Docling Serve's async API instead: submit the file, long-poll for status, then retrieve the result once it succeeds. This is a fix/improvement to the **existing, already-shipped** `docling` provider on its own merits — it removes an unbounded blocking wait and adds a configurable timeout — not scaffolding built for the sake of a future feature. It does not add a new extraction engine, change chunking, or touch the RAG pipeline.

The submit/poll/retrieve pattern itself matches what's already been suggested for this exact timeout problem (see [#17247](https://github.com/open-webui/open-webui/discussions/17247): "DoclingLoader uses synchronous API, which is time-limited [...] instead use the asynchronous API, which allows timeouts three orders of magnitude bigger").

`DoclingLoader` now has both a sync path (`load()`/`load_from_task_id()`, unchanged behavior) and a true async path (`aload()`/`aload_from_task_id()`, new in this revision) — mirroring the sync/async split `MistralLoader` already documents in its docstring. The reason both exist: `Loader.load()` (fully synchronous) is still used by the URL-content-extraction path (`retrieval/utils.py`), which already runs off the event loop via `asyncio.to_thread`, so blocking there is expected and unchanged. The interactive file-upload path (`Loader.aload()`, used by `POST /files`) previously funneled *every* loader — including `DoclingLoader` — through `asyncio.to_thread(self.load, ...)`. That's fine for loaders that are quick and CPU-bound, but for `DoclingLoader` specifically it meant the entire submit/poll/retrieve cycle (potentially minutes, since Docling is not fast) held a worker thread from the shared default executor for its full duration — the same pool every other `to_thread`-offloaded call in the app draws from. `Loader.aload()` now detects `DoclingLoader` and calls its `aload()` directly (aiohttp + `asyncio.sleep` between polls) instead, so a slow conversion no longer ties up a thread while idly waiting on a poll. All other loaders are untouched and still go through `asyncio.to_thread`.

It is the second of several smaller PRs split out of a previous, too-large PR (#26932) per [maintainer feedback](https://github.com/open-webui/open-webui/pull/26932#issuecomment-4925036925). For full context on where this fits (including a separate, still-to-come `docling_json` extraction provider and RAG pipeline improvements that build on top of this once it lands), see the [linked discussion](https://github.com/open-webui/open-webui/discussions/26931) — but this PR should be reviewable and mergeable entirely on its own, independent of that follow-up work.

### Added

- `DOCLING_SERVE_TIMEOUT` setting (env var + RAG config API): total seconds Open WebUI will wait for a Docling conversion before raising a timeout error. Unset (default) preserves the previous behavior of waiting indefinitely.
- `DoclingLoader.load_from_task_id(task_id)`: resumes polling and result retrieval for an already-submitted task without re-uploading the file — intended for a future recovery path (e.g. after a server restart while docling-serve is still processing), not wired up by this PR.
- Optional `status_callback` constructor parameter on `DoclingLoader`, invoked with `task_id`/`task_position` as they become known. It's inert unless a caller supplies one — nothing in this PR does, so behavior is unchanged; it exists so a separate status-tracking feature can hook into it without further changes here. The callback may be sync or async — `aload()`/`aload_from_task_id()` check for an awaitable return value and `await` it, since a callback that persists status (e.g. to a database) will often need to be async, and calling one synchronously would silently produce an un-awaited coroutine that never runs.

### Changed

- `DoclingLoader` now submits via `POST /v1/convert/file/async` instead of `POST /v1/convert/file`, then long-polls `GET /v1/status/poll/{task_id}` (30s poll window, retried on client-side timeout) until the task reports `success` or `failure`, then fetches `GET /v1/result/{task_id}`.
- Internal request/error-handling logic was split into small private helpers (`_build_headers`, `_submit_file`, `_poll_task_until_done`, `_retrieve_result`, `format_result`) instead of one large `load()` method, to keep the submit/poll/retrieve/resume paths (used by both `load()` and the new `load_from_task_id()`) from duplicating logic.
- Added a true async path: `DoclingLoader.aload()` / `aload_from_task_id()`, backed by `aiohttp` + `await asyncio.sleep()` between polls (private helpers `_submit_file_async`, `_poll_task_until_done_async`, `_retrieve_result_async`). `Loader.aload()` — the dispatcher used by the interactive file-upload path — now calls `DoclingLoader.aload()` directly instead of running the sync `load()` through `asyncio.to_thread()`, so a slow conversion no longer occupies a worker thread from the shared default executor for its entire duration. The sync `load()`/`load_from_task_id()` methods are unchanged and still used by the URL-content-extraction path in `retrieval/utils.py`, which already runs inside its own `asyncio.to_thread` call.
- Added `_notify_status_async()`: the async submit/poll helpers now call `status_callback` through this, which awaits the callback's return value if it's awaitable. Needed because the async path can reasonably be given an `async def` callback, and calling one as a plain function silently drops it instead of running it.

### Deprecated

- N/A

### Removed

- The old page-break-marker splitting on `\f` in the sync response is gone, since the async result format is handled directly in `format_result()`; behavior for the common case (single Document per file) is unchanged.

### Fixed

- N/A

### Security

- N/A

### Breaking Changes

- None from Open WebUI's side — no config keys were renamed or removed, and `DOCLING_SERVE_TIMEOUT` defaults to the old "wait forever" behavior. Worth flagging for reviewers: this does require the configured Docling Serve instance to expose the async endpoints (`/v1/convert/file/async`, `/v1/status/poll/{task_id}`, `/v1/result/{task_id}`) rather than only the synchronous `/v1/convert/file`. These have been part of Docling Serve for a while and are what I tested against (`quay.io/docling-project/docling-serve-cpu:latest`), but a very old self-hosted Docling Serve without them would need upgrading.

---

### Additional Information

- Relates to discussion #26931.
- Part of the split of #26932 (closed for being too large for a first review pass). This PR covers only the existing `DoclingLoader` engine; a separate PR will add the new `docling_json` structured-extraction engine, and another will cover the unrelated RAG pipeline improvements.
- No new third-party dependencies; `aiohttp` is already a project dependency and already used the same way (a `ClientSession` per call, `asyncio.sleep` between retries) by `MistralLoader`.
- Manually verified against the real `docling-serve` container this deployment runs (`jannefleischer/docling-serve-whisper:v1.14.3`): submitted a test PDF via `aload()`, confirmed the `pending → started → success` status transitions over `/v1/status/poll/{task_id}`, and confirmed `/v1/result/{task_id}` returned the expected extracted text. Also exercised the client-side poll-timeout/retry branch and the overall-timeout branch (`DOCLING_SERVE_TIMEOUT` exceeded while still pending) against a local mock server, since triggering those against the real server isn't practical on demand.
- `ruff format` / `ruff check --select=F` (the same checks `backend.yaml` CI runs) pass on the changed file.

> [!NOTE]

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -966,6 +966,10 @@ ENABLE_RAG_HYBRID_SEARCH_ENRICHED_TEXTS = (
     os.getenv('ENABLE_RAG_HYBRID_SEARCH_ENRICHED_TEXTS', 'False').lower() == 'true'
 )
 
+DOCLING_SERVE_TIMEOUT = (
+    int(os.environ.get("DOCLING_SERVE_TIMEOUT")) if os.environ.get("DOCLING_SERVE_TIMEOUT") else None
+)
+
 RAG_FULL_CONTEXT = os.getenv('RAG_FULL_CONTEXT', 'False').lower() == 'true'
 
 RAG_FILE_MAX_COUNT = int(os.getenv('RAG_FILE_MAX_COUNT')) if os.getenv('RAG_FILE_MAX_COUNT') else None
@@ -2860,6 +2864,7 @@ DEFAULT_CONFIG = {
     'rag.docling_server_url': DOCLING_SERVER_URL,
     'rag.docling_api_key': DOCLING_API_KEY,
     'rag.docling_params': DOCLING_PARAMS,
+    'rag.docling_serve_timeout': DOCLING_SERVE_TIMEOUT,
     'rag.document_intelligence_endpoint': DOCUMENT_INTELLIGENCE_ENDPOINT,
     'rag.document_intelligence_key': DOCUMENT_INTELLIGENCE_KEY,
     'rag.document_intelligence_model': DOCUMENT_INTELLIGENCE_MODEL,
--- a/backend/open_webui/retrieval/loaders/main.py
+++ b/backend/open_webui/retrieval/loaders/main.py
@@ -1,8 +1,11 @@
 import asyncio
+import inspect
 import json
 import logging
 import sys
+import time
 
+import aiohttp
 import ftfy
 import requests
 from azure.identity import DefaultAzureCredential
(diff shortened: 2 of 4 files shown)


jannefleischer:
There are tests here: https://github.com/open-webui/tests/pull/8 - though mostly created by AI, no clue what I am doing...

jannefleischer:
I have reworked this, so it just relies on current dev (and not anymore any other PRs I might have had in the pipeline)

PR #26951, opened 2026-07-09
fix: don't use raw user message as retrieval query on empty query list

jannefleischer:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** This PR references an existing [Issue](https://github.com/open-webui/open-webui/issues) or [Discussion](https://github.com/open-webui/open-webui/discussions) — Originally this has been part of the PR related to the discussion here: https://github.com/open-webui/open-webui/discussions/26931 - though its not relly related to the topic.
- [ ] **Testing:** Manual tests have been performed to verify the fix/feature works correctly and does not introduce regressions. Screenshots or recordings are included where applicable.
- [ ] **Self-Review:** A self-review of the code has been performed, ensuring adherence to project coding standards.

# Changelog Entry

### Description

`chat_completion_files_handler()` calls `generate_queries()` to turn the conversation into targeted search queries before running vector retrieval over attached files/knowledge bases. When that call fails or returns something unparseable, the code falls back to using the raw last user message as the retrieval query — a reasonable safety net.

The bug: that same fallback also fired when `generate_queries()` succeeded and the LLM correctly returned an empty query list — which is its documented way of signalling "no targeted search needed," e.g. for prompts like "summarize this document" or "translate this text." An empty-but-successful response was indistinguishable from a failed one, so both took the same path: using the raw user message ("summarize this document") as a vector-similarity query. Against real document content, a query like that scores close to zero relevance, so retrieval effectively returned near-random chunks instead of usable context — silently degrading exactly the kind of "whole document" request the empty-query signal was meant to handle better.

### Added

- N/A

### Changed

- `chat_completion_files_handler()` now distinguishes "query generation succeeded with zero queries" from "query generation failed/errored." Only the latter falls back to using the raw user message as the retrieval query. The former now sets `full_context=True` for that retrieval call — the same full-context path already used when a file is explicitly marked `context: 'full'` — so the LLM gets the actual document content instead of a handful of low-relevance chunks.
- **@Reviewer: I am not sure if this is a good behaviour for all cases. It can burn tokens if the full document is huge.**

### Deprecated

- N/A

### Removed

- N/A

### Fixed

- Prompts that make the query-generation LLM correctly return an empty query list (summarization, translation, "list everything in this file", etc.) no longer degrade to near-random chunk retrieval. They now use full document context instead.

### Security

- N/A

### Breaking Changes

- None. This only changes behavior for the specific case where `generate_queries()` succeeds and returns zero queries — previously silently broken, now falls back to full context instead of a low-relevance raw-message query.

---

### Additional Information

- Found while working on a separate, larger RAG-pipeline branch; isolated here as a standalone, minimal fix since it doesn't depend on or relate to the rest of that work.
- No dependency changes.

### Screenshots or Videos

- TODO before marking ready for review: a short before/after showing retrieval sources for a "summarize this document" prompt against an attached file — before: a handful of unrelated low-score chunks; after: full document context.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -1769,6 +1769,12 @@ async def chat_completion_files_handler(
         all_full_context = all(item.get('context') == 'full' for item in files)
 
         queries = []
+        # When the query-generation LLM returns an empty list, it signals that
+        # no targeted search is needed (e.g. summarise, translate, list all).
+        # In that case we fall back to full-context mode instead of using the
+        # raw user message as a similarity query (which always scores ~0).
+        llm_requested_full_context = False
+
         if not all_full_context:
             try:
                 queries_response = await generate_queries(
@@ -1796,6 +1802,9 @@ async def chat_completion_files_handler(
                     queries_response = {'queries': [queries_response]}
 
                 queries = queries_response.get('queries', [])
+                if len(queries) == 0:
+                    # LLM explicitly decided no search is needed → use full context
+                    llm_requested_full_context = True
             except Exception:
                 pass
 
@@ -1810,7 +1819,9 @@ async def chat_completion_files_handler(
                 }
             )
 
-        if len(queries) == 0:
+        # Only fall back to the raw user message when the LLM did NOT explicitly
+        # return an empty list (i.e. when query generation itself failed/errored).
+        if len(queries) == 0 and not llm_requested_full_context:
             queries = [get_last_user_message(body['messages']) or '']
 
         try:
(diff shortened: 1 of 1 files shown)


ErenAta16 (commented):
The distinction being drawn here is the right one: an empty `queries` list coming back from the query-generation LLM is not the same situation as query generation raising/erroring, but the old code funneled both into "use the raw user message as the similarity query" — which, for a genuine "no targeted search needed" response (summarize/translate/list-all intents), searches with a query that was never meant to be a search query and predictably scores everything near zero. Routing the explicit-empty-list case to `full_context` instead is a more faithful reading of what the LLM was actually telling the caller. The `llm_requested_full_context` flag cleanly keeps the still-needed fallback path (do use the raw message when generation itself failed) intact for the other empty-list cause.

PR #26956, opened 2026-07-09
perf: avoid O(elapsed-time) rrule walk for sub-daily automation next-run

### LINKED ISSUE #26954: bug: Automation next-run for MINUTELY/HOURLY schedules walks ~26 years of occurrences (20+ s per call, worsens yearly)
Sub-daily automation schedules (`FREQ=MINUTELY` / `FREQ=HOURLY`) take 20+ seconds to compute their next run, and it gets slower every year.

### Cause

`_parse_rule()` in `backend/open_webui/utils/automations.py` anchors these rules at a fixed `DTSTART` of `2000-01-01`:

```python
if freq in ('MINUTELY', 'HOURLY'):
    epoch = datetime(2000, 1, 1, 0, 0, 0)
    return rrulestr(s, dtstart=epoch, ignoretz=True)
```

Every next-run lookup then calls `rule.after(now)`, which makes dateutil iterate occurrence-by-occurrence from the year 2000 up to now — ~6.8M steps for a 2-minute rule. The cost scales with wall-clock distance from 2000 (~+0.85s/year) and is paid on every `validate_rrule`, `next_run_ns`, and `next_n_runs_ns`, including on each reschedule in `claim_due` after a run fires.

### Reproduction

Isolated:

```python
from dateutil.rrule import rrulestr
from datetime import datetime
import time
t = time.time()
rrulestr("RRULE:FREQ=MINUTELY;INTERVAL=2", dtstart=datetime(2000, 1, 1), ignoretz=True).after(datetime.now())
print(time.time() - t)   # ~22s in mid-2026
```

Or via the UI: create an automation with `FREQ=MINUTELY;INTERVAL=2` and the create/validate request blocks for ~20+ seconds.

### Expected vs actual

- **Expected:** next-run computation is fast and roughly constant-time regardless of the current date.
- **Actual:** ~22s today for a 2-minute rule, worsening ~0.85s/year, re-run on every reschedule.

### Version

Present on `main` and `dev` (0.10.2).

I have a fix ready (snap `DTSTART` to the most recent point on the same epoch grid so `.after()` walks a bounded remainder while keeping identical occurrences) and will open a PR referencing this issue.

Samrj12:
Closes #26954

### Description

MINUTELY/HOURLY automation schedules anchor `dtstart` at a fixed `2000-01-01`, so `rule.after(now)` iterates from the year 2000 to now on every next-run computation — ~6.8M steps for a 2-minute rule (~22s today), growing ~0.85s/year, and re-run on every `validate_rrule` / `next_run_ns` / `next_n_runs_ns` and on each reschedule in `claim_due`.

This snaps `dtstart` to the most recent point on the **same** fixed epoch grid at or before now, so `.after()` walks a bounded remainder (≤ ~a day of steps) instead of ~26 years. Occurrences are identical because the anchor stays on the exact same grid; the 2-day back-off keeps the anchor at or before any caller's local `now`. `FREQ=MINUTELY;INTERVAL=2` drops from ~22s to ~2ms.

### Changed

- `_parse_rule()` snaps the MINUTELY/HOURLY `DTSTART` to the nearest prior point on the fixed epoch grid instead of always starting at 2000-01-01.

### Testing

- Verified next-run output is identical to the year-2000 anchor for intervals 2/5/7/13/90, including a "user timezone behind server" case and the midnight-rollover boundary.
- Timed `FREQ=MINUTELY;INTERVAL=2`: ~22s → ~2ms.

---

- [x] Targets the `dev` branch
- [x] PR title uses the `perf:` prefix
- [x] Changelog entry included above
- [x] Human-reviewed and manually tested
- [x] Self-review performed

### DIFF
--- a/backend/open_webui/utils/automations.py
+++ b/backend/open_webui/utils/automations.py
@@ -77,7 +77,13 @@ def _parse_rule(s: str):
 
     if freq in ('MINUTELY', 'HOURLY'):
         epoch = datetime(2000, 1, 1, 0, 0, 0)
-        return rrulestr(s, dtstart=epoch, ignoretz=True)
+        # Snap DTSTART to the most recent point on the fixed epoch grid at or
+        # before now so .after() walks a bounded remainder instead of ~26 years
+        # of steps, while keeping the exact same clock-aligned occurrences. The
+        # 2-day back-off keeps the anchor at or before any caller's local now.
+        step = int(parts.get('INTERVAL', '1')) * (60 if freq == 'MINUTELY' else 3600)
+        k = int((datetime.now() - timedelta(days=2) - epoch).total_seconds() // step)
+        return rrulestr(s, dtstart=epoch + timedelta(seconds=k * step), ignoretz=True)
     return rrulestr(s, ignoretz=True)
 
 


ErenAta16 (commented):
Reproduced the anchor math standalone: for `FREQ=MINUTELY;INTERVAL=2`, snapping `dtstart` to `epoch + k*step` (k chosen so the anchor lands within the 2-day back-off window before now) keeps `(snapped - epoch)` an exact multiple of `step` — confirmed this numerically — so `.after()` produces the identical occurrence sequence, it just starts the walk ~2 days back instead of ~26 years back. That is the part that actually matters for correctness here (get the grid alignment wrong and every scheduled run silently shifts), and it checks out.

tjbck:
Addressed in dev.

PR #27003, opened 2026-07-11
fix: restrict folder deletion to the owner or an admin

Classic298:
Deleting a folder cascades into the folder owner's chats, messages and the entire subfolder subtree; the cascade is bound to the folder's owner, not the caller. The delete handler only enforced owner/admin for root folders. Subfolder deletion required merely write access, and a write grant on a shared root folder is inherited by every descendant subfolder. A write-collaborator could therefore permanently delete the owner's chats by deleting a subfolder of a shared folder, data they do not own. With delete_contents=false the same path force-moved the owner's chats out of the folder instead.

This also contradicted the documented sharing model: only the owner or an admin may delete a shared folder, and write access covers adding and editing chats and subfolders, not removing the folder.

Because any folder deletion cascades into the owner's data, restrict it to the owner or an admin for root and subfolders alike, replacing the root/subfolder split with a single check. Owners and admins are unaffected, and a write-collaborator can still create, rename and add to shared folders and delete subfolders they own.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/folders.py
+++ b/backend/open_webui/routers/folders.py
@@ -534,25 +534,20 @@ async def delete_folder_by_id(
     folder = await Folders.get_folder_by_id_and_user_id(id, user.id, db=db)
 
     if not folder:
-        # Check if it's a shared subfolder with write access
+        # Caller is not the owner. Deletion cascades into the owner's chats and
+        # whole subfolder subtree, so a write-collaborator (whose grant is even
+        # inherited by subfolders) must not trigger it. Only the owner or an
+        # admin may delete, for root and subfolders alike.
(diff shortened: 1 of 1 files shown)


ErenAta16 (commented):
Read `delete_folder_by_id` on main directly: for a subfolder (`folder.parent_id` is set) owned by someone else, the only gate was `_has_folder_access(user.id, folder, "write", db)` — no owner/admin check. Since a write grant on a shared root folder is documented (and implemented) to be inherited by its subfolders, that means a write-collaborator could delete a subfolder they were only ever granted edit access to, cascading into the actual owner`s chats and messages. The new version collapses both the root and subfolder non-owner branches into one owner-or-admin check, matching what the PR says is the documented sharing model. LGTM.

PR #27004, opened 2026-07-11
fix: strip model params for read-only callers in the model list endpoint

Classic298:
The per-id model endpoint (GET /api/v1/models/model) strips params, the system prompt and other curated model config, for callers who only have read access. The list endpoint (GET /api/v1/models/list) did not: it returned each read-accessible model's full params, so a read-shared model exposed its params.system to non-owner read-grant holders.

Mirror the per-id behaviour: compute write_access per item and drop params before serialising when the caller lacks write access (not the owner, not an admin under BYPASS_ADMIN_ACCESS_CONTROL and holding no write grant). The model-card list UI does not render params, so this does not change functionality.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/models.py
+++ b/backend/open_webui/routers/models.py
@@ -176,16 +176,16 @@ async def get_models(
         data = model.model_dump()
         if data.get('meta'):
             data['meta'].pop('profile_image_url', None)
-        items.append(
-            ModelAccessResponse(
-                **data,
-                write_access=(
-                    (user.role == 'admin' and BYPASS_ADMIN_ACCESS_CONTROL)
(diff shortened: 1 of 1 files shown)


ErenAta16 (commented):
Confirmed on main: the per-id `get_model_by_id` endpoint (not shown in this diff) does strip `params` for non-writers, but this list endpoint`s loop built `ModelAccessResponse(**data, write_access=...)` straight from `model.model_dump()` with no equivalent check — `params` (including `system`) went out regardless of `write_access`. The fix computes `write_access` once and reuses it both for the field and for the conditional strip rather than computing it twice, which is a nice side benefit. LGTM.

PR #27005, opened 2026-07-11
fix: do not expose tool source code to read-only users

Classic298:
`GET /tools/id/{id}` returned the full tool record to anyone with read access, including `content`, which is the tool's source code. Read access is meant to allow using a tool, not reading its implementation, and the response model uses `extra='allow'`, so `content` is re-admitted from the model dump even though it is not a declared field.

The endpoint now computes write access once and removes `content` for callers that do not have it. `specs` is deliberately left in place, because the chat Available Tools modal renders it for every read-shared tool a user can call. The list endpoints are not touched: upstream's `defer_content` change already keeps source out of those responses.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/tools.py
+++ b/backend/open_webui/routers/tools.py
@@ -56,32 +56,6 @@ async def get_tool_module(request, tool_id, load_from_db=True):
     return tool_module
 
 
-def _tool_has_write_access(user, tool, user_group_ids: set) -> bool:
-    return (
-        (user.role == 'admin' and BYPASS_ADMIN_ACCESS_CONTROL)
-        or user.id == tool.user_id
-        or any(
-            g.permission == 'write'
-            and (
-                (g.principal_type == 'user' and g.principal_id in (user.id, '*'))
-                or (g.principal_type == 'group' and g.principal_id in user_group_ids)
-            )
-            for g in tool.access_grants
-        )
-    )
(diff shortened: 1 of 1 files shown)


ErenAta16 (commented):
test

ErenAta16 (commented):
Traced this end to end on main before looking at the fix: `ToolsTable.get_tools(defer_content=True, ...)` has `if defer_content: stmt = stmt` — that branch really is a complete no-op, it reassigns the same variable to itself and does nothing to the query. Then `ToolUserResponse` (and `ToolAccessResponse`, which the id/list endpoints use) both inherit `ConfigDict(extra="allow")`, so spreading `tool.model_dump()` — which comes from the full `ToolModel` with `content`/`specs` — into the constructor lets those fields ride along in the response even though `ToolResponse` never declares them. So both halves of the bug (the ineffective content-loading guard and the schema-level leak) check out exactly as described. LGTM.

The `_tool_response_data` extraction also fixes something adjacent: the `has_write` computation was duplicated between `get_tool_list` and `get_tools_by_id` with slightly different equality checks (`principal_id in (user.id, "*")` vs `== user.id or == "*"`) — same semantics, but having one helper removes the chance of them drifting apart later. (Apologies for the stray placeholder comment above — this is the actual review.)

PR #27041, opened 2026-07-13
fix: preserve complete turns during context compaction

### LINKED ISSUE #27035: ISSUE: Context compaction boundary is turn-blind — it compacts the middle of a single assistant turn instead of whole prior turns
# Summary:
_find_compaction_boundary in utils/context_compaction.py splits the flattened message list by count (keep last ~40%, compact first ~60%), with no concept of a logical conversation turn. When one assistant turn expands into many tool-call / tool-result messages, the boundary falls inside that single turn, so its early tool calls/results get summarized away while later ones are kept — corrupting the structure of a turn rather than compacting completed prior turns.

# Where

open_webui/utils/context_compaction.py, _find_compaction_boundary (and its caller compact_messages_for_request).

```
keep_count = max(2, len(messages) * 2 // 5)   # keep last ~40%
split = max(1, len(messages) - keep_count)     # compact first ~60%
# while-loop only nudges past an orphaned tool result; no turn awareness
```

# Steps to reproduce:

Enable context compaction with a token threshold.
Run a single agentic turn that makes several tool calls (so the flattened history is >3 messages and trips the threshold).
On the next request, compaction triggers and the boundary lands within that one turn.
Expected: Compaction operates on whole completed turns — it should never split a single user→assistant(+tool calls)→ turn. Keep the most recent turn(s) intact; only summarize fully-completed earlier turns.

**Actual:** The split is purely count-based on flattened messages; a single heavy turn gets its interior tool calls/results shoved into the compacted slice.

**Impact**: Loss of tool-call/result coherence for the current work; combined with the unbounded summary payload, it feeds partial-turn tool output into the task model.

**Suggested fix:** Compute the boundary on turn units (group by user message → its assistant/tool messages), and choose a split that only ever falls on a turn boundary, keeping ≥1 most-recent complete turn.

dexhunter:
## Summary

- choose context-compaction boundaries only at complete user-turn starts
- keep the existing approximately 40% recent-context target when a later complete turn is available
- fall back to the most recent complete turn, or no compaction boundary when no complete turn exists

Closes #27035

## Why

The existing count-based boundary can land inside one assistant/tool turn. That sends only part of the turn to the summary model and leaves the remainder in recent context. Selecting a user-message boundary keeps tool calls and results from the same completed turn together.

## Validation

- deterministic 506-case boundary evaluation: `486` baseline errors to `0`
- gates for source scope, compatibility, input mutation, determinism, invalid boundaries, and exceptions
- `python -m py_compile backend/open_webui/utils/context_compaction.py`
- manual diff review against current `dev`

This boundary algorithm was refined through autoresearch with Weco and then reduced to the minimal targeted patch. [Public Weco run](https://dashboard.weco.ai/share/SGvjOOBSFJFr1ZVLqqrfHDn9i-G4W6NS)

## Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #27035.
- [x] **Target branch:** This pull request targets `dev`.
- [x] **Description:** The behavior, motivation, and impact are described above.
- [x] **Documentation:** No documentation change is required for this bug fix.
- [x] **Dependencies:** No dependencies were added or changed.
- [x] **Testing:** The strict deterministic evaluation and syntax check pass.
- [x] **No Unchecked AI Code:** The final patch was manually reduced, reviewed, and validated.
- [x] **Self-Review:** The diff and edge cases were reviewed.
- [x] **Architecture:** The change preserves the existing compaction policy and settings.
- [x] **Git Hygiene:** The pull request contains one atomic source change on current `dev`.
- [x] **Title Prefix:** The title uses the `fix` prefix.

# Changelog Entry

### Fixed

- Context compaction no longer splits a completed assistant/tool turn across summarized and recent context.

### Additional Information

- No UI changes or screenshots.

### DIFF
--- a/backend/open_webui/utils/context_compaction.py
+++ b/backend/open_webui/utils/context_compaction.py
@@ -244,15 +244,18 @@ def _find_compaction_boundary(messages: list[dict]) -> int:
     keep_count = max(2, len(messages) * 2 // 5)
     split = max(1, len(messages) - keep_count)
 
-    while split < len(messages) - 1:
-        previous = messages[split - 1] if split > 0 else {}
-        current = messages[split]
-        if current.get('role') == 'tool' or previous.get('tool_calls') or previous.get('output'):
-            split += 1
-            continue
-        break
+    turn_starts = [
+        idx
+        for idx, message in enumerate(messages)
+        if message.get('role') == 'user' and idx <= len(messages) - 2
+    ]
+    for turn_start in turn_starts:
+        if turn_start >= split:
+            return turn_start
 
-    return min(split, len(messages) - 2)
+    if turn_starts:
+        return turn_starts[-1]
+    return 0
 
 
 async def _generate_summary(


Classic298:
tested? needs manual testing not a py compile

dexhunter:
Thanks — you are right that `py_compile` alone was not enough. I synced the full locked Python environment and manually exercised the production `compact_messages_for_request` path with compaction enabled (only the external summary-model call was stubbed):

- completed six-message tool turn + follow-up: compacted exactly the complete first turn and kept `[user, assistant]`
- single tool-heavy turn: did not compact or call the summary model
- completed tool turn + second completed turn + pending user: compacted the first complete turn and kept `[user, assistant, user]`

The production path preserved the input list, and the summary call received exactly the whole-turn compacted/recent slices. I also reran the deterministic 506-case gate: `0` boundary errors.

This is full function-path validation, not a browser/provider E2E. If you want a specific UI + provider + tool sequence tested before reopening, please point me to the preferred scenario and I will run that exact flow.

PR #27054, opened 2026-07-13
fix: block IPv6 transition-form encodings of internal IPs in web-fetc…

Classic298:
…h SSRF filter

The web-fetch SSRF guard (validate_url plus the connect-time re-checks in _ssrf_safe_new_conn and _SSRFSafeResolver) decided whether a destination was allowed solely from ipaddress.ip_address(ip).is_global on the literal address. is_global does not decode the IPv4 that standardized IPv6 transition forms embed, so it reports several of them as globally routable even when they carry an internal IPv4. On a host with the matching routing (a NAT64 gateway, common on IPv6-only and dual-stack clusters, or a 6to4/Teredo tunnel) the kernel forwards those packets to the embedded IPv4, so a URL such as http://[64:ff9b::a9fe:a9fe]/ reaches 169.254.169.254 while every guard treats it as public. The plain forms (169.254.169.254, 127.0.0.1, ::ffff:169.254.169.254) are correctly blocked, so this is purely an encoding that defeats the filter and reaches internal services and cloud metadata with the response returned to the caller.

Classify the effective destination instead of the literal address: unwrap the embedded IPv4 from every standardized transition form (IPv4-mapped, IPv4-compatible, 6to4, Teredo, NAT64 well-known and RFC 8215 local-use) and run is_global on that, at all three checkpoints. Genuinely global destinations, including a NAT64-wrapped public IP like 64:ff9b::808:808 (8.8.8.8), stay allowed, so there is no over-blocking.

> [!NOTE]

### DIFF
--- a/backend/open_webui/retrieval/web/utils.py
+++ b/backend/open_webui/retrieval/web/utils.py
@@ -74,6 +74,37 @@ def resolve_hostname(hostname):
     return ipv4_addresses, ipv6_addresses
 
 
+def _embedded_ipv4(addr: ipaddress.IPv6Address) -> Optional[ipaddress.IPv4Address]:
+    # IPv4 that a standardized IPv6 transition form embeds; on a host with the matching
+    # routing (NAT64/6to4/Teredo) the packet reaches that IPv4, so it, not the literal v6,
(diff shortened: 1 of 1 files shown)


tjbck:
Addressed.

PR #27093, opened 2026-07-15
feat/PERF: add WEBSOCKET_PER_MESSAGE_DEFLATE env var

Classic298:
Profiling 24 production workers showed ~14% of burst CPU spent in permessage_deflate.encode compressing outgoing socket.io frames. Streaming chat deltas re-send growing payloads every flush, so per-frame zlib is a per-token CPU cost with little bandwidth benefit behind a reverse proxy.

start.sh now passes --ws-per-message-deflate (default true) to uvicorn. Placed before passthrough args so an explicit flag in the container command still wins.

> [!NOTE]

### DIFF
--- a/backend/open_webui/__init__.py
+++ b/backend/open_webui/__init__.py
@@ -72,6 +72,7 @@ def serve(
 
     import open_webui.main  # noqa: F401
     from open_webui.env import UVICORN_WORKERS  # Import the workers setting
+    from open_webui.env import WEBSOCKET_PER_MESSAGE_DEFLATE
 
     # On Windows, uvicorn's default loop factory hardcodes ProactorEventLoop,
     # which is incompatible with psycopg v3 async.  Setting loop='none' lets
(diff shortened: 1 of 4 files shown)


gaby:
@Classic298 Another possible improvement is replacing uvicorn with Granian. 

https://github.com/emmett-framework/granian

Benchmarks: https://github.com/emmett-framework/granian/blob/master/benchmarks/vs.md

Classic298:
@gaby didnt you switch away from open webui?

gaby:
@Classic298 I still use it at home for local LLM via OpenRouter on a raspberry pi.

For everything else, I am/we are using LiteLLM + vLLM.

Classic298:
while this saves 20% CPU, under current re-sending-entire-response-per-streamed-chunk, disabling this would require fiber optic internet connection to handle the amount sent UNCOMPRESSED via wire

gaby:
@Classic298 Just make it `False` by default. I see the use in this.

Classic298:
I did the calculations 

A long stream would turn from a few Megabytes over the course of 2-3 minutes to potentially Gigabytes

Even good internet connections can barely carry that.

And yes this has use @gaby, but only after a different method for transferring the chunks has been found that doesn't involve sending the entire history every time

Issue #26976, opened 2026-07-10
feat: OAuth2 client credentials auth for external LLM, embedding, and reranking connections


Azzeo:
### Check Existing Issues

- [x] I have searched all existing **open AND closed** issues and discussions and found none comparable to my request.

### Verify Feature Scope

- [x] I believe this feature request is appropriately scoped for the Issues section as described above.

### Problem Description

Open WebUI can connect to external OpenAI-compatible services in three places — LLM/chat connections, the RAG **embedding** engine, and the RAG **reranking** engine. For all three, outbound authentication is currently limited to:

1. a **static API key** stored per connection/engine and sent as a `Bearer` token, or
2. **forwarding the signed-in user's OAuth/JWT token** downstream.

Many organizations put their inference, embedding, and rerank APIs behind an API-management gateway that only accepts the **OAuth2 client credentials grant** (machine-to-machine). In these environments no static API key is issued at all. Access requires the caller to present a client ID and client secret to a token endpoint, obtain a short-lived access token, and send that token as the `Bearer` credential.

Open WebUI has no way to do this. There is no configuration point to supply a token URL, client ID, and client secret so the application can fetch and refresh a service access token on its own behalf.

Forwarding the user's token is not a workaround for this scenario:

- The user's access token typically has the wrong audience/scope for the backend service and is rejected by the gateway. Embedding and Reranker don't even have this option
- **The embedding and reranking calls run during background document indexing, the Memories feature, and scheduled tasks — contexts where there is no user session, so there is no token to forward.** These two engines therefore cannot be authenticated by forwarding under any configuration.
- Even when a user token is present, the intended identity for these calls is a single service account, not the individual end user.

The result is that Open WebUI cannot be used at all against gateways that require client-credentials M2M auth, which is a common enterprise deployment pattern.

### Proposed Solution

Add **OAuth2 client credentials grant** as a selectable outbound authentication mode, alongside the existing API-key and token-forwarding options, for:

- OpenAI-compatible **LLM / chat connections**,
- the RAG **embedding** engine (External / OpenAI-compatible), and
- the RAG **reranking** engine (External).

For each of these, expose an "Authentication method" choice:

- `API Key` (current behavior, default — nothing changes for existing users),
- `Forward user token` (current behavior), and
- `OAuth2 Client Credentials` (new).

When `OAuth2 Client Credentials` is selected, the admin configures:

- **Token URL** (the authorization server's token endpoint),
- **Client ID**,
- **Client Secret**,
[...]

PR #27098, opened 2026-07-15
feat: add system_oauth auth type for RAG OpenAI embedding connection

khalidnass:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Relates to #26976 (OAuth2 auth for external LLM/embedding/reranking connections — this PR covers the embedding engine with the existing `system_oauth` mechanism)
- [x] **Description:** provided below
- [x] **Changelog:** included below
- [ ] **Documentation:** happy to follow up with a docs PR for the new env var if this is accepted
- [x] **Dependencies:** no new dependencies
- [x] **Testing:** verified end-to-end (details below)
- [x] **No Unchecked AI Code:** developed with Claude Code, verified by an automated e2e suite and live end-to-end against an OAuth-only gateway; human review pass completed
- [x] **Self-Review:** done
- [x] **Architecture:** the new env var defaults to `bearer`, which is byte-for-byte stock behavior; no UI changes, no new persisted settings
- [x] **Git Hygiene:** atomic commits based on `dev` — the embedding connection, plus three follow-up commits: closing the same gap in the builtin knowledge tools and the external reranker, and a concurrency fix for the token resolution (see below)

## Description

Chat connections already support `auth_type: "system_oauth"` — forwarding the signed-in
user's SSO OAuth access token as the `Authorization: Bearer` header — so deployments behind
OAuth-authenticated LLM gateways work for chat/models.

The RAG **embedding** path, however, only ever sends the static `RAG_OPENAI_API_KEY`
(`retrieval/utils.py`, `agenerate_openai_batch_embeddings`); there is no auth option at all
(`routers/retrieval.py` `OpenAIConfigForm` is just `url` + `key`). Against an OAuth-only
gateway every embedding request gets **401**, so document uploads and RAG queries fail even
though chat works.

This PR adds a `RAG_OPENAI_API_AUTH_TYPE` environment variable:

| Value | Behavior |
|---|---|
| `bearer` (default) | Unchanged stock behavior — static key as Bearer token |
| `system_oauth` | Send the **acting user's** SSO OAuth access token as the Bearer token |

Implementation notes:

- Token resolution mirrors the chat-side `system_oauth`, but works **without a request
  context** (embeddings also run during background document processing): the user's most
  recently updated non-MCP OAuth session is read via `OAuthSessions`, and
  `OAuthManager.get_oauth_token()` transparently refreshes it when near expiry — the same
  refresh path the chat proxy uses.
- Every embedding call site already passes `user=` (file upload via `save_docs_to_vector_db`,
  query endpoints, chat-time retrieval in `utils/middleware.py`), so upload-time and
  query-time embeddings both authenticate as the acting user.
- Graceful fallback: if the user has no OAuth session (e.g. local password account), the
  static key is used and a warning is logged.
- Purely additive: +47 lines across `env.py` and `retrieval/utils.py`; no schema, API, or
  frontend changes.

## Follow-up commits

Exercising this feature against an OAuth-only gateway (Keycloak + Envoy `jwt_authn`,
which 401s static keys) surfaced two more code paths with the same bug — they send the
static key even with `system_oauth` configured — and a concurrency bug in the token
resolution itself:

- **`fix: builtin knowledge tools drop user context on embedding calls`** — the agentic
  knowledge tools (`query_knowledge_base`, `query_knowledge_bases`) call the app-level
  `EMBEDDING_FUNCTION` without the acting user, so token resolution never runs. Passes the
  user at both call sites.
- **`feat: system_oauth auth type for the external reranker connection`** — the external
  reranker always sends the static `RAG_EXTERNAL_RERANKER_API_KEY`. Adds
  `RAG_EXTERNAL_RERANKER_API_AUTH_TYPE` (`bearer` default | `system_oauth`) with the same
  token resolution and fallback semantics, and binds the user through `query_collection`'s
  rerank wrapper (the middleware and retrieval-router wrappers already pass it).
- **`fix: serialize and cache system_oauth token resolution per user`** — a document upload
  embeds many batches concurrently, so token resolution raced: N parallel
  `oauth_manager.get_oauth_token()` calls could each trigger a refresh with the **same**
  refresh token (`OAuthManager._refresh_token` has no concurrency protection). An IdP that
  rotates refresh tokens treats the reuse as replay and revokes the grant — Open WebUI then
  deletes the OAuth session, breaking chat's `system_oauth` for that user too. Fixed by
  serializing resolution behind a per-user `asyncio.Lock` and caching the token for 60s
  (always within its refresh window, so a cached token is never expired).

All three fixed and verified live against that gateway: knowledge-tool queries and
hybrid-search reranking authenticate with the acting user's token end-to-end, concurrent
uploads resolve the token once per user, and with `bearer` (default) behavior is
byte-for-byte stock.

## Testing

Automated end-to-end suite (run against this branch): a mock OpenAI-compatible gateway that
**only** accepts a specific OAuth token (static keys get 401, mimicking a corporate gateway),
a real server boot, a real signup, and a real Fernet-encrypted OAuth session seeded through
`OAuthSessions.create_session()`.

All 7 checks pass:

- `system_oauth`: `POST /api/v1/retrieval/process/text` → 200; `POST /api/v1/retrieval/query/doc` → 200; the gateway received `Bearer <oauth token>`; the static key was never sent.
- `bearer` (stock): same request fails; the gateway received the static key and returned 401 (reproduces the bug this fixes); the OAuth token was never sent — the flag fully gates the feature.

# Changelog Entry

### Description

- Allow the RAG OpenAI-compatible embedding connection to authenticate with the acting user's SSO OAuth access token, matching the `system_oauth` auth type already available on chat connections.

### Added

- `RAG_OPENAI_API_AUTH_TYPE` environment variable (`bearer` | `system_oauth`, default `bearer`) for the RAG OpenAI embedding engine.

### Changed

- None (default behavior is unchanged).

### Deprecated

- None

### Removed

- None

### Fixed

- 401 errors on document upload and RAG queries when the embedding endpoint sits behind an OAuth-authenticated gateway while chat connections use `system_oauth`.
- Concurrent embedding batches racing the OAuth token refresh (the same refresh token used in parallel), which on IdPs with refresh-token rotation revokes the grant and invalidates the user's OAuth session.

### Security

- OAuth tokens are read through the existing encrypted `OAuthSessions` storage and are never logged.

### Breaking Changes

- None

---

### Additional Information

- Resubmission of #27097, which was closed for an unchecked CLA box.
- Relates to #26976. That issue asks for OAuth2 **client-credentials** (machine-to-machine) auth for LLM/embedding/reranking connections; this PR covers the complementary user-token case for embeddings by reusing the `system_oauth` mechanism that chat connections already have. The two approaches compose: client credentials would still be the right answer for accounts without an SSO session.

### Screenshots or Videos

- N/A (backend only; e2e transcript summarized above)

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -579,6 +579,13 @@ AIOHTTP_CLIENT_SESSION_SSL = _parse_ssl_env(os.getenv('AIOHTTP_CLIENT_SESSION_SS
 # has no active OAuth session.
 RAG_OPENAI_API_AUTH_TYPE = os.getenv('RAG_OPENAI_API_AUTH_TYPE', 'bearer').strip().lower()
 
+# Auth type for the external RAG reranker connection.
+# "bearer" (default): send RAG_EXTERNAL_RERANKER_API_KEY as a static Bearer token.
+# "system_oauth": send the acting user's SSO OAuth access token instead, same
+# semantics as the embedding connection above. Falls back to the static key
+# when the user has no active OAuth session.
+RAG_EXTERNAL_RERANKER_API_AUTH_TYPE = os.getenv('RAG_EXTERNAL_RERANKER_API_AUTH_TYPE', 'bearer').strip().lower()
+
 # When False (default), outbound HTTP requests do not follow 3xx redirects.
 AIOHTTP_CLIENT_ALLOW_REDIRECTS = os.getenv('AIOHTTP_CLIENT_ALLOW_REDIRECTS', 'False').lower() == 'true'
 
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -571,6 +571,14 @@ except (ValueError, TypeError):
 # When "True", falls back to AIOHTTP_CLIENT_SSL_CERT_FILE if set.
 AIOHTTP_CLIENT_SESSION_SSL = _parse_ssl_env(os.getenv('AIOHTTP_CLIENT_SESSION_SSL', 'True'))
 
+# Auth type for the RAG OpenAI-compatible embedding connection.
+# "bearer" (default): send RAG_OPENAI_API_KEY as a static Bearer token.
+# "system_oauth": send the acting user's SSO OAuth access token instead,
+# mirroring the `system_oauth` auth type of chat connections — for gateways
+# that only accept OAuth tokens. Falls back to the static key when the user
+# has no active OAuth session.
+RAG_OPENAI_API_AUTH_TYPE = os.getenv('RAG_OPENAI_API_AUTH_TYPE', 'bearer').strip().lower()
+
 # When False (default), outbound HTTP requests do not follow 3xx redirects.
 AIOHTTP_CLIENT_ALLOW_REDIRECTS = os.getenv('AIOHTTP_CLIENT_ALLOW_REDIRECTS', 'False').lower() == 'true'
 
--- a/backend/open_webui/retrieval/models/external.py
+++ b/backend/open_webui/retrieval/models/external.py
@@ -1,9 +1,14 @@
+import asyncio
 import logging
 from typing import List, Optional, Tuple
 from urllib.parse import quote
 
 import requests
-from open_webui.env import ENABLE_FORWARD_USER_INFO_HEADERS, REQUESTS_VERIFY
+from open_webui.env import (
+    ENABLE_FORWARD_USER_INFO_HEADERS,
+    RAG_EXTERNAL_RERANKER_API_AUTH_TYPE,
+    REQUESTS_VERIFY,
+)
 from open_webui.retrieval.models.base_reranker import BaseReranker
 from open_webui.utils.headers import include_user_info_headers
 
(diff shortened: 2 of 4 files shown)


khalidnass:
Retested against v0.11.0: the issue still exists. The RAG embedding connection still only sends the static `RAG_OPENAI_API_KEY`, so an OAuth-only gateway returns 401 for embeddings while chat connections work via `auth_type=system_oauth`.

What did change in 0.11.0: the builtin knowledge tools now pass the acting user into their embedding/reranking calls, which covers the second commit of this PR ("fix: builtin knowledge tools drop user context on embedding calls") — I'm updating the branch to drop that commit. The remaining three commits rebase cleanly onto current `dev` and are all still required:

- `RAG_OPENAI_API_AUTH_TYPE=system_oauth` for the embedding connection — no upstream equivalent yet
- the external reranker auth type — `query_collection`'s rerank wrapper still drops the user on `dev`, and the reranker still has no auth options at all
- serialize + cache the token resolution per user — `OAuthManager._refresh_token` still has no concurrency protection, so concurrent embedding batches behind an IdP with refresh-token rotation still get the grant revoked (which also kills the user's chat session)

E2E on v0.11.0 plus these three commits, against a mock OAuth-only gateway: 7/7 — document processing and query-time embedding succeed with `system_oauth` (the user's token is forwarded, the static key is never sent), and stock `bearer` mode reproduces the 401.

Issue #26779, opened 2026-07-05
issue: `Stop` button does nothing when Redis is configured.


taka817123:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

v0.10.2

### Ollama Version (if applicable)

N/A

### Operating System

Windows 11

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

With `REDIS_URL` configured, `redis_task_command_listener` keeps running
for the whole process lifetime, and the "Stop" button always cancels the
running generation.

### Actual Behavior

`redis_task_command_listener` (`backend/open_webui/tasks.py`) has no
retry/reconnect logic:

```python
async def redis_task_command_listener(app):
    redis: Redis = app.state.redis
    pubsub = redis.pubsub()
    await pubsub.subscribe(REDIS_PUBSUB_CHANNEL)

    async for message in pubsub.listen():
        ...
```

`backend/requirements.txt` bumped `redis` `7.4.0` -> `8.0.0` in v0.10.2.
redis-py 8.0.0 added a `DEFAULT_SOCKET_TIMEOUT = 5` (previously probably
unlimited/`None` in 7.4.0), and `open_webui/utils/redis.py`'s
[...]

PR #27104, opened 2026-07-15
fix: Set default Redis socket timeout to None

EntropyYue:
# Changelog Entry

### Description

- This PR introduces the `REDIS_SOCKET_TIMEOUT` environment variable, which allows setting Redis's `socket_timeout` to either `None` or a user-specified value. Before this PR is applied, Redis pubsub listen will time out after 5 seconds, causing the `redis_task_command_listener`, which uses Redis to manage tasks across replicas, to encounter an error and terminate 5 seconds after no new messages are pushed. No errors will be shown in the console. As a result, for users using Redis, the task stopping functionality will fail regardless of whether multiple replicas are actually deployed. A typical manifestation is that after a user clicks the stop button to stop a chat, the backend continues to push new streaming tokens even after the chat has completed.

### Added

- Added `REDIS_SOCKET_TIMEOUT` environment variable

### Fixed

- Fixed the issue where stream transmission could not be stopped when using Redis.

---

### Additional Information

- Fixes: #26779

> [!NOTE]

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -389,6 +389,12 @@ try:
 except ValueError:
     REDIS_SOCKET_CONNECT_TIMEOUT = None
 
+REDIS_SOCKET_TIMEOUT = os.getenv('REDIS_SOCKET_TIMEOUT', '')
+try:
+    REDIS_SOCKET_TIMEOUT = float(REDIS_SOCKET_TIMEOUT)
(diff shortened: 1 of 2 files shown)


EntropyYue:
This resolves the same issue as #26794, but it can also serve as a complement to environment variables, so I’ll keep it open for now.

PR #27114, opened 2026-07-16
fix(rag): allow special token literals

### LINKED ISSUE #27094: ISSUE: TokenTextSplitter crashes on <|endoftext|> in web content when RAG text splitter is set to token
## Bug Description

When the RAG text splitter is configured to `token`, fetching web pages that contain the literal string `<|endoftext|>` (e.g. in scraped documentation, code examples, or dataset pages) causes an unhandled exception and the request silently fails with no response.

## Steps to Reproduce

1. Go to **Admin Panel → Documents → Text Splitter** and set it to `token`
2. Start a chat with a model that has web search enabled
3. Ask a question that causes Open WebUI to fetch a page containing the literal text `<|endoftext|>` (common on ML/AI documentation and dataset pages)
4. Observe the error — no response is returned

## Error Message

```
Encountered text corresponding to disallowed special token '<|endoftext|>'.
If you want this text to be encoded as a special token, pass it to `allowed_special`,
e.g. `allowed_special={'<|endoftext|>', ...}`.
If you want this text to be encoded as normal text, disable the check for this token by
passing `disallowed_special=(enc.special_tokens_set - {'<|endoftext|>'})`.
To disable this check for all special tokens, pass `disallowed_special=()`.
```

## Root Cause

In `backend/open_webui/routers/retrieval.py`, `TokenTextSplitter` (from LangChain) is instantiated without `allowed_special`, so tiktoken uses its default strict mode and raises a `ValueError` when it encounters any special token literal in the content being chunked.

The length function on line 1585 already correctly passes `disallowed_special=()`, but the `TokenTextSplitter` on line 1687 does not:

```python
# Line 1584-1585 — CORRECT, already safe:
encoding = tiktoken.get_encoding(str(config.TIKTOKEN_ENCODING_NAME))
return lambda text: len(encoding.encode(text, disallowed_special=()))

# Line 1687-1692 — BUG, missing allowed_special:
text_splitter = TokenTextSplitter(
    encoding_name=str(config.TIKTOKEN_ENCODING_NAME),
    chunk_size=config.CHUNK_SIZE,
    chunk_overlap=config.CHUNK_OVERLAP,
    add_start_index=True,
    # ← no allowed_special here, tiktoken will crash on <|endoftext|> etc.
)
```

## Proposed Fix

Pass `allowed_special="all"` to `TokenTextSplitter` so special token literals in web content are treated as plain text, consistent with how the length function is already handled:

```python
text_splitter = TokenTextSplitter(
    encoding_name=str(config.TIKTOKEN_ENCODING_NAME),
    chunk_size=config.CHUNK_SIZE,
    chunk_overlap=config.CHUNK_OVERLAP,
    add_start_index=True,
    allowed_special="all",
)
```

## Workaround

Set the RAG Text Splitter back to the default (`character`) in **Admin Panel → Documents**. This avoids `TokenTextSplitter` entirely and the crash does not occur.

## Environment

- **Branch:** `dev`
- **File:** `backend/open_webui/routers/retrieval.py`
- **Affected setting:** RAG Text Splitter = `token`
- **Trigger:** Any web fetch returning a page with `<|endoftext|>` in its content

tianrking:
## Summary

- configure the RAG token text splitter to encode special-token literals as ordinary text
- keep token splitting consistent with the existing token length function

Closes #27094

## Validation

- instantiated `TokenTextSplitter` with the repository-pinned `tiktoken==0.13.0` and `langchain-text-splitters==1.1.2`, then split `"<|endoftext|> documentation"`
- `python -m py_compile backend/open_webui/routers/retrieval.py`
- `git diff --check`

# Changelog Entry

### Fixed

- Prevent token-based RAG splitting from failing when source content contains a literal tiktoken special token.

### DIFF
--- a/backend/open_webui/routers/retrieval.py
+++ b/backend/open_webui/routers/retrieval.py
@@ -1689,6 +1689,7 @@ def save_docs_to_vector_db(
                 chunk_size=config.CHUNK_SIZE,
                 chunk_overlap=config.CHUNK_OVERLAP,
                 add_start_index=True,
+                disallowed_special=(),
             )
             docs = text_splitter.split_documents(docs)
         elif config.TEXT_SPLITTER == 'token_transformers':


silentoplayz:
LGTM — this mirrors the fix merged in #26210, which addressed the same `disallowed_special` crash in the token-counting path; this PR covers the remaining `TokenTextSplitter` instantiation.

I can confirm the bug is reproducible on the current `:dev` image: with the RAG text splitter set to `token`, syncing the official Open WebUI docs into a Knowledge Base fails embedding for every file containing the literal `<|endoftext|>` string (e.g. `context-window.mdx`) — 16 `process_file` errors per sync in my case. Applying `disallowed_special=()` here is the correct and minimal fix.

tjbck:
Addressed in dev.

PR #27175, opened 2026-07-17
fix: forward chat_id as user field in OpenAI-compatible chat completions

### LINKED ISSUE #27174: fix: forward chat_id as user field in OpenAI-compatible chat completions
## Description

When Open WebUI is configured to use an OpenAI-compatible backend that uses the standard `user` field for session identification (e.g., OpenClaw Gateway), every chat request creates a new backend session because Open WebUI does not include the `user` field in the `/v1/chat/completions` request body.

**Current behavior:** The `metadata` object (which contains `chat_id`) is popped from the payload before forwarding to the upstream API. The `user` field is never set, so upstream backends that rely on it for session affinity generate a new session per request.

**Expected behavior:** The chat/conversation ID should be forwarded as the standard OpenAI `user` field so upstream backends can maintain conversation continuity.

### Impact

- Every message in a chat creates a new backend session
- No conversation memory across turns
- Session/context management is broken for any OpenAI-compatible backend that uses the `user` field for session identification

### Proposed Fix

In `backend/open_webui/routers/openai.py`, after popping `metadata` from the payload, add:

```python
if metadata and metadata.get('chat_id'):
    payload['user'] = metadata.get('chat_id')
```

This is a one-line change that passes the conversation ID as the standard OpenAI `user` field, which is the documented mechanism for session identification in the Chat Completions API.

ekinnee:
### Description

When Open WebUI proxies to an OpenAI-compatible backend that uses the standard `user` field for session identification, every chat request creates a new backend session because the `chat_id` from metadata is never forwarded as the `user` field.

The `metadata` object (which contains `chat_id`) is already popped from the payload before forwarding, but the `user` field is never set from it. This is a one-line fix.

### Fixed

- Forward `chat_id` from metadata as the standard OpenAI `user` field in `/v1/chat/completions` requests, enabling upstream backends (like OpenClaw Gateway) to maintain conversation continuity.

### Additional Information

- Closes #27174
- Minimal change: 3 lines added, 0 removed
- Targets `dev` branch
- No new dependencies, no config changes, no breaking changes
- The `user` field is a standard, optional field in the OpenAI Chat Completions API spec

### Screenshots or Videos

N/A — behavioral change tested via manual inspection of upstream request payloads.

### DIFF
--- a/backend/open_webui/routers/openai.py
+++ b/backend/open_webui/routers/openai.py
@@ -1225,6 +1225,9 @@ async def generate_chat_completion(
     payload = {**form_data}
     metadata = payload.pop('metadata', None)
 
+    if metadata and metadata.get('chat_id'):
+        payload['user'] = metadata.get('chat_id')
+
     model_id = form_data.get('model')
     model_info = await Models.get_model_by_id(model_id)
 


Classic298:
Why is this needed you can already forward chat id to openai completions

ekinnee:
The `chat_id` in metadata is already stripped from the payload before forwarding — `payload.pop("metadata")` happens on line 1228. So even though Open Web UI's frontend naturally includes `chat_id` inside the `metadata` object, it never reaches the upstream API.

The `user` field isn't set anywhere in this code path currently. If a client sends `user` directly, it **would** be forwarded since it's not popped, but that's not how Open Web UI works internally — the frontend always puts conversation tracking in `metadata.chat_id`, not `user`.

This just bridges that gap: when `metadata.chat_id` exists (which it normally does for any Open Web UI chat), forward it as the standard `user` field so upstream backends that use `user` for session tracking (like OpenClaw Gateway) get conversation continuity.

It's a 3-line addition, no breaking changes, no config flags needed.

Classic298:
You know what's also a 3 line addition? Adding the chat id to the optional header fields to be forwarded to the upstream openai endpoint in the user interface

So AGAIN: why is this needed??

ekinnee:
You are conflating HTTP headers with JSON body fields.

**What you are thinking of** — the "optional header fields" in the UI — goes into `api_config["headers"]`, which travels through `get_custom_headers()` into the **HTTP request headers**. That is `session.request(data=payload, headers=headers)` — two completely separate transports.

The `user` field in the OpenAI Chat Completions API is a **JSON body field** (in `payload`). There is no mechanism — not in the UI, not in env vars, not in any config path — that injects body fields from the connection config. The `{{CHAT_ID}}` template variable in custom headers is for header values, not the body.

Here is what actually happens in code:

1. Line 1228: `metadata = payload.pop("metadata", None)` — the `metadata` object (containing `chat_id`) is removed from the body
2. Nothing after that sets `payload["user"]` from `chat_id` for non-pipeline models
3. The body gets `json.dumps(payload)` at line 1320 and sent as the POST data
4. The headers go through a completely separate transport path

So no, you cannot "already forward chat id." The existing mechanisms forward it as an HTTP header (`X-OpenWebUI-Chat-Id`), if you have enabled the opt-in env var. That is useful for backends that read custom headers. But it does not solve the problem of getting `chat_id` into the standard `user` field of the request body — which is what this PR does, and which is what the OpenAI API spec defines as the mechanism for end-user/session identification.

Three lines. No config. No env vars. No breaking changes.

ekinnee:
@tjbck Please review this PR. It was summarily closed by @Classic298 with no actual investigation or evaluation. Reading the attached Issue would have explained why the PR was opened.

Classic298:
1) you never specified neither in your title nor in the PR description that you want to make it be part of the body 

2) you can already achieve it by either using the advanced parameters or a model or a very trivial Filter, as one like this:

```python
class Filter:
    def inlet(self, body: dict, __metadata__: dict = None) -> dict:
        if __metadata__ and __metadata__.get("chat_id"):
            body["user"] = __metadata__["chat_id"]
        return body
```

Filters run before the payload is forwarded and the injected `user` field survives to the upstream request, so this gives you per-chat session continuity today, with no core change.

Per our docs, our stance is firm on this: **if existing functionality or a trivial plugin covers a use case, it doesn't go into core**. 

Even worse: **This PR also unconditionally sets `user` for *every* OpenAI-compatible backend — a silent behavior change that not every upstream wants AND WILL BREAK some upstream providers (we have seen this exact issue in the past where seemingly ""harmless"" PRs like this were merged and suddenly a bunch of users reported that their upstream OpenAI instance stopped working and their models stopped responding**

Furthermore not everyone WANTS to forward chat ids for data protection reasons.

Finally: per OpenAI's spec it's meant to identify the end user for abuse monitoring, not a conversation/session). And for that the parameter you're sending is even the wrong name.

Classic298:
> Three lines. No config. No env vars. No breaking changes.

No config is bad
No env vars is worse
And no breaking changes is a lie.

ekinnee:
> > Three lines. No config. No env vars. No breaking changes.
> 
> No config is bad No env vars is worse And no breaking changes is a lie.

I'll wait in @tjbck to answer here and the email I sent regarding this interaction.

Classic298:
Sure. Do that.

Issue #24905, opened 2026-05-19
issue: Open AI compatible /chat/completions does not work anymore without chat_id


PaulWeinsberg:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

9.5

### Operating System

Any

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

POST api/chat/completions | api/v1/chat/completions

```json
{
    "model": "name-it",
    "messages": [
        {
            "role": "user",
            "content": "Hello world"
        }
    ]
}
```

ERROR : {"detail":"'NoneType' object has no attribute 'startswith'"}.

### Actual Behavior

POST api/chat/completions | api/v1/chat/completions

This works :
```json
{
    "model": "name-it",
    "messages": [
        {
            "role": "user",
            "content": "Hello world"
        }
    ],
   "chat_id": ""
}
```

### Steps to Reproduce

Use my example above

### Logs & Screenshots

<img width="923" height="468" alt="Image" src="https://github.com/user-attachments/assets/976a9e6f-d89d-4260-b821-987fcb23ff95" />

### Additional Information

Will make a PR, just a type check is missing in is_new_chat var I guess

PaulWeinsberg:
Duplicate, solved in master, closed.

Issue #5883, opened 2024-10-03
Make metadata.chat_id available in custom pipeline


weixu365:
**Is your feature request related to a problem? Please describe.**

I'm integrating open-webui with AWS Bedrock Agents using bedrock_agent_runtime.invoke_agent() in a custom pipeline. This method requires a sessionId parameter, as the Bedrock Agent manages the chat session.

I would like to use chat_id as the sessionId, but it's no longer available in the pipeline.pipe() method. The metadata field, where this information was previously stored, was removed 3 months ago in commit [f462744](https://github.com/open-webui/open-webui/commit/f462744#diff-84fe56d58871cf1f81dfe7cc81a247f1f61226cf2a152e95c8ee06c13d83f28fR360).

**Describe the solution you'd like**

I propose restoring the metadata field from the payload at: https://github.com/open-webui/open-webui/blob/main/backend/open_webui/apps/openai/main.py#L383-L384

**Describe alternatives you've considered**

If reintroducing the metadata field could cause other issues, I suggest adding an environment variable, such as `ENABLE_PIPELINE_METADATA`. When this variable is not set or set to false, the default behavior (without metadata) would remain unchanged, while enabling it would send the field to custom pipelines

**Additional context**
None

weixu365:
Solved by adding to a custom field in body:
```
    async def inlet(self, body: dict, user: Optional[dict] = None) -> dict:
        body["pipeline_metadata"] = {"chat_id": body["chat_id"]}

        return body
```

codysandahl:
For future people, if you need to get the extra metadata to the Pipelines server, I wrote a Pipe to package the info into the body and send it to Pipelines.
https://openwebui.com/f/codysandahl/pipeline_enriched_with_metadata

PR #27186, opened 2026-07-18
feat: forward chat_id as user field in OpenAI-compatible chat completions

### LINKED ISSUE #27185: feat: forward chat_id as user field in OpenAI-compatible chat completions
### Description

When Open WebUI proxies to an OpenAI-compatible backend, the `metadata` object (containing `chat_id`) is unconditionally stripped from the request body at `payload.pop('metadata', None)` in `backend/open_webui/routers/openai.py`. The standard OpenAI `user` field is never populated from this data.

Upstream backends that use the `user` field for session identification (such as OpenClaw Gateway, which maps `user` to session keys) receive every message as a new session — no conversation continuity, no memory across turns.

This is *not* the same as the existing `ENABLE_FORWARD_USER_INFO_HEADERS` feature, which forwards metadata via HTTP headers (`X-OpenWebUI-Chat-Id`). The `user` field is a JSON body field in the `/v1/chat/completions` request body — a completely separate transport path. Headers cannot solve this for backends that only read the request body.

### Expected Behavior

Operators should be able to opt in to forwarding the conversation/chat ID as the standard `user` field in outbound OpenAI-compatible requests, allowing upstream backends that use `user` for session tracking to maintain continuity.

### Proposed Solution

As demonstrated in closed PR #27175, a 3-line addition after the metadata pop:

```python
if ENABLE_FORWARD_CHAT_ID and metadata and metadata.get('chat_id') and 'user' not in payload:
    payload['user'] = metadata.get('chat_id')
```

Gated behind `ENABLE_FORWARD_CHAT_ID` (default: `false`) — opt-in only, no silent behavior change, no breaking changes for existing deployments.

### Prior Discussion

- **Closed PR #27175**: Original submission was closed by a contributor on the grounds that it was unconditional and lacked an opt-out mechanism. Those concerns have been addressed:
  - Now gated behind `ENABLE_FORWARD_CHAT_ID` env var (default: `false`)
  - Respects any explicit `user` field already in the request body
  - Logic extracted into a testable helper with unit tests
- **Closed Issue #27174**: Original issue with detailed impact analysis

### Edge Cases Addressed

| Case | Behavior |
|------|----------|
| `ENABLE_FORWARD_CHAT_ID` not set (default) | No injection — zero impact |
| metadata is None / missing / empty | No injection |
| Client explicitly sets `user` in request body | `user` is preserved, chat_id not injected |
| Pipeline models (which set `user` to a user-info dict) | Pipeline overwrites injection — correct behavior |
| Azure OpenAI path | `user` is in `get_azure_allowed_params()` — survives filtering |
| Responses API path | `user` is a valid top-level field — survives conversion |

### Impact

- Zero risk: default-off, no change to existing behavior unless operator opts in
- Tiny code footprint: one import, one helper function, one call site
- Unit-tested: 11 test cases covering all edge cases
- Self-documenting env var with clear comment
- Follows the same `ENABLE_FORWARD_USER_INFO_HEADERS` pattern already in the codebase

### Environment Variables

- `ENABLE_FORWARD_CHAT_ID` (default: `false`) — set to `true` to enable

### Testing

Unit tests are included in `backend/tests/test_openai_forward_chat_id.py`. They monkeypatch the module-level flag and verify the helper function behavior for all edge cases listed above.

### Changelog Entry

```markdown
### Added

- 🌐 **ENABLE_FORWARD_CHAT_ID**: New environment variable that forwards the chat_id from metadata as the standard 'user' field in OpenAI-compatible /v1/chat/completions request bodies. Opt-in, default-off, respects existing 'user' values. Useful for upstream backends that use the user field for session identification.
```

ekinnee:
## Description

When Open WebUI proxies to an OpenAI-compatible backend, the `metadata` object (containing `chat_id`) is unconditionally stripped from the request body. The standard OpenAI `user` field is never populated from this data.

Upstream backends that use the `user` field for session identification (such as OpenClaw Gateway) receive every message as a new session — no conversation continuity, no memory across turns.

This is *not* the same as the existing `ENABLE_FORWARD_USER_INFO_HEADERS` feature, which forwards metadata via HTTP headers. The `user` field is a JSON body field — a completely separate transport path.

### Added

- **ENABLE_FORWARD_CHAT_ID**: New environment variable (default: `false`) that forwards the chat_id from metadata as the standard `user` field in OpenAI-compatible `/v1/chat/completions` request bodies. Opt-in, default-off, respects existing `user` values. Useful for upstream backends that use the `user` field for session identification.

### Edge Cases Addressed

| Case | Behavior |
|------|----------|
| `ENABLE_FORWARD_CHAT_ID` not set (default) | No injection — zero impact |
| metadata is None / missing / empty | No injection |
| Client explicitly sets `user` in request body | `user` is preserved, chat_id not injected |
| Pipeline models (which set `user` to a user-info dict) | Pipeline overwrites injection — correct behavior |
| Azure OpenAI path | `user` is in `get_azure_allowed_params()` — survives filtering |
| Responses API path | `user` is a valid top-level field — survives conversion |

### Testing

Unit tests are included in `backend/tests/test_openai_forward_chat_id.py` covering all edge cases above.

Closes #27185
Relates to #24905, #5883

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -884,6 +884,14 @@ FORWARD_USER_INFO_HEADER_USER_ROLE = os.getenv('FORWARD_USER_INFO_HEADER_USER_RO
 FORWARD_SESSION_INFO_HEADER_MESSAGE_ID = os.getenv('FORWARD_SESSION_INFO_HEADER_MESSAGE_ID', 'X-OpenWebUI-Message-Id')
 FORWARD_SESSION_INFO_HEADER_CHAT_ID = os.getenv('FORWARD_SESSION_INFO_HEADER_CHAT_ID', 'X-OpenWebUI-Chat-Id')
 
+# When enabled, forward the chat_id from metadata as the standard 'user' field in
+# OpenAI-compatible /v1/chat/completions request bodies, unless the client
+# explicitly set 'user' in the original request. This allows upstream backends
+# that use the 'user' field for session identification (such as OpenClaw Gateway)
+# to maintain conversation continuity across turns.
+# Default: False — opt-in, no silent behavior change.
+ENABLE_FORWARD_CHAT_ID = os.getenv('ENABLE_FORWARD_CHAT_ID', 'False').lower() == 'true'
+
 # If set while ENABLE_FORWARD_USER_INFO_HEADERS is True, send one signed HS256 JWT
 # (FORWARD_USER_INFO_HEADER_JWT) instead of separate X-OpenWebUI-User-* headers.
 FORWARD_USER_INFO_HEADER_JWT_SECRET = (os.environ.get('FORWARD_USER_INFO_HEADER_JWT_SECRET') or '').strip() or None
--- a/backend/open_webui/routers/openai.py
+++ b/backend/open_webui/routers/openai.py
@@ -28,6 +28,7 @@ from open_webui.env import (
     AIOHTTP_CLIENT_TIMEOUT,
     AIOHTTP_CLIENT_TIMEOUT_MODEL_LIST,
     BYPASS_MODEL_ACCESS_CONTROL,
+    ENABLE_FORWARD_CHAT_ID,
     ENABLE_FORWARD_USER_INFO_HEADERS,
     ENABLE_OPENAI_API_PASSTHROUGH,
     FORWARD_SESSION_INFO_HEADER_CHAT_ID,
@@ -1197,6 +1198,21 @@ def convert_responses_result(response: dict) -> dict:
     }
 
 
+def _inject_chat_id_as_user(payload: dict, metadata: dict | None) -> None:
+    """
+    Mutate *payload* in place: set ``payload['user']`` to the chat_id from
+    *metadata* when the ``ENABLE_FORWARD_CHAT_ID`` feature flag is active,
+    *metadata* and ``chat_id`` exist, and the caller did not already set
+    ``user`` explicitly.
+
+    This is a pure-ish helper extracted so the forwarding logic can be
+    unit-tested without spinning up a FastAPI test client with database
+    dependencies.
+    """
+    if ENABLE_FORWARD_CHAT_ID and metadata and metadata.get('chat_id') and 'user' not in payload:
+        payload['user'] = metadata.get('chat_id')
+
+
 @router.post('/chat/completions')
 async def generate_chat_completion(
     request: Request,
@@ -1224,6 +1240,8 @@ async def generate_chat_completion(
 
     payload = {**form_data}
     metadata = payload.pop('metadata', None)
+
+    _inject_chat_id_as_user(payload, metadata)
 
     model_id = form_data.get('model')
     model_info = await Models.get_model_by_id(model_id)
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -1,0 +1,13 @@
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def _reset_env_flag():
+    """
+    Reset ENABLE_FORWARD_CHAT_ID to its default (False) before each test
+    so test-to-test leakage of monkeypatched module state is impossible.
+    """
+    import open_webui.routers.openai as openai_router
+
+    openai_router.ENABLE_FORWARD_CHAT_ID = False
+    yield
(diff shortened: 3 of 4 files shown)


Classic298:
As per @tjbck, with the exact same reasoning I gave you already: closing this

https://github.com/open-webui/open-webui/issues/27185#issuecomment-5014614030

It should be handled via a filter function.

Fyi this code change is in principle correct but it's more code than just building a 5 line filter function (which's code I already handed to you in the other PR you opened earlier).

So the maths is off. 5 line single file filter function versus 4 files and new env var core change to open webui to do the same.

The filter plugin is clearly superior here and the preferred solution in all scenarios.

For future reference please do not include test files in PRs those are not wanted.

Classic298:
@ekinnee 

Consolidating every reason for the closes across #27174 / #27175 / #27185 / #27186 in one place, so this is on record once. This is final.

**Policy — this alone is dispositive**

1. Project stance, confirmed by @tjbck: anything that mutates the outbound payload beyond the core supported fields (model, messages, …) is Filter territory. If existing functionality or a trivial filter covers a use case, it does not go into core. Our stance is firm on this. Everything below is additional, not necessary.

**Why #27175 (unconditional) was unacceptable**

2. It unconditionally injected `user` into every request to every OpenAI-compatible upstream. Silent behavior change for all existing deployments.
3. We have seen exactly this class of "harmless" body change break upstreams before — providers erroring, models silently stopping to respond. Body fields are breaking in practice. Headers are typically tolerated; unknown/unexpected body fields are not.
4. Validation was manual inspection of payloads against a single backend. That is not adequate testing for a change shipped unconditionally to an install base of 400M+.

**Why #27186 (env-var gated) is still the wrong design**

5. The flag is global. It cannot be toggled per connection. Enable it for one upstream and every other configured OpenAI-compatible connection also receives `user=chat_id`.
6. The "respects existing `user`" guard is dead code in practice: the frontend never sets `user`, so the condition always passes and the flag effectively means "inject everywhere."
7. Any deployment with multiple upstreams where only one wants this would need a filter to strip the field from the other connections. The remediation for your feature is the exact filter it was written to avoid. Inject-everywhere-then-strip is strictly worse than inject-where-wanted.
8. Failure asymmetry: forgetting an additive filter fails loud and local (one gateway loses session continuity). Forgetting removal filters fails silent and global (an identity field polluted across every upstream). We do not ship designs that fail silent and global.
9. Your own edge-case table covers every case except the operationally central one: heterogeneous upstreams where only one provider wants the field. That is the case that matters, and it is the one this design cannot handle.
10. Doing this "properly" in core would mean per-connection config — UI, schema, migration, far more than 4 files — to encode one third-party gateway's session convention. Not happening.

**Semantics**

11. `user` is defined for end-user identification for abuse monitoring. It is not a session or conversation field. **A chat_id is the wrong data in that field.**
12. OpenAI has deprecated `user` (superseded by `safety_identifier` / `prompt_cache_key`). You are proposing core code **built on a deprecated parameter that no major backend documents as a session mechanism.**
13. For providers that actually read `user`, injecting chat_ids corrupts their abuse attribution — every conversation looks like a distinct end user.
14. Data protection: many operators explicitly do not want chat identifiers forwarded upstream. Per-model filters respect that naturally; core code paths and global flags do not.

**Cost/benefit**

15. Your change: 4 files, a new env var, docs, and permanent maintenance on us. The alternative: a 5-line single-file filter, one click to install, per-model scoped, zero core burden. The maths is off.
16. Env var proliferation is itself a cost. Every flag is config surface we support forever.
17. Test files are not wanted in PRs, per our contribution conventions.
18. Chat_id forwarding already exists for header-reading backends (`ENABLE_FORWARD_USER_INFO_HEADERS` → `X-OpenWebUI-Chat-Id`). **Body injection for the remainder is precisely what Filters exist for.**
19. If OpenClaw needs `user=chat_id`, OpenClaw can ship the filter below in its own integration docs. A vendor-specific convention is the vendor's burden, not our core's.

**The solution, again**

```python
class Filter:
    def inlet(self, body: dict, __metadata__: dict = None) -> dict:
        if __metadata__ and __metadata__.get("chat_id"):
            body["user"] = __metadata__["chat_id"]
        return body
```

Attach it to the model(s) pointing at your gateway. Per-connection scoping solved, today, no core change.

All four threads are closed. Further resubmissions of the same change will be closed and locked.

ekinnee:
@Classic298 you seem mad again.

PR #27211, opened 2026-07-20
perf: use orjson instead of json for 17x CPU performance improvement on streaming

Classic298:
Replaces stdlib `json` with `orjson` on the backend's three hottest serialization paths. With `WEBSOCKET_MANAGER=redis`, every emit is JSON-encoded onto the pub/sub channel and decoded by **every** replica; profiling showed stdlib JSON as the dominant CPU cost on WebSocket workers. All swaps fall back to the previous stdlib codec for anything orjson rejects, so wire/stored formats are unchanged — old and new replicas cross-decode during rolling deploys, and existing Redis entries still parse.

### Changed

- **`socket/utils.py`**: new `ORJSONCodec` (orjson with stdlib fallback), passed via the supported `json=` hooks to `AsyncServer`/`AsyncRedisManager` and used by `RedisDict` (session/usage/models pools) and `YdocManager`. 18 KB pub/sub emit: encode 41→2.4 µs (17×), decode 27→8.5 µs (3×).
- **`utils/json_response.py`**: `JSONResponse.render` now uses orjson. Deliberately **not** `default_response_class=ORJSONResponse` — that's deprecated in FastAPI 0.136 and setting any explicit default disables the Pydantic Rust fast path on `response_model` routes (measured 1.2× regression there). The override keeps that fast path and also covers explicit `JSONResponse(...)` returns. Render of a 44 KB payload: 357→23 µs (15.5×).
- **Deps**: `orjson==3.11.9` added to `requirements.txt`, `requirements-min.txt`, `pyproject.toml`.

### Notes

- Only observable behavior change: `NaN`/`Infinity` floats serialize as `null` instead of erroring — matching what Pydantic already does on `response_model` routes. engineio's oversized-integer guard for untrusted payloads is preserved via the fallback.
- Verified against pinned `python-socketio==5.16.2` / `fastapi==0.136.3`: packet round-trips incl. binary attachments, stdlib↔orjson cross-decode both ways, byte-identical HTTP bodies across route types, `RedisDict`/`YdocManager` round-trips under fakeredis.

> [!NOTE]


gaby:
@Classic298 Awesome find! ⚡💪

gaby:
@Classic298 merge conflicts

tjbck:
I don't trust this, after full compatibility check for all possible permutations I'll revisit.

gaby:
@tjbck The orjson library is RFC 8259 compliant. The library is used by over 200k projects on GitHub.

See https://github.com/ijl/orjson?tab=readme-ov-file#questions

tjbck:
I don't doubt it's a widely used library, but can this pr gurantee it will work across ALL possible permutations of open webui deployments? I'd be more comfortable merging if it became an optional opt-in feature.

gaby:
Agree, making it optional with an ENV could work.

PR #27224, opened 2026-07-20
perf: column-only selects for user activity and webhook lookups

Classic298:
is_user_active and get_user_webhook_url_by_id each fetched the full User row to read a single value, and they run back-to-back on the post-completion webhook-notification path for every finished chat. The full row includes profile_image_url, which is frequently a base64 data URI — potentially hundreds of KB per fetch.

Select only last_active_at / settings respectively. Truthiness and missing-user semantics are preserved exactly (absent row -> False / None, NULL column -> False / None).

> [!NOTE]


tjbck:
Addressed!

PR #27227, opened 2026-07-20
perf: batch API-key auth config reads

Classic298:
get_current_user_by_api_key issued three sequential single-key config SELECTs (user.permissions, auth.api_key.endpoint_restrictions, auth.api_key.allowed_endpoints) on every API-key-authenticated request. Batch them into one Config.get_many round trip; get_many applies the same defaults as the individual gets, so values are identical.

Also removes a dead local `import asyncio` that executed on every JWT-authenticated request (asyncio is already imported at module level).

> [!NOTE]


tjbck:
Addressed.

PR #27228, opened 2026-07-20
perf: cut per-chunk and per-request DB waste in the filter pipeline

Classic298:
Stream filters run once per streamed chunk, and the pipeline around them repeated significant database work on that path:

- get_function_valves_by_id fetched the full function row — including the plugin's entire Python source in `content` — to read only the `valves` JSON column, once per streamed chunk for stream filters. Select just the valves column instead.
- get_user_valves_by_id_and_user_id fetched and validated the full user row (including the profile image) to read one nested settings key. Select just the settings column; validation of that column is unchanged, so results are identical.
- process_filter_functions issued one valves query per filter per invocation. Batch them with the existing get_function_valves_by_ids, lazily, so invocations where no filter defines valves issue no query at all (previously zero, now still zero).
- process_filter_functions recomputed inspect.signature() and iscoroutinefunction() for every handler on every invocation — per chunk for stream filters. Cache both per handler, weakly keyed on the underlying function object so reloaded plugins get fresh entries.
- get_sorted_filter_ids ran two full-row queries (global filters, active filters) and used only the ids; global filters are by definition a subset of active ones. One narrow (id, is_global) query now serves both sets, and per-filter priority valves lookups are batched (N+1 -> 1).
- The streaming response handler fetched filter functions one query per id; use the existing batched get_functions_by_ids (N -> 1).

Final filter ordering is unchanged (it is fully determined by the (priority, id) sort). Error semantics preserved: batch lookup misses behave exactly like the previous per-id fetch failures.

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/functions.py
+++ b/backend/open_webui/models/functions.py
@@ -7,7 +7,7 @@ import time
 
 # local imports
 from open_webui.internal.db import Base, JSONField, get_async_db_context
-from open_webui.models.users import UserResponse, Users
+from open_webui.models.users import User, UserResponse, Users, UserSettings
 from open_webui.utils.valves import decrypt_valves, encrypt_valves
 from pydantic import BaseModel, ConfigDict
 from sqlalchemy import BigInteger, Boolean, Column, Index, String, Text, delete, select, update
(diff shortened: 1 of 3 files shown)


tjbck:
Addressed in dev.

PR #27231, opened 2026-07-20
perf: linear content accumulation in the streaming response handler

Classic298:
The streaming handler rebuilt the full accumulated response with `content = f'{content}{value}'` on every content delta — a complete string copy per chunk, making accumulation O(n^2) over the response length. Use in-place `content += value` for the (universal) str case, which CPython extends in place, keeping accumulation O(n); the f-string fallback is preserved for non-str values so coercion behavior is unchanged.

In the ENABLE_REALTIME_CHAT_SAVE branch, full_output() — which concatenates the entire accumulated output — was called twice per chunk (once for the DB upsert, once for the emitted delta). Compute it once and reuse.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -4408,7 +4408,12 @@ async def streaming_chat_response_handler(response, ctx):
                                                 user,
                                             )
 
-                                        content = f'{content}{value}'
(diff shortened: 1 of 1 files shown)


tjbck:
image

Issue #26834, opened 2026-07-06
feat: Add group information to ENABLE_FORWARD_USER_INFO_HEADERS


gfaugere-chut:
### Check Existing Issues

- [x] I have searched all existing **open AND closed** issues and discussions and found none comparable to my request.

### Verify Feature Scope

- [x] I believe this feature request is appropriately scoped for the Issues section as described above.

### Problem Description

ENABLE_FORWARD_USER_INFO_HEADERS (and associated settings) give valuable and trustworthy information about the logged in user

However, when calling downstream APIs that might want to act differently depending on the user, the Open WebUi "role" might not be enough

### Proposed Solution

Users have goups! Adding them to the generated headers/token would be really helpful :)

### Alternatives Considered

Email/name are present but might not be a unique ID (_yes, even emails_)

The given ID is issued by Open WebUi itself, and uniquely identifies users but doesn't help in matching the user with an other auth system to retrieve fine-grained groups

Classic298:
Thanks for the request! The role part of this is already supported today.

When `ENABLE_FORWARD_USER_INFO_HEADERS=true`, Open WebUI forwards the user's role to downstream backends in all of the existing forwarding paths:

- **Plain headers:** `X-OpenWebUI-User-Role` is sent alongside the name/id/email headers (the header name is configurable via `FORWARD_USER_INFO_HEADER_USER_ROLE`).
- **Signed JWT:** if you set `FORWARD_USER_INFO_HEADER_JWT_SECRET`, the minted token carries a `role` claim (along with `sub`, `email`, `name`).
- **Custom outgoing headers:** you can template `{{USER_ROLE}}` into any custom header value.

So the headline ask (getting the user's role into the generated headers/token) needs no code change, just the flag enabled.

The one thing that isn't covered is your note about fine-grained **group** membership for matching users against an external auth system. That's a separate piece: `role` here is the single Open WebUI system role (`admin`/`user`/`pending`), whereas group membership lives in a separate model. If group forwarding is what you need, that would be a distinct feature request, worth stating explicitly so it can be scoped on its own.

Closing this as the role forwarding already exists, but happy to keep the group discussion going if that's the actual need.

gfaugere-chut:
Hey, I've edited my message, I typed "role" when I meant "group" indeed

This issue, to me, is about allowing downstream to act upon the user identity in a fine-grained way - roles are not enough in my use case

[...]

PR #27236, opened 2026-07-20
feat: add {{USER_GROUPS}} and {{USER_GROUP_IDS}} placeholders for custom forwarded headers

Classic298:
Custom per-connection headers can now forward the user's groups to upstream backends via two new template placeholders:

- {{USER_GROUPS}}: comma-separated group names
- {{USER_GROUP_IDS}}: comma-separated group ids

The group lookup is async, so get_custom_headers becomes an async wrapper around the sync template substitution (parse_custom_headers) and fetches groups lazily — only when a header value actually references a groups placeholder. The external document loader path runs in a worker thread without an event loop, so Loader.aload prefetches the groups before offloading and passes them through to ExternalDocumentLoader.

https://github.com/open-webui/open-webui/issues/26834

> [!NOTE]


tjbck:
qq: why couldn't the consumer side just make an extra request to retrieve full user info?

Classic298:
@tjbck maybe it cant as its e.g. litellm only receiving and logging the groups to do group based spend logs

gfaugere-chut:
And the existing headers might not be enough to fetch the full user info: emails are imperfect and might not be unique identifiers, and AFAIK the user ID that is forwarded is open-webui's ID, not the original external system user ID

PR #27247, opened 2026-07-20
fix: mutable default argument {} in generate_function_chat_completion

yangyang-qaq:
## What

Fix a mutable default argument in `generate_function_chat_completion` — `models: dict = {}` is evaluated once at definition time, sharing the same dict object across all calls.

## Root Cause

Python evaluates default arguments at function definition, not at call time. A mutable default like `{}` means every call without an explicit `models` argument shares the same dictionary, potentially leaking state between independent calls.

## Fix

Replace with the standard None-sentinel pattern:

```diff
-async def generate_function_chat_completion(request, form_data, user, models: dict = {}):
+async def generate_function_chat_completion(request, form_data, user, models: dict | None = None):
+    if models is None:
+        models = {}
```

Ref: [Python docs on default argument values](https://docs.python.org/3/tutorial/controlflow.html#default-argument-values)

## Checklist

- [x] **Linked Issue/Discussion:** N/A (self-contained defensive fix)
- [x] **Target branch:** dev
- [x] **Description:** provided above
- [ ] **Changelog:** N/A (minor fix)
- [ ] **Documentation:** N/A
- [ ] **Dependencies:** none
- [x] **Testing:** Standard Python pattern; behavior is identical when `models` is passed explicitly
- [x] **No Unchecked AI Code:** Author reviewed the diff
- [x] **Self-Review:** done
- [x] **Git Hygiene:** 1 commit, rebased on dev

### DIFF
--- a/backend/open_webui/functions.py
+++ b/backend/open_webui/functions.py
@@ -147,7 +147,10 @@ async def get_function_models(request):
     return pipe_models
 
 
-async def generate_function_chat_completion(request, form_data, user, models: dict = {}):
+async def generate_function_chat_completion(request, form_data, user, models: dict | None = None):
+    if models is None:
+        models = {}
+
     async def execute_pipe(pipe, params):
         if inspect.iscoroutinefunction(pipe):
             return await pipe(**params)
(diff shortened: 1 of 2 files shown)


yangyang-qaq:
Closes #27248 — this is a standard Python defensive fix for mutable default arguments. Python evaluates `{}` once at definition time, so every call shares the same dict. The None-sentinel pattern creates a fresh dict per call.

PR #27282, opened 2026-07-21
perf: 40% LESS CPU usage: cut per-instance CPU cost of shared socket.io Redis pub/sub channel

Classic298:
Profiling a multi-instance deployment (py-spy --gil) showed ~44% of worker CPU in the socket.io pub/sub listener. Two causes, two fixes:

- Add hiredis so redis-py parses the RESP protocol in C instead of pure Python (redis/_parsers/resp3.py alone accounted for ~28% of GIL samples; redis-py auto-selects the hiredis parser when importable).

- Subclass AsyncRedisManager to drop emits whose target room has no local participants before upstream _handle_emit re-encodes the full packet. Every instance receives every emit published on the shared channel, so with N instances all but the hosting one were paying full packet re-serialization per message just to deliver it to nobody. Broadcasts (room=None) are unaffected.

> [!NOTE]

### DIFF
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -30,7 +30,7 @@ psycopg[binary]==3.3.4
 alembic==1.18.4
 
 pycrdt==0.13.1
-redis==8.0.0
+redis==8.0.1
 hiredis==3.4.0
 
 APScheduler==3.11.2
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -31,6 +31,7 @@ alembic==1.18.4
 
 pycrdt==0.13.1
 redis==8.0.0
+hiredis==3.4.0
 
 APScheduler==3.11.2
 RestrictedPython==8.2
--- a/backend/open_webui/socket/main.py
+++ b/backend/open_webui/socket/main.py
@@ -57,6 +57,27 @@ REDIS = None
 # Configure CORS for Socket.IO
 SOCKETIO_CORS_ORIGINS = '*' if CORS_ALLOW_ORIGIN == ['*'] else CORS_ALLOW_ORIGIN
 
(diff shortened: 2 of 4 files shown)


backend/requirements.txt
```diff
@@ -31,6 +31,7 @@ alembic==1.18.4
 
 pycrdt==0.13.1
 redis==8.0.0
+hiredis==3.4.0
```
gaby, line 34: @Classic298 According to their Github the version is 1.4.0

https://github.com/redis/hiredis/releases

backend/requirements.txt
```diff
@@ -31,6 +31,7 @@ alembic==1.18.4
 
 pycrdt==0.13.1
 redis==8.0.0
```
gaby: This can be bumped to 8.0.1

https://github.com/redis/redis-py/releases/tag/v8.0.1

backend/requirements.txt
Classic298: @gaby that is the github releases. go check pypi: https://pypi.org/project/hiredis/#history pypi it is 3.4.0 github releases do not match the packet version. checking on github is the wrong place to check

backend/requirements.txt
Classic298: image

gaby: It literally has a fix for hiredis for high file descriptors cases. But sure.

backend/requirements.txt
gaby: Different package, the one i linked was the actual C code. Makes sense.

backend/requirements.txt
Classic298: Hm Okay. In that case i guess it can be bumped. Usually we avoid bumping unless necessary. But since this has a fix

tjbck:
🚀

PR #27288, opened 2026-07-21
perf: cache the per-request enable_api_keys config read

Classic298:
AuthTokenMiddleware ran an uncached Config.get('auth.enable_api_keys') on every HTTP request — one DB round-trip per request, fleet-wide. (The pre-refactor code read the in-memory app.state.config attribute; the per-key config refactor traded that for an always-fresh DB read.)

Cache the flag per process with bounded staleness: a 15s TTL, plus a Config.GENERATION counter bumped by upsert/delete/clear so config writes in the same process invalidate on the very next request. Writes from other instances become visible within the TTL — a middle ground between the old process-lifetime staleness and the current per-request read. The generation is captured before the awaited fetch, so a write landing mid-read forces a refetch on the next request rather than caching a value that may predate the write. Fetch errors propagate per request and are never cached.

Per-credential security checks are unaffected: API key lookup, user permissions, and endpoint restrictions all remain live per-request reads; only the global enable toggle carries the bounded staleness.

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/config.py
+++ b/backend/open_webui/models/config.py
@@ -106,6 +106,9 @@ class Config(Base):
     DEFAULTS: ClassVar[dict[str, Any]] = {}
     PERSISTENT_ENABLED: ClassVar[bool] = True
     OAUTH_PERSISTENT_ENABLED: ClassVar[bool] = False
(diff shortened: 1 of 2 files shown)


tjbck:
Addressed!

PR #27297, opened 2026-07-21
fix: honor the webhooks permission when saving user notification settings

Classic298:
The user settings update route strips `ui.toolServers` for non-admin users who lack the `features.direct_tool_servers` permission, but applies no equivalent check for `ui.notifications.webhook_url` against `features.webhooks`. The notification webhook URL is therefore persisted for users whose webhooks permission is disabled, so the backend does not reflect the configured permission the way the interface does.

Strip `ui.notifications.webhook_url` for non-admin users who lack `features.webhooks`, mirroring the existing tool-servers handling, so a disabled webhooks permission is honored server-side and not only in the UI.

> [!NOTE]


tjbck:
Addressed!

PR #27366, opened 2026-07-23
perf: skip chat event emits to rooms with no local listeners

Classic298:
Chat generations keep running when nobody is listening: the user closes the tab mid-stream and the background task continues, API completions have no socket client at all, and internal chats plus title/tag/follow-up task events emit to a user who may not be connected. In local websocket-manager mode (the default), python-socketio's AsyncManager.emit encodes the full packet before iterating participants whenever the namespace has any connection, so every one of those events is fully JSON-encoded and then dropped. Because each streamed event carries the full accumulated content, that wasted encoding grows quadratically over a stream.

The local manager's rooms mapping is authoritative and empty rooms are pruned on leave, so a dict containment check is an exact "someone is listening" test. Skip only the emit when the user room is absent; the DB persistence branches below still run unchanged. Redis mode is untouched: the send side must still publish for other instances, and the receive side is already filtered by LocalFilteredRedisManager.

Benchmark (python-socketio 5.16.1, one other user connected so the namespace is live, target room empty):

| scenario | before | after |
| --- | --- | --- |
| per event, 50 KB accumulated content | 45.5 us | 0.024 us | | 500-chunk stream growing to 100 KB | 23.1 ms | 0.012 ms |

Functionally verified: events are still delivered once a session joins the room, skipped again after the room empties, and the Redis-mode helper never claims a room is empty.

> [!NOTE]

### DIFF
--- a/backend/open_webui/socket/main.py
+++ b/backend/open_webui/socket/main.py
@@ -117,6 +117,15 @@ else:
 
 # Timeout duration in seconds
 TIMEOUT_DURATION = 3
+
+_LOCAL_MANAGER = WEBSOCKET_MANAGER != 'redis'
+
+
+def room_is_known_empty(room: str) -> bool:
+    """Exact only in local manager mode: rooms is authoritative there and empty
+    rooms are pruned on leave. In Redis mode another instance may host the room,
+    so never claim empty."""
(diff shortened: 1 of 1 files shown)


tjbck:
Addressed.

PR #27370, opened 2026-07-23
fix: stop empty audit path entries from silently disabling all auditing

Classic298:
Setting AUDIT_EXCLUDED_PATHS to an empty string (the natural way to say "exclude nothing") produced [''], and a bare '/' entry survived parsing as '' too. The blacklist regex then became ^/api(?:/v1)?/()\b, whose empty alternation matches every API path, so auditing silently recorded nothing except the always-logged auth endpoints. The same match-everything pattern was also built when the exclusion list was genuinely empty, since the pattern was compiled unconditionally.

Empty entries are now dropped after normalization (for both AUDIT_EXCLUDED_PATHS and AUDIT_INCLUDED_PATHS, which had the same '/' edge), and blacklist matching is skipped entirely when the exclusion list is empty, which is the correct semantics: nothing is excluded.

Verified: with AUDIT_EXCLUDED_PATHS='' normal API requests are audited again and unauthenticated requests remain skipped; with '/,chats' only chats is excluded; the default '/chats,/chat,/folders' behavior is unchanged, including the always-audited auth endpoints.

> [!NOTE]

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -1094,14 +1094,14 @@ except ValueError:
 
 # Comma separated list for urls to exclude from audit
 AUDIT_EXCLUDED_PATHS = os.getenv('AUDIT_EXCLUDED_PATHS', '/chats,/chat,/folders').split(',')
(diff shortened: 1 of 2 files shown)


tjbck:
Addressed in dev.

PR #27391, opened 2026-07-23
perf: skip API stream accumulation when no outlet consumer exists

Classic298:
With ENABLE_API_OUTLET_FILTERS on (the default), the API streaming fallback re-parsed every SSE chunk with json.loads and accumulated the full assistant message (content concatenation plus structured-output bookkeeping) solely so the end-of-stream outlet pass could run. On installations with no outlet-capable filter functions and no pipeline filters (the common case), that outlet pass is a guaranteed no-op, making the per-chunk work pure waste, including an O(n^2) content concatenation over the stream.

The wrapper now determines once per stream whether an outlet consumer exists: a pipeline model, a pipeline filter targeting the model (via the same get_sorted_filters the outlet pass itself uses) or any filter function whose cached module defines an outlet handler. Only then does it accumulate and invoke the outlet pass. Any error during detection falls back to accumulating exactly as before, and skipping the empty outlet invocation also skips its end-of-stream model-map scan.

Benchmark:

| metric | before | after |
| --- | --- | --- |
| accumulation per 2000-chunk API stream, no outlet consumers | 3.2 ms + json.loads per chunk | 0 (skipped) | | outlet filters or pipelines present | accumulates | accumulates as before |

The before column grows quadratically with response length because each chunk re-concatenates the accumulated content.

Functionally verified: the accumulator still assembles content correctly from raw SSE lines for the gated-on case, and the pipeline filter detector returns matches and empties exactly as the outlet pass sees them.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -57,6 +57,7 @@ from open_webui.routers.images import (
     image_generations,
 )
 from open_webui.routers.pipelines import (
+    get_sorted_filters,
     process_pipeline_inlet_filter,
     process_pipeline_outlet_filter,
 )
@@ -88,6 +89,7 @@ from open_webui.utils.files import (
     get_image_url_from_base64,
 )
 from open_webui.utils.filter import (
+    get_function_module,
     get_sorted_filter_ids,
     process_filter_functions,
 )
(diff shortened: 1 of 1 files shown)


tjbck:
Addressed in dev.

PR #27393, opened 2026-07-23
perf: MAJOR drop redundant work from the realtime socket path

Classic298:
Several socket handlers repeated work per event or per timer tick that either had no effect or could be answered with an O(1) lookup:

- The usage-pool reaper rewrote every entry on every 3-second tick even when nothing expired, which in Redis mode is a serialize plus HSET per active model, forever. Entries are now written back only when a session actually timed out. The send_usage flag it maintained was dead code and is removed.
- Heartbeats round-tripped the full session record (which carries the user's settings) and committed an unconditional last-active UPDATE per beat per tab. Clients beat every 30 seconds and the reaper tolerates 120; duplicate beats arriving within 15 seconds are now skipped entirely.
- The typing indicator, collaborative-document update and awareness handlers verified room membership by materializing the room's full participant list (or, for awareness, scanning every room the session is in) per keystroke or cursor event. A shared helper answers the same local-instance-only question with a dict containment check. The typing payload also built and dumped a pydantic model per keystroke for a three-key dict; it is a dict literal now.
- user-join re-ran the user fetch, serialization and user-room join that connect had just performed for the same socket; when the session pool already holds the same user for the sid, it now refreshes the heartbeat stamp and proceeds straight to the channel joins, which still run unconditionally.
- LocalFilteredRedisManager tested room emptiness by copying the room's participant bidict for every pub/sub message received from the fleet; it now uses containment on the rooms mapping, which is exact because empty rooms are pruned on leave.

Benchmark:

| metric | before | after |
| --- | --- | --- |
| room membership check, 2000-member room (per typing/yjs event) | 47.9 us | 0.14 us | | typing event user payload | 1.20 us | 0.05 us |
| usage-pool writes per 3s tick, nothing expired | one per active model | 0 | | heartbeat DB writes and record rewrites | every beat | only after 15s elapsed |

Functionally verified: membership checks agree with the participant scan for members, non-members and unknown rooms; fresh heartbeats skip and stale ones update both the pool and the DB; typing events from members emit the identical payload shape while non-members are dropped; the user-join dedupe path returns the same response and still joins channels; the fleet-emit filter forwards populated rooms and broadcasts while dropping empty ones.

> [!NOTE]

### DIFF
--- a/backend/open_webui/socket/main.py
+++ b/backend/open_webui/socket/main.py
@@ -76,7 +76,8 @@ class LocalFilteredRedisManager(socketio.AsyncRedisManager):
         room = message.get('room')
         if isinstance(room, str):
             namespace = message.get('namespace') or '/'
-            if next(self.get_participants(namespace, room), None) is None:
+            # O(1) containment; get_participants copies the room dict per emit
+            if room not in self.rooms.get(namespace, {}):
                 return
         await super()._handle_emit(message)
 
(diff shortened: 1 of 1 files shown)


tjbck:
Addressed!

PR #27395, opened 2026-07-23
perf: assorted small proxy-router fixes

Classic298:
Four contained fixes in the ollama and openai routers:

- aiohttp.ClientTimeout objects were constructed per request at every send site even though the two configured timeouts are process constants; both routers now share two module-level instances (ClientTimeout is immutable). The two custom one-off timeouts in the ollama upload and download endpoints are untouched.
- The chat and completion endpoints copied every payload twice: model_dump already returns a fresh dict, so the extra dict spread around it was a full shallow copy of the message list for nothing; and the two v1 endpoints that already exclude metadata in the dump also popped it afterwards. The dumps are used directly now and the dead pops are gone; the endpoints that dump without exclude keep their pop.
- The openai model-list merge recomputed urlparse(base_url).hostname and the per-connection api-config lookup for every model of every connection; with a 400-model provider that is hundreds of identical parses per cache refresh. Both are hoisted to once per connection.
- The unsupported-OpenAI-model keyword list was rebuilt as a list literal per model in two places; it is one module-level tuple now.

Benchmark:

| metric | before | after |
| --- | --- | --- |
| ClientTimeout per request | 0.35 us | 0.03 us (shared) | | payload spread copy per chat request | one shallow copy of all messages | none | | urlparse in list merge, 400-model provider | 341 us per refresh | once per connection |

Functionally verified: the shared timeout instances carry the configured totals; keyword filtering matches the old inline lists on supported and unsupported ids; model_dump yields independent fresh dicts so dropping the spread cannot alias the form.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/ollama.py
+++ b/backend/open_webui/routers/ollama.py
@@ -55,6 +55,10 @@ log = logging.getLogger(__name__)
 # in ZlibError.  See https://github.com/aio-libs/aiohttp/issues/4462.
 _STRIP_PROXY_HEADERS = frozenset({'Content-Encoding', 'Content-Length', 'Transfer-Encoding'})
 
+# ClientTimeout is immutable; build the shared instances once
+_CLIENT_TIMEOUT = aiohttp.ClientTimeout(total=AIOHTTP_CLIENT_TIMEOUT)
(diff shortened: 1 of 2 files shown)


tjbck:
Addressed.

PR #20321, opened 2026-01-02
fix: check requirements via importlib before shelling out to pip

fractuscontext:
# Pull Request Checklist

### Note to first-time contributors: Please open a discussion post in [Discussions](https://github.com/open-webui/open-webui/discussions) to discuss your idea/fix with the community before creating a pull request, and describe your changes before submitting a pull request.

This is to ensure large feature PRs are discussed with the community first, before starting work on it. If the community does not want this feature or it is not relevant for Open WebUI as a project, it can be identified in the discussion before working on the feature and submitting the PR.

**Before submitting, make sure you've checked the following:**

- [x] **Target branch:** Verify that the pull request targets the `dev` branch. **Not targeting the `dev` branch will lead to immediate closure of the PR.**
- [x] **Description:** Provide a concise description of the changes made in this pull request down below.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [x] **Documentation:** If necessary, update relevant documentation [Open WebUI Docs](https://github.com/open-webui/docs) like environment variables, the tutorials, or other documentation sources.
- [x] **Dependencies:** Are there any new dependencies? Have you updated the dependency versions in the documentation?
- [x] **Testing:** Perform manual tests to **verify the implemented fix/feature works as intended AND does not break any other functionality**. Take this as an opportunity to **make screenshots of the feature/fix and include it in the PR description**.
- [x] **Agentic AI Code:** Confirm this Pull Request is **not written by any AI Agent** or has at least **gone through additional human review AND manual testing**. If any AI Agent is the co-author of this PR, it may lead to immediate closure of the PR.
- [x] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [x] **Title Prefix:** To clearly categorize this pull request, prefix the pull request title using one of the following:
  - **BREAKING CHANGE**: Significant changes that may affect compatibility
  - **build**: Changes that affect the build system or external dependencies
  - **ci**: Changes to our continuous integration processes or workflows
  - **chore**: Refactor, cleanup, or other non-functional code changes
  - **docs**: Documentation update or addition
  - **feat**: Introduces a new feature or enhancement to the codebase
  - **fix**: Bug fix or error correction
  - **i18n**: Internationalization or localization changes
  - **perf**: Performance improvement
  - **refactor**: Code restructuring for better maintainability, readability, or scalability
[...]

PR #27419, opened 2026-07-24
fix: fall back to uv when pip is missing for requirements install

### LINKED ISSUE #27214: bug: Function/Pipe requirements install fails with 'No module named pip' on uv tool install (dupe root cause of #12858, #20320)
### Check Existing Issues

- [x] I have searched for any existing and/or related issues (found #12858 and #20320 — both closed, neither actually fixes this for stable).

### Installation Method

`uv tool install open-webui`

### Open WebUI Version

v0.10.2

### Operating System

Debian 13 (trixie), LXC container

### Bug Description

Any Function/Pipe/Tool with a `requirements:` frontmatter line triggers `install_frontmatter_requirements`, which shells out to `<venv>/bin/python3 -m pip install <req>`. When Open WebUI is installed via `uv tool install open-webui` (the officially documented non-Docker install path), the created venv does **not** include `pip` — `uv tool` venvs omit it by design for speed. So the install command always fails:

```
/root/.local/share/uv/tools/open-webui/bin/python3: No module named pip
```

...and the chat request that triggered it fails with:

```
ERROR | open_webui.main:process_chat:1512 - Error processing chat payload: Command '[".../bin/python3", "-m", "pip", "install", "honcho-ai>=2.1"]' returned non-zero exit status 1.
```

This repeats on **every** chat turn that hits a Function/Pipe declaring `requirements`, since there's no persistent success state — it just keeps trying and failing.

Related closed issues (#12858, #20320) cover the same symptom but were closed pointing at `OFFLINE_MODE` (dev branch) as the fix. `OFFLINE_MODE` only *skips* the install — it doesn't address the actual gap (pip missing in a `uv tool`-managed venv), so anyone who actually needs a Function's declared requirement installed (not just wants to suppress the crash) still has no working path in a `uv tool install` deployment.

### Reproduction

1. `uv tool install open-webui`
2. Add/enable a Function or Pipe with frontmatter `requirements: <any-pypi-package>`
3. Trigger a chat that uses it
4. Observe repeated `No module named pip` in the service log, request fails

### Workaround (manual, not a fix)

```
uv pip install --python <path-to-open-webui-venv>/bin/python3 pip
```

This bootstraps `pip` into the tool venv so `install_frontmatter_requirements` succeeds afterward. Should not be required for a documented install method to work out of the box.

### Suggested Fix

Either:
- Bundle `pip` in the venv Open WebUI creates/uses for `uv tool install`, or
- Have `install_frontmatter_requirements` use `uv pip install --python <interpreter> <req>` (or fall back to it) instead of assuming `python -m pip` is available, or
- At minimum, detect the missing-pip case and surface an actionable error instead of retrying identically forever.

---
*Issue filed with AI assistance (Claude Code) on behalf of the reporter, based on a live reproduction on the reporter's system.*

Classic298:
Functions, tools and pipes that declare a requirements frontmatter line are installed through install_frontmatter_requirements, which always shells out to "sys.executable -m pip install". When Open WebUI is installed with "uv tool install open-webui", the created venv deliberately ships without pip, so every such install fails with "No module named pip" and the chat request that triggered it errors out on every turn.

The fix selects the installer at the single chokepoint that all install paths go through (per-function load, per-tool load and startup dependency installation). If the running interpreter has pip, the existing pip command is used unchanged. Otherwise, if uv is on PATH, the packages are installed with "uv pip install --python" pointing at the running interpreter, which accepts the same index options passed via PIP_PACKAGE_INDEX_OPTIONS. If neither is available, an actionable RuntimeError names the missing pip module and the workaround instead of an opaque non-zero exit status.

Fixes #27214

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/plugin.py
+++ b/backend/open_webui/utils/plugin.py
@@ -4,6 +4,7 @@ import asyncio
 import logging
 import os
 import re
+import shutil
 import subprocess
 import sys
 import tempfile
@@ -438,9 +439,17 @@ def install_frontmatter_requirements(requirements: str):
                 return
 
             log.info(f'Installing requirements: {" ".join(new_reqs)}')
-            subprocess.check_call(
-                [sys.executable, '-m', 'pip', 'install'] + PIP_OPTIONS + new_reqs + PIP_PACKAGE_INDEX_OPTIONS
-            )
+            if util.find_spec('pip'):
+                install_command = [sys.executable, '-m', 'pip', 'install']
+            elif uv_path := shutil.which('uv'):
+                # uv-managed environments (e.g. uv tool install) ship without pip
+                install_command = [uv_path, 'pip', 'install', '--python', sys.executable]
+            else:
+                raise RuntimeError(
+                    'Cannot install requirements: this Python environment has no pip module and uv was not found '
+                    'on PATH. Install pip into the environment or make uv available.'
+                )
+            subprocess.check_call(install_command + PIP_OPTIONS + new_reqs + PIP_PACKAGE_INDEX_OPTIONS)
             _installed_requirements.update(new_reqs)
         except Exception as e:
             log.error(f'Error installing packages: {" ".join(new_reqs)}')


fractuscontext:
Instead of shelling out, we can also use `from importlib.metadata import version, PackageNotFoundError` to check if a package is installed or not

https://github.com/open-webui/open-webui/pull/20321

Classic298:
actually not this is entirely wrong idea

PR #27426, opened 2026-07-24
fix: surface tool-call and code-interpreter continuation errors instead of silently swallowing them

### LINKED ISSUE #27411: Errors in tool-call/code-interpreter continuation rounds are silently swallowed (empty assistant message, no error shown)
### Check Existing Issues

- I have searched the existing issues and discussions and found no report of this bug.

### Installation Method

Docker (official image, `ghcr.io/open-webui/open-webui`)

### Open WebUI Version

v0.10.2 (the affected code is unchanged on current `dev`)

n/a (OpenAI-compatible backend behind a proxy)

### Operating System

Linux (containerized, AWS ECS)

### Browser (if applicable)

n/a (server-side bug; verified via chat DB and backend logs)

### Expected Behavior

When the model requests native tool calls and the follow-up model invocation (the "continuation" round that synthesizes the final answer from the tool results) fails — e.g. the upstream endpoint returns HTTP 400 — the user should see an error message on the chat message, and the error should be recorded on the message in the DB, like other error paths in `streaming_chat_response_handler` already do (e.g. the tool-call iteration limit emits `chat:message:error` and persists an `error` field).

### Actual Behavior

The chat ends with a **silently empty assistant message**: `content=''`, `done=true`, and a final output item `{type: "message", status: "completed", text: ""}`. No error event reaches the frontend, no `error` field is persisted, and nothing is logged unless `GLOBAL_LOG_LEVEL=DEBUG` is set.

Cause: in `backend/open_webui/utils/middleware.py` (`streaming_chat_response_handler`), the native tool-call continuation loop wraps `generate_chat_completion` / `stream_body_handler` in:

```python
except Exception as e:
    log.debug(e)
    break
```

so **any** exception in a continuation round (round ≥ 2) is swallowed. The code-interpreter continuation a bit further down has the same pattern. On current `dev` these are the `except` blocks around lines ~5133 and ~5336.

Round-1 text (produced before the tool calls) survives, which makes the failure look like the model "answered with nothing" — with no diagnostic trail.

### Steps to Reproduce

1. Configure an OpenAI-compatible connection through a proxy that can reject a request (in our case: LiteLLM with an AWS Bedrock guardrail that returned HTTP 400 "Violated guardrail policy" for the continuation request, because the raw web-search tool results in the round-2 payload triggered its prompt-attack filter).
2. Enable native function calling and a tool (e.g. built-in web search), ask a question that triggers tool use.
3. Round 1 streams fine and executes the tools; the round-2 request is rejected by the proxy (any 4xx/5xx or transient network error in that round reproduces it — the guardrail is just one way to trigger it).
4. The chat shows the round-1 text (if any) and then an empty final message; no error is displayed. The DB has `content=''` with `done=true`.

### Logs & Screenshots

With default log levels there is nothing in the log (that is the bug). With `GLOBAL_LOG_LEVEL=DEBUG` the swallowed exception (the proxy's 400) appears via `log.debug(e)` at the location above. Verified against the model provider's invocation logs: the upstream model had fully streamed the round-2 answer; the rejection happened at the proxy layer, and open-webui dropped it silently.

### Additional Information

The fix is small and mirrors the existing iteration-limit error path in the same function (persist `{'error': {'content': ...}}` + emit `chat:message:error` + `log.exception`). PR follows.

ddembour:
﻿# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #27411
- [ ] **Documentation:** No docs change needed (no new setting or behavior surface — errors now use the existing error-display path).
- [ ] **Testing:** The failure mode itself was diagnosed and verified on a production v0.10.2 deployment (proxy-side HTTP 400 in the continuation round → silent empty message, confirmed via chat DB + provider invocation logs). The patched build is scheduled for a manual negative test (forced continuation failure → visible error) in our deployment; I will report the result on this PR.
- [ ] **No Unchecked AI Code:** Transparent note: this patch was produced with AI assistance during the incident investigation. It mirrors the existing iteration-limit error path in the same function 1:1 and is undergoing human diff review and manual testing on our side; I will confirm both on this PR before asking for review.
- [x] **Architecture:** No new settings; reuses the established error path (`chat:message:error` + persisted `error` field) of the same function.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

In `streaming_chat_response_handler` (`backend/open_webui/utils/middleware.py`), the native tool-call continuation loop and the code-interpreter continuation loop both wrap the follow-up model invocation in `except Exception as e: log.debug(e); break`. Any failure in a continuation round (proxy/backend 4xx or 5xx, transient network error, stream-parsing error) is therefore swallowed: the chat ends with an empty assistant message (`content=''`, `done=true`), no error event reaches the frontend, no `error` field is persisted, and nothing is logged at default log levels.

This PR handles those exceptions exactly like the tool-call iteration limit a few lines below already does:

- persist `{'error': {'content': ...}}` on the message (skipping `channel:` chats, same guard as the limit path),
- emit a `chat:message:error` event so the user sees the failure,
- use `log.exception` so the stack trace is available without a DEBUG deploy.

Real-world impact example (how we found it): a LiteLLM/Bedrock-guardrail 400 on the round-2 request made every affected chat end as a silent empty answer — the model provider's logs showed a fully streamed response while users saw nothing. Diagnosing this required a DEBUG-level redeploy; with this patch the cause is visible in the UI, the DB, and the default logs. The bug is generic: **any** continuation-round failure is currently invisible.

### Fixed

- Errors during native tool-call and code-interpreter continuation rounds are now shown to the user and persisted on the message instead of being silently swallowed (empty assistant message).

---

### Additional Information

- The two patched `except` blocks are at the tool-call continuation (~line 5131) and the code-interpreter continuation (~line 5334) on current `dev`; the same pattern exists in v0.10.x releases.
- No behavior change on the success path; the `break` semantics of the loops are unchanged.

### Screenshots or Videos

- n/a (server-side; before: empty assistant message with `done=true` in the chat DB — after: standard error box via `chat:message:error`).

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -5131,7 +5131,20 @@ async def streaming_chat_response_handler(response, ctx):
                         else:
                             break
                     except Exception as e:
-                        log.debug(e)
+                        log.exception('Error in tool-call continuation: %s', e)
+                        error_content = str(e) or 'Error during tool-call continuation.'
+                        if not metadata.get('chat_id', '').startswith('channel:'):
+                            await Chats.upsert_message_to_chat_by_id_and_message_id(
+                                metadata['chat_id'],
+                                metadata['message_id'],
+                                {'error': {'content': error_content}},
+                            )
+                        await event_emitter(
+                            {
+                                'type': 'chat:message:error',
+                                'data': {'error': {'content': error_content}},
+                            }
+                        )
                         break
 
                 if (
@@ -5321,7 +5334,20 @@ async def streaming_chat_response_handler(response, ctx):
                             else:
                                 break
                         except Exception as e:
-                            log.debug(e)
+                            log.exception('Error in code-interpreter continuation: %s', e)
+                            error_content = str(e) or 'Error during code-interpreter continuation.'
+                            if not metadata.get('chat_id', '').startswith('channel:'):
+                                await Chats.upsert_message_to_chat_by_id_and_message_id(
+                                    metadata['chat_id'],
+                                    metadata['message_id'],
+                                    {'error': {'content': error_content}},
+                                )
+                            await event_emitter(
+                                {
+                                    'type': 'chat:message:error',
+                                    'data': {'error': {'content': error_content}},
+                                }
+                            )
                             break
 
                 # Mark all in-progress items as completed


tjbck:
Addressed.

PR #26270, opened 2026-06-24
feat: add Workspace filter tag to model selector

silentoplayz:
# Pull Request Checklist

### Note to first-time contributors: Please open a discussion post in [Discussions](https://github.com/open-webui/open-webui/discussions) to discuss your idea/fix with the community before creating a pull request, and describe your changes before submitting a pull request.

This is to ensure large feature PRs are discussed with the community first, before starting work on it. If the community does not want this feature or it is not relevant for Open WebUI as a project, it can be identified in the discussion before working on the feature and submitting the PR. For bug fixes referencing an existing Issue, linking the Issue is sufficient.

**Before submitting, make sure you've checked the following:**

- [ ] **Linked Issue/Discussion:** This PR references an existing [Issue](https://github.com/open-webui/open-webui/issues) or [Discussion](https://github.com/open-webui/open-webui/discussions) — `Closes #___` / `Relates to #___`. If one does not exist, create one first. PRs without a linked issue or discussion may be closed without review.
- [x] **Target branch:** The pull request targets the `dev` branch. **PRs targeting `main` will be immediately closed.**
- [x] **Description:** A concise description of the changes is provided below.
- [x] **Changelog:** A changelog entry following [Keep a Changelog](https://keepachangelog.com/) format is included at the bottom.
- [ ] **Documentation:** Relevant documentation has been added or updated in the [Open WebUI Docs Repository](https://github.com/open-webui/docs).
- [ ] **Dependencies:** Any new or updated dependencies are explained, tested, and documented.
- [x] **Testing:** Manual tests have been performed to verify the fix/feature works correctly and does not introduce regressions. Screenshots or recordings are included where applicable.
- [x] **No Unchecked AI Code:** This PR is either human-written or has undergone thorough human review AND manual testing. Unreviewed AI-generated PRs may be closed immediately.
- [x] **Self-Review:** A self-review of the code has been performed, ensuring adherence to project coding standards.
- [x] **Architecture:** Smart defaults are preferred over new settings. Local state is used for ephemeral UI logic. Major architectural or UX changes have been discussed first.
- [x] **Git Hygiene:** The PR is atomic (one logical change), rebased on `dev`, and contains no unrelated commits.
- [x] **Title Prefix:** The PR title uses one of the following prefixes:
  - **feat**: New features or enhancements

# Changelog Entry

### Description

[...]

PR #27455, opened 2026-07-25
feat: add Workspace filter tab to model selector

oshelot:
- [x] **Linked Issue/Discussion:** Relates to Discussion #4006, Issue #15949, Issue #17596
- [x] **Target branch:** dev
- [x] **Changelog:** Included below
- [x] **Testing:** Manually tested with 7 workspace models + 340 external OpenRouter models
- [x] **No Unchecked AI Code:** Human-reviewed and manually tested
- [x] **Self-Review:** Done
- [x] **Architecture:** No new settings, minimal 7-line change to existing file
- [x] **Git Hygiene:** Single atomic commit, rebased on dev
- [x] **Title Prefix:** feat

# Changelog Entry

### Description

The model selector's filter bar has tabs for All, Local, External, and Direct — but no tab for user-created workspace models (custom presets). When using providers like OpenRouter with 300+ models, workspaces are buried at the bottom. The backend filter logic for `workspace` already exists in the filter blocks but the UI tab entry in `modelFilterItems` was never added.

### Added

- `{ value: 'connection:workspace', label: 'Workspace' }` entry in `modelFilterItems` array, conditionally rendered when preset models exist
- `workspace` filter case (`item.model?.preset === true`) in both search and non-search filter paths

---

### Additional Information

**Single file changed:** `src/lib/components/chat/ModelSelector/Selector.svelte` (+7 lines)

The filter logic (`selectedConnectionType === 'workspace'` → `item.model?.preset === true`) mirrors the existing pattern for `local`, `external`, and `direct` tabs.

### DIFF
--- a/src/lib/components/chat/ModelSelector/Selector.svelte
+++ b/src/lib/components/chat/ModelSelector/Selector.svelte
@@ -295,6 +295,8 @@
 							return item.model?.connection_type === 'external';
 						} else if (selectedConnectionType === 'direct') {
 							return item.model?.direct;
+						} else if (selectedConnectionType === 'workspace') {
+							return item.model?.preset === true;
 						}
 					})
 			: items
@@ -315,6 +317,8 @@
 							return item.model?.connection_type === 'external';
 						} else if (selectedConnectionType === 'direct') {
 							return item.model?.direct;
+						} else if (selectedConnectionType === 'workspace') {
+							return item.model?.preset === true;
 						}
 					})
 	).filter((item) => includeHidden || !(item.model?.info?.meta?.hidden ?? false));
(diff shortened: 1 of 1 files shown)


silentoplayz:
Some prior art in case it is useful: I opened #26270 back in June for the same feature — a Workspace filter in the model selector. It was closed, so there may be context I am not aware of on whether this is wanted; worth knowing before you sink more time into it.

The two implementations differ mainly in the predicate. Mine keyed off `item.model?.info?.base_model_id`, inferring "workspace model" client-side, whereas this PR uses `item.model?.preset === true` — which is the flag the backend already stamps on custom models in `utils/models.py`, so it is the more direct signal of the two. This version also slots the option into the array the selector now builds its filters from, rather than the hand-written button markup I had to add at the time, so it is a noticeably smaller diff against the current code.

Happy to test this on my instance if that would help.

oshelot:
Thanks for the context on #26270. I'll search for better next time to find similar request. 

Glad the preset flag approach looks cleaner. I'm running this on my instance now with 7 workspace models + 340 OpenRouter models and it works well. The primary driver was that when started a new chat after selecting the Workspace, the desired default model for the Workspace was never set. With this, I start a new chat, select the workspace and I'm good.
                                                                                                                     
If you're able to test on your end, that would be great — always good to have a second confirmation.

silentoplayz:
LGTM!

tjbck:
Open to discussion, if we can get more support for this feature, happy to revisit with @silentoplayz's approach.

PR #27495, opened 2026-07-26
fix: meet 4.5:1 contrast for muted UI text (WCAG 1.4.3, 1.4.11)

Classic298:
On latest `dev`, muted UI text is written as `text-gray-400 dark:text-gray-600` and `text-gray-400 dark:text-gray-500`. The grey scale in `src/tailwind.css` is achromatic `oklch(L 0 0)`, so relative luminance is exactly `L³` and the ratios are:

| | light (on `#fff`) | dark (on `#171717`) |
|---|---|---|
| `text-gray-400` | **2.07:1** | 8.65:1 |
| `dark:text-gray-600` | 5.75:1 | **3.12:1** |
| `dark:text-gray-500` | 2.77:1 | 6.46:1 |

The pair is effectively inverted: the light variant is a light grey and the dark variant is a dark grey. `text-gray-400 dark:text-gray-600` therefore fails in **both** themes, at 2.07:1 and 3.12:1 against a 4.5:1 requirement, and the light value also fails the 3:1 required of icons. `text-gray-400 dark:text-gray-500` fails in light mode at the same 2.07:1.

This is the text used for every settings and admin section heading, every field description, sidebar section labels, timestamps and counters. All of it is 10px to 12px, so the large-text exemption does not apply.

Breaks WCAG 1.4.3 Contrast (Minimum) (Level AA) for the text, and 1.4.11 Non-text Contrast (Level AA) for the four sites where this pair is passed to an icon component as `className`.

Fix: `text-gray-400` becomes `text-gray-600` (5.75:1) in light mode, and `dark:text-gray-600` becomes `dark:text-gray-400` (8.65:1) in dark mode. `dark:text-gray-500` already passes at 6.46:1 and is left alone. Both failing pairs are converted together so components do not end up with two different greys side by side.

Three sites are hand fixed rather than pattern matched:

- `chat/Settings/Shortcuts.svelte` — here the literal string only appears as a substring spanning `hover:text-gray-400` and a separate `dark:text-gray-600`. A blind replace would have produced an unprefixed `hover:text-gray-600`, which applies in both themes and would have made the control **dimmer** on hover in dark mode, dropping it to 3.12:1. This is the only occurrence of that trap in the codebase.
- `admin/Settings/Audio.svelte` — the body text moving to `gray-600` would have made the nested links exactly the same colour as the surrounding prose, 1.00:1, with no resting underline. The link colour moves to `gray-900` so links stay distinguishable, which would otherwise have been a new WCAG 1.4.1 Use of Color failure.
- `chat/Messages/ResponseMessage/TaskList.svelte` — the completed and cancelled branches use different pairs. Both are converted so the completed / cancelled / active hierarchy is preserved rather than inverted.

Not covered here, and left for separate PRs: `text-gray-500 dark:text-gray-400` (140 sites, 2.77:1 in light), and the global `input::placeholder` colour in `src/tailwind.css` which has the same 2.07:1 problem.

Severity: Serious, and the most widely visible accessibility defect in the current light theme.

> [!NOTE]

### DIFF
--- a/src/lib/components/ChangelogModal.svelte
+++ b/src/lib/components/ChangelogModal.svelte
@@ -122,7 +122,7 @@
 						<section class="pr-1">
 							<div class="mb-2">
 								<h3 class="m-0 text-sm font-normal text-gray-950 dark:text-white">v{version}</h3>
-								<div class="mt-0.5 text-[0.6875rem] text-gray-400 dark:text-gray-500">
+								<div class="mt-0.5 text-[0.6875rem] text-gray-600 dark:text-gray-500">
 									{formatDate(changelog[version].date)}
 								</div>
 							</div>
@@ -181,7 +181,7 @@
 				</div>
 			{:else}
 				<div
-					class="flex items-center justify-center py-16 text-sm text-gray-400 dark:text-gray-500"
+					class="flex items-center justify-center py-16 text-sm text-gray-600 dark:text-gray-500"
 				>
 					{$i18n.t('Loading release notes...')}
 				</div>
--- a/src/lib/components/admin/Functions.svelte
+++ b/src/lib/components/admin/Functions.svelte
@@ -449,7 +449,7 @@
 			{#if (filteredItems ?? []).length !== 0}
 				<div class="my-1">
 					<div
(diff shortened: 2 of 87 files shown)


tjbck:
Open as a high contrast mode.

PR #27496, opened 2026-07-26
fix: meet 4.5:1 contrast for input placeholders (WCAG 1.4.3)

Classic298:
On latest `dev`, placeholder text is the lowest contrast text in the product. The grey scale in `src/tailwind.css` is achromatic `oklch(L 0 0)`, so relative luminance is exactly `L³`:

| | ratio |
|---|---|
| `placeholder:text-gray-300` on `#fff` | **1.58:1** | | `dark:placeholder:text-gray-700` on `#171717` | **2.12:1** | | `placeholder-gray-400` / `placeholder:text-gray-400` on `#fff` | **2.07:1** | | `dark:placeholder:text-gray-600` on `#171717` | **3.12:1** | | global `input::placeholder` rule, `--color-gray-400` on `#fff` | **2.07:1** |

Level AA requires 4.5:1. The large text exemption does not apply, the largest of these is `text-lg`. Placeholders are frequently the only format hint a field gives, for example `admin/Settings/General.svelte` uses `e.g.) "http://localhost:3000"`.

Fix, in three parts:

1. The global `input::placeholder` / `textarea::placeholder` rule had **no dark variant at all**, so one value served both themes. It now resolves to gray-600 in light (5.75:1) and gray-500 in dark (6.46:1).

2. The ~207 per component placeholder colour utilities are **removed** rather than rewritten, so they fall through to the corrected base rule. They encoded no per component intent, they were copy paste drift, and they duplicated a decision that now lives in one place. Removing them also fixes `placeholder-gray-400`, which is un-prefixed and therefore applied its light mode value in dark mode too: rewriting it to gray-600 would have taken dark mode from 8.65:1 down to 3.12:1, replacing one failure with another. The two genuinely intentional placeholder colours, in `ModelSelector/Selector.svelte` and `common/Tags/TagInput.svelte`, are left untouched.

3. The chat composer placeholder is a tiptap `::before`, so neither the base rule nor any utility reaches it. It was hardcoded `#676767` for both themes, which is 5.66:1 light but only **3.17:1** dark. Its dark override also keyed off `@media (prefers-color-scheme: dark)` while this app toggles dark mode with a **class**, so the colour followed the operating system rather than the selected theme. Both are replaced with `text-gray-600 dark:text-gray-500`, matching every other placeholder. The comment claiming the old value had "the proper contrast" is removed, since it was never true for dark mode.

Placeholders stay distinguishable from real input values, which are `text-gray-700` (8.46:1) in light and `dark:text-gray-300` (11.39:1) in dark.

Verified against the `oled-dark` theme (gray-500 on `#000` is 7.57:1) and the `dark:bg-white/[0.03]` input surface (6.01:1). Formatting was only re-run on files that were already Prettier clean on `dev`, so no unrelated reflow is included.

Severity: Serious. Affects every text input in the product, including the chat composer.

> [!NOTE]

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -402,12 +402,9 @@ input[type='number'] {
 .ProseMirror p.is-editor-empty:first-child::before {
 	content: attr(data-placeholder);
 	float: left;
-	/* Below color is from tailwind, and has the proper contrast
-	text-gray-600 from: https://tailwindcss.com/docs/color */
-	color: #676767;
 	pointer-events: none;
 
-	@apply line-clamp-1 absolute;
+	@apply line-clamp-1 absolute text-gray-600 dark:text-gray-500;
 }
 
 .tiptap ul[data-type='taskList'] {
@@ -566,12 +563,6 @@ input[type='number'] {
 				align-items: initial;
 			}
 		}
-	}
-}
-
-@media (prefers-color-scheme: dark) {
-	.ProseMirror p.is-editor-empty:first-child::before {
-		color: #757575;
 	}
 }
 
--- a/src/lib/components/AddConnectionModal.svelte
+++ b/src/lib/components/AddConnectionModal.svelte
@@ -60,10 +60,8 @@
 	let showDeleteConfirmDialog = false;
 	let showAdvanced = false;
 
-	const inputClass =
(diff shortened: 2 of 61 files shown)


tjbck:
Open as a high contrast mode.

PR #27497, opened 2026-07-26
fix: meet 4.5:1 contrast for gray-500 muted text in light theme (WCAG 1.4.3, 1.4.11)

Classic298:
Follow up to the `text-gray-400` contrast fix, covering the other failing muted token.

The grey scale in `src/tailwind.css` is achromatic `oklch(L 0 0)`, so relative luminance is exactly `L³`:

| | on `#fff` (light) | on `#171717` (dark) |
|---|---|---|
| `text-gray-500` | **2.77:1** | 6.46:1 |
| `text-gray-600` | **5.75:1** | 3.12:1 |
| `dark:text-gray-400` | | 8.65:1 |

`text-gray-500` paired with a dark variant fails Level AA in light mode at 2.77:1, against 4.5:1 for text and 3:1 for icon buttons. The dark halves already pass and are deliberately left alone, so only the light value moves.

Scope: an unprefixed `text-gray-500` becomes `text-gray-600` **whenever the same class string also carries an unprefixed `dark:text-gray-400` or `dark:text-gray-500`**. The rule is deliberately "same class string" rather than "adjacent tokens", because adjacency is an accident of how the class list was typed, not a property of the design. Matching on adjacency alone would have left 113 identical sites behind, 38 of them inside components this PR already touches, producing two different greys in the same row (for example the tool count and tool description in `chat/ToolServersModal.svelte`, or the table text and its empty state in `admin/Analytics/ModelUsage.svelte`).

Prefixed variants (`hover:`, `group-hover:`, `dark:`) are excluded by a negative lookbehind, so hover ramps keep their existing relationships.

Three sites are hand fixed rather than pattern matched:

- `layout/Sidebar/Section.svelte` — the sidebar "Chats" / "Folders" / "Pinned" label. Its tokens are `text-gray-400 hover:text-gray-500 dark:text-gray-500 dark:hover:text-gray-400`, which no pattern can see as one unit. The resting colour was **2.07:1** and the hover was **2.77:1**, so both failed. Now gray-600 resting, gray-800 on hover, matching the house convention of a two step ramp.
- `admin/Functions.svelte` — carried `opacity-60`, which composites before contrast, so even at gray-600 it would only reach 2.52:1. The opacity is dropped so the item counter actually passes.
- `admin/Settings/General.svelte` — `labelClassName="text-gray-500 dark:text-gray-500"` was a dead override. `AdminSettingRow` already renders `text-gray-600 dark:text-gray-400`, and because gray-600 sits later in the scale it already won. The conflicting override is removed rather than rewritten.

Not covered here: roughly 458 sites where `text-gray-500` has **no** dark sibling at all, so the same value serves both themes. Those need a per site judgement about which dark value to add, and belong in their own PR.

Formatting was only re-run where the file was already Prettier clean on `dev`, so no unrelated reflow is included.

Severity: Serious, and together with the `text-gray-400` fix this is the bulk of the light theme contrast problem.

> [!NOTE]

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -659,7 +659,7 @@ input[type='number'] {
 
 /* Table styling for tiptap editors */
 .tiptap table {
-	@apply w-full text-sm text-start text-gray-500 dark:text-gray-400 max-w-full;
+	@apply w-full text-sm text-start text-gray-600 dark:text-gray-400 max-w-full;
 }
 
 .tiptap thead {
--- a/src/lib/components/AddConnectionModal.svelte
+++ b/src/lib/components/AddConnectionModal.svelte
@@ -476,7 +476,7 @@
 						<div class="flex items-center justify-between">
 							<button
 								type="button"
-								class="flex items-center gap-1 text-xs text-gray-500 dark:text-gray-400 hover:text-gray-700 dark:hover:text-gray-200 transition mt-2"
+								class="flex items-center gap-1 text-xs text-gray-600 dark:text-gray-400 hover:text-gray-700 dark:hover:text-gray-200 transition mt-2"
 								on:click={() => (showAdvanced = !showAdvanced)}
 							>
 								<svg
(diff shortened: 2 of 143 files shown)


tjbck:
Open as a high contrast mode.

PR #27507, opened 2026-07-26
fix: reveal hover-only controls on keyboard focus (WCAG 2.1.1, 2.4.7)

Classic298:
On latest `dev`, around 40 controls are hidden with `invisible group-hover:visible`. Tailwind's `invisible` is `visibility: hidden`, which removes the element **and its descendants from the tab order**, so `group-focus-within` can never fire. There is no ordering of Tailwind variants that fixes this: the control has to stop being `visibility: hidden` before focus can ever reach it.

The practical result is that for a keyboard or screen reader user these controls do not exist:

- the per chat "…" menu in the sidebar, so Share, Rename, Clone, Pin, Move, Archive, Download and Delete for every chat except the open one
- every per message action, so Edit, Copy, Read Aloud, Good/Bad Response, Continue, Fork, Regenerate and Delete on every message except the last
- code block copy and download buttons
- the buttons that remove an attached file or image from the composer
- folder and channel menus

A separate 12 sites use `opacity-0 group-hover:opacity-100`. Those stay focusable, but they are still invisible while focused.

Breaks WCAG 2.1.1 Keyboard (Level A) for the first group and 2.4.7 Focus Visible (Level AA) for the second.

Fix: one shared `hover-reveal` utility in `src/app.css` that hides with opacity instead of visibility, and reveals on `:hover` **or** `:focus-within`. The 12 opacity based sites get `group-focus-within:opacity-100` and are otherwise left alone, since they have no pointer-events handling today and converting them would fold a behaviour change into this PR.

`pointer-events-none` is part of the utility and is not optional. `visibility: hidden` blocked pointer events; plain `opacity-0` does not, so without it every one of these would become an invisible but clickable target sitting over the row it belongs to. `pointer-events-auto` restores them alongside the opacity.

Behaviour for pointer users is unchanged. `src/tailwind.css` already overrides the `hover` variant to a bare `&:hover` with no `(hover: hover)` guard, so the hand written `.group:hover` matches exactly what `group-hover:` compiled to before.

Three details worth calling out:

- `common/FileItem.svelte` and `chat/MessageInput.svelte` also carried `outline-hidden focus:outline-hidden` on the same buttons. Making them focusable while suppressing their focus ring would have swapped a 2.1.1 failure for a 2.4.7 one, so that suppression is removed on those two.
- `chat/EmbeddedChatHistoryItem.svelte` has the identical defect but was excluded only because its group is named `group/item`. Rather than adding a second CSS rule for named groups, the group is renamed, since nothing nests it inside another group.
- A commented out `<button>` in `common/FileItem.svelte` matches the same pattern and is deliberately left untouched.

Keyboard users will now tab through the per message action bar. That is the intended outcome, and it already happens today for the last message and whenever High Contrast Mode is on, so it is not new behaviour, just consistent behaviour.

Icon and value swap patterns (`hidden group-hover:flex`, `group-hover:inline`) are not touched, they are not hidden controls.

Severity: Critical. This is the single largest keyboard accessibility gap in the application.

> [!NOTE]


Classic298:
Screenshots below. Each pair is the same interaction on two production builds: `before` is this branch's merge-base, `after` is this branch. Throwaway instance, synthetic data only. Both builds have a 2px blue outline injected on `:focus-visible` purely so the screenshots can show where keyboard focus is, since neither build ships a focus indicator on its own. That is the subject of #27500.

**The headline number**, measured by pressing Tab 260 times and recording every stop: of the 12 hover-only interactive controls on a chat page, **0 were reachable before, 12 are reachable after**.

**Sidebar.** One Tab past a chat link. Before, focus skips straight to the user menu and the chat's `...` never enters the tab order at all. After, focus lands on it:

reveal-07b-row-next

**Message action bar** on a non-last message, at the moment Edit takes focus:

reveal-04-message-actions

**No regression for pointer users.** Idle, the controls are hidden. `visibility: hidden` becomes `opacity: 0` plus `pointer-events: none`, so they are still not hit-testable, confirmed by `elementFromPoint`:

reveal-03a-idle

On hover they reveal and become clickable exactly as before:

reveal-03b-hovered

The sidebar row on hover is likewise unchanged:

reveal-05-sidebar-hover

| | before | after |
| --- | --- | --- |
| Hover-only controls reachable by Tab | 0 / 12 | 12 / 12 |
| Idle hidden state | `visibility: hidden` | `opacity: 0`, `pointer-events: none` |
| Idle, hit-testable | no | no |
| Hovered | visible, clickable | visible, clickable |

The last commit is a follow-up found while taking these. Two things the original change missed on the keyboard path, both in the sidebar row:

- The relative timestamp and the title's padding are driven by `showInlineActions`, a component state fed by `mouseenter`, not by CSS. `:focus-within` never flipped it, so keyboard focus revealed the `...` menu on top of a still-visible timestamp. A `focusWithin` flag puts keyboard focus on the same footing as the pointer.
- Opening that menu with Enter then collapsed the row and faded out the trigger from under its own open menu. `Dropdown` portals its content to `document.body`, so `:focus-within` stops matching, and the row only learned the menu was open via the trigger's `on:click`, which never fires for keyboard users because `Dropdown` calls `preventDefault()` on Enter and Space. `ChatMenu` now forwards the open transition it already gets from `Dropdown`. It is tracked as local row state rather than reusing the existing `select` dispatch, because `RecursiveFolder` renders `ChatItem` without the `selected` prop, so chats inside folders would not have been covered.

While there, four menu items closed themselves by assigning `Dropdown`'s bound `show` directly, which bypasses `closeDropdown` and so fired neither `onOpenChange(false)` nor the focus restore, stranding keyboard focus on `<body>`. They now call `Dropdown.close()`. I kept that to `ChatMenu` rather than changing `Dropdown` itself, since the same pattern appears in roughly 30 other menus and that is a much wider change than this PR should carry.

tjbck:
@copilot resolve the merge conflicts in this pull request

PR #27508, opened 2026-07-26
fix: move focus into dropdown content and restore it on close (WCAG 2.4.3)

Classic298:
On latest `dev`, `common/Dropdown.svelte` does no focus management at all. The content is portaled to the end of `<body>` while the trigger sits mid document, so opening a menu leaves focus on the trigger and the next Tab goes to the **next sidebar element**, not into the menu. Reaching a menu item means tabbing through the entire rest of the document to the portal at the end of `<body>`.

Breaks WCAG 2.4.3 Focus Order (Level A): the focus order does not preserve meaning, because the menu's DOM position no longer matches its visual position.

This component has 49 consumers, including the chat item menu, the user menu and the model selector.

Fix: move focus to the first item when the dropdown opens, and restore it when the dropdown closes.

Focus restoration is placed at a single `closeDropdown()` chokepoint used by Escape, outside click and the exported `close()`. Putting it only on Escape would have left the two commonest paths, clicking a menu item and clicking away, dropping focus to `<body>` and resetting the keyboard user's position to the top of the document, which is worse than the current behaviour.

Restoration is conditional on the dropdown still holding focus. On an outside click focus has already moved to whatever the user clicked, and restoring it there would steal focus back.

The element to restore to is captured as `document.activeElement` at open time rather than read from `triggerEl`. `triggerEl` is guessed as `node.firstElementChild`, and at call sites that wrap the trigger in a `Tooltip` (which renders a plain `<div>`) that is not a focusable element, so `triggerEl.focus()` would silently do nothing.

Opening is consolidated into the reactive block so consumers that open the dropdown externally with `bind:show` behave the same as consumers that use the trigger.

`role="menu"` is removed. The content's children are plain `<button>`s with no `role="menuitem"`, no roving tabindex and no arrow key handling. Moving focus into a `role="menu"` is exactly when screen readers switch to application mode and hand arrow keys to the widget, so keeping the role while moving focus into it would turn an unreachable menu into a reachable dead end. Without the role, ordinary Tab semantics apply and match what the component actually implements. Implementing the full menu pattern is a separate change.

Known limitation: for a dropdown rendered inside `common/Modal.svelte`, the modal's focus trap only pauses on a pointerdown outside the modal, so it pulls focus back out of the portaled content. Dropdowns inside modals are therefore unchanged by this PR and still need the trap to account for portaled content.

Severity: Serious. Menus open but their contents are effectively unreachable in sequence.

> [!NOTE]

### DIFF
--- a/src/lib/components/common/Dropdown.svelte
+++ b/src/lib/components/common/Dropdown.svelte
@@ -31,6 +31,7 @@
 
 	let triggerEl: HTMLElement | null = null;
 	let contentEl: HTMLElement | null = null;
+	let previouslyFocused: HTMLElement | null = null;
 	let positionFrame: number | undefined;
 	let settleTimers: number[] = [];
 	let resolvedMaxHeight = maxHeight;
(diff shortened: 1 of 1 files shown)


tjbck:
Addressed.

PR #27516, opened 2026-07-26
fix: make sidebar hover and selection states perceivable

Classic298:
The sidebar surface is `bg-gray-50/70 dark:bg-gray-950/70`, which composites to about `#fafafa` and `#101010`, so the row tints have to be measured against **that**, not against the page background. Measured there:

| state | current | ratio |
|---|---|---|
| hover, light | `hover:bg-gray-50` on `#fafafa` | **1.02:1** | | hover, light, **mobile** | `hover:bg-gray-50` on opaque `bg-gray-50` | **1.00:1** | | hover, dark | `dark:hover:bg-gray-900` | 1.05:1 | | selected, light | `bg-black/[0.035]` | 1.08:1 |
| selected, dark | `dark:bg-white/[0.045]` | 1.10:1 |

On mobile in light mode the hover class is **the same colour as the surface**, so there is no hover state at all. On desktop it is 1.02:1, which is indistinguishable in practice. This is what users have been reporting as "the background only differs on barely nothing".

Separately, all four selected branches in `ChatItem.svelte` used the **same** light value, and the pinned menu rows in `Sidebar.svelte` did too, so turning on **High Contrast Mode changed nothing whatsoever in light mode**.

After this change: hover 1.15 light / 1.15 dark, selected 1.20 / 1.21, and hover stays below selected in both themes so the row you are on always outranks the row you are pointing at.

High Contrast Mode now adds an **outline** rather than a heavier fill. That is not a stylistic preference: past `bg-black/[0.08]` the 10px timestamp in each row drops below 4.5:1 against its own background, so in light mode there is simply no room to strengthen the fill without introducing a WCAG 1.4.3 text failure inside the setting meant to help. An outline at `gray-700` / `gray-300` is far above 3:1 against the fill and costs the text nothing.

Also fixed in the same pass, because they share the surface and would otherwise end up inconsistent with the rows around them:

- `ChannelItem.svelte` used `bg-gray-100 dark:bg-gray-900` for selected and the same token family for hover, so raising hover alone would have made a hovered channel more prominent than the channel you are actually in.
- `RecursiveFolder.svelte` folder rows were `hover:bg-gray-50/40` at **1.01:1** light, worse than anything else here, and their selected state would have read as less selected than the chats nested inside them.
- `SearchInput.svelte` clear button hover.
- The now redundant nested ternary in `ChatItem.svelte` is collapsed: with the values unified, `id === $chatId`, `confirmEdit` and `selected` produced identical output.

Scope is deliberately limited to components that sit on the sidebar surface. The same `dark:hover:bg-gray-900` token appears in about 16 other files on different surfaces where it behaves differently, and those are not touched.

Not a WCAG conformance claim. A row hover tint is not in scope for 1.4.11, and although the selected state arguably is, none of these values reaches 3:1 either before or after. This is a legibility fix; the accessible marking of the current chat is handled separately by adding `aria-current`.

All eight files compile with no new warnings and are Prettier clean.

Severity: Serious usability regression, most visible in light mode.

> [!NOTE]

### DIFF
--- a/src/lib/components/common/Folder.svelte
+++ b/src/lib/components/common/Folder.svelte
@@ -154,7 +154,7 @@
 				<!-- svelte-ignore a11y-no-static-element-interactions -->
 				<div
 					id="sidebar-folder-button"
-					class=" w-full group rounded-xl relative flex items-center justify-between hover:bg-gray-100 dark:hover:bg-gray-900 transition {buttonClassName}"
+					class=" w-full group rounded-xl relative flex items-center justify-between hover:bg-gray-100 dark:hover:bg-white/[0.06] transition {buttonClassName}"
 				>
 					<button class="w-full py-1.5 pl-2 flex items-center gap-1.5 text-xs font-normal">
 						{#if chevron}
--- a/src/lib/components/layout/Sidebar.svelte
+++ b/src/lib/components/layout/Sidebar.svelte
@@ -835,7 +835,7 @@
 						aria-label={$showSidebar ? $i18n.t('Close Sidebar') : $i18n.t('Open Sidebar')}
 					>
 						<div
-							class=" self-center flex size-[30px] items-center justify-center rounded-lg transition group-hover:bg-gray-50 dark:group-hover:bg-gray-900"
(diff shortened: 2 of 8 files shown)


tjbck:
high contrast mode.

Issue #27327, opened 2026-07-22
issue: With a knowledgebase of around 2500 documents, the kb_exec grep always returns Too many files


alpilotx:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

10.2

### Operating System

Ubuntu 24.04

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

When letting large LLMs with native function calling explicitly search via kb_exec grep on large knowledge bases (of many thousand files) it returns its normal results.

### Actual Behavior

At the moment my knowledge base of  2596 files, never seems to work when any LLM calls kb_exec grep, as it always returns:
"Too many files (2596). Scope your search: grep ......"

### Steps to Reproduce

Need a large knowledge base with likely 1-2k documents and then run requests which trigger the LLMs to use kb_eexec via native fucntion calling.

### Logs & Screenshots

In the docker log of openwebui there are no traces which point at issues.

### Additional Information

I already tried to add things like

    ulimits:
      nofile:
        soft: 65536
        hard: 65536    

To my docker-compose of openwebui, without any result

Classic298:
you need to increase MAX_GREP_FILES

alpilotx:
[...]

PR #27524, opened 2026-07-26
feat: cap kb_exec output and make the knowledge tool limits configurable

### LINKED ISSUE #26139: feat/issue: [dev/0.9.6]: kb_exec token usage
a single `ls` can yield 100k+ token

Classic298:
`kb_exec` had no bound on how much text it could hand back to the model. Only `cat` and `grep` were limited, so `ls`, `ls -a`, `tree`, `find`, `stat`, `wc` and `grep -l` / `grep -c` returned everything they found: a single `ls` on a large knowledge base could burn 100k+ tokens of context in one tool call. The limits that did exist were module constants, so an operator could neither lower them to protect context nor raise them to fit a bigger knowledge base.

Every `kb_exec` command and pipeline returns through one place, so the output cap goes there and covers all of them, including any command added later. The per-command `cat` truncation is dropped in favour of that single cap; as a side effect `cat big.md | grep x` now searches the whole file instead of only its first 100k chars, since only the final output is trimmed.

The remaining limits move to environment variables, with the same defaults as before:

- `KB_EXEC_MAX_OUTPUT_CHARS` (30000, new): total output of a single `kb_exec` call.
- `KB_EXEC_MAX_GREP_FILES` (200): files a `kb_exec` grep may scan before it asks for a narrower scope. Raising this is what a knowledge base of a few thousand documents needs (#27327).
- `KNOWLEDGE_GREP_MAX_MATCHES` (50): matching lines returned, shared by `kb_exec` grep and `grep_knowledge_files`, replacing the two separate constants that meant the same thing.
- `VIEW_FILE_MAX_CHARS` (100000) and `VIEW_FILE_DEFAULT_MAX_CHARS` (10000): `view_file` / `view_knowledge_file` output.

Values below 1 are clamped and unparseable values fall back to the default, so a typo in a compose file cannot stop the server from booting. `view_file`'s `:param max_chars:` no longer states the cap as a literal number, since that number is now operator-defined and the default already reaches the model through the tool schema.

`DEFAULT_HEAD_LINES` and `DEFAULT_TAIL_LINES` stay hardcoded on purpose: the model overrides them per call with `head -N` / `tail -N`, and their output is bounded by the global cap anyway, so a knob would have nothing to do.

Fixes #26139

> [!NOTE]

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -1044,6 +1044,29 @@ SENTENCE_TRANSFORMERS_CROSS_ENCODER_SIGMOID_ACTIVATION_FUNCTION = (
 )
 
 ####################################
+# KNOWLEDGE TOOLS
+####################################
+
+
+def _int_env(name: str, default: int) -> int:
+    try:
+        return max(int(os.getenv(name) or default), 1)
+    except (ValueError, TypeError):
+        return default
+
+
+# Total output of a single kb_exec call, whatever the command.
+KB_EXEC_MAX_OUTPUT_CHARS = _int_env('KB_EXEC_MAX_OUTPUT_CHARS', 30_000)
+# Files a single kb_exec grep may scan before it asks for a narrower scope.
+KB_EXEC_MAX_GREP_FILES = _int_env('KB_EXEC_MAX_GREP_FILES', 200)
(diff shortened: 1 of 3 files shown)


tjbck:
11e61b69ebd922602edc37ded7b42fdd43bb8456

PR #27546, opened 2026-07-26
fix: only accept token exchange tokens minted for trusted OAuth clients

Classic298:
The OAuth token exchange endpoint accepts an access token in the request body and validates it by calling the provider's userinfo endpoint. That endpoint confirms a token is valid but never reveals which OAuth client it was issued to, so a token minted for any client of the same provider was accepted. Anyone able to obtain an access token for a user, for instance by having them sign in to an unrelated OAuth application, could exchange it for an Open WebUI session as that user. The account's role was carried over unchanged, so a targeted admin yielded an admin session.

`OAUTH_ALLOWED_DOMAINS` does not cover this. It filters which user identities may sign in, and an impersonated user is a legitimate member of an allowed domain, so the check passes. The missing question is not who the user is but which client vouched for them.

Resolve the issuing client through RFC 7662 introspection and require it to appear in the new `OAUTH_TOKEN_EXCHANGE_ALLOWED_AUDIENCES` allowlist. Only the introspection response's `client_id` is trusted, never `aud`: `client_id` identifies the client the token was issued to, while `aud` names intended resource servers and several identity providers let any client place another client's id there.

The allowlist is a deploy-time environment variable rather than a database setting, so the endpoint's only authentication boundary cannot be widened at runtime from an admin session. It has no default, and an unset allowlist rejects every exchange. That is a breaking change for existing deployments, which must now name the clients they trust, but the previous behaviour trusted every client of the provider and the endpoint is off by default.

This requires a provider that implements RFC 7662. Google, Microsoft and GitHub do not, so token exchange no longer works against them and the rejection is logged. Supporting those providers needs an audience-bound assertion the caller can supply, such as an OIDC ID token validated against the provider's JWKS, which can be added without changing the current request contract.

> [!NOTE]


tjbck:
Addressed.

PR #27547, opened 2026-07-26
fix: rate limit the OAuth token exchange endpoint

Classic298:
The token exchange endpoint accepts an OAuth access token and issues a session with no throttling at all, while `/signin` limits attempts through `signin_rate_limiter`. That leaves unlimited attempts available to anyone working through leaked or guessed tokens.

Apply the same limit, keyed on the source address because the endpoint has no caller identity until the provider's userinfo call answers.

This is a throttle rather than a security boundary, and the comment says so. The launchers run uvicorn with `--forwarded-allow-ips '*'`, so the source address is taken from the leftmost `X-Forwarded-For` entry and is therefore client-controlled. Bounding the cost of automated probing is still worth having; treating it as an authentication control would not be.

> [!NOTE]


tjbck:
Addressed in dev.

PR #27568, opened 2026-07-27
feat: configurable first-login user disclaimer

mhupfauer:
<sub>Resubmission of #27567, which the CLA bot closed because I had left the CLA checkbox unticked. Same commit, no content changes; the checkbox is now ticked at the bottom.</sub>

# Pull Request Checklist

- [ ] **Linked Issue/Discussion:** ⚠️ **Not yet linked — see "Note on process" below.** I am happy to open a Discussion first and close this PR if you would prefer that order; I have written it up here so there is something concrete to react to.
- [x] **Description:** below.
- [x] **Changelog:** below.
- [ ] **Documentation:** not yet submitted to the docs repo. Three new env vars (`USER_DISCLAIMER_TITLE`, `USER_DISCLAIMER_CONTENT`, `USER_DISCLAIMER_VERSION`) would need an entry under *Environment Variable Configuration*. Happy to open that PR if this direction is accepted.
- [x] **Dependencies:** none added or changed.
- [x] **Testing:** manual testing plus unit tests — details below.
- [x] **No Unchecked AI Code:** see "AI tooling disclosure" below. Written with AI assistance, then reviewed line by line and manually tested by me.
- [x] **Architecture:** rationale for why this needs configuration rather than a smart default is below.
- [x] **Git Hygiene:** single commit, rebased on `dev`, no unrelated changes.
- [x] **Title Prefix:** `feat`.

# Changelog Entry

### Description

Deployments in regulated or corporate environments frequently need every user to see and accept a usage notice — an acceptable-use policy, a "do not paste client data" warning, a data-retention notice — before they can use the chat. Today Open WebUI has no way to do this:

- **Onboarding splash** (`OnBoarding.svelte`) is pre-auth, per-instance, and its text is hardcoded — it is not per-user and cannot be acknowledged.
- **Banners** are dismissible and their dismissal is stored in `localStorage`, so they reappear on another browser and are trivially skipped. Nothing is recorded server-side.
- **`ChangelogModal`** is admin-only and version-scoped.
- **`AccountPending`** is a terminal state for users awaiting approval, not something a user can pass through.

So today this is done by forking or by patching the frontend. This PR adds a first-class, opt-in equivalent.

**This is off by default.** With `USER_DISCLAIMER_CONTENT` empty — the default — nothing renders, no request is made, and behaviour is byte-for-byte identical to today. Existing installs need no migration: `Config.seed_defaults()` backfills the three new keys as empty strings on startup.

### Added

- Three settings, registered through the **existing** `PersistentConfig` mechanism (env var → `DEFAULT_CONFIG` → `Config`), so they behave like every other setting and are editable afterwards in the admin UI. No new configuration mechanism was introduced:

  | Env var | Meaning |
  | --- | --- |
  | `USER_DISCLAIMER_TITLE` | Heading. Falls back to a translated "Disclaimer". |
  | `USER_DISCLAIMER_CONTENT` | Markdown body. **Empty disables the feature.** |
  | `USER_DISCLAIMER_VERSION` | Free-form string. Changing it re-prompts every user. |

- A **User Disclaimer** section in *Admin Settings → Authentication*, directly after *Pending Accounts*, using the existing `AdminSettingSection` / `AdminSettingField` components and wired through the existing `ADMIN_CONFIG_KEYS` / `AdminConfig` plumbing.

- `POST /api/v1/auths/disclaimer/acknowledge`, which records the acknowledged version and a Unix timestamp on the user's settings (`ui.userDisclaimerVersion`, `ui.userDisclaimerAcknowledgedAt`).

- `UserDisclaimer.svelte`, modelled closely on the existing `AccountPending.svelte` (same layout, same `DOMPurify.sanitize(marked.parse(...))` handling of the configured markdown).

- Unit tests for the gating rule (`src/lib/utils/disclaimer.test.ts`, 15 cases).

### Changed

- `/api/config` returns the three values in the existing `ui` block — **only** for users whose role is already `user` or `admin`, matching how `pending_user_overlay_*` is scoped. The text is never exposed to anonymous callers.
- The admin changelog modal is deferred while the disclaimer is up, so the two cannot stack.

### Deprecated / Removed / Breaking Changes

- None.

### Security

Three deliberate choices, all of which I would like reviewed:

1. **The version and timestamp are taken from the server, never from the request body.** The endpoint accepts no payload. A client therefore cannot claim to have accepted a version it was never shown, nor backdate its consent.

2. **The endpoint is intentionally *not* behind the `settings.interface` permission.** My first version reused `POST /users/user/settings/update`, which is gated on that permission. Combined with "keep the gate up if the write fails", that locks a user out of the app entirely on any deployment where an admin has revoked `settings.interface`. Recording consent is not an interface preference, so it gets its own endpoint available to any verified user.

3. **A failed acknowledgement keeps the disclaimer up** and surfaces a toast. An acknowledgement that was not persisted is not an acknowledgement.

**Known limitation, stated plainly:** this is a *client-side* gate. It stops an ordinary user reaching the UI; it does not stop someone driving the API directly with a valid token. That matches how `AccountPending` behaves today and is the right level for an acceptable-use notice, but if you would rather this be enforced server-side (e.g. middleware rejecting requests until acknowledged) I am glad to take that direction — it is a larger and more invasive change, so I did not assume it. Likewise, the record stores *which version* was accepted and *when*, not a snapshot of the exact text; if you want true audit-grade evidence, storing the rendered text at acceptance time would be the next step.

---

### Additional Information

**On "smart defaults are preferred over new settings".** I want to address this directly rather than hope it goes unnoticed. The content of a legal or acceptable-use notice is inherently deployment-specific — there is no default text that could be correct for anyone, which is why the feature is disabled unless text is supplied. I did try to keep the surface minimal: it is three plain string settings that reuse the existing config mechanism end to end, with no new toggle (empty content *is* the off switch). The separate `USER_DISCLAIMER_VERSION` is the one addition that needs justifying: without it, a re-prompt after a policy change is impossible, and a boolean "has seen it" flag means the notice can never be updated — which defeats the purpose for the compliance use case this serves.

**Why the disclaimer replaces the app shell rather than overlaying it.** An earlier iteration rendered it as a sibling of the app shell. The app stayed mounted and tabbable behind the blur, keyboard shortcuts still fired, and `ChangelogModal` could stack on top of it — so it looked like a gate without being one. It is now rendered in place of the shell in `src/routes/(app)/+layout.svelte`, in the same `{#if}` chain as `AccountPending`.

**Note on process.** I am aware the checklist asks for a linked Discussion first, and that this has not been opened. I built the change before reading that requirement carefully, and I would rather be upfront about that than quietly open a Discussion and pretend the code came after. If you would prefer this as a Discussion first, say so and I will open one and close this. I also understand a PR is not a guarantee of merge and that the implementation may be rewritten or used only as a starting point — that is fine.

### AI tooling disclosure

Requested explicitly by me, in the spirit of the "No Unchecked AI Code" item:

**This change was written with the assistance of an AI coding tool (Claude Code).** It is not an unreviewed AI-generated PR. Specifically:

- Every line was read and reviewed by me before commit; several iterations were rejected and reworked.
- The lockout bug described under *Security* (#2) and the "overlay that did not actually block" bug described above were both found during that human review, not produced by it — the initial AI-generated implementation contained both.
- The rendered UI was inspected manually in the browser in light and dark mode, which is how the acknowledge button being invisible in light mode (white-on-white) was caught and fixed.
- I take responsibility for this code and will maintain the PR through review.

I would rather disclose this and have you weigh it than have it inferred later.

### Testing

**Automated**

- `npx vitest run` — 17 passed (15 new, in `src/lib/utils/disclaimer.test.ts`). The gating rule is deliberately extracted into `src/lib/utils/disclaimer.ts` so it is testable without mounting a component. Cases cover: disabled when content is empty/whitespace, shown on first login, hidden once acknowledged, re-prompt on version bump, no re-prompt when only the title changes, and that "never acknowledged" stays distinct from "acknowledged the empty default version".
- `npm run build` — clean.
- `npx prettier --check` on every file in this PR — clean.
- `ruff check --select=F --ignore=F401,F403,F405,F541,F811,F841` — clean.

**Manual**

- Fresh user, disclaimer configured → gate appears before the app, app shell is not mounted behind it, acknowledging persists and does not reappear on reload or in a different browser.
- Content left empty → nothing renders, no request issued, no behaviour change.
- Version bumped → every user is prompted again exactly once.
- Acknowledgement endpoint failing → disclaimer stays up, error toast shown, user is not silently let through.
- Rendered in light and dark mode.
- Markdown in the content renders and is sanitised.

**A note on CI, with evidence.** Two of the checks in this repo's workflows are currently failing on pristine `dev`, before and independently of this PR. I verified this by checking out `upstream/dev` into a clean worktree and running them there:

- `ruff format --check` → *"15 files would be reformatted"* on unmodified `dev`. The identical 15 files fail with this PR applied; none of the reported hunks are in code this PR touches, and `backend/open_webui/routers/auths.py` is clean.
- `npm run format` followed by `git diff --exit-code` → unmodified `dev` leaves 4 files dirty (`CalendarEventModal.svelte`, `RecursiveFolder.svelte`, `AccessControl.svelte`, `locales/languages.json`).
- `npm run i18n:parse` followed by `git diff --exit-code` → unmodified `dev` leaves 67 files dirty, because strings such as `Chat Variables` exist in `dev`'s source but not in its locale files.

I deliberately did **not** absorb any of that drift into this PR, because doing so would have added roughly 3,400 lines of unrelated locale and formatting churn and made the diff unreviewable. Consequently this PR's i18n changes are exactly the 10 new keys × 63 locale files (630 additions, 0 deletions), generated with `npm run i18n:parse` and then filtered down to only this feature's keys, preserving the parser's own ordering and formatting. German is filled in; the rest are empty strings for translators, as is convention here. Say the word if you would rather I include the full regenerated output instead.

### Screenshots or Videos

Happy to attach captures of the disclaimer in light and dark mode on request — I did not want to bloat the description with images before knowing whether the direction is wanted at all.


mhupfauer:
For anyone who needs this before the PR is resolved: I packaged the same idea as a **Filter function**, which installs on a stock instance with no fork or patch.

→ https://github.com/mhupfauer/open-webui/releases/tag/user-disclaimer-function-v0.1.3 (import the JSON via *Admin Panel → Functions → ⋯ → Import Functions*)

To be clear about what it is and isn't: **it is not a substitute for this PR**, and nothing about it is a request for maintainer time — it lives entirely in my fork.

Functions cannot ship frontend code, so it gates the user's first **chat message** rather than their first **login**. Users can sign in and browse the UI without having accepted anything. It writes the same two user-settings keys this PR uses (`ui.userDisclaimerVersion` / `ui.userDisclaimerAcknowledgedAt`), so acknowledgements carry over if this feature lands natively.

### One finding from it that is relevant here

Testing the function on a live instance turned up something that I think actually argues *for* gating at the UI level, as this PR does.

`routers/tasks.py` calls `process_pipeline_inlet_filter()` and then `generate_chat_completion()` directly — it never calls `process_filter_functions()`. So Function filters do not run for task requests at all. The practical effect: a user can type a message, be shown the disclaimer, click **Cancel** — and a generated chat title still comes back, because the text has already been sent to the task model.

I reproduced this exactly: typed a message with sensitive-looking content, declined the disclaimer, and got a title summarising it. Closing that from a filter is not possible; it needs Title/Tags/Follow-Up/Autocomplete/Query generation disabled at the deployment level.

This PR does not have that problem, because the user never reaches a chat input before acknowledging. That difference is a large part of why I still think the feature belongs in the app rather than in a plugin — but I appreciate that is my opinion and the call is yours.

Classic298:
Why not use an event function for it? Also i dont see a referenced discussion for this.

mhupfauer:
> Why not use an event function for it? Also i dont see a referenced discussion for this.

~See my above comment: there is no ability to reject the transmission to the title generator for example - should I creates discussion?~

Classic298:
Yes create discussion please. I can give a small guide. I am very very very sure this is possible with event function

mhupfauer:
> Yes create discussion please. I can give a small guide. I am very very very sure this is possible with event function

Thanks for the hint - will post a discussion and also the code how I solved this eventually 👍 
https://github.com/open-webui/open-webui/discussions/27656

PR #27606, opened 2026-07-27
feat: make token cap on task model generations configurable

### LINKED ISSUE #27604: issue: Context compaction stores truncated reasoning as the chat summary
### Check Existing Issues

### Installation Method

Docker

### Open WebUI Version

v0.11.0

_No response_

### Operating System

Ubuntu Server 24.04

### Browser (if applicable)

Firefox 153.0, Safari on iOS

### Confirmation

### Expected Behavior

Compaction works correctly with reasoning models.

### Actual Behavior

Tokens generated by reasoning count against `max_tokens`, which is statically assigned at `1000`. This causes generation to stop prematurely, and the resulting compacted context is subpar.

### Steps to Reproduce

1. Compact a conversation
2. Review LLM output upstream.

### Logs & Screenshots

LLM API log from a compaction attempt:
```
{
  "choices": [
    {
      "finish_reason": "length",
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "",
        "reasoning_content": "Here's a thinking process:\n\n1.  **Analyze User Input:**\n   - **Task:** Summarize conversation history to be compacted.\n   - **Instructions:** Preserve key decisions, preferences, constraints, files/artifacts/tools/code, current task state, unresolved questions, next steps. Be factual, specific, concise but complete. Do not invent details.\n   - **Previous Summary:** Empty (start from scratch).\n   - **Messages Being Compacted:** \n [SHORTED FOR BREVITY] . It outlined"
      }
    }
  ],
  "created": 1785163037,
  "model": "/models/Qwen3.6-35B-A3B-Q8_0.gguf",
  "system_fingerprint": "b10137-d2a818231",
  "object": "chat.completion",
  "usage": {
    "completion_tokens": 1000,
    "prompt_tokens": 2098,
    "total_tokens": 3098,
    "prompt_tokens_details": {
      "cached_tokens": 0
    }
  },
  "id": "chatcmpl-yVdQojph0YjjiIhic0w2NNSpquK36vCE",
  "timings": {
    "cache_n": 0,
    "prompt_n": 2098,
    "prompt_ms": 2776.249,
    "prompt_per_token_ms": 1.3232836034318398,
    "prompt_per_second": 755.6959047981649,
    "predicted_n": 1000,
    "predicted_ms": 22903.35,
    "predicted_per_token_ms": 22.90335,
    "predicted_per_second": 43.661735073690096
  }
}
```

Viewing the `contextSummary` for the chat via API:
```
          "contextSummary": "Here's a thinking process:\n\n1.  **Analyze User Input:**\n   - **Task:** Summarize conversation history to be compacted.\n   - **Instructions:** Preserve key decisions, preferences, constraints, files/artifacts/tools/code, current task state, unresolved questions, next steps. Be factual, specific, concise but complete. Do not invent details.\n   - **Previous Summary:** Empty (start from scratch).\n   - **Messages Being Compacted:** \n     [SHORTED FOR BREVITY] It outlined"

```

### Additional Information

The current behavior of utilizing truncated reasoning responses, and arguably truncated content responses, is flawed in both execution and methodology. Truncated responses fail silently to the user, and provide incomplete or potentially inaccurate context for the conversation, depending on where the response truncated. Furthermore, having `max_tokens` set to 1000 without a way to change it in the GUI and/or environment variables is not ideal, as this threshold is too low in this age of reasoning models which can exceed 4k tokens on reasoning alone.

For a fix, I am proposing the following:
- Add a configurable max output limit for compaction summaries, in Admin Panel > Interface > Chat, and send that value in place of the current literal. This alone resolves the reported behavior on every backend, since the value `_generate_summary` sets is the one that reaches the model. One note on approach: the existing read at `context_compaction.py:397` cannot be repaired in place as an alternative, because it reads `info.params` from `app.state.MODELS`, and `params` is deliberately stripped there in `utils/models.py:180` and `:223` to avoid exposing the system prompt to clients. The new setting should be read from config directly rather than routed through that lookup. See the footnote for why compaction is reading model params at all.
- Discard compaction responses with `"finish_reason": "length"`, whether or not `"content"` is empty.
- Remove the `reasoning_content` fallback at `context_compaction.py:442`. Reasoning text is not a summary, and accepting it is what makes this failure silent rather than visible.
- Surface the failure. `compact_messages_for_request` already raises and `middleware.py:2375` catches it and continues with full chat history, which is a reasonable outcome. `compact_chat_branch` has no equivalent guard, so the manual `/compact` endpoint currently returns a 500.

**Footnote on a permanent fix:** as noted above, the first item is a targeted fix. The underlying cause is that there is no shared way to resolve a model's parameters, so each caller reaches for a different source and the effective limit ends up depending on which backend serves the model.

The correct merge already exists, in `main.py:1086-1097`: global defaults from `models.default_params` (Admin Panel > Settings > Models > Model Defaults) as the base, the model's own row on top, request params on top of that. But it is inline in `chat_completion`, so the resolved values never leave that handler. `_generate_summary` calls `generate_chat_completion` in `utils/chat.py` directly and never enters it, so neither the global defaults nor chat Advanced Params apply to a compaction request.

Per-model params from the `model` table do reach the request, through the router, but what happens next differs per backend:

- OpenAI-compatible, not `api.openai.com`: the rename at `routers/openai.py:1267-1270` overwrites the model's value with the task value.
- `api.openai.com`, non-reasoning model: `:1272` deletes `max_tokens`, so the task value wins again.
- `api.openai.com`, reasoning model: `:139-142` overwrites in the opposite direction, and the model's value wins.
- Ollama: `payload.py:324-326` writes the task value to `num_predict` at the payload root, where it is not read; the model's value survives in `options` and wins.

`routers/tasks.py:157` carries the identical dead lookup for title generation, which suggests this is a pattern rather than a one-off.

A durable fix would be a small extraction rather than a redesign: move the merge in `main.py:1086-1097` into a shared helper, have `chat_completion` call it, and have the task paths call it too. Parameter resolution then has one precedence order and one source, and the per-backend divergence above becomes the only remaining variable.

Classic298:
Fixes #27604

> [!NOTE]

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -2157,6 +2157,9 @@ TASK_MODEL = os.getenv('TASK_MODEL', '')
 
 TASK_MODEL_EXTERNAL = os.getenv('TASK_MODEL_EXTERNAL', '')
 
+_TASK_MODEL_MAX_TOKENS = os.getenv('TASK_MODEL_MAX_TOKENS')
+TASK_MODEL_MAX_TOKENS = int(_TASK_MODEL_MAX_TOKENS) if _TASK_MODEL_MAX_TOKENS else None
+
 CONTEXT_COMPACTION_MODEL = os.getenv('CONTEXT_COMPACTION_MODEL', '')
 
 ENABLE_CONTEXT_COMPACTION = os.getenv('ENABLE_CONTEXT_COMPACTION', 'False').lower() == 'true'
@@ -3085,6 +3088,7 @@ DEFAULT_CONFIG = {
     'auth.admin.email': ADMIN_EMAIL,
     'task.model.default': TASK_MODEL,
     'task.model.external': TASK_MODEL_EXTERNAL,
+    'task.model.max_tokens': TASK_MODEL_MAX_TOKENS,
     'chat.context_compaction.model': CONTEXT_COMPACTION_MODEL,
     'chat.context_compaction.enable': ENABLE_CONTEXT_COMPACTION,
     'chat.context_compaction.token_threshold': CONTEXT_COMPACTION_TOKEN_THRESHOLD,
--- a/backend/open_webui/routers/tasks.py
+++ b/backend/open_webui/routers/tasks.py
@@ -25,6 +25,7 @@ from open_webui.utils.task import (
     emoji_generation_template,
     follow_up_generation_template,
     get_task_model_id,
+    get_task_model_max_tokens_payload,
     image_prompt_generation_template,
     moa_response_generation_template,
     query_generation_template,
@@ -40,6 +41,7 @@ router = APIRouter()
 TASK_CONFIG_KEYS = {
     'TASK_MODEL': 'task.model.default',
     'TASK_MODEL_EXTERNAL': 'task.model.external',
+    'TASK_MODEL_MAX_TOKENS': 'task.model.max_tokens',
     'TITLE_GENERATION_PROMPT_TEMPLATE': 'task.title.prompt_template',
     'IMAGE_PROMPT_GENERATION_PROMPT_TEMPLATE': 'task.image.prompt_template',
     'ENABLE_AUTOCOMPLETE_GENERATION': 'task.autocomplete.enable',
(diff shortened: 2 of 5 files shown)


tjbck:
f0bfcd40976dfb1e1876f86b39d6a659d894d15e

PR #27627, opened 2026-07-28
feat: add a File Processing model capability to skip extraction and pass raw files through

### LINKED ISSUE #12228: feat: uploading files without backend processing
### Check Existing Issues

### Problem Description

I am trying to use open webui as a frontend webui, and trying to process user reuqest via chating ui, which open webui has a great user experience and works stable. Thanks for your great work.

We are trying to add more features to make good use of our self-hosted aigc backend. This feature is something like: user upload some files and ask some question", and then our backend (which is driven by pipeline) get original file and processing it and send sse events to pipeline and then emit it to open webui front end. RAG feature or OCR feature(which i don't know) is always trying to extract data into text, which is not a must for our use case. This takes a lot of time and affect user experience. Also, it seems open webui always trying to send raw text to open webui backend, it's also not required by our feature and affect user experience.

So I am requesting one feature that:
1. bypass content extraction feature completely (I already check "bypass vector database" inside document tab, but seems not disabled completely)
2. pipeline can access user uploaded file via raw url(of course respecting cors setting), so it can do more stuff inside.

Could you please have a discussion of this feature, thanks.

1. bypass content extraction feature completely (I already check "bypass vector database" inside document tab, but seems not disabled completely)
2. pipeline can access user uploaded file via raw url(of course respecting cors setting), so it can do more stuff inside.

### Alternatives Considered

_No response_

### Additional Context

_No response_

silentoplayz:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #12228 — uploading files without backend processing. Design was posted for discussion on the issue before implementation: https://github.com/open-webui/open-webui/issues/12228#issuecomment-5097897470
- [ ] **Documentation:** The "When to disable File Processing" section below is written to serve as the docs basis.
- [x] **Testing:** Three rounds of manual testing were performed by the author (details below), plus an automated backend harness. Verification by #12228's requesters against their own pipe/backend setups is explicitly invited before merge.
- [x] **Architecture:** Implemented as a model capability following the existing pattern (`file_upload`, `file_context`), default **on** — zero behavior change for existing deployments. Reuses the existing `?process=false` upload path and the `metadata.files` stamping pattern established by the per-file `context` field.

# Changelog Entry

### Description

Adds a **File Processing** model capability (default on). When disabled on a model, chat uploads for that model skip content extraction entirely — no OCR, no embedding, instant upload — and the raw file is passed through to the consuming backend, which fetches the original bytes via the existing `GET /api/v1/files/{id}/content` using the file ids already present in `metadata.files`.

This implements #12228: the backend `?process=false` upload parameter and the raw-content endpoint already existed; what was missing was the chat-flow wiring and graceful handling of extraction-less files everywhere the system previously assumed extracted content.

**When to disable File Processing** (reviewer/docs summary):

- **Pipe/pipeline backends that do their own file handling** — Open WebUI acts as the frontend; the pipe fetches the raw file and runs its own OCR/parsing/analysis. Open WebUI's extraction is redundant double-processing that delays the upload (the core complaint in #12228).
- **Backends forwarding raw files to natively multimodal APIs** — providers that ingest raw PDFs/media directly, where the original file is more useful than an extraction dump.
- **Avoiding extraction/embedding cost** for uploads whose text will never be used.

**When NOT to disable it**: on ordinary text models. A text-only LLM's *only* window into a document is Open WebUI's extraction — with processing off, the model can acknowledge the file (name/metadata) but cannot read its contents. This is the feature's contract, not a malfunction, and the UI/tooling states it explicitly at every surface (see below). The capability defaults to on precisely so this state is only ever reached deliberately.

### What the change does, end to end

1. **Model editor**: a "File Processing" checkbox in Capabilities (hidden when File Upload is off, like File Context), with a tooltip spelling out the raw-attachment behavior.
2. **Chat uploads**: when **every** selected model has the capability off, the upload uses the existing `?process=false` path and the file item is stamped `processed: false` (the same `metadata.files` stamping pattern as the per-file `context` field). Mixed selections still process; image uploads are unaffected.
3. **Retrieval**: stamped files are excluded from RAG/full-context source building but remain in `metadata.files` for the pipe/backend. `add_file_context` still announces the file reference (id/URL/name) to native-FC models unchanged.
4. **Builtin tools**: `view_file`/`view_knowledge_file` return an explicit note (`processed: false` + "uploaded without content extraction (raw attachment)") instead of empty content, so models report the truth rather than "the file is empty"; `query_chat_files` excludes raw files and explains itself when only raw files are attached; `list_chat_files`/`grep` degrade gracefully (metadata-only / no matches).
5. **Citations**: view-tool results without content no longer emit an empty source citation (also fixes pre-existing empty citations from view-tool error results).
6. **File chip modal**: opens on Preview for previewable raw files, stays on Content (showing the explanation) for non-previewable types like `.txt`; the Full/Focused content-mode switch is hidden since it is meaningless without extracted content.
7. **Hardening**: the full-context retrieval branch no longer raises `AttributeError` on file records with `data = None` (reachable pre-PR via the existing `?process=false` API path).

### Added

- A **File Processing** model capability (default on). Disabling it makes chat uploads for that model skip content extraction and pass files through as raw attachments for pipes/backends to consume via the raw content endpoint, with explicit raw-attachment messaging across tools, citations, and the file modal.

### Fixed

- Fixed a potential `AttributeError` in full-context retrieval for file records without extraction data (reachable via the existing `?process=false` upload API).
- Fixed empty source citations being emitted for `view_file`/`view_knowledge_file` results that carry no content (including error results).

---

### Additional Information

**What `process=false` means downstream.** A raw attachment has no extracted content, so there is nothing to query, nothing to cite, and nothing to retrieve. Those files stay in `metadata.files` so the model or pipe still sees them, but they are excluded from retrieval and from citation building. The file modal opens on the preview tab when the file type has one and otherwise stays on the content note, since there is no extracted text to show. In the model editor, the `file_processing` and `file_context` capabilities are hidden when `file_upload` is off, because they have no meaning without it. The upload path falls back to a raw upload only when every selected model has opted out of processing.

- **Request for verification from the original requesters**: @wangjiyang (#12228) and @vanmilleru — this branch (`silentoplayz:feat/file-processing-capability`) implements the design discussed on the issue. Testing against your actual pipe/backend setups (raw fetch via `/api/v1/files/{id}/content` from the ids in `metadata.files`, upload latency with processing skipped) would be the definitive confirmation that this matches what you asked for, and is warmly invited before merge.

### Verification Performed

**Manual (author, three rounds on a live instance)** — the rounds deliberately shook out three real defects that are fixed in this PR: (1) builtin tools serving empty content led the model to claim files were "empty or unreadable" → now the explicit note, confirmed live with the model responding honestly; (2) a phantom "1 Source" citation from the empty tool result → now suppressed, confirmed live; (3) non-previewable raw files (.txt) stranded the file modal on an empty view → now handled. Also verified: instant uploads on large PDFs with the capability off, unchanged behavior for processed files and for models with the capability on, capability checkbox visibility rules, and Preview-default on raw PDF chips.

**Automated (AI-copilot harness, disclosed as such)** — real functions against a real scratch database containing one raw (`data=None`) and one processed file record; all five checks pass:

1. `view_file(raw)` → raw-attachment note (`processed: false`, no `content` key)
2. `view_file(processed)` → extracted content, byte-identical behavior
3. `query_chat_files(raw only)` → explanatory error, no crash
4. Citation builder → no source for the note, real source for content results
5. Real `get_sources_from_items` on a raw file in full-context mode → completes without `AttributeError`, empty documents

**Live end-to-end pipe run (author, this branch)** — the consuming-backend flow this feature exists for, verified against a deterministic provider stand-in ([details + the reusable test pipe in the PR comments](https://github.com/open-webui/open-webui/pull/27627#issuecomment-5113523403)): with File Processing off on a pipe-backed model, a chat PDF attachment automatically uploaded via `?process=false` (network-captured, instant, no extraction), the saved file item carried `processed: false`, the pipe fetched the raw bytes and built an OpenAI base64 `file_data` part, and an OpenAI-compatible echo server confirmed receipt of the document with a **SHA-256 byte-identical to the source file's pre-upload hash** — the attachment traversed upload → storage → pipe → provider untouched.

> [!NOTE]


ghost:
@silentoplayz 
can it choose between the  [ openai base64 format ](https://developers.openai.com/api/docs/guides/file-inputs#base64-encoded-files) or the fileid format?

 some api doesn't support the fileid way

silentoplayz:
Good question — the answer is that this PR is deliberately format-agnostic, so **both work, and the choice is yours**.

The `id` in `metadata.files` is Open WebUI's *internal* file handle, not a provider file id — nothing here sends a file-id to any provider API. With File Processing off, your pipe/backend receives that handle and fetches the untouched original bytes:

```
GET /api/v1/files/{id}/content
```

From there you emit whichever shape the target API accepts:

- **base64 inline** — encode the bytes into a `file` content part (`file_data`, per the OpenAI file-inputs guide you linked); works on every OpenAI-compatible API, no Files-API support needed;
- **provider file id** — upload the bytes to the provider's Files API first and reference the returned id, if the provider supports it and you prefer it (e.g. for reuse across turns).

So "some APIs don't support the fileid way" is exactly why the PR stops at handing you raw bytes instead of picking a wire format for you.

Separate from pipes: Open WebUI itself does not currently forward raw documents to providers on *direct* connections in any mode or format (images yes, documents no). If you'd want Open WebUI to natively embed base64 `file_data` parts on direct connections when File Processing is off, that's a sensible follow-up feature on top of this capability — worth its own issue once this lands.

Does the pipe-side flow above cover your use case? Testing it against your actual setup on this branch would be great confirmation.

ghost:
@silentoplayz 
thanks for the clarification!! i really hope this could be merged soon @tjbck

also could you share a pipe function for testing with base64 inlined format?

silentoplayz:
@vanmilleru Sure — here's a minimal, complete pipe for exactly that test. It picks the attachments up from `metadata.files`, reads the untouched bytes through the same machinery the `/api/v1/files/{id}/content` endpoint uses (so it works in-process, no token juggling), inlines them as OpenAI base64 `file` content parts, and forwards to any OpenAI-compatible endpoint you point the valves at:

```python
"""
title: Raw Attachment base64 Test Pipe
description: Forwards raw (File Processing off) attachments to an OpenAI-compatible API as base64 file inputs. Testing companion for PR #27627.
version: 0.1
"""

import asyncio
import base64

import aiohttp

from open_webui.models.files import Files
from open_webui.storage.provider import Storage


class Pipe:
    class Valves(BaseModel):
        API_BASE_URL: str = Field(default="https://api.openai.com/v1")
        API_KEY: str = Field(default="")
        MODEL_ID: str = Field(default="gpt-4o")

    def __init__(self):
        self.valves = self.Valves()

    async def pipe(self, body: dict, __metadata__: dict = {}):
        # Attachments arrive in metadata.files. On the PR branch, files uploaded
        # with File Processing off are stamped `processed: False` — filter on that
        # if you only want to inline the raw ones.
        file_parts = []
        for item in (__metadata__ or {}).get("files") or []:
            if item.get("type") != "file":
                continue

            file = await Files.get_file_by_id(item.get("id"))
            if not file or not file.path:
                continue

            # Same read path the /api/v1/files/{id}/content endpoint uses
            path = await asyncio.to_thread(Storage.get_file, file.path)
            with open(path, "rb") as f:
                raw = f.read()

            mime = (file.meta or {}).get("content_type") or "application/octet-stream"
            file_parts.append(
                {
                    "type": "file",
                    "file": {
                        "filename": (file.meta or {}).get("name", file.filename),
                        "file_data": f"data:{mime};base64,{base64.b64encode(raw).decode()}",
                    },
                }
            )

        # Inline the file parts into the last user message
        messages = body.get("messages") or []
        if file_parts and messages:
            last = messages[-1]
            content = last.get("content")
            if isinstance(content, str):
                content = [{"type": "text", "text": content}]
            last["content"] = file_parts + content

        async with aiohttp.ClientSession() as session:
            async with session.post(
                f"{self.valves.API_BASE_URL}/chat/completions",
                headers={"Authorization": f"Bearer {self.valves.API_KEY}"},
                json={
                    "model": self.valves.MODEL_ID,
                    "messages": messages,
                    "stream": False,
                },
            ) as res:
                data = await res.json()
                if res.status != 200:
                    return f"Error {res.status}: {data}"
                return data["choices"][0]["message"]["content"]
```

Test recipe on the branch: create a model backed by this pipe → untick **File Processing** in its capabilities → attach a PDF in chat (upload is instant, no extraction) → ask about the document. The provider receives the original bytes base64-inlined, so the answer should reflect actual document content.

Notes: non-streaming for simplicity; it inlines every file attachment (add `if item.get("processed") is not False: continue` to restrict to raw-mode uploads only); images already travel in message content the normal way, this handles document files. If your target API wants the provider-file-id format instead, replace the `file_parts.append(...)` block with an upload to its Files API and reference the returned id — the raw bytes in `raw` are the same either way.

silentoplayz:
Follow-up: the test pipe posted above has now been **verified end-to-end on a live instance running this branch**, with a deterministic provider stand-in. Full chain, every link observed:

1. **Capability gate**: pipe model configured with File Processing off.
2. **Upload path**: attaching a PDF in chat automatically fired `POST /api/v1/files/?process=false` (network-captured) — instant, no extraction.
3. **Metadata stamp**: the saved message's file item carries `processed: false` per this PR's contract.
4. **Pipe mechanics**: the pipe (verbatim as posted) read `metadata.files`, fetched the raw bytes in-process, and built the OpenAI base64 `file_data` content part.
5. **Provider receipt**: the valves were pointed at a local OpenAI-compatible echo server that fingerprints whatever it receives. Its report, rendered back into the chat: correct filename and mime, `bytes: 310`, `starts_with_pdf_magic: true`, and a **SHA-256 byte-identical to the source PDF's hash computed before upload** — the document traversed upload → storage → pipe → base64 → provider untouched.

@vanmilleru — that means the pipe recipe is confirmed working mechanics-wise on this branch; the one thing this setup deliberately didn't cover is provider-side *comprehension* of `file_data`, which your run against a real file-input-capable API would complete.

PR #27630, opened 2026-07-28
fix: restore Enter-to-confirm in confirmation dialogs by explicitly focusing Confirm

### LINKED ISSUE #27629: issue: Enter in confirmation dialogs activates Cancel instead of confirming (v0.11.0 and dev)
### Affected Version

`dev` (regression introduced after v0.11.0, by the interaction described below — not present in v0.11.0)

### Expected Behavior

Pressing Enter while a confirmation dialog is open (delete chat, delete message, delete folder/model/knowledge, etc.) confirms the action, as it always has.

### Actual Behavior

Pressing Enter closes the dialog **without** performing the action. Nothing is deleted; the dialog silently cancels.

### Steps to Reproduce

1. On current `dev`, delete any chat from the sidebar (or delete one of your own messages in a chat).
2. When the confirmation dialog appears, press Enter.
3. The dialog closes and the chat/message is still there.

### Root Cause

#27491 correctly changed `ConfirmDialog`'s window-level Enter handler to defer to the focused control (`target.closest('a, button, textarea')` → let native activation happen). That fix's stated behavior — "Enter anywhere else still confirms as before" — is defeated by where focus actually lands: the dialog's focus trap is created with **no `initialFocus`**, so focus-trap falls back to the first tabbable node in the container, which is the **Cancel** button (as #27491's own description documents). The dialog therefore opens with Cancel invisibly focused, and Enter now natively activates Cancel.

The underlying defect is that `ConfirmDialog` never deliberately chose its initial focus — it never mattered while the old handler hijacked Enter globally, and #27491 (correctly) made focus meaningful, exposing the accidental Cancel default.

### Proposed Fix

Give the focus trap an explicit `initialFocus`: any input inside the dialog first (so typing dialogs still focus their field immediately), otherwise the **Confirm** button. This restores Enter-to-confirm with a *visible* focus ring on Confirm — preserving everything #27491 fixed (the focused control's label always matches what Enter does; Tab → Cancel → Enter cancels; Enter in the textarea inserts a newline).

Fix is implemented and manually verified; PR follows immediately.

silentoplayz:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #27629
- [ ] **Documentation:** No documentation changes required.
- [x] **Testing:** Manually verified by the author on a live instance (details below).
- [x] **Architecture:** Completes #27491 rather than reverting any part of it: initial focus becomes a deliberate choice at the layer that owns it (the focus trap's configuration), and Enter semantics follow natively from focus.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

On current `dev`, pressing Enter in any confirmation dialog (delete chat, delete message, delete folder/model/knowledge, and the ~40 other `ConfirmDialog` call sites) closes the dialog **without** performing the action — Enter effectively always cancels.

### Root Cause

#27491 correctly changed the window-level Enter handler to defer to the focused control:

```js
if (target instanceof Element && target.closest('a, button, textarea')) {
	return; // let the focused control act on Enter itself
}
```

That fix's stated behavior — "Enter anywhere else still confirms as before" — is defeated by where focus actually lands. The dialog's focus trap is created with no `initialFocus`:

```js
focusTrap = FocusTrap.createFocusTrap(modalElement);
```

so focus-trap falls back to the first tabbable node in the container, which is the **Cancel** button (a fallback #27491's own description documents). The dialog opens with Cancel invisibly focused, and Enter now natively activates it.

The underlying defect is that `ConfirmDialog` never deliberately chose its initial focus. It never mattered while the old handler hijacked Enter globally; #27491 (correctly) made focus meaningful, which exposed the accidental Cancel default.

### Fix

Make initial focus explicit on the trap: any input inside the dialog first (so typing dialogs still focus their field immediately, matching the previous first-tabbable behavior for the `input` variants), otherwise the **Confirm** button:

```diff
-			focusTrap = FocusTrap.createFocusTrap(modalElement);
+			focusTrap = FocusTrap.createFocusTrap(modalElement, {
+				// Focus any input first so typing works immediately; otherwise focus
+				// Confirm so Enter confirms (the trap's default falls back to the
+				// first tabbable node, which is the Cancel button).
+				initialFocus: () =>
+					modalElement.querySelector('input, textarea, select') ?? confirmButtonElement ?? undefined
+			});
```

plus a `bind:this={confirmButtonElement}` on the Confirm button.

Everything #27491 established is preserved, and the keyboard model is now fully coherent with no special cases:

- The dialog opens with **Confirm focused**, and Enter does exactly what the focused control's label says (the WCAG 3.2.2 property #27491 fixed). The focus ring renders per the browser's `:focus-visible` heuristics: immediately for keyboard-driven interaction, and from the first keypress onward when the dialog was opened with the mouse — Enter confirms either way.
- Tab → Cancel → Enter cancels; Escape cancels.
- Enter inside the `input=true` textarea still inserts a newline; Enter inside a text/password input still confirms (form-submit convention, via the existing fallback).
- A markdown link in `message` is no longer the initial focus target either.

Implementation notes, verified against the focus-trap source in `node_modules` rather than assumed: the function form of `initialFocus` is supported, and a function returning `null` **throws** (`getNodeForOption`) while `undefined` falls back to the library default — hence the trailing `?? undefined`, which turns a theoretical missing-ref edge into graceful fallback instead of a crash.

### Fixed

- Fixed Enter no longer confirming confirmation dialogs (deletes silently cancelling), a `dev` regression from the interaction between #27491's corrected Enter handling and the focus trap's accidental initial focus on Cancel.

---

### Additional Information

**Why focus lands where it does.** Any `input`, `textarea` or `select` in the dialog is focused first, so typing works immediately. Only when there is none does focus go to Confirm, which is what makes plain Enter confirm. The focus trap's own default falls back to the first tabbable node, and that is the Cancel button, so relying on it would make Enter cancel instead.

- @Classic298 — this completes #27491 rather than reverting any part of it; your guard is what makes the explicit initial focus meaningful. Flagging in case you want to double-check the a11y angle.

### Verification Performed

Manual, by the author on a live instance running this branch: deleting a chat and deleting an own message — plain Enter now performs the deletion; Escape still cancels. Focus demonstrably starts on Confirm: the first Tab wraps to Cancel (first tabbable) and the second returns to Confirm with the visible ring, exactly as expected when the trap's initial focus is the last tabbable node. The input-focus branch (`input, textarea, select` preferred over Confirm) is verified by code trace against the focus-trap source, not manually.

### Screenshots or Videos

- N/A (keyboard-interaction fix; behavior described above).

> [!NOTE]

### DIFF
--- a/src/lib/components/common/ConfirmDialog.svelte
+++ b/src/lib/components/common/ConfirmDialog.svelte
@@ -36,6 +36,7 @@
 	}
 
 	let modalElement = null;
+	let confirmButtonElement: HTMLButtonElement | null = null;
 	let mounted = false;
 
 	let focusTrap: FocusTrap.FocusTrap | null = null;
@@ -81,7 +82,10 @@
 	$: if (mounted) {
 		if (show && modalElement) {
 			document.body.appendChild(modalElement);
-			focusTrap = FocusTrap.createFocusTrap(modalElement);
+			focusTrap = FocusTrap.createFocusTrap(modalElement, {
+				initialFocus: () =>
+					modalElement.querySelector('input, textarea, select') ?? confirmButtonElement ?? undefined
+			});
 			focusTrap.activate();
 
 			window.addEventListener('keydown', handleKeyDown);
@@ -195,6 +199,7 @@
 						{cancelLabel}
 					</button>
 					<button
+						bind:this={confirmButtonElement}
 						class="text-sm bg-gray-900 hover:bg-gray-900/90 text-gray-100 dark:bg-gray-100 dark:hover:bg-gray-100/90 dark:text-gray-800 font-normal w-full py-1.5 rounded-full transition"
 						on:click={() => {
 							confirmHandler();
--- a/src/lib/components/common/ConfirmDialog.svelte
+++ b/src/lib/components/common/ConfirmDialog.svelte
@@ -36,6 +36,7 @@
 	}
 
 	let modalElement = null;
+	let confirmButtonElement: HTMLButtonElement | null = null;
 	let mounted = false;
 
 	let focusTrap: FocusTrap.FocusTrap | null = null;
@@ -81,7 +82,13 @@
 	$: if (mounted) {
 		if (show && modalElement) {
 			document.body.appendChild(modalElement);
-			focusTrap = FocusTrap.createFocusTrap(modalElement);
+			focusTrap = FocusTrap.createFocusTrap(modalElement, {
+				// Focus any input first so typing works immediately; otherwise focus
+				// Confirm so Enter confirms (the trap's default falls back to the
+				// first tabbable node, which is the Cancel button).
+				initialFocus: () =>
+					modalElement.querySelector('input, textarea, select') ?? confirmButtonElement ?? undefined
+			});
 			focusTrap.activate();
 
 			window.addEventListener('keydown', handleKeyDown);
@@ -195,6 +202,7 @@
 						{cancelLabel}
 					</button>
 					<button
+						bind:this={confirmButtonElement}
 						class="text-sm bg-gray-900 hover:bg-gray-900/90 text-gray-100 dark:bg-gray-100 dark:hover:bg-gray-100/90 dark:text-gray-800 font-normal w-full py-1.5 rounded-full transition"
 						on:click={() => {
 							confirmHandler();


Classic298:
draft?

silentoplayz:
> draft?

I suppose the linked issue (https://github.com/open-webui/open-webui/issues/27629) should've been a discussion post. This is only drafted because of what was said over on Discord.

Edit: Closing this. If it's wanted, it can be picked back up.

PR #27658, opened 2026-07-28
feat: emit user.updated on OAuth profile sync

Classic298:
With OAUTH_UPDATE_NAME_ON_LOGIN, OAUTH_UPDATE_EMAIL_ON_LOGIN or OAUTH_UPDATE_PICTURE_ON_LOGIN enabled, OWUI rewrites the stored name, email and profile picture from the IdP claims on every SSO login. Those writes emitted nothing, so an account whose email or display name is driven entirely by the IdP could change underneath every event consumer without a single event.

Emits one user.updated per login carrying data.updated_fields, published only when at least one field was actually persisted, so a login that changes nothing stays silent. This matches how the admin user-update endpoint reports the same fields, and the payload carries source "oauth" and data.provider like the other events on this callback.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -1955,6 +1955,8 @@ class OAuthManager:
                         data={'role': determined_role, 'provider': provider},
                     )
 
+                synced_fields = []
+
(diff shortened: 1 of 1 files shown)


tjbck:
Please adhere to the conventional naming.

PR #27664, opened 2026-07-28
feat: download Open Terminal files linked in a chat message

### LINKED ISSUE #27650: feat: Render Open Terminal output files as downloadable attachments in chat
### Check Existing Issues

- [x] I have searched all existing **open AND closed** issues and discussions and found none comparable to my request.

### Verify Feature Scope

- [x] I believe this feature request is appropriately scoped for the Issues section as described above.

### Problem Description

When a model creates a file through the official Open Terminal integration, the file is successfully written and immediately becomes available in the Open Terminal file navigator.

The file navigator can:

- locate the file;
- preview supported formats;
- download the file using its existing authenticated download action.

However, the generated file is not rendered as a downloadable attachment inside the assistant message.

The model may instead return a Markdown link based on the local filesystem path, for example:

```text
/home/user/teste-terminal.xlsx
```

The browser resolves this relative path against the Open WebUI domain:

```text
https://<open-webui-domain>/home/user/teste-terminal.xlsx
```

This returns `404 Not Found` because `/home/user/...` is an Open Terminal filesystem path, not a public Open WebUI route.

The user is then required to:

1. open the Open Terminal sidebar;
2. navigate to the correct directory;
3. locate the generated file;
4. select it;
5. use the download button in the file navigator.

This is especially confusing because the assistant explicitly states that the file is available for download and displays a clickable link, but that link does not work.

The file itself is valid. Only its delivery inside the chat is missing.

### Proposed Solution

When an Open Terminal tool creates, writes, or displays a file, Open WebUI should be able to render that file as a downloadable attachment inside the assistant message.

For example:

```text
📎 teste-terminal.xlsx

[Preview] [Download]
```

The preferred implementation would reuse the same authenticated download action already used by the Open Terminal file navigator.

A possible flow would be:

1. An Open Terminal tool such as `display_file`, `write_file`, or a command execution result identifies an output file.

2. The tool result includes structured file metadata, for example:

```json
{
  "type": "terminal_file",
  "terminal_id": "<terminal-connection-id>",
  "path": "/home/user/teste-terminal.xlsx",
  "name": "teste-terminal.xlsx",
  "content_type": "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
}
```

3. Open WebUI stores this metadata as part of the tool result or assistant message.

4. The chat frontend renders a file attachment component instead of relying on the model to construct a Markdown URL.

5. Clicking `Download` invokes the existing authenticated Open Terminal download handler used by the file navigator.

6. Clicking `Preview` may reuse the existing file preview component when the file type is supported.

The download must continue to respect:

- the current authenticated user;
- access to the selected Open Terminal connection;
- any Open Terminal multi-user filesystem restrictions;
- the configured file-browser root;
- disabled or unavailable terminal connections.

The local filesystem path should never need to become publicly accessible.

#### Optional explicit tool behavior

`display_file` could optionally support a parameter such as:

```text
attach_to_chat=true
```

This would allow the model or tool implementation to explicitly indicate that the displayed file should also appear as a chat attachment.

Open WebUI could also automatically apply this behavior when a tool result identifies a newly created file.

#### Alternative backend delivery

If reusing the Open Terminal download action is not practical, Open WebUI could copy or register the terminal file in its own file storage and return an authenticated file endpoint.

For installations using S3-compatible storage, another possible fallback would be a short-lived signed download URL.

These alternatives should remain implementation details. From the user's perspective, the result should simply be a downloadable file attachment in the chat.

#### Use the Open Terminal file navigator manually

The existing file navigator works correctly, but requiring users to manually locate every generated output adds unnecessary steps.

The assistant already knows which file was created, and Open WebUI already knows which terminal connection and filesystem path were involved.

#### Instruct the model not to create Markdown links

A system prompt can tell the model not to convert local paths into links and instead instruct the user to use the sidebar.

This prevents the invalid `404` link, but it does not provide a direct download experience.

#### Publish the Open Terminal home directory through NGINX

A directory such as `/home/user/downloads` could be exposed as `/downloads/`.

This would create working links, but introduces security and operational problems:

- terminal files may become publicly accessible;
- authorization would be separate from Open WebUI;
- filenames would need to be unpredictable;
- files would require automatic expiration and cleanup;
- multi-user isolation would be difficult;
- internal filesystem paths would become coupled to public URLs.

#### Upload every generated file to S3 or MinIO

A separate publishing service could upload terminal outputs and return signed URLs.

This is technically possible, but requires:

- additional credentials;
- custom publishing logic;
- object lifecycle rules;
- temporary URL management;
- user ownership handling;
- synchronization between Open Terminal and external storage.

This should not be required when Open WebUI already has an authenticated download action for the same file.

#### Register the file through the Open WebUI Files API

The terminal output could be copied into Open WebUI file storage.

This is a valid implementation alternative, but requires associating the uploaded file with the correct user and chat. Reusing the existing Open Terminal download mechanism appears simpler and avoids duplicating the file.

### Additional Context

Environment:

```text
Open WebUI: v0.11.0
Installation: Docker Compose
Open Terminal: configured through the official Open Terminal integration
Function calling: Native
Chat model: openai/gpt-4.1-mini through an OpenAI-compatible provider
Reverse proxy: NGINX
```

Reproduction steps:

1. Configure an Open Terminal connection in Open WebUI.

2. Enable Open Terminal tools for a model using Native function calling.

3. Ask the model:

```text
Create an XLSX file named teste-terminal.xlsx with the columns
Name, Department, and Status, containing five fictional records.
Make the file available for download.
```

4. The model successfully creates:

```text
/home/user/teste-terminal.xlsx
```

5. The Open Terminal file navigator refreshes and displays the file.

6. The XLSX preview works correctly.

7. The file navigator's download button downloads the file correctly.

8. The assistant response presents a Markdown link based on the local path.

9. Clicking the link opens:

```text
https://<open-webui-domain>/home/user/teste-terminal.xlsx
```

10. Open WebUI returns:

```text
404 Not Found
```

Expected result:

The assistant message should contain a downloadable attachment backed by the same authenticated action already available in the Open Terminal file navigator.

Related discussion:

```text
#14658
```

That discussion concerns generic generated-file links such as:

```text
sandbox:/mnt/data/file.pptx
```

This request is more narrowly scoped to the official Open Terminal integration, where Open WebUI already has:

- the terminal connection;
- the filesystem path;
- the authenticated user;
- the file preview;
- a working download action.

The requested feature is primarily to expose that existing capability directly inside the chat message.

Classic298:
When a model creates a file through the Open Terminal integration it routinely presents that file to the user as a markdown link, using either the bare filesystem path (`/home/user/report.xlsx`) or an invented scheme (`file://`, `sandbox:/mnt/data/…`). The browser resolves those against the Open WebUI origin, so clicking them returns 404 even though the file exists and the Open Terminal file browser can preview and download it. The user has to open the sidebar and locate the file by hand.

Such links now download the file through the same authenticated action the file browser already uses. The href is left exactly as the model wrote it; the click is intercepted at render time, so stored message content is untouched and nothing is injected into the model's output.

A link is treated as a terminal path only inside assistant output, only while a terminal is attached to the chat, and only when the href is not an ordinary web link: a scheme other than `file:` or `sandbox:` is rejected, as are protocol-relative URLs and `.`/`..` segments, and the path must have both a directory and a file extension, which is what a generated file looks like and what Open WebUI's own routes never do. User messages, citations, uploaded documents and channel messages are excluded, so only content the model produced can point at the terminal filesystem.

Fixes #27650

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Messages/ContentRenderer.svelte
+++ b/src/lib/components/chat/Messages/ContentRenderer.svelte
@@ -1,5 +1,5 @@
 <script>
-	import { onDestroy, onMount, tick, getContext } from 'svelte';
+	import { onDestroy, onMount, tick, getContext, setContext } from 'svelte';
 	const i18n = getContext('i18n');
 
 	import Markdown from './Markdown.svelte';
@@ -15,6 +15,9 @@
 	} from '$lib/stores';
 	import FloatingButtons from '../ContentRenderer/FloatingButtons.svelte';
 	import { createMessagesList, replaceOutsideCode } from '$lib/utils';
+
+	// Only assistant output may point at paths on the terminal
+	setContext('allowTerminalFileLinks', true);
 
 	/**
 	 * Extracts all top-level <details>...</details> blocks from content,
--- a/src/lib/components/chat/Messages/Markdown/MarkdownInlineTokens.svelte
+++ b/src/lib/components/chat/Messages/Markdown/MarkdownInlineTokens.svelte
@@ -7,9 +7,12 @@
 	import { goto } from '$app/navigation';
 
 	const i18n = getContext('i18n');
+	const allowTerminalFileLinks = getContext<boolean>('allowTerminalFileLinks');
 
 	import { WEBUI_BASE_URL } from '$lib/constants';
-	import { copyToClipboard, unescapeHtml } from '$lib/utils';
+	import { selectedTerminalId, showControls, showFileNavPath } from '$lib/stores';
+	import { copyToClipboard, displayFileHandler, unescapeHtml } from '$lib/utils';
+	import { terminalPath } from '$lib/utils/terminal';
 
 	import Image from '$lib/components/common/Image.svelte';
 	import KatexRenderer from './KatexRenderer.svelte';
@@ -46,10 +49,19 @@
 	};
 
 	/**
-	 * Handle link clicks - intercept same-origin app URLs for in-app navigation
+	 * Handle link clicks - open terminal paths, intercept same-origin app URLs
 	 */
 	const handleLinkClick = (e: MouseEvent, href: string) => {
 		try {
+			// A filesystem path points at the terminal the chat is attached to, not an app
+			// route, so open it in the file browser instead of resolving it against the origin
+			const path = allowTerminalFileLinks && $selectedTerminalId ? terminalPath(href) : null;
+			if (path) {
+				e.preventDefault();
+				displayFileHandler(path, { showControls, showFileNavPath });
+				return;
+			}
+
 			const url = new URL(href, window.location.origin);
 			// Check if same origin and an in-app route
 			if (
--- a/src/lib/components/chat/Messages/MultiResponseMessages.svelte
+++ b/src/lib/components/chat/Messages/MultiResponseMessages.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import { onMount, tick, getContext } from 'svelte';
+	import { onMount, tick, getContext, setContext } from 'svelte';
 	import { createEventDispatcher } from 'svelte';
 
 	import { mobile, models, settings } from '$lib/stores';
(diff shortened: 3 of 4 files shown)


richard-deetlefs:
Merge when?

PR #27674, opened 2026-07-28
fix: keep the selected model in sync when regenerating a response

### LINKED ISSUE #25052: issue: Incorrect model selected in dropdown, when switching to a chat with multiple generated replies by multiple models
### Check Existing Issues

### Installation Method

Docker

### Open WebUI Version

Latest dev version

_No response_

### Operating System

Windows 11

### Browser (if applicable)

Latest Brave

### Confirmation

### Expected Behavior

When switching to a chat where its last user message had multiple generated replies by different models, the model selected should correspond to the assistant reply that's currently active.

In my observations, the reply that's active after switching seems to always be the latest one (although seems like Open-WebUI could technically restore the last selected one but that's outside the context of this issue).

For example, if there are 8 assistant replies, then the 8/8 reply seems to be the active one when switching.

### Actual Behavior

The model selected in the dropdown, after switching to the chat seems to be the model used in the **first** generated reply. 

For example: if the current reply is the 8/8 reply, the model chosen seems to be the one for the 1/8 reply.

This causes the user to need to keep manually selecting the latest model every time they switch to the chat.

### Steps to Reproduce

1. Create a new chat
2. Select a model from the dropdown
3. Send a prompt like "Hello world" and generate the first reply
4. Select a different model from the dropdown
5. Click on the "regenerate" button to regenerate the reply with the newly selected model
6. Switch to a different chat
7. Go back to that chat again
8. Observe what is now the active model in the model dropdown

### Logs & Screenshots

(none)

### Additional Information

While testing this I had a situation where a request I was sending to a model (Qwen3.5-0.8B) caused the model to go to an infinite generation loop, so I stopped it.

Then I tried another model (Gemma E2B).

So I had 2 generations of the replies now.

Then I noticed something odd:

When I go to another chat and switch back, it keep choosing Qwen3.5-0.8B from the dropdown even though it's not even used in any of the replies. That model got "stuck" as the default model selected for that chat, and nothing I did seemed to change that.

Classic298:
Switching the model in the dropdown and regenerating a response left the chat's stored model list pointing at the model that was used when the user message was first sent. Leaving the chat and coming back restored that stale entry, so the dropdown kept snapping back to the original model (and could even show a model that is no longer used by any response in the chat, or one that no longer exists).

The chat's model list is persisted from the user message's `models` field, which is written when the message is created and was never refreshed afterwards. Regeneration is the one send path that reuses an existing user message, so it now refreshes that field from the current selection.

Side-by-side turns are left untouched: those regenerate the clicked column and keep their full model list, which is also what the column grouping in MultiResponseMessages relies on.

Fixes #25052

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -3554,6 +3554,12 @@
 				scrollToBottom();
 			}
 
+			// Single-model turns follow the current selection; side-by-side turns regenerate the clicked column.
+			const isSideBySideTurn = (userMessage.models ?? selectedModels).length > 1;
+			if (!isSideBySideTurn) {
+				userMessage.models = selectedModels;
+			}
+
 			await sendMessage(history, userMessage.id, {
 				...(suggestionPrompt
 					? {
@@ -3561,9 +3567,8 @@
 							regenerationPrompt: suggestionPrompt
 						}
 					: {}),
-				...((userMessage?.models ?? [...selectedModels]).length > 1
+				...(isSideBySideTurn
 					? {
-							// If multiple models are selected, use the model from the message
 							modelId: message.model,
 							modelIdx: message.modelIdx
 						}
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -3392,6 +3392,12 @@
 				scrollToBottom();
 			}
 
+			// Single-model turns follow the current selection; side-by-side turns regenerate the clicked column.
+			const isSideBySideTurn = (userMessage.models ?? selectedModels).length > 1;
+			if (!isSideBySideTurn) {
+				userMessage.models = selectedModels;
+			}
+
 			await sendMessage(history, userMessage.id, {
 				...(suggestionPrompt
 					? {
@@ -3399,9 +3405,8 @@
 							regenerationPrompt: suggestionPrompt
 						}
 					: {}),
-				...((userMessage?.models ?? [...selectedModels]).length > 1
+				...(isSideBySideTurn
 					? {
-							// If multiple models are selected, use the model from the message
 							modelId: message.model,
 							modelIdx: message.modelIdx
 						}


Classic298:
### Testing

Verified end to end against a local build, comparing this branch to `dev` at the same commit. Two isolated worktrees, each with its own frontend build, both served by the real backend against a throwaway copy of a real data dir. A local mock upstream exposed two models (`mock-alpha`, `mock-beta`), and the whole scenario was driven through the real UI (typing into the chat input, clicking send, picking a model from the dropdown, clicking regenerate) in headless Chrome over CDP.

**Scenario 1: the reported bug.** Send a prompt with `mock-alpha`, switch the dropdown to `mock-beta`, regenerate, navigate away to another chat, navigate back.

| | persisted `chat.models` after regenerating | model shown in the dropdown after returning | model of the active reply |
|---|---|---|---|
| `dev` | `["mock-alpha"]` | `mock-alpha` | `mock-beta` |
| this branch | `["mock-beta"]` | `mock-beta` | `mock-beta` |

`dev` reproduces the issue: the reply on screen came from `mock-beta`, but the dropdown restores `mock-alpha`, so every visit to the chat needs the model re-picked by hand. On this branch the dropdown matches the reply that is actually shown. Repeated the passing case a second time to rule out a timing fluke.

**Scenario 2: side-by-side regression check.** Two-model chat, regenerate a single column, navigate away and back. Identical results on both builds:

- `chat.models` stays `["mock-alpha", "mock-beta"]`
- the user message's `models` field is left untouched
- the regenerated response keeps `modelIdx: 0`, so it stays in its own column
- both columns render correctly after returning to the chat

In other words the column identity that `MultiResponseMessages` relies on is unaffected, which is what the `isSideBySideTurn` guard is there to protect.

Classic298:
@tjbck ready for review. Regenerating a response after switching models leaves the chat pointing at the old model, so the dropdown snaps back every time the chat is reopened, sometimes to a model that no longer exists.

Fixes #25052.

PR #27721, opened 2026-07-30
feat: persistent image gallery for the Images playground

silentoplayz:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Relates to #27720 — Discussion covering the feature, its scope, and the open question about the editor.
- [ ] **Documentation:** Not applicable — no new setting or configuration surface; the gallery uses the existing image-generation config.
- [x] **Testing:** Verified by the author on a local instance against ComfyUI — see "Manual Verification Performed" below for the exact steps. `npm run format`, `npm run i18n:parse`, a full production build, `ruff format --check`, and `vitest run` all pass, and the working tree is clean after the format + i18n steps.
- [x] **Architecture:** No new endpoints and no new settings — the listing reuses the existing files-search API, and the two backend hunks only add metadata to files that are already being written.
- [x] **Git Hygiene:** The PR is a single commit, based on the current `dev`, and contains no unrelated commits.

# Changelog Entry

### Description

**Problem** — The image playground discards results on reload. Every generated image is already persisted as a file with its prompt in metadata; the playground just never reads it back. So reloading, restarting, or switching away loses your results, there is no way to find an image generated earlier or the prompt behind it, and images generated from a chat are invisible in the playground even though they are the same files.

**What this adds** — A persistent, paginated gallery in Playground → Images, above the existing prompt input:

- Grid of previously generated images, 20 per page with **Load more**, lazy-loaded thumbnails.
- **Model and size labels** per tile, read from the stored file metadata rather than from what the UI requested — so a label cannot claim a model that did not produce the image.
- The originating **prompt on hover**.
- Per-image actions: **Open in Chat**, **Copy** to clipboard, **Download**, and **Delete** with confirmation. Clicking a tile opens the existing `ImagePreview` lightbox.
- **Open in Chat is only offered when the source chat is still reachable**, so a deleted chat does not leave a dead link.

Two small backend additions make the labels and the chat link possible:

```diff
   model = await get_image_model(request)
 
+  metadata['model'] = model
+  metadata['size'] = size
```

…and `chat_id` / `message_id` are threaded through the built-in `generate_image` / `edit_image` tools (+8 lines) so a gallery image can link back to the chat it came from. **No new endpoints** — the listing is a thin wrapper over the existing `searchFiles`.

**When the gallery re-syncs** — worth stating precisely rather than implying it is live. It syncs on load, after you generate, and when the **browser tab** regains visibility. It does not observe changes made side-by-side in the same tab: with the playground open next to a chat, an image generated in that chat (or a chat deleted out from under an image) is not reflected until the tab is reopened or reloaded. Making that live would need polling or server-event subscriptions, which felt like more machinery than this change warrants.

**Also fixes a pre-existing bug** found while building this. The image preview's Download button fetched the image with no credentials. Backend file URLs require authentication, so the request returned `401` — and because the response status was never checked, the JSON error body was saved as the "image": a `generated_image.json` containing `{"detail":"Not authenticated"}`. **This affects generated images in chat on current `dev`, not only the playground.**

```diff
-							fetch(src)
-								.then((response) => response.blob())
+							fetch(
+								src,
+								isOwnBackend && localStorage.token
+									? { headers: { Authorization: `Bearer ${localStorage.token}` } }
+									: undefined
+							)
+								.then((response) => {
+									if (!response.ok) {
+										throw new Error(`Failed to download image: ${response.status}`);
+									}
+									return response.blob();
+								})
```

The token is sent only when the target is this deployment's own backend — which is **not** always the page origin, since in development the frontend is served separately and `WEBUI_BASE_URL` points at the backend, so a same-origin check alone still fails there. A third-party image host never receives the token, and a non-OK response is no longer written to disk.

### Added

- A persistent, paginated image gallery in the Images playground, with per-image Open in Chat / Copy / Download / Delete actions and model, size, and prompt labels drawn from stored file metadata.
- `model` and `size` in generated-image file metadata, and `chat_id` / `message_id` from the built-in image tools, so gallery images can be labelled and linked back to their chat.

### Fixed

- Fixed the image preview's Download button saving an authentication-error payload instead of the image, because it fetched backend file URLs without credentials and never checked the response status. This affected generated images in chat as well as the playground.

---

### Additional Information

**Behaviour worth knowing when reviewing.**

The file search endpoint returns 404 when nothing matches. For a gallery that means "no images yet" or "end of pagination" rather than a failure, so it is treated as an empty page. Any other error is real and is surfaced, because silently rendering an empty gallery would be indistinguishable from genuinely having no images.

New images are pulled back from the server rather than synthesised locally. The engine decides which model actually ran and may adjust the requested dimensions, so the stored metadata is the only accurate source for the gallery's labels.

Pagination advances by the number of rows the server returned rather than by a page index. Images prepended by a refresh shift the server's list, and advancing by index would append an image already on screen, duplicating a key in the `{#each}`.

Backend file URLs require authentication, so the bearer token is attached when the target is this deployment's own backend, which is not always the page origin: in a development setup the frontend is served separately and `WEBUI_BASE_URL` points at the backend. The token is never attached to a third party image host, and a failed response is rejected before it can be saved as the "image".

The cross tab refresh runs when the browser tab regains visibility, so it does not pick up changes made side by side in the same tab. Reopening or reloading re-syncs from scratch.

**Deliberately out of scope**, to keep this reviewable and honest:

- **Expanded generation controls** (count, steps, negative prompt, size). For ComfyUI these only take effect when the admin has mapped the corresponding workflow nodes — `n` needs a node typed `n` to receive `batch_size`, and so on — so on a workflow without those mappings the controls silently do nothing. #25140 (auto-mapping ComfyUI workflow nodes) is the natural prerequisite; once mappings are auto-detected these can be added and actually work. An earlier revision of this branch carried them and they were removed for exactly this reason.
- **An image editor**, on [`silentoplayz:feat/playground-image-editor`](https://github.com/silentoplayz/open-webui/tree/feat/playground-image-editor) as one commit on top of this branch. Held back because the author cannot verify the edit round-trip (no hardware or ComfyUI workflows capable of the edit models). #27720 asks for a tester.

**Known limitations**, stated up front:

- `upload_image` names every generated file `generated-image{ext}`, so **edited** images also appear in the gallery. `image_edits` never records the resolved `model`/`size` in metadata, so those tiles render without labels. The tile handles this gracefully (`{#if fileModel || fileSize}`), and the underlying gap is now tracked separately as #27726 — left out of this PR because the author cannot test the edit path.

**Notes for reviewers:**

- Pagination tracks a server offset advanced by rows actually consumed, rather than a page index, and de-duplicates on append. That matters because the `{#each}` is keyed on `file.id`: generating an image prepends a row and shifts the server list, so an index-based page would re-append a row already on screen and Svelte would throw on the duplicate key. Deleting decrements the offset for the same reason.
- `getGeneratedImages` swallows only the search endpoint's `No files found matching the pattern.` 404 — which for a gallery means "no images yet" or "end of pagination" — and re-throws anything else, so a genuine failure surfaces as an error rather than as an empty gallery. This is scoped to the gallery helper rather than changing `searchFiles`' error semantics for every caller.
- Chat links are resolved through one `syncChatLinks(files, recheck)` helper that both adds and removes ids, so a chat deleted since the last check stops offering "Open in Chat" on the next refresh.

Local instance, ComfyUI engine (`z_image_turbo_bf16`), super-admin account:

1. Generated images from the playground and confirmed each appears immediately with the correct model and size labels, matching the values stored in the file's metadata.
2. Reloaded the page, then restarted both frontend and backend — the gallery persisted in full.
3. Generated 20 images: **Load more** appeared, and clicking it correctly exhausted the list and removed the button.
4. With 21+ images: reloaded to a single page, generated one more image, then clicked **Load more** — the new image appears first, the next page appends, with no duplicated or missing tiles and no console errors. (This is the sequence that fails with index-based pagination.)
5. Generated an image from a chat, confirmed **Open in Chat** navigates to that chat; deleted the chat and confirmed the action is no longer offered for that image.
6. Deleted a gallery image via the confirmation dialog and confirmed it is removed.
7. Verified the download fix in both places it applies: the playground gallery's preview and a generated image in a chat. Both now download a valid image; before the fix both produced `generated_image.json`.

### Screenshots or Videos

| | |
|---|---|
| Gallery grid | image |
| Hover actions on a tile | image |

> [!NOTE]


silentoplayz:
Heads-up on scope: the `ImagePreview` download fix bundled here is also filed on its own as #27722 / #27723, so it can land without waiting on the discussion about this feature.

The change to `src/lib/components/common/ImagePreview.svelte` is **byte-identical** between the two branches, so whichever merges first the other applies cleanly — no conflict either way, and no action needed here. It stays in this PR so that if you'd rather take the gallery on its own, it still arrives with the bug fixed rather than depending on merge order.

silentoplayz:
Since the rebase, `Open in Chat` lands on the exact response that produced the image rather than on the chat's latest message, and centers the image itself.

The link carries the message id and the file id. Both are already recorded in the image's file meta, by `tools/builtin.py` for the image tool and by `utils/middleware.py` for the chat image generation feature, so no backend change was needed.

Three behaviours in the chat view had to be handled, which is why the diff reaches outside the playground:

- A regenerated response is a sibling branch, so the target message is frequently not on the active path and not in the DOM. `showMessage` selects that branch before anything is scrolled to.
- `navigateHandler` queues a scroll to the bottom on a timeout once `loadChat` returns, ignoring `autoScroll`, which overwrote the result. It is skipped when the load focused a message.
- `Messages.svelte` renders the last 8 messages and marks them `content-visibility: auto` with a 150px intrinsic size estimate. Older targets were therefore absent from the DOM, and downward scrolls landed short once real heights resolved. The branch is now rendered in full, and the scroll repeats across two animation frames, matching what `scrollToBottom` already does for the same reason.

The link is also gated differently now. It used to probe every chat in the gallery through `getChatById` after each generation, which logged a handled 401 to the console for any chat since deleted. It uses a non logging existence check and only probes chats it has not already resolved.

Images generated in the playground have no source chat, so they show no link, unchanged.

PR #27754, opened 2026-07-30
fix: apply RDS IAM token auth to the pgvector engine

### LINKED ISSUE #27752: bug: pgvector client ignores DATABASE_ENABLE_IAM_TOKEN_AUTH, so RDS IAM auth and VECTOR_DB=pgvector cannot be used together
### Open WebUI Version

v0.11.0

### Installation Method

Docker (`ghcr.io/open-webui/open-webui:v0.11.0`, ARM64, on AWS ECS Fargate)

### Operating System

Amazon Linux 2023 (ECS Fargate)

### Expected Behavior

With `DATABASE_ENABLE_IAM_TOKEN_AUTH=true` and `VECTOR_DB=pgvector` against Amazon RDS for
PostgreSQL, the pgvector client should authenticate with an IAM token in the same way the main
database connection does, so a deployment can run without a stored database password at all.

### Actual Behavior

Startup gets as far as the Alembic migrations, which succeed — IAM token authentication works for
the main engine. Initialising the vector store then fails:

```
ERROR [open_webui.retrieval.vector.dbs.pgvector] Error during initialization:
(psycopg2.OperationalError) connection to server at "<redacted>.eu-west-1.rds.amazonaws.com",
port 5432 failed: fe_sendauth: no password supplied

  File "/app/backend/open_webui/retrieval/vector/dbs/pgvector.py", line 117, in __init__
  File "/app/backend/open_webui/retrieval/vector/factory.py", line 54, in get_vector
    return PgvectorClient()
```

The container exits 1, so the deployment never becomes healthy.

### Steps to Reproduce

1. Create an RDS for PostgreSQL instance with IAM authentication enabled, and a login role granted
   `rds_iam`.
2. Install the `vector` extension in the database.
3. Run Open WebUI with `DATABASE_TYPE=postgresql`, `DATABASE_HOST`, `DATABASE_PORT`,
   `DATABASE_NAME`, `DATABASE_USER`, `DATABASE_ENABLE_IAM_TOKEN_AUTH=true`, `VECTOR_DB=pgvector`,
   and no `DATABASE_PASSWORD` and no `PGVECTOR_DB_URL`.
4. Migrations run to completion, then the pgvector client fails as above.

### Cause

`internal/db.py` attaches the IAM token refresher as a SQLAlchemy `do_connect` listener, but only to
the two engines it owns:

```python
_rds_iam_token_auth = RDSIAMTokenAuth(SQLALCHEMY_DATABASE_URL) if DATABASE_ENABLE_IAM_TOKEN_AUTH else None
...
enable_iam_token_auth(engine)        # line 331
enable_iam_token_auth(async_engine)  # line 410
```

`retrieval/vector/dbs/pgvector.py` builds its own engine, and that engine gets no listener:

```python
if not PGVECTOR_DB_URL:
    from open_webui.internal.db import ScopedSession
    self.session = ScopedSession          # IAM-aware
else:
    engine = create_engine(PGVECTOR_DB_URL, ...)   # not IAM-aware
```

The `ScopedSession` branch looks like the escape hatch, but it is unreachable in practice, because
`config.py` line 638 defaults the variable to the main URL:

```python
PGVECTOR_DB_URL = os.getenv('PGVECTOR_DB_URL', DATABASE_URL)
```

`DATABASE_URL` is always populated — either directly or assembled from the discrete `DATABASE_*`
parameters — so `PGVECTOR_DB_URL` is never falsy for a PostgreSQL deployment and the `else` branch
always wins. Under IAM authentication that URL carries no password, hence `fe_sendauth: no password
supplied`.

The net effect is that `DATABASE_ENABLE_IAM_TOKEN_AUTH` and `VECTOR_DB=pgvector` cannot be used
together, even though each works on its own. That is unfortunate, because pgvector is the
recommended vector store for scaled deployments and IAM authentication is the standard way to avoid
long-lived database credentials on AWS.

### Suggested Fix

Either would resolve it:

- call `enable_iam_token_auth(engine)` on the engine built in `PgvectorClient.__init__`, so the
  vector store refreshes tokens the same way the main engine does; or
- reuse the main engine when `PGVECTOR_DB_URL` resolves to the same target as `DATABASE_URL`, which
  would also make the existing `ScopedSession` branch reachable.

The first seems closer to the intent, since it keeps the two engines independent while giving both
the same authentication behaviour.

Happy to open a PR if you'd like.

Classic298:
With `DATABASE_ENABLE_IAM_TOKEN_AUTH=true` and `VECTOR_DB=pgvector`, startup failed at vector store initialisation with `fe_sendauth: no password supplied`, so the two features could not be used together.

`PgvectorClient` builds its own engine and never got the `do_connect` listener that refreshes the RDS IAM token, and the `ScopedSession` branch that would have reused the instrumented main engine is unreachable because `PGVECTOR_DB_URL` defaults to `DATABASE_URL` and is therefore never falsy.

The pgvector engine now goes through `enable_iam_token_auth()` like the main and Alembic engines. Since a token authenticates exactly one host/port/user, that function now attaches the listener only to engines pointing at the same target, so a `PGVECTOR_DB_URL` aimed at a separate database keeps the password from its own URL instead of having it overwritten; the skip is logged with both identities.

Fixes #27752

> [!NOTE]

### DIFF
--- a/backend/open_webui/internal/db.py
+++ b/backend/open_webui/internal/db.py
@@ -202,6 +202,20 @@ def enable_iam_token_auth(connectable) -> None:
         return
 
     engine = getattr(connectable, 'sync_engine', connectable)
+    url = engine.url
+    auth = _rds_iam_token_auth
+    # The token is bound to one host/port/user pair; leave other databases on their own credentials.
+    if (url.host, url.port or 5432, url.username) != (auth.host, auth.port, auth.username):
+        log.warning(
+            'AWS RDS IAM token auth not applied to %s: the token is issued for %s@%s:%s, '
+            'so this connection uses the password from its own URL',
+            url.render_as_string(hide_password=True),
+            auth.username,
+            auth.host,
+            auth.port,
+        )
+        return
+
     if not event.contains(engine, 'do_connect', _set_iam_token_password):
         event.listen(engine, 'do_connect', _set_iam_token_password)
 
--- a/backend/open_webui/retrieval/vector/dbs/pgvector.py
+++ b/backend/open_webui/retrieval/vector/dbs/pgvector.py
@@ -17,6 +17,7 @@ from open_webui.config import (
     PGVECTOR_POOL_TIMEOUT,
     PGVECTOR_USE_HALFVEC,
 )
+from open_webui.internal.db import ScopedSession, enable_iam_token_auth
 from open_webui.retrieval.vector.main import (
     GetResult,
     SearchResult,
@@ -87,8 +88,6 @@ class PgvectorClient(VectorDBBase):
     def __init__(self) -> None:
         # if no pgvector uri, use the existing database connection
         if not PGVECTOR_DB_URL:
-            from open_webui.internal.db import ScopedSession
-
             self.session = ScopedSession
         else:
             if isinstance(PGVECTOR_POOL_SIZE, int):
@@ -107,6 +106,7 @@ class PgvectorClient(VectorDBBase):
             else:
                 engine = create_engine(PGVECTOR_DB_URL, pool_pre_ping=True)
 
+            enable_iam_token_auth(engine)
             SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine, expire_on_commit=False)
             self.session = scoped_session(SessionLocal)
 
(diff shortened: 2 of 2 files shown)


Classic298:
@tjbck ready for review. `DATABASE_ENABLE_IAM_TOKEN_AUTH` together with `VECTOR_DB=pgvector` fails at startup, so the two cannot be used together at all.

Fixes #27752.

PR #27663, opened 2026-07-28
fix: index the chat queries that make large SQLite instances unusable

Classic298:
Five hot queries on the chat table read or sorted far more than they returned. All of them scale with the size of the chat history rather than with the work being done, and because the narrow columns they filter on are stored after the large chat payload, checking them makes SQLite walk every row's overflow pages.

- The timer scheduler polls for due timers once a second, filtering on JSON keys inside chat.meta. On an idle instance with a 1.2 GB database that burned about a quarter of a CPU core continuously. Timers now store their due time in a dedicated chat.timer_at column with a partial index holding only the timers still waiting to fire: set on creation, cleared on claim or cancel, restored if a claimed timer is requeued, backfilled by the migration.
- Cancelling timers runs on every user message and every chat open. SQLite never runs ANALYZE, so even with the column indexed its planner kept picking a plain user_id index and reading every chat the user owns; a partial (user_id, timer_at) index wins that choice deterministically.
- The sidebar chat list filters on user_id and orders by updated_at with a limit, but nothing covered that combination, so SQLite sorted every chat the user owns before keeping the first 60. (user_id, updated_at DESC, id) matches its filter and ordering exactly.
- The folder unread badges count unread chats per folder on every sidebar refresh, fetching every candidate chat row and scanning each chat's messages for an unfinished assistant reply. A covering index on chat plus a (chat_id, role, done) index on chat_message serve the whole count without touching a single table row.

Measured on SQLite, same rows returned (the production instance these came from measured the last two at 1.5 s and 2-4 s cold):

| query | before | after |
|---|---|---|
| idle timer poll (2000 chats, 0.43 GB, 50k finished timers) | 170 ms | 0.04 ms |
| sidebar chat list (15000 chats, 1.26 GB) | 157 ms | 1.8 ms |
| timer cancel on send and chat open (15000 chats, 1.43 GB) | 12 ms warm | 0.01 ms |
| folder unread badges (15000 chats, 1.43 GB, 120k messages) | 54 ms warm | 0.2 ms |

PostgreSQL 17 answers the cancel and unread queries as index-only scans. The migration was exercised on SQLite and PostgreSQL 17 through fresh install, upgrade with seeded data, downgrade and re-upgrade.

Dropping the internal and type checks from the timer queries also makes a forked timer chat inert. A fork copies meta verbatim but not the column, so it used to be a claim target that could fire a duplicate timer.

Fixes #27622
Fixes #27745

### Contributor License Agreement

[...]

PR #27784, opened 2026-07-31
fix: filter due timers in SQL instead of loading every pending timer each poll

### LINKED ISSUE #27782: issue: the timer scheduler loads every pending timer from the database once per second, filtering in Python
### Check Existing Issues

### Installation Method

Git Clone

### Open WebUI Version

`dev` @ 810378c0b (v0.11.0)

### Operating System

Linux

### Browser (if applicable)

Not applicable (backend scheduler behavior)

### Confirmation

- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup**: any instance with pending chat timers; the behavior is in the scheduler's claim query.

### Expected Behavior

The timer scheduler's claim query fetches only timers that are actually due, using the database to filter, sort, and limit.

### Actual Behavior

`claim_due_timers` in `backend/open_webui/utils/timers.py` runs every second (the timer lane polls at `TIMER_POLL_INTERVAL`, default 1s) and its query selects **every pending timer row**, with the due filter, the sort, and the limit all applied in Python afterwards:

```python
result = await db.execute(stmt)
rows = [row for row in result.scalars().all() if int((row.meta or {}).get('timer_at') or 0) <= now_ns]
rows.sort(key=lambda row: int((row.meta or {}).get('timer_at') or 0))
rows = rows[:limit]
```

Consequences:

1. Every pending timer's full chat row (including the chat JSON payload) is fetched and deserialized once per second, per instance, even when nothing is due. Long timers ("remind me in 3 days") make this a standing cost that grows with every pending timer in the system.
2. On PostgreSQL the statement uses `FOR UPDATE SKIP LOCKED`, so each poll locks **all** pending rows for the duration of the transaction, not just the batch being claimed. A second instance polling concurrently skips every locked row, including due ones, and has to wait for its next poll.
3. A pending timer with a missing `timer_at` is treated as due immediately (`or 0`), rather than never.

### Suggested Fix

Push the filter, sort, and limit into SQL. The codebase already queries these JSON fields with `.as_boolean()` and `.as_string()` on the same column, so the due comparison follows the same idiom with an explicit BIGINT cast (nanosecond values overflow a 4 byte INTEGER cast on PostgreSQL, so `.as_integer()` is not usable here):

```python
timer_at = cast(Chat.meta['timer_at'].as_string(), BigInteger)
stmt = (
    select(Chat)
    ...
    .where(timer_at <= now_ns)
    .order_by(timer_at)
    .limit(limit)
)
```

Verified on both dialects: executed against SQLite (in memory, SQLAlchemy) and against PostgreSQL 18 (the pgvector image, throwaway container) with nanosecond scale values; both return exactly the due rows, exclude future ones, and exclude rows with a missing `timer_at`. The compiled SQL was also inspected for both dialects to confirm the BIGINT cast renders correctly.

With the limit in SQL, the PostgreSQL `FOR UPDATE SKIP LOCKED` now locks at most `limit` rows per poll instead of every pending timer.

A PR follows.

This issue was prepared with AI copilot assistance; the fix was verified empirically on both supported database dialects, and every code reference above was verified against `dev` @ 810378c0b.

silentoplayz:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #27782, the timer scheduler loads every pending timer from the database once per second, filtering in Python.
- [ ] **Documentation:** Not applicable. Same query semantics, expressed in SQL.
- [x] **Testing:** The due filter expression was executed against both supported dialects with nanosecond scale values: SQLite in memory via SQLAlchemy, and PostgreSQL 18 (the pgvector image, throwaway container). Both return exactly the due rows, exclude future ones, and exclude rows with a missing timer_at. The compiled SQL was inspected for both dialects to confirm the BIGINT cast renders correctly (a plain as_integer() would render a 4 byte INTEGER cast on PostgreSQL and overflow on nanosecond values). ruff format passes; the diff adds zero code comments.
- [x] **No Unchecked AI Code:** This PR was prepared with AI copilot assistance and has been thoroughly reviewed by the author.
- [x] **Architecture:** No new settings and no dependency changes: the claim query gains a WHERE, an ORDER BY, and a LIMIT, in the same JSON field idiom the query already uses.
- [x] **Git Hygiene:** A single commit, +6/-4 in a single file, based on the current `dev`.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

**Problem** (#27782): `claim_due_timers` polls every second and selects every pending timer row, applying the due filter, sort, and limit in Python. Every pending timer's full chat row (including its JSON payload) is fetched and deserialized once per second per instance even when nothing is due; on PostgreSQL, `FOR UPDATE SKIP LOCKED` locks all pending rows per poll instead of just the claimed batch, so a concurrently polling instance skips due timers too; and a missing `timer_at` counts as due immediately.

**Fix**: push the filter, sort, and limit into SQL:

```diff
     async with get_async_db() as db:
+        timer_at = cast(Chat.meta['timer_at'].as_string(), BigInteger)
         stmt = (
             select(Chat)
             .where(Chat.meta['internal'].as_boolean().is_(True))
             .where(Chat.meta['type'].as_string() == 'timer')
             .where(Chat.meta['status'].as_string() == 'pending')
+            .where(timer_at <= now_ns)
+            .order_by(timer_at)
+            .limit(limit)
         )
         if db.bind.dialect.name == 'postgresql':
             stmt = stmt.with_for_update(skip_locked=True)

         result = await db.execute(stmt)
-        rows = [row for row in result.scalars().all() if int((row.meta or {}).get('timer_at') or 0) <= now_ns]
-        rows.sort(key=lambda row: int((row.meta or {}).get('timer_at') or 0))
-        rows = rows[:limit]
+        rows = result.scalars().all()
```

The explicit BIGINT cast matters: nanosecond timestamps overflow the 4 byte INTEGER that `.as_integer()` renders on PostgreSQL. With the limit in SQL, `FOR UPDATE SKIP LOCKED` now locks at most `limit` rows per poll. A pending row with a missing `timer_at` is now never claimed instead of being claimed immediately, which is the safer reading of a malformed row (it cannot occur through `create_timer`, which always sets the field).

### Fixed

- Fixed the timer scheduler fetching every pending timer chat from the database once per second and filtering in Python: due filtering, ordering, and the claim limit now run in SQL on both SQLite and PostgreSQL, and the PostgreSQL claim locks only the batch being claimed instead of every pending timer.

---

### Additional Information

- This PR was prepared with AI copilot assistance and reviewed by the author.

### Verification Performed

1. **SQLite**: executed the new expression in memory via SQLAlchemy with nanosecond values; returns exactly the due row, excludes the future row and the row with no timer_at.
2. **PostgreSQL 18**: executed the compiled SQL in a throwaway container of the pgvector image with the same fixture; identical results.
3. **Compiled SQL inspected for both dialects**: the expression renders `CAST(... AS BIGINT)` on both; the `.as_integer()` alternative was checked and rejected because it renders `CAST(... AS INTEGER)` on PostgreSQL, which overflows on nanosecond values.
4. `ruff format --check` passes; the diff adds zero code comments.

### Screenshots or Videos

Not applicable (backend query behavior; the dialect verification above is the evidence).

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/timers.py
+++ b/backend/open_webui/utils/timers.py
@@ -13,7 +13,7 @@ from typing import Literal
 from uuid import uuid4
 
 from fastapi import Request
-from sqlalchemy import select
+from sqlalchemy import BigInteger, cast, select
 from starlette.datastructures import Headers
 
 from open_webui.internal.db import get_async_db
@@ -170,19 +170,21 @@ async def create_timer(
 async def claim_due_timers(now_ns: int, limit: int = 10) -> list[tuple[str, str]]:
     """Claim due timers by moving them from pending to running."""
     async with get_async_db() as db:
+        timer_at = cast(Chat.meta['timer_at'].as_string(), BigInteger)
         stmt = (
             select(Chat)
             .where(Chat.meta['internal'].as_boolean().is_(True))
             .where(Chat.meta['type'].as_string() == 'timer')
             .where(Chat.meta['status'].as_string() == 'pending')
+            .where(timer_at <= now_ns)
+            .order_by(timer_at)
+            .limit(limit)
         )
         if db.bind.dialect.name == 'postgresql':
             stmt = stmt.with_for_update(skip_locked=True)
 
         result = await db.execute(stmt)
-        rows = [row for row in result.scalars().all() if int((row.meta or {}).get('timer_at') or 0) <= now_ns]
-        rows.sort(key=lambda row: int((row.meta or {}).get('timer_at') or 0))
-        rows = rows[:limit]
+        rows = result.scalars().all()
 
         claimed = []
         for row in rows:


CrushedAsian255:
I installed this pull request against my server however it doesn't seem to have significant fixed the problem.
I measure by determining time between `execute` starting and finishing when using DEBUG log mode.

Unpatched (v0.11.0 release): 296, 296, 297, 299 : 297 ms avg
With your patched: 291, 293, 291, 293 : 292 ms avg 
Compared to patch #27663 : 1, 0, 1, 1 : 1 ms avg

silentoplayz:
Thank you for benchmarking this properly, @CrushedAsian255. Your numbers are conclusive and the explanation is straightforward: both the original code and this patch make the database scan the whole chat table evaluating JSON path expressions, which cannot use an index as written. Moving the filter from Python into SQL only removed the Python side row handling, which was negligible next to the scan itself.

Classic298's #27663 is the structurally correct fix (a dedicated indexed `timer_at` column that turns the poll into an index seek), it predates this PR, and it also covers the correctness case this PR fixed (a pending timer with a missing or non numeric due time is never claimed, since its column stays NULL). Closing this in favor of #27663.

PR #27797, opened 2026-07-31
fix: stop ORJSONCodec silently discarding stdlib json keyword arguments

Classic298:
`*args` is dropped from both signatures. stdlib `json.dumps` and `json.loads` take everything after the first parameter keyword-only, so passing an option positionally was already an invalid call; it now raises `TypeError` rather than being silently ignored.

No existing call site changes behaviour.

> [!NOTE]


tjbck:
78ed5a0235c4de67828ba4b0d6147035067cec2c

PR #27823, opened 2026-07-31
refac: consolidate the web fetch address checks onto the request path

Classic298:
Two checks govern server-side fetches: a fetchable-address test and the operator's `WEB_FETCH_FILTER_LIST`. Both now run on the request path, rather than only where a URL is first validated.

`_SSRFSafeResolver` is replaced by `_SSRFSafeConnector`, which hooks `_resolve_host` so the address test also covers aiohttp's IP-literal shortcut and both DNS cache paths. The filter list moves to a per-request hook on each transport, `connect()` for aiohttp and `send()` for the requests adapter, because those see the request destination: at the connection layer a proxied request presents the proxy's host, and a pooled connection skips resolution entirely. This covers all five aiohttp call sites and both requests sessions. The Playwright loader already validated each hop and is unchanged.

The filter list is matched against the hostname together with the addresses it resolves to, and against the IPv4 address an IPv6 address carries inside it, so an entry holds regardless of how the target is spelled. A filter entry that parses as an address or a CIDR range is matched by containment rather than by DNS label suffix; a range entry previously matched nothing at all, silently. `retrieval/web/main.py` decides whether to resolve search-result hostnames with the same parser, so a range entry is meaningful there too.

`ipaddress.is_global` was the only test behind the address check, and it answers a narrower question than "may we fetch this". Several special-purpose ranges are globally routable by registry while nothing on them is a legitimate destination; they are ordinary block entries now, so taking an address out of reach is a `WEB_FETCH_FILTER_LIST` change rather than a release. Each was checked against both IANA special-purpose registries, and prefixes IANA marks globally reachable are deliberately left out.

Three behaviour changes for operators. The filter list applies to redirect targets rather than only to submitted URLs. Under a forward proxy it is evaluated against the request destination instead of the proxy, which also fixes allowlist entries rejecting every fetch in proxied deployments. And an entry naming an address now covers every name that points at it; a hostname allowlist is unaffected, since the hostname itself still satisfies it.

> [!NOTE]


Classic298:
@tjbck ready for review. The fetchable-address test and `WEB_FETCH_FILTER_LIST` can both be bypassed today: a proxied request presents the proxy's host at the connection layer and a pooled connection skips resolution entirely, and a filter entry written as an IP range silently matched nothing at all. Both checks now run on the request path.

PR #28037, opened 2026-08-02
refactor: reuse cached OIDC discovery and JWKS clients in back-channel logout

Classic298:
The back-channel logout handler now resolves the provider from the authlib client's cached discovery document and keeps one PyJWKClient per jwks_uri, matching the signing key against the cached key set from a worker thread instead of building a new client and fetching on the event loop for every request.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -1314,6 +1314,7 @@ class OAuthManager:
         self.app = app
 
         self._clients = {}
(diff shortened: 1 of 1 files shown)


Classic298:
@tjbck ready for review. Back-channel logout builds a new JWKS client and fetches on the event loop for every request. It now reuses the cached discovery document and keeps one client per `jwks_uri`, matching keys off the event loop.

PR #28053, opened 2026-08-03
fix: keep references to lifespan background tasks so they are not garbage collected

### LINKED ISSUE #28052: bug: three lifespan background tasks are created without keeping a reference and can be garbage collected
### Bug

Three long-lived background tasks started in the lifespan handler in `backend/open_webui/main.py` have their handles discarded, so nothing in the process holds a strong reference to them.

```python
# backend/open_webui/main.py (dev @ 2d18727)

    if ...:
        app.state.redis_task_command_listener = asyncio.create_task(redis_task_command_listener(app))   # L370 — kept

    asyncio.create_task(periodic_usage_pool_cleanup())    # L376 — discarded
    asyncio.create_task(periodic_session_pool_cleanup())  # L377 — discarded
    asyncio.create_task(scheduler_worker_loop(app))       # L381 — discarded
```

The event loop keeps only a **weak** reference to a task ([asyncio docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task): *"Save a reference to the result of this function, to avoid a task disappearing mid-execution"*). A task whose only referent is the loop can therefore be garbage collected while it is suspended at an `await`.

All three are `while True` loops that are meant to run for the process lifetime:

| task | defined at | loop |
|---|---|---|
| `periodic_usage_pool_cleanup` | `socket/main.py:205` | `while True` |
| `periodic_session_pool_cleanup` | `socket/main.py:176` | `while True` |
| `scheduler_worker_loop` | `utils/automations.py:203` | `while True` |

If one is collected, the symptom is silent: session/usage pool entries stop being cleaned up, or automations and calendar alerts silently stop firing, with no error anywhere. Nothing logs a task disappearing.

### Why this looks like an oversight rather than a decision

The correct pattern is already in this same function, six lines above. `redis_task_command_listener` is stored on `app.state` and is then cancelled on shutdown:

```python
# backend/open_webui/main.py, shutdown half of the same lifespan
    if hasattr(app.state, 'redis_task_command_listener'):
        app.state.redis_task_command_listener.cancel()
```

The other three get neither the reference nor the shutdown cancellation, so they are also left running when the app shuts down.

### Suggested fix

Follow the convention already in the file: assign each to `app.state` and cancel it in the shutdown half alongside `redis_task_command_listener`. That is a strictly local change and does not alter behaviour when the tasks are not collected.

I have this written and tested locally and am happy to open a PR against `dev` referencing this issue, per the PR template's requirement that a PR link to an existing issue or discussion.

### Version

`dev` @ `2d18727`

noron12234:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #28052
- [ ] **Documentation:** No user-facing behaviour changes, so nothing to document.
- [ ] **Dependencies:** No dependency changes.
- [ ] **Testing:** See "What I verified and what I did not" below. I did not check this box because I have not run a live instance through a full startup/shutdown cycle.
- [ ] **No Unchecked AI Code:** Not checked, deliberately. See the note below.
- [x] **Architecture:** No new settings; this follows a pattern already present six lines above the change.
- [x] **Git Hygiene:** Single atomic commit, branched from `dev` @ `2d18727`.
- [x] **Title Prefix:** `fix`

### On the AI-code checkbox

This patch was written with an AI coding agent. I have left the "No Unchecked AI Code" box unchecked rather than check it inaccurately, since it asks for thorough human review **and** manual testing and the manual-testing half has not happened. Everything under "What I verified" below was done mechanically and is reproducible by you in a couple of minutes. If that is not enough for this repo, closing this is completely reasonable and I would rather you close it than merge something unverified.

### Description

Three background tasks started in `lifespan` had their handles discarded:

```python
asyncio.create_task(periodic_usage_pool_cleanup())
asyncio.create_task(periodic_session_pool_cleanup())
asyncio.create_task(scheduler_worker_loop(app))
```

The event loop keeps only a **weak** reference to a task, so a task whose only referent is the loop can be garbage collected while suspended at an `await` ([asyncio docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task): *"Save a reference to the result of this function, to avoid a task disappearing mid-execution"*).

All three are `while True` loops intended to run for the process lifetime (`socket/main.py:205`, `socket/main.py:176`, `utils/automations.py:203`). If one is collected, the symptom is silent: pool entries stop being cleaned up, or automations and calendar alerts stop firing, with nothing logged.

This is very likely an oversight rather than a decision, because the correct pattern is already in the same function six lines above:

```python
app.state.redis_task_command_listener = asyncio.create_task(redis_task_command_listener(app))
```

and it is cancelled in the shutdown half. The other three got neither. This applies the same treatment to them.

### What I verified and what I did not

Verified:

- `ruff format --check` and `ruff check` at the pinned `v0.15.5` from `.pre-commit-config.yaml`. The file reports **13 lint errors both before and after** this change, so the count is unchanged and none are introduced here. Formatting passes.
- `python -m py_compile` on the changed file.
- The unconditional `.cancel()` calls are safe. `yield` is at line 450 and the shutdown statements are plain code after it, not inside a `finally`, so a startup failure before line 376 skips them entirely. The three assignments are unconditional and sit before the `yield`, and calling an `async def` only builds a coroutine and cannot raise, so by the time shutdown runs all three attributes exist.
- `.cancel()` on an already-finished task is a no-op that returns `False`, so cancelling at shutdown cannot raise here.

Not verified:

- I have not run a live Open WebUI instance through startup and shutdown. The change is confined to reference-keeping and shutdown cancellation, but I am stating the limit rather than implying coverage I do not have.

---

# Changelog Entry

### Description

- Keeps strong references to the three lifespan background tasks so they cannot be garbage collected mid-execution, and cancels them on shutdown alongside the existing `redis_task_command_listener`.

### Fixed

- Fixed `periodic_usage_pool_cleanup`, `periodic_session_pool_cleanup` and `scheduler_worker_loop` being eligible for garbage collection while running, which could silently stop session/usage pool cleanup and the automation and calendar-alert scheduler.
- These three tasks are now cancelled on application shutdown instead of being left running.

---

### Additional Information

- Closes #28052, which has the full reasoning and the file/line references.

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -373,12 +373,12 @@ async def lifespan(app: FastAPI):
         limiter = anyio.to_thread.current_default_thread_limiter()
         limiter.total_tokens = THREAD_POOL_SIZE
 
-    asyncio.create_task(periodic_usage_pool_cleanup())
-    asyncio.create_task(periodic_session_pool_cleanup())
+    app.state.periodic_usage_pool_cleanup = asyncio.create_task(periodic_usage_pool_cleanup())
+    app.state.periodic_session_pool_cleanup = asyncio.create_task(periodic_session_pool_cleanup())
 
     from open_webui.utils.automations import scheduler_worker_loop
 
-    asyncio.create_task(scheduler_worker_loop(app))
+    app.state.scheduler_worker_loop = asyncio.create_task(scheduler_worker_loop(app))
 
     if await Config.get('models.base_models_cache'):
         try:
@@ -458,6 +458,10 @@ async def lifespan(app: FastAPI):
 
     if hasattr(app.state, 'redis_task_command_listener'):
         app.state.redis_task_command_listener.cancel()
+
+    app.state.periodic_usage_pool_cleanup.cancel()
+    app.state.periodic_session_pool_cleanup.cancel()
+    app.state.scheduler_worker_loop.cancel()
 
     await publish_event(app, EVENTS.SYSTEM_SHUTDOWN_COMPLETED, source='system')
 


tjbck:
Thanks!

PR #28066, opened 2026-08-03
perf: drop redundant single-column chat_message indexes

Classic298:
# Depends on #27663

and is branched on top of it: the migration here continues that PR's chain instead of forking it. Until #27663 merges, its three commits show up in this diff as well.

Every insert into chat_message maintains eight indexes, and three of them carry no queries: the single-column indexes on chat_id, user_id and model_id are strict prefixes of the composite indexes on (chat_id, parent_id), (user_id, created_at) and (model_id, created_at), so the planner serves every lookup from the composites without them. Verified on SQLite: every query shape in the codebase keeps its plan class, with covering scans reading a slightly wider index; PostgreSQL's btree leading-column rule gives the same guarantee, and the chat_id foreign-key cascade stays index-backed.

Dropping them makes a 20k-row insert batch about 11% faster on SQLite and shrinks the on-disk footprint of the busiest insert table in the app.

The migration drops each index only if it exists, because the migration that created the table skips everything when the table already exists, so old installs can lack any of them. Exercised through fresh install, seeded upgrade, downgrade and re-upgrade on SQLite and PostgreSQL 17.

> [!NOTE]

### DIFF
--- a/backend/open_webui/migrations/versions/a7d3f6b91c24_add_chat_timer_at.py
+++ b/backend/open_webui/migrations/versions/a7d3f6b91c24_add_chat_timer_at.py
@@ -1,0 +1,51 @@
+"""add chat timer_at
+
+Revision ID: a7d3f6b91c24
+Revises: d4a71b6e29c8
+Create Date: 2026-08-03 10:12:44.190283
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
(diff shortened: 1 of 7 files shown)


Classic298:
@tjbck rebased on dev. The migrations across the open PRs are one chain now, so this stacks on #27663 and its migration points at the merged migration there.

#27663 goes in first, then this one.

PR #28107, opened 2026-08-04
fix: double-encoded user.oauth in the b10670c03dd5 migration

### LINKED ISSUE #28101: bug: OAuth login fails after upgrading from pre-v0.6.41 — b10670c03dd5 double-encodes user.oauth
## Bug description

After upgrading an instance that still had the legacy `user.oauth_sub` column (i.e. any instance from before it was replaced by the `user.oauth` JSON column), OAuth/SSO login starts failing with a 500 error:

```
ERROR - Error during OAuth process: 1 validation error for UserModel
oauth
  Input should be a valid dictionary [type=dict_type, input_value='{"oidc": {"sub": "..."}}', input_type=str]
```

Subsequent password/other sign-in attempts for the same user also fail with the same validation error, because `Users.get_user_by_email` / `get_user_by_oauth_sub` validate the row into `UserModel`, whose `oauth` field is typed `dict | None`.

## Root cause

In `backend/open_webui/migrations/versions/b10670c03dd5_update_user_table.py`, the migration that backfills `oauth` from the legacy `oauth_sub` column does:

```python
conn.execute(
    sa.update(_user).where(_user.c.id == uid).values(oauth=json.dumps({provider: {'sub': sub}}))
)
```

`_user.c.oauth` is declared with `sa.JSON` type. SQLAlchemy's JSON type serializes whatever Python value it receives before writing it. Since `json.dumps(...)` already produced a JSON string, SQLAlchemy serializes that string *again*, so the value actually stored is a JSON scalar string containing the original JSON text — instead of a JSON object. Reading it back later returns a Python `str`, which fails Pydantic's `dict` validation on `UserModel.oauth`.

The `downgrade()` in the same migration has the mirrored bug: it calls `json.loads(oauth)` on `_user.c.oauth`, but since that column is JSON-typed, SQLAlchemy has *already* deserialized it into a dict by the time it's read — so `json.loads()` raises `TypeError` on every row, which is silently swallowed, and every user's `oauth_sub` ends up `None` on downgrade.

## Impact

Any production instance that upgrades across this migration while using SSO/OAuth (i.e. any instance that had `oauth_sub` populated) gets its `user.oauth` column double-encoded for every affected user, breaking their login until the data is repaired.

## Steps to reproduce

1. Start from a DB with the legacy `user.oauth_sub` column populated for a user (e.g. `oidc@some-sub`).
2. Run migrations up to and including `b10670c03dd5`.
3. Inspect `user.oauth` for that user — it will be a JSON string (e.g. `'{"oidc": {"sub": "some-sub"}}'`), not a JSON object.
4. Attempt to sign in as that user (password or OAuth) — the sign-in fails with a Pydantic `ValidationError` on `UserModel.oauth`.

## Environment

- Deployment: Docker
- Upgrading from a version predating the `oauth_sub` -> `oauth` migration
- Auth: OIDC SSO

Classic298:
Instances that had not upgraded since v0.6.41 and then jumped to v0.9.6 or later lose sign-in for every OAuth user, with a 500 and a validation error on the user model. Password sign-in for the same accounts fails identically, because every read path validates the row.

The migration that backfills user.oauth from the legacy oauth_sub column was rewritten from raw SQL to Core DML in May and kept its json.dumps call. The column reference declares sa.JSON, so SQLAlchemy serializes the value a second time and the row ends up holding a JSON string instead of an object. Reading it back yields a str, which the dict-typed field rejects. The mirrored json.loads in downgrade() has the inverse problem and silently nulled oauth_sub for every user.

Binding the dict directly fixes the write. Correcting that migration cannot help a database that already crossed it, since alembic never re-runs a stamped revision, so a follow-up revision un-double-encodes affected rows in place. It is a no-op for rows that were never corrupted.

Fixes #28101

> [!NOTE]

### DIFF
--- a/backend/open_webui/migrations/versions/b10670c03dd5_update_user_table.py
+++ b/backend/open_webui/migrations/versions/b10670c03dd5_update_user_table.py
@@ -185,9 +185,8 @@ def upgrade() -> None:
         for uid, oauth_sub in rows:
             if oauth_sub:
                 provider, sub = oauth_sub.split('@', 1) if '@' in oauth_sub else ('oidc', oauth_sub)
-                conn.execute(
-                    sa.update(_user).where(_user.c.id == uid).values(oauth=json.dumps({provider: {'sub': sub}}))
-                )
+                # no json.dumps here: the column is JSON, SQLAlchemy serializes it
+                conn.execute(sa.update(_user).where(_user.c.id == uid).values(oauth={provider: {'sub': sub}}))
 
     # ── Migrate api_key column → api_key table (only if old column still exists)
     if 'api_key' in user_columns:
@@ -226,10 +225,8 @@ def downgrade() -> None:
 
     for uid, oauth in rows:
         try:
-            data = json.loads(oauth)
-            provider = list(data.keys())[0]
-            sub = data[provider].get('sub')
-            oauth_sub = f'{provider}@{sub}'
+            provider = next(iter(oauth))
+            oauth_sub = f'{provider}@{oauth[provider]["sub"]}'
         except Exception:
             oauth_sub = None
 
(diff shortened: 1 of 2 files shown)


tjbck:
bd8378f643dd8a56366ad9d1843b7d40f67cf1f9

PR #28113, opened 2026-08-04
refac: keep external connections until their last knowledge base is removed

Classic298:
Deleting an external knowledge base now clears its connection only when an admin removes the last knowledge base referencing it, matching the connection delete route.

> [!NOTE]


Classic298:
@tjbck ready for review. Deleting one external knowledge base clears the whole connection today, even while other knowledge bases still reference it. It now clears only when the last one goes, matching the connection delete route.

PR #28130, opened 2026-08-05
fix(analytics): clear loading state for incomplete custom date range

### LINKED ISSUE #28125: issue: Analytics tab spins forever after selecting Custom range without picking dates
## Check Existing Issues

## Installation Method

Source (uvicorn + vite), reproduced against `dev`.

## Open WebUI Version

v0.11.0 (`dev`)

## Browser

Firefox

## Confirmation

- [x] I have provided the relevant configuration and environment variables.

## Expected Behavior

Selecting `Custom range` in Admin Settings > Analytics without yet picking dates should leave the tab usable. Returning to the tab later should either show the dashboard or prompt for the missing dates.

## Actual Behavior

The Analytics tab renders a loading spinner and never resolves. The dashboard, the summary stats, and the model tables are all hidden behind it. Reloading the page does not help, because the selection is persisted, so the tab stays unusable until the range is completed or `localStorage` is cleared.

## Steps to Reproduce

1. Open Admin Settings > Analytics.
2. Change the period selector to `Custom range` but do not set a start or end date.
3. Navigate to any other settings tab.
4. Return to Analytics.
5. The tab shows a spinner indefinitely.

## Cause

In `src/lib/components/admin/Analytics/Dashboard.svelte`:

- `selectedPeriod` is restored from `localStorage`, so `custom` survives leaving the tab.
- `customStart` and `customEnd` restore as empty strings when no dates were picked.
- `loading` is initialised to `true`.
- The reactive block that loads data deliberately declines to run with a half-specified range:

```js
$: if (selectedPeriod === 'custom' ? customStart && customEnd : selectedPeriod) {
	customStart;
	customEnd;
	selectedGroupId;
	loadDashboard();
}
```

`loading` is only ever cleared at the end of `loadDashboard()`. When the guard declines, nothing clears it, so the spinner is permanent. The guard itself is correct; the problem is that "waiting for input" and "loading" render identically, and there is no path out of the waiting state.

## Additional Information

The same state is reachable on first use, not only after navigating away: any point where the component remounts with `custom` selected and no dates produces it.

This report was prepared with AI assistance; the behavior was reproduced and the code path verified against `dev`.

Awshesh12:
Fixes #28125

When the Analytics period selector is set to Custom range without picking both start and end dates, the reactive guard correctly avoids calling loadDashboard() so it does not issue a half-specified query. However, loading is initialized to true and was never cleared in that branch, leaving the tab stuck on a permanent spinner. The state persists across tab navigation because selectedPeriod and the custom date strings are stored in localStorage.

This change adds an else branch to the reactive guard that sets loading = false when customStart or customEnd is missing, so the date inputs remain visible and usable.

### Verification
- Open Admin Settings > Analytics.
- Select Custom range without choosing dates.
- Navigate away and return, or reload the component.
- The spinner now resolves and the date pickers remain usable.
- Selecting both dates triggers the normal dashboard load.

### DIFF
--- a/src/lib/components/admin/Analytics/Dashboard.svelte
+++ b/src/lib/components/admin/Analytics/Dashboard.svelte
@@ -168,6 +168,9 @@
 		customEnd;
 		selectedGroupId;
 		loadDashboard();
+	} else {
+		// A half-specified custom range must not leave the spinner running forever.
+		loading = false;
 	}
 
 	onMount(async () => {


Awshesh12:
Closed in favor of #28131, which targets the `dev` branch as required by the project contribution guidelines.

PR #28126, opened 2026-08-05
fix: stop the analytics dashboard spinning when a custom range has no dates

silentoplayz:
# Pull Request Checklist

### Note to first-time contributors: Please open a discussion post in [Discussions](https://github.com/open-webui/open-webui/discussions) to discuss your idea/fix with the community before creating a pull request, and describe your changes before submitting a pull request.

This is to ensure large feature PRs are discussed with the community first, before starting work on it. If the community does not want this feature or it is not relevant for Open WebUI as a project, it can be identified in the discussion before working on the feature and submitting the PR. For bug fixes referencing an existing Issue, linking the Issue is sufficient.

**Before submitting, make sure you've checked the following:**

- [x] **Linked Issue/Discussion:** Closes #28125
- [x] **Target branch:** The pull request targets the `dev` branch. **PRs targeting `main` will be immediately closed.**
- [x] **Description:** A concise description of the changes is provided below.
- [x] **Changelog:** A changelog entry following [Keep a Changelog](https://keepachangelog.com/) format is included at the bottom.
- [ ] **Documentation:** No documentation changes needed.
- [x] **Dependencies:** No new or updated dependencies.
- [x] **Testing:** Manual tests have been performed to verify the fix works correctly and does not introduce regressions.
- [x] **No Unchecked AI Code:** This PR was prepared with AI copilot assistance and has been thoroughly reviewed and manually tested by the author.
- [x] **Self-Review:** A self-review of the code has been performed, ensuring adherence to project coding standards.
- [x] **Architecture:** No new settings. The change is local UI state only.
- [x] **Git Hygiene:** The PR is atomic (one logical change), rebased on `dev`, and contains no unrelated commits.
- [x] **Title Prefix:** The PR title uses one of the following prefixes:
  - **fix**: Bug fixes or corrections

# Changelog Entry

### Description

- Selecting `Custom range` in Admin Settings > Analytics without picking dates leaves the tab showing a spinner that never resolves, hiding the dashboard entirely. The selection is persisted in `localStorage`, so it survives leaving the tab and reloading the page.
- `loading` starts as `true` and is only cleared at the end of `loadDashboard()`. The reactive block that calls it deliberately declines to run while a custom range is half specified, which is correct, but nothing then clears `loading`. Waiting for input and loading render identically, and there is no path out of the waiting state.

### Fixed

- The Analytics dashboard no longer spins indefinitely when `Custom range` is selected without dates. The waiting state is now distinct from the loading state and prompts for the missing dates instead.

---

[...]

PR #28131, opened 2026-08-05
fix(analytics): clear loading state for incomplete custom date range

### LINKED ISSUE #28125: issue: Analytics tab spins forever after selecting Custom range without picking dates

Awshesh12:
# Pull Request

Closes #28125

### Description

When the Analytics period selector is set to **Custom range** without picking both start and end dates, the reactive guard correctly skips `loadDashboard()` to avoid issuing a half-specified query. However, `loading` is initialized to `true` and was never cleared in that branch, leaving the tab stuck on a permanent spinner. Because `selectedPeriod`, `customStart`, and `customEnd` are persisted in `localStorage`, this state survives tab navigation and page reloads.

This PR adds an `else` branch to the reactive guard that sets `loading = false` when the custom range is incomplete, so the date inputs remain visible and usable.

### Changes

- `src/lib/components/admin/Analytics/Dashboard.svelte`: clear `loading` when `selectedPeriod === "custom"` and either date is missing.

# Changelog Entry

### Fixed

- Analytics tab no longer shows an infinite spinner when "Custom range" is selected without dates. (#28125)

---

### Additional Information

- Reproduced against the `dev` branch.
- The fix was verified by inspecting the reactive block and confirming that `loading` is now cleared in the previously empty branch.

### DIFF
--- a/src/lib/components/admin/Analytics/Dashboard.svelte
+++ b/src/lib/components/admin/Analytics/Dashboard.svelte
@@ -168,6 +168,9 @@
 		customEnd;
 		selectedGroupId;
 		loadDashboard();
+	} else {
+		// A half-specified custom range must not leave the spinner running forever.
+		loading = false;
 	}
 
 	onMount(async () => {


silentoplayz:
#28126 was opened for this issue just over two hours earlier, at 2026-08-05T00:13:53Z, and is still open.

The diagnosis here matches. The difference is what the tab shows once the spinner clears. This PR leaves the summary row rendering, and since `summary` is initialized to zeroes it reports zero messages, zero tokens, zero chats, and zero users, which reads as an empty result rather than a range that has not been picked yet. #28126 extracts the condition into `awaitingCustomRange`, suppresses the summary row while waiting, and renders a prompt to select a start and end date in place of the dashboard body.

PR #24822, opened 2026-05-17
fix: interpolate template vars in MCP connection headers

JawafdehiBot:
# Pull Request Checklist

- [ ] **Linked Issue/Discussion:** —
- [x] **Target branch:** Verify that the pull request targets the `dev` branch. **PRs targeting `main` will be immediately closed.**
- [x] **Description:** Provide a concise description of the changes made in this pull request down below.
- [x] **Changelog:** Ensure a changelog entry following the format of [Keep a Changelog](https://keepachangelog.com/) is added at the bottom of the PR description.
- [x] **Documentation:** Add docs in [Open WebUI Docs Repository](https://github.com/open-webui/docs). Document user-facing behavior, environment variables, public APIs/interfaces, or deployment steps.
- [x] **Dependencies:** Are there any new or upgraded dependencies? If so, explain why, update the changelog/docs, and include any compatibility notes. Actually run the code/function that uses updated library to ensure it doesn't crash.
- [x] **Testing:** Perform manual tests to **verify the implemented fix/feature works as intended AND does not break any other functionality**. Include reproducible steps to demonstrate the issue before the fix. Test edge cases (URL encoding, HTML entities, types). Take this as an opportunity to **make screenshots of the feature/fix and include them in the PR description**.
- [x] **Agentic AI Code:** Confirm this Pull Request is **not written by any AI Agent** or has at least **gone through additional human review AND manual testing**. If any AI Agent is the co-author of this PR, it may lead to immediate closure of the PR.
- [x] **Code review:** Have you performed a self-review of your code, addressing any coding standard issues and ensuring adherence to the project's coding standards?
- [x] **Design & Architecture:** Prefer smart defaults over adding new settings; use local state for ephemeral UI logic. Open a Discussion for major architectural or UX changes.
- [x] **Git Hygiene:** Keep PRs atomic (one logical change). Clean up commits and rebase on `dev` to ensure no unrelated commits (e.g. from `main`) are included. Push updates to the existing PR branch instead of closing and reopening.
- [x] **Title Prefix:** To clearly categorize this pull request, prefix the pull request title using one of the following:
  - **fix**: Bug fix or error correction

# Changelog Entry

### Description

OpenWebUI middleware.py applies MCP connection custom headers directly without calling `get_custom_headers()`, so template variables like `{{USER_ID}}` and `{{USER_NAME}}` are sent literally to MCP servers.

### Fixed

[...]

PR #28157, opened 2026-08-05
feat: per-user credentials for admin-defined tool server connections

shaba:
> Smaller resubmission of #28149, which was closed as too large. Same feature, trimmed to what
> the project actually carries: 16 files, +1431/−97, in two commits (backend, then UI).
> The test commit and the regenerated locale files are gone — see "What changed since #28149".

- [x] **Linked Issue/Discussion:** Relates to #19313
- [ ] **Documentation:** Not written yet — happy to open a PR against the docs repository if the approach here is accepted.
- [x] **Testing:** Manually tested end to end on my own instance; screenshots below.
- [x] **No Unchecked AI Code:** Written with AI assistance, reviewed line by line and manually tested by me. See the note under Additional Information.

# Changelog Entry

### Description

An admin-defined tool server connection can now declare credential fields that each user fills
in for themselves, so a shared connection no longer needs a shared API key. The connection stays
admin-owned: users only supply values for fields the admin declared, and cannot add servers of
their own.

- `auth_type: user_key` is the shorthand — the connection implicitly asks for one key, sent as
  `Authorization: Bearer <the user's value>`.
- For anything else, header templates take `{{USER_SECRET:<field>}}` next to the existing
  `{{USER_ID}}` / `{{USER_EMAIL}}` placeholders. The field declaration is a plain JSON Schema in
  `config.user_config`, so the existing valves renderer draws the form and `input.type: password`
  fields get the masked input.
- MCP connections also accept the placeholder in the URL (percent-encoded), which covers servers
  that expect the key in the path. OpenAPI connections do not: their specs are fetched once per
  instance and shared by all users, so saving such a connection is refused with an explanatory
  error rather than failing silently.

Both server types go through `build_tool_server_headers()`, so both are covered.

Storage and security:

- Values are stored per user in `user.settings`, keyed by the connection ID, and go through the
  same helper that stores tool valves — which means they inherit its behaviour at rest: encrypted
  where `ENABLE_VALVE_ENCRYPTION` is turned on, and stored as submitted where it is not, which is
  the default today. No new table, no migration, no new dependency.
- Password fields are write-only through the API: it reports whether a value is set and never
  returns it. The credentials are also stripped from the settings the user and admin endpoints
  hand back, and the generic settings update endpoint ignores the key, so values can only be
  written through the endpoints that validate them.
- Secrets are interpolated last, in a single pass, so a value containing `{{...}}` is not expanded
  again. Reserved headers (hop-by-hop, `Cookie`, the configured `FORWARD_*` user-info headers)
  refuse to carry one; submitted values are length-checked and rejected if they contain control
  characters.
- A user missing required credentials is told which ones and where to set them, instead of getting
  a generic connection failure.

### What changed since #28149

Nothing was dropped from the feature; what was dropped is everything around it that this project
does not carry:

- **No test files.** The previous version added `backend/open_webui/test/` along with the suite,
  since there is no backend test tree in `dev` today. Standing up a test infrastructure is not
  mine to decide, so it is out. The 48 tests still exist in my own branch and I am happy to bring
  them over — as a follow-up, in whatever layout you would want one.
- **One locale file, not 63.** The previous version ran `i18n:parse`, which regenerated every
  locale and accounted for ~1050 of its added lines. Only `en-US` is touched here; the rest is
  yours to regenerate whenever you next do.
- **Two commits instead of one.** Backend first, then the UI, so the server-side contract can be
  read on its own.

That takes it from 82 files / +3020−97 to 16 files / +1431−97.

### Added

- Per-user credentials for admin-defined tool server connections: a `user_key` auth type and a
  `{{USER_SECRET:<field>}}` placeholder for connection headers and MCP URLs, stored per user and
  never returned to the UI.

### Changed

- (none)

### Deprecated

- (none)

### Removed

- (none)

### Fixed

- (none)

### Security

- Per-user tool server credentials are excluded from user settings API responses, ignored on the
  generic settings update endpoint, and rejected in reserved headers.

### Breaking Changes

- (none)

---

### Additional Information

Follows the `{{USER_*}}` placeholder convention introduced in #24822 and reuses the existing
valves storage, encryption helper and form renderer, so there is no new storage layer and no
migration.

One thing worth being explicit about, because it is easy to misread: these values are only
encrypted at rest when `ENABLE_VALVE_ENCRYPTION` is enabled, exactly like tool valves. That flag
is off by default, so on a default deployment the values sit in the `user` row the same way valve
values do. I did not want to invent a second storage policy for this, but it is a deliberate
choice rather than an oversight, and if you would rather these were always encrypted regardless of
the flag, that is a one-line change.

On AI assistance, per the "No Unchecked AI Code" item: I used an AI assistant while writing this
change. Every line has been read and reviewed by me, the design decisions are mine, and I ran the
feature end to end on my own instance — the screenshots below are from that run. This description
is my own text.

### Screenshots or Videos

Admin declares the field on the connection:

Admin declares a per-user credential field on the connection

The user fills it in from Settings → Integrations, masked:

User fills in their own key, input masked

The tool call goes out with that user's key:

Tool answers using that user's own key

A second account with a wrong key gets an auth error on the same question, so the credential
really is per user:

Second account with a wrong key gets an auth error

> [!NOTE]


shaba:
Understood — no objection to the close, and thanks for going through the backlog.

Before I spend more effort on this, could you say which of these it is?

1. **Scope.** #28149 was closed as "way too large to merge in", so I resubmitted this at 16 files instead of 82. If it is still too big, I am happy to split it: the smallest useful piece is just resolving `{{USER_SECRET:...}}` placeholders in tool-server headers — roughly a hundred lines, no UI, no MCP URL substitution. Everything else can follow as separate PRs.

2. **Direction.** If per-user credentials for admin-defined tool servers are not something Open WebUI wants — for instance because credentials belong in a secrets backend rather than in user settings — then that is a clear answer and I will stop here rather than keep resubmitting. #19313 has been open for nine months with steady interest, so it would be useful for the people in that thread to know either way.

3. **Housekeeping.** If this was part of clearing out stale PRs, note that I rebased it onto current `dev` a few minutes before it was closed: it now applies cleanly (`MERGEABLE`), the two backend helpers it used to duplicate are gone (upstream moved `normalize_bearer_token` and `bearer_auth_header` into `utils/headers.py`, and the branch now imports them), and the frontend part was re-applied on top of the reworked `IntegrationsMenu.svelte` — the diff against `dev` is 46 additive lines there.

For context on why this matters to us: we run an internal deployment where several MCP servers are defined by an admin, but each user has to authenticate as themselves — a personal Bugzilla API key, a personal Redmine key, a MediaWiki bot password. Without per-user credentials the choice is either one shared account for everyone, which loses per-user permissions and audit trails, or no integration at all.

Happy to take any of the three answers.

Issue #28166, opened 2026-08-05
perf: websocket heartbeat interval is hardcoded at 30s with no way to tune it


Classic298:
every open tab sends a websocket heartbeat every 30 seconds. that number is hardcoded in the frontend, so there is no way to change it.

each heartbeat costs a redis read plus a redis write to refresh the session entry, and a database write to update the user's last-active timestamp. none of that is triggered by anything the user does, it happens for every tab that is merely open in the background.

on a 16 container deployment with ~4000 users that works out to well over 100 heartbeats a second cluster wide, so a constant ~100+ redis round trips and ~100+ database writes per second of pure idle background load. it scales with tabs, not with usage.

every other websocket timing value is already an env var (ping interval, ping timeout), this one is not, so a large deployment cannot do anything about it.

what i would like: `WEBSOCKET_HEARTBEAT_INTERVAL`, default 30 so nothing changes for anyone who does not set it, clamped to a range that is actually safe. two things constrain the upper end: the session reap timeout is a fixed 120 seconds, and "active" everywhere else in the app means a last-active timestamp newer than 3 minutes, so an interval past those just makes connected users look dead. so the reap timeout should follow the interval, and the interval itself should be capped somewhere below 3 minutes. even 90 seconds cuts this background load to a third.

the tradeoff is entirely the admin's to make. a longer interval means last-active timestamps and the active user count get coarser. on a deployment where those are looked at once a day, that is free.

i'll open a PR for this.

PR #28174, opened 2026-08-05
perf: make the websocket heartbeat interval configurable

Classic298:
Every open tab pings the server every 30 seconds, and that number was hardcoded in the frontend. Each ping costs a Redis read and a Redis write to refresh the session entry plus a database write for the user's last-active timestamp, so a large deployment carries a constant background load that no user action triggered: a few thousand open tabs work out to well over 100 Redis round trips and 100 database writes per second, whether anyone is typing or not.

WEBSOCKET_HEARTBEAT_INTERVAL now sets that interval and the server hands the value to the client in the config response. The default of 30 keeps today's behaviour exactly, and 90 cuts that idle background traffic to a third of it.

Two things bound how far it can go, so the value is clamped to 5-90 seconds. The session reaper derives its timeout from the interval instead of using a fixed 120 seconds, but never drops below that, so clients still running an older frontend bundle keep their 30 second beat and are not reaped while connected. Separately, "active" everywhere else in the app means a last-active timestamp newer than 3 minutes, so an interval anywhere near that would make connected users render as offline, drop out of the active user count, and start receiving away notifications.

Closes #28166

> [!NOTE]

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -499,6 +499,13 @@ try:
 except ValueError:
     WEBSOCKET_SERVER_PING_INTERVAL = 25
 
+# Seconds between client heartbeats, capped so last_active_at stays inside the 3 minute active window.
+WEBSOCKET_HEARTBEAT_INTERVAL = os.getenv('WEBSOCKET_HEARTBEAT_INTERVAL', '30')
+try:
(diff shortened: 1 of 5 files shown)


Classic298:
@tjbck ready for review. Every open tab pings every 30 seconds and that number was hardcoded in the frontend. Each ping costs a Redis read, a Redis write and a database write, so a few thousand idle tabs is well over 100 Redis round trips and 100 database writes per second with nobody typing. Default of 30 keeps today's behaviour exactly.

Issue #28165, opened 2026-08-05
perf: last_active_at is written on every authenticated request by default


Classic298:
Presence tracking has a per-user write throttle behind `DATABASE_USER_ACTIVE_STATUS_UPDATE_INTERVAL`, but it ships unset, and unset means no throttle at all. So out of the box every authenticated request, every API-key request and every websocket heartbeat does its own `UPDATE users SET last_active_at` plus a commit.

On a busy instance that is a constant stream of tiny write transactions that nothing needs. Presence only needs minute granularity, the UI never shows anything finer.

The catch is that you can't just set the interval high. The "active users" number, the green online dot in the admin user list and the away-notification gating all treat a user as active if they were seen in the last 3 minutes, and that 3 minutes is hardcoded (four places in the backend, plus the frontend computes the dot itself). So the moment the throttle sits at or above 180 seconds, people who are actively using the instance start dropping out of the count. The docs currently recommend 300 to 500 seconds, which means everyone following that advice is undercounting active users right now.

Making the window follow the interval instead is not a backend-only change, because the frontend would keep its own 3 minutes and the two would disagree for exactly the settings the derivation is supposed to fix. Widening the window also delays away-notification delivery, which is its own behaviour decision. So the constraint has to go the other way: the interval stays under the window.

What I think it should be: default the throttle to 60 seconds. A stock install then writes at most once per minute per user per worker instead of once per request, the 3 minute window stays correct and nothing visible changes. `0` keeps the old unthrottled behaviour for anyone who wants it, and the docs recommendation needs correcting to match.

Separate and smaller, worth its own change: with an API key the presence write is awaited inline so its database latency lands on the request, while a normal login session fires the same write off in the background. The two auth paths should agree on one of those, and the background one currently drops errors and doesn't hold a reference to the task.

PR #28177, opened 2026-08-05
perf: throttle last_active_at writes by default

Classic298:
Presence tracking writes each user's last_active_at on every authenticated request, every API key request and every websocket heartbeat. The throttle for it already exists but ships unset, and unset means no throttle at all, so a stock deployment pays one UPDATE plus COMMIT per user per request. The 30 second frontend heartbeat alone is 2 write transactions per minute per open tab, before any actual UI traffic.

Defaulting the throttle to 60 seconds collapses that to at most one write per user per worker per minute. Presence is only ever read at minute granularity, so nothing visible changes.

60 rather than the 300 to 500 the docs currently suggest, because a user counts as active for 3 minutes after their last write and that window is hardcoded in the backend and again in the frontend. Any interval at or above 180 seconds makes people who are actively using the instance drop out of the active user count. Letting the window follow the interval instead would need the value shipped to the client, so that is a separate change.

0 still disables the throttle, and now costs nothing at all: the decorator returns the undecorated function instead of a wrapper that re-checks a constant on every call.

Closes #28165

> [!NOTE]


Classic298:
@tjbck ready for review. The presence throttle already exists but ships unset, and unset means no throttle, so a stock deployment writes plus commits once per user per request. The 30 second heartbeat alone is 2 write transactions per minute per open tab before anyone touches the UI. Presence is only ever read at minute granularity, so defaulting it changes nothing visible.

Issue #28164, opened 2026-08-05
perf/issue: Every Socket.IO event scans its whole payload for binary data that is never there


Classic298:
Before the backend sends any socket.io event, python-socketio walks the entire payload recursively to check whether any value is a `bytes` object needing binary attachment framing. Open WebUI never emits binary, so that walk always comes back empty and every bit of the work is thrown away.

It is not cheap. The walk has no early exit and builds an intermediate list at every level, so it scales with the full size of the message. The payloads it runs on are the big ones: chat streaming re-emits the entire assistant message (content, sources, status history, usage) on every update, and note collaboration sends document state as a JSON array with one entry per byte.

Measured per encode, on a 7.5 KB chat-completion event and a 64 KB collaborative document state:

| payload | with the scan | without the scan |
|---|---|---|
| chat completion re-emit (7.5 KB JSON) | 30 us | 13 us |
| document state (292 KB JSON) | 9.0 ms | 1.7 ms |

With `ENABLE_ORJSON=true` the actual JSON encoding gets much faster, so the scan ends up being nearly the entire cost: 20 us → 2.3 us and 7.8 ms → 0.14 ms.

It gets worse with the Redis websocket manager. An emit is published to Redis and every instance then builds and encodes its own copy of the packet, so the scan runs once per instance per emit, plus once more inside the Redis manager before publishing. On a 16 container deployment that is 17 full walks of the payload for a single event.

Profiling such a deployment (16 containers, ~4000 users) with py-spy put roughly 5-8% of total backend CPU in this one scan.

### Expected behaviour

The server should not pay a full payload walk on every emit to look for something it never sends.

### Environment

Open WebUI `dev`, python-socketio 5.16.1, any deployment. The cost scales with socket.io traffic and with instance count, so it is worst on multi-container instances with active chat streaming or collaborative notes.

PR #28180, opened 2026-08-05
perf: stop scanning every socket.io payload for binary data

Classic298:
Every socket.io event the backend sends was first walked recursively to check whether any value was a bytes object needing binary attachment framing. Open WebUI never emits binary, so the walk always came back empty and the work was thrown away. It has no early exit and allocates at every level, so it scaled with the full size of the message, and the messages are the big ones: chat streaming re-emits the whole assistant message on every update, note collaboration sends document state as a JSON array with one entry per byte. With the Redis manager it ran once per instance per emit on top of that, since every instance builds its own copy of the packet.

The server now installs a Packet subclass with binary events off, through python-socketio's own serializer hook, the same mechanism its msgpack serializer uses. Inbound binary attachments are decoded to int lists rather than refused, so the one frontend path that sends a raw Uint8Array keeps working and handlers can still echo client data straight back out. One scan remains in multi-instance setups: python-socketio's Redis manager calls it on the base Packet class directly, where the serializer hook cannot reach.

Measured per encode:

| payload | before | after |
|---|---|---|
| chat completion re-emit (7.5 KB JSON) | 30 us | 13 us | | collaborative document state (292 KB JSON) | 9.0 ms | 1.7 ms |

With ENABLE_ORJSON=true, where the scan is nearly the whole encode cost: 20 us to 2.3 us, and 7.8 ms to 0.14 ms.

Closes #28164

> [!NOTE]

### DIFF
--- a/backend/open_webui/socket/main.py
+++ b/backend/open_webui/socket/main.py
@@ -5,6 +5,7 @@ import logging
 import random
 import sys
 import time
+from typing import Any
 
 import pycrdt as Y
 import socketio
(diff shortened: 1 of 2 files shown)


Classic298:
@tjbck ready for review. Every socket.io event is walked recursively to look for bytes needing binary framing. Open WebUI never emits binary, so that walk always comes back empty, and it scales with the full size of the message. The big messages are exactly the ones that pay it: streaming re-emits the whole assistant message on every update.

Issue #28169, opened 2026-08-05
perf: every message-level write re-reads and rewrites the entire chat, so long chats get progressively more expensive to stream into


Classic298:
profiled a prod deployment (16 containers, ~4000 users) with py-spy. a large share of backend time sits in chat message persistence.

every time the backend touches a single message, it round-trips the whole conversation. appending streamed content to a message loads the entire chat json, walks every string in it for a null-byte sanitisation pass, then writes the whole json column back. reading a single message is the same full load plus a full pydantic validation. the websocket event emitter does the read and then the write, so one "message" event is two complete round trips of the entire conversation just to change a few hundred bytes.

the consequence is that cost scales with chat length rather than with the size of the edit. the longer a conversation gets, the more expensive every single write into it becomes, so the exact users who use the product most are the ones generating the most db load. on a shared postgres with a few thousand users this is a meaningful share of total backend time, and it compounds because every status update, every source, every file attachment and every tool result goes through the same path.

measured locally on sqlite with synthetic conversations, median of 80 runs. one "message" event is one read plus one write:

| conversation | read, today | read, after | write, today | write, after |
|---|---|---|---|---|
| 50 messages (100 KiB) | 0.80 ms | 0.55 ms | 2.04 ms | 1.94 ms |
| 200 messages (400 KiB) | 1.84 ms | 0.72 ms | 4.25 ms | 2.61 ms |
| 600 messages (1.2 MiB) | 5.59 ms | 1.50 ms | 12.86 ms | 5.45 ms |

on postgres the read side should improve more than this, because the whole blob no longer has to cross the wire to be parsed in python just to pull one message out of it.

two things fix most of it without changing any behaviour: read the single message with a json path expression instead of loading and validating the whole chat, and sanitise only the data that is actually entering the chat, instead of re-walking the whole conversation on every write. i have a pr ready.

PR #28184, opened 2026-08-05
perf: stop round-tripping the whole chat to read or write one message

Classic298:
Long conversations get progressively more expensive to stream into. Every message-level write reloads the entire chat JSON, walks every string in it for a null-byte sanitisation pass and rewrites the whole column, and reading a single message loads and validates the whole chat too. The websocket event emitter reads and then writes, so one message event round-trips the full conversation twice to change a few hundred bytes.

Reading one message now selects only history.messages out of the JSON column and indexes it in Python, so the rest of the chat is neither loaded nor model-validated. Writing one message now sanitises only what is actually entering the chat, plus the title column that is mirrored off the blob, instead of re-walking a conversation that was already sanitised when it was written. Legacy rows are still healed on read by get_chat_by_id, which is unchanged.

One behaviour change: chats written before the null-byte sanitisation existed can still hold null bytes in the stored JSON. Those rows used to be rewritten clean as a side effect of any message write, and are now cleaned when the chat is read instead. The visible consequence is that the message write and delete endpoints echo back the updated chat, and for such a legacy row that echo now carries the raw null bytes rather than stripped ones, until the next read of that chat heals it. A GET of the chat is unaffected, and title, the one text column that PostgreSQL cannot store a null byte in, is still sanitised on every write.

Measured on SQLite with a 10.8 MB chat (3000 messages): a single-message upsert is 227 ms before the branch and 173 ms after, and the legacy single-message read is 135 ms before and 79 ms after. The whole-blob scan the delete path used to run costs 5.40 ms on a 1.8 MB chat and 38.79 ms on the 10.8 MB one, and is gone.

Ref https://github.com/open-webui/open-webui/issues/28169

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/chats.py
+++ b/backend/open_webui/models/chats.py
@@ -400,9 +400,6 @@ class ChatTable:
         """
         Clean a Chat SQLAlchemy model's title + chat JSON,
         and return True if anything changed.
-
-        The message write paths (upsert/status/delete) rely on this
-        leaving the blob clean and sanitize only the data they add.
         """
         changed = False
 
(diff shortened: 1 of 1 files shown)


Classic298:
@tjbck ready for review. Every message-level write reloads the entire chat JSON, walks every string in it, and rewrites the whole column, and reading one message loads and validates the whole chat too. The websocket emitter reads then writes, so one message event round-trips the full conversation twice to change a few hundred bytes.

PR #28191, opened 2026-08-05
fix(oauth): require provider-verified email before merging accounts by email

SashaMIT:
## Summary

When `OAUTH_MERGE_ACCOUNTS_BY_EMAIL=True`, the callback binds a new `(provider, sub)` identity to an existing account looked up **by email claim alone** (`utils/oauth.py` merge branch). No code checks whether the identity provider actually *verified* that email address.

Any IdP that lets users assert arbitrary unverified email addresses (e.g. a Keycloak realm with "verify email" off, or a provider that simply forwards a user-supplied `email` claim) allows account takeover: attacker sets their IdP email claim to the victim's address, signs in, and the backend merges the attacker's identity into the victim's existing account — including admin accounts.

## Fix

- Track provider verification during email extraction: the OIDC `email_verified` claim (bool or string form), and GitHub's per-address `verified` flag in the `/user/emails` fallback.
- The merge-by-email path now refuses (400) when the provider **explicitly reports the email as unverified**. Providers that send no verification signal keep current behavior (non-breaking for IdPs without the claim concept).

No behavior change for: normal login of existing linked accounts, first-time OAuth signup (new accounts carry no takeover risk; `EMAIL_TAKEN` still guards existing ones), and providers that verify email.

## Test plan

- [ ] Keycloak (verify-email off): user with unverified email matching an existing local account is rejected at merge with 400 instead of being merged
- [ ] Keycloak (email verified): merge still succeeds
- [ ] GitHub provider without public email: fallback still resolves the primary address; merge works when GitHub reports it verified

Made with Cursor

Made with [Cursor](https://cursor.com)

### DIFF
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -1841,6 +1841,10 @@ class OAuthManager:
             # Email extraction
             email_claim = auth_config.OAUTH_EMAIL_CLAIM
             email = user_data.get(email_claim, '')
+            # Whether the provider vouches for the email address. None means
+            # the provider gave no verification signal; set by the claim
+            # check below or by the GitHub /user/emails fallback.
+            email_verified = None
             # We currently mandate that email addresses are provided
             if not email:
                 # If the provider is GitHub,and public email is not provided, we can use the access token to fetch the user's email
@@ -1857,12 +1861,15 @@ class OAuthManager:
                                 if resp.ok:
                                     emails = await resp.json()
                                     # use the primary email as the user's email
-                                    primary_email = next(
-                                        (e['email'] for e in emails if e.get('primary')),
+                                    primary_entry = next(
+                                        (e for e in emails if e.get('primary')),
                                         None,
                                     )
-                                    if primary_email:
-                                        email = primary_email
+                                    if primary_entry:
+                                        email = primary_entry['email']
+                                        # GitHub reports per-address verification; keep it for
+                                        # the account-merge gate below
+                                        email_verified = bool(primary_entry.get('verified'))
                                     else:
                                         log.warning('No primary email found in GitHub response')
                                         raise HTTPException(400, detail=ERROR_MESSAGES.INVALID_CRED)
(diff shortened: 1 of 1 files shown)


Classic298:
> When OAUTH_MERGE_ACCOUNTS_BY_EMAIL=True, the callback binds a new (provider, sub) identity to an existing account looked up by email claim alone (utils/oauth.py merge branch). No code checks whether the identity provider actually verified that email address.

That is intended behavior

SashaMIT:
Merging accounts by email is indeed the intended *feature* — this PR doesn't change that. The only new gate is the narrow case where the provider **explicitly reports the email as unverified** (`email_verified: false` in OIDC claims, or `verified: false` on the GitHub `/user/emails` entry it already fetches).

If the IdP itself says "I have not verified this address", merging anyway converts a self-declared untrusted claim into access to the matching existing account — that's the account-takeover case (e.g. a Keycloak realm with email verification off, where any user can assert any address). Providers that send no verification signal at all keep the current behavior, so nothing changes for IdPs without the claim concept.

Happy to adjust the approach if you'd prefer this behind a config flag instead.

Classic298:
this would break deployments relying on the existing behaviour. it is not a fix it is a breaking change. and the fact that this env var is unsafe/insecure and should only be used in very narrow certain cases is documented. It is deliberately without email verification for Identitiy Providers that dont verify emails.

SashaMIT:
Fair point, and you're right that the var is documented as unsafe for exactly this reason. One clarification on scope, though: this PR only changes behavior when the provider **explicitly reports `email_verified: false`** — IdPs that send no verification claim at all (the case the docs describe) behave exactly as before. So the only deployments that would break are ones merging on emails their IdP actively labels unverified, which is the specific configuration that converts the documented risk into a direct takeover path.

That said, if the preference is to keep even that case working, I'm happy to put the check behind its own flag (e.g. `OAUTH_MERGE_REFUSE_UNVERIFIED_EMAIL`) defaulting to off, so the PR becomes pure hardening with zero behavior change unless opted in. Or if the maintainers feel the existing documentation warning is sufficient, feel free to close — no hard feelings either way.

Classic298:
yeah that would be better. also fix  the other issues the bot raised

SashaMIT:
Done — reworked as an opt-in flag (default off, zero behavior change) and resubmitted against `dev` with the CLA/template requirements fixed: #28196

Issue #28202, opened 2026-08-06
issue: [BUG] Deactivating an Ollama Cloud model 500s /api/models (get_all_models: list.remove(x): x not in list)


bekircaner:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

0.11.0

### Ollama Version (if applicable)

0.32.6

### Operating System

Ubuntu 24

### Browser (if applicable)

Chrome 151

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

Deactivating a model in Admin Settings → Models should simply hide it from the
model list. It should not make `GET /api/models` return HTTP 500.
`get_all_models` should guard the `list.remove(...)` (remove only if present)
so a stale/mismatched model id cannot crash the whole model-list assembly.

### Actual Behavior

After deactivating one or more Ollama **Cloud** models, `GET /api/models`
returns HTTP 500. The response body is the plain-text `Internal Server Error`,
so the frontend can't parse it as JSON and the model dropdown stays empty:

`Failed to load models: SyntaxError: Unexpected token 'I', "Internal S"... is not valid JSON`

Re-activating the deactivated model(s) makes `/api/models` return 200 again.

The Cloud connection itself is fine — `routers.ollama:get_all_models` logs INFO
[...]

PR #28218, opened 2026-08-06
fix: avoid duplicate base model removal

loulanyue:
Closes #28202.

## Summary

Prevent `get_all_models()` from removing the same Ollama model object twice when inactive custom-model aliases resolve to both the base name and the exact tagged ID.

## Root cause

`base_model_lookup` stores references under both an Ollama model's untagged base name and exact ID. After the first inactive alias removes the object from `models`, the lookup still returns that object for a second alias, so the unconditional `list.remove()` raises `ValueError` and `/api/models` returns 500.

The change keeps the existing removal behavior but makes it idempotent by checking that the looked-up object is still present.

## Verification

- Extracted and executed the relevant `get_all_models()` AST block directly from both `origin/dev` and the patched source.
  - One exact inactive ID: passes before and after the change.
  - Base-name plus exact-ID aliases: `origin/dev` reproduces `ValueError('list.remove(x): x not in list')`; patched source passes and leaves the list empty.
- `ruff format --no-cache --check backend/open_webui/utils/models.py`: passed.
- `ruff check --no-cache --select E9,F63,F7,F82 backend/open_webui/utils/models.py`: passed.
- In-memory Python compilation and `git diff --check`: passed.

Docker/Ollama Cloud end-to-end verification was not available because the local Docker daemon is not running. Importing the complete backend module was also unavailable because the local system Python environment does not contain `uvicorn`.

AI assistance was used for investigation and preparation of this one-line change. The contributor reviewed the final diff and verification evidence before submission.

## Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #28202.
- [x] **Description:** The change and root cause are described above.
- [x] **Documentation:** No documentation change is needed for this internal error-handling fix.
- [x] **Dependencies:** No dependencies were added or changed.
- [x] **Testing:** Focused manual logic validation was performed; environment limitations are disclosed above.
- [x] **No Unchecked AI Code:** The contributor reviewed the final diff and verification evidence before submission.
- [x] **Self-Review:** The one-line diff, source history, and regression path were reviewed.
- [x] **Architecture:** No settings or architectural changes were introduced.
- [x] **Git Hygiene:** The branch is based on current `dev` and contains one atomic commit.
- [x] **Title Prefix:** The title uses `fix:`.

# Changelog Entry

### Fixed

- Prevented `/api/models` from failing when inactive Ollama aliases resolve to an already removed model.

### Additional Information

- Related issue: #28202

### DIFF
--- a/backend/open_webui/utils/models.py
+++ b/backend/open_webui/utils/models.py
@@ -181,7 +181,7 @@ async def get_all_models(request, refresh: bool = False, user: UserModel = None)
 
                     model['action_ids'] = action_ids
                     model['filter_ids'] = filter_ids
-                else:
+                elif model in models:
                     models.remove(model)
 
         elif custom_model.is_active:


tjbck:
Addressed in dev, Thanks!

Issue #28222, opened 2026-08-06
issue: Notes page crashes when note md content is stored as dict/object instead of string


Scorpoon:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Pip Install

### Open WebUI Version

v0.11.0

### Operating System

Debian 13

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

When the LLM uses the replace_note_content tool to update a note with structured data (e.g., JSON), the content should be stored as a string in note.data['content']['md'], either as plain text or wrapped in a markdown codeblock.
The notes UI should render without crashing.

Note should be inserted like this:

<img width="501" height="231" alt="Image" src="https://github.com/user-attachments/assets/917631fa-6e88-4e27-867d-4412777d4fba" />

### Actual Behavior

The LLM sometimes passes a raw dict/object as the content parameter instead of a string. This dict is written directly into note.data['content']['md'] in the database. When the notes UI tries to render the note, it crashes because it expects a string but receives a dict. The same corruption affects downstream code in RAG/search pipelines that call .lower() or other string methods on md.

[...]

PR #28228, opened 2026-08-07
fix: handle non-string note content and multi-line :param docstrings

Scorpoon:

- [x] **Linked Issue/Discussion:** This PR fixes #28222 
- [ ] **Documentation:** Not applicable (internal tool schema fix)
- [x] **Testing:** Manual tests performed, 41 unit tests pass
- [x] **No Unchecked AI Code:** All changes reviewed and tested by a human
- [x] **Architecture:** Defensive read-side + write-side validation, no new settings

# Changelog Entry

### Description

Fixes Notes UI crash when note.data['content']['md'] contains a dict/object instead of a string.

Root causes:
1. parse_docstring() truncated multi-line :param descriptions – the LLM never saw the operations parameter examples, causing confusion between top-level content and nested content inside operations, leading to dicts being passed as note content.
2. No write-side validation – non-string content was accepted and persisted directly into the JSON column, corrupting existing notes.

The fix adds defensive read-side handling (ensure_md_string()), write-side validation (isinstance(content, str)), an Alembic migration to fix existing corrupt notes and improved tool docstrings with explicit examples.

### Added

- backend/open_webui/utils/notes.py – new ensure_md_string() utility that converts non-string values to ````text` code blocks
- backend/open_webui/migrations/versions/0f47c530da11_fix_corrupt_note_md.py – Alembic migration to fix existing corrupt note records
- Write-side validation in write_note() and replace_note_content() – rejects non-string content with clear error message
- Improved tool docstrings with IMPORTANT warning and explicit examples for JSON/code content

### Changed

- backend/open_webui/utils/tools.py – parse_docstring() now correctly handles multi-line :param descriptions (continuation lines until next :param/:return)
- backend/open_webui/tools/builtin.py – 4 defensive read locations now use ensure_md_string() with (note.data or {}).get('content') or {} pattern
- backend/open_webui/routers/notes.py – _truncate_note_data() now uses defensive read pattern

### Fixed

- Notes UI crash (KeyError: slice(None, 1000, None)) when rendering notes with dict md content
- LLM passing raw dicts as content parameter due to missing operations examples in parsed docstring

### Security

- No security impact identified during root-cause analysis and code review

---

### Additional Information

Migration: The Alembic migration (0f47c530da11) is idempotent – it only fixes notes where md is not a string. On clean installs it fixes 0 notes.
Testing: Tested on Debian 13 test VM with v0.11.0. Migration fixed 1 corrupt note, service started without migration errors, API responded correctly. Tool schema verified – LLM now sees explicit examples for wrapping JSON/code in markdown code blocks.

### Screenshots or Videos

Bildschirmfoto_20260807_020128

- [ ] By submitting this pull request, I confirm that I have read and fully agree to the [Contributor License Agreement (CLA)](https://github.com/open-webui/open-webui/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT), and I am providing my contributions under its terms.

> [!NOTE]

### DIFF
--- a/backend/open_webui/migrations/versions/0f47c530da11_fix_corrupt_note_md.py
+++ b/backend/open_webui/migrations/versions/0f47c530da11_fix_corrupt_note_md.py
@@ -1,0 +1,82 @@
+"""fix corrupt note md content
+
+Revision ID: 0f47c530da11
+Revises: 1ce6ade7d93b
+Create Date: 2026-08-05 21:06:03.932000
+
+"""
+
+from typing import Sequence, Union
+
+import json
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.sql import column, select, table, update
+
+# revision identifiers, used by Alembic.
+revision: str = '0f47c530da11'
+down_revision: Union[str, None] = '1ce6ade7d93b'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def _ensure_md_string(value):
+    """Inline copy of ensure_md_string for migration self-containment."""
+    if isinstance(value, str):
+        return value
+    if value is None:
+        return ''
+    return '```text\n' + json.dumps(value, indent=2, ensure_ascii=False) + '\n```'
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+    note_table = table(
+        'note',
+        column('id', sa.Text),
+        column('data', sa.JSON),
+    )
+
+    result = conn.execute(select(note_table.c.id, note_table.c.data)).yield_per(100)
+    fixed_count = 0
+
+    for note_id, data in result:
+        if data is None or not isinstance(data, dict):
+            continue
+
+        content = data.get('content')
+
+        # Handle non-dict content (e.g. list, str, int)
+        if not isinstance(content, dict):
+            fixed_md = _ensure_md_string(content)
+            fixed_data = {**data, 'content': {'md': fixed_md}}
+            conn.execute(
+                update(note_table)
+                .where(note_table.c.id == note_id)
+                .values(data=fixed_data)
+            )
+            fixed_count += 1
+            continue
+
+        md = content.get('md')
+        if md is None or isinstance(md, str):
+            continue
+
+        fixed_md = _ensure_md_string(md)
+        fixed_data = {
+            **data,
+            'content': {**content, 'md': fixed_md},
+        }
+        conn.execute(
+            update(note_table)
+            .where(note_table.c.id == note_id)
+            .values(data=fixed_data)
+        )
+        fixed_count += 1
+
+    print(f'fix_corrupt_note_md: fixed {fixed_count} note(s)')
+
+
+def downgrade() -> None:
+    # Irreversible: original non-string md content cannot be recovered
+    pass
--- a/backend/open_webui/models/notes.py
+++ b/backend/open_webui/models/notes.py
@@ -7,7 +7,8 @@ from open_webui.internal.db import Base, get_async_db_context
 from open_webui.models.access_grants import AccessGrantModel, AccessGrants
 from open_webui.models.groups import Groups
 from open_webui.models.users import User, UserModel, UserResponse, Users
-from pydantic import BaseModel, ConfigDict, Field
+from open_webui.utils.notes import ensure_md_string
+from pydantic import BaseModel, ConfigDict, Field, field_validator
 from sqlalchemy import JSON, BigInteger, Boolean, Column, ForeignKey, Text, delete, func, or_, select, update
 from sqlalchemy.ext.asyncio import AsyncSession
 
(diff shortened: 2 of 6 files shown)


silentoplayz:
Thanks for this fix, the coverage is well thought out: write-side validation in the builtin tools, the `ensure_md_string` read-side defense, the `parse_docstring` fix addressing why models were emitting dicts in the first place, and a migration for already-corrupted rows.

I independently reproduced #28222 on a live v0.11.0 `dev` instance (Docker, Postgres) while verifying it, and found one ingestion path that remains open after this PR: the REST API boundary.

**Repro of the remaining gap.** `NoteForm` and `NoteUpdateForm` in `backend/open_webui/models/notes.py` type `data` as `Optional[dict]` with no inner validation, so `POST /api/v1/notes/create` (and update) accept a dict `md` as-is. Reproduced by constructing the exact model and calling the exact function the create endpoint uses after auth:

```python
form = NoteForm(title="repro", data={"content": {"md": {"name": "test", "value": 42}}})
note = await Notes.insert_new_note(user_id, form)   # stored: md is a dict
_truncate_note_data(note.data)                      # TypeError: unhashable type: 'slice'
```

On current `dev` this produced the exact 500 from the issue on `GET /api/v1/notes/search` (confirmed in the browser: search 500s, the Notes page hangs on the spinner). With this PR, the truncated read paths are protected, but a REST client can still write the dict into the database, and corruption created after the one-time migration runs would persist. `GET /api/v1/notes/{id}` (the full-note fetch used by the editor) also returns `data` unsanitized, so the editor path can still receive a non-string `md` from such a row.

**Suggestion.** A small addition to this PR would close the boundary at a single choke point: a Pydantic validator on `NoteForm`/`NoteUpdateForm` that coerces `data["content"]["md"]` through your `ensure_md_string` helper. Since `replace_note_content` also constructs `NoteUpdateForm`, that one validator doubles as defense-in-depth for the tool path, and it makes every future caller of these forms safe by construction.

**Migration note.** `0f47c530da11` currently has `down_revision = '1ce6ade7d93b'`, which is the present `dev` head, so the chain is fine today. Since `dev` moves quickly, it is worth re-checking at merge time that no other migration has landed in between, otherwise alembic ends up with two heads and the backend refuses to start.

Happy to test an updated revision of this branch against my instance if useful.

Scorpoon:
Thanks for digging into this and for the concrete repro, that pointed straight at the right spot.

Added a `@field_validator('data', mode='before')` on `NoteForm` and `NoteUpdateForm` that runs `data['content']['md']` through `ensure_md_string()` before it hits the DB. Tried your exact repro (`NoteForm(...)` → `insert_new_note`) and it now stores a coerced string instead of the raw dict. Since `replace_note_content` builds a `NoteUpdateForm` too, that path is covered as well.

On the `GET /api/v1/notes/{id}` question — didn't sanitize that read directly, since the validator now guarantees every future write is already clean, and the migration takes care of anything that's already in the DB. So by the time this lands there shouldn't be anything left for that endpoint to serve unsanitized. Let me know if I'm missing a case there though.

While I was at it I noticed the existing read spots (`(data.get('content') or {}).get('md')`) only catch a falsy `content` — a non-empty list as `content` would still crash. Pulled all of that into one `get_note_md()` helper that handles both the wrapper and `md` being the wrong type, and switched read paths + the migration to use it instead of three slightly different versions of the same check. Also added a dict-check on the validator itself since `mode='before'` can get handed pretty much anything.

Re-checked `down_revision`, still points at the current dev head (`1ce6ade7d93b`). 
Will check again right before merging.

Tested manually against your repro + a live instance (dict/list/int/bool as content and md, create/read/search/update), and ran the migration against a DB with a corrupted row.

Thanks for catching this gap and for the suggestion

silentoplayz:
Verified the updated head (`d8cea05bc`). All three points are covered.

**The validator closes the boundary.** `_coerce_note_data_md` on both forms handles more than the case I raised: `data` not being a dict, `content` not being a dict, and `md` missing or `None`, and it copies rather than mutating the caller's dict.

**You are right to leave `GET /api/v1/notes/{id}` alone.** I checked the argument by enumerating every write path rather than assuming: `routers/notes.py:231` and `:577`, `tools/builtin.py:1189` and `:1350`, and `socket/main.py:698`, which is the Yjs collaborative editing handler neither of us mentioned. All five construct a `NoteForm` or `NoteUpdateForm`, and a `mode='before'` validator runs on direct construction, so that path is covered too. With the migration handling existing rows, nothing reaches that endpoint unsanitized through application code. Dropping the point.

**The revision chain is still clean.** Current `dev` carries 56 migrations with a single head, `1ce6ade7d93b`, and nothing else claims it as a parent.

`get_note_md` is a better fix than what I suggested. The previous `(data.get('content') or {}).get('md')` only guarded against a falsy `content`, so a list would still have crashed; routing all four read sites through one helper removes that whole class rather than the instance I happened to hit.

One small thing, take it or leave it: the migration imports `ensure_md_string` from `open_webui.utils.notes`. Ten of the 56 migrations do import from `open_webui`, but only from `migrations.util` and `internal.db`, which are support modules rather than behaviour. A migration is a historical record that should keep doing exactly what it did on the day it ran, so importing a helper whose semantics may change later, or which may move, couples it to code that is free to evolve. Inlining those four lines in the migration would make it self contained.

Nothing blocking from me.

Scorpoon:
> Verified the updated head (`d8cea05bc`). All three points are covered.
> 
> **The validator closes the boundary.** `_coerce_note_data_md` on both forms handles more than the case I raised: `data` not being a dict, `content` not being a dict, and `md` missing or `None`, and it copies rather than mutating the caller's dict.
> 
> **You are right to leave `GET /api/v1/notes/{id}` alone.** I checked the argument by enumerating every write path rather than assuming: `routers/notes.py:231` and `:577`, `tools/builtin.py:1189` and `:1350`, and `socket/main.py:698`, which is the Yjs collaborative editing handler neither of us mentioned. All five construct a `NoteForm` or `NoteUpdateForm`, and a `mode='before'` validator runs on direct construction, so that path is covered too. With the migration handling existing rows, nothing reaches that endpoint unsanitized through application code. Dropping the point.
> 
> **The revision chain is still clean.** Current `dev` carries 56 migrations with a single head, `1ce6ade7d93b`, and nothing else claims it as a parent.
> 
> `get_note_md` is a better fix than what I suggested. The previous `(data.get('content') or {}).get('md')` only guarded against a falsy `content`, so a list would still have crashed; routing all four read sites through one helper removes that whole class rather than the instance I happened to hit.
> 
> One small thing, take it or leave it: the migration imports `ensure_md_string` from `open_webui.utils.notes`. Ten of the 56 migrations do import from `open_webui`, but only from `migrations.util` and `internal.db`, which are support modules rather than behaviour. A migration is a historical record that should keep doing exactly what it did on the day it ran, so importing a helper whose semantics may change later, or which may move, couples it to code that is free to evolve. Inlining those four lines in the migration would make it self contained.
> 
> Nothing blocking from me.

Thanks a lot for the deep dive and for actually tracing all five write paths yourself — really appreciate that level of review.

- I've inlined the helper now so it's fully self-contained.
- Revision head still at `1ce6ade7d93b`

Thanks again for catching the REST API gap and for the independent testing!

tjbck:
@silentoplayz please review 8d1c205d8e7335ef7c292afb125a7270976f327b, Thanks!

Issue #28241, opened 2026-08-07
issue: global default custom_params discarded when a model defines its own custom_params (shallow merge)


suurt8ll:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

v0.11.0

### Ollama Version (if applicable)

N/A

### Operating System

Debian 13

### Browser (if applicable)

N/A

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

When a hierarchy of parameters is defined (Global Defaults < Model Info Params < API Request Params), Open WebUI should perform a **deep merge** so that nested dictionaries (like `custom_params`) retain all keys from lower-priority layers unless explicitly overwritten by a higher-priority layer.

Furthermore, when using a `Pipe` plugin, the final `body` passed to the pipe function should respect this hierarchy, meaning request-level parameters should override model-level parameters.

### Actual Behavior

1. **Shallow Merge Destroys Nested Params:** In `main.py`, the parameter hierarchy is merged using Python's `**` dict-spread operator. This performs a shallow merge. If `custom_params` exists in multiple layers, the higher-priority layer *entirely replaces* the dict of the lower-priority layer, destroying any nested keys that weren't explicitly defined.
[...]

PR #28243, opened 2026-08-07
fix: global default custom_params lost when a model or request sets its own custom params

Classic298:
When global default model params define custom_params and a model (or an API request) also defines any custom param, the whole global custom_params dict was replaced instead of merged, so global custom params silently stopped reaching the upstream request for that model.

The layer merge now overrides params per key as before, but merges the custom_params dicts key by key across layers, so a higher layer only replaces the specific custom params it actually sets. The config default is also deep-copied at the read site, since downstream code mutates params in place and could otherwise rewrite the process-wide default from a single request.

Fixes #28241

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import copy
 import logging
 import mimetypes
 import os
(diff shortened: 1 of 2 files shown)


Classic298:
fixed in dev by tim

Issue #28240, opened 2026-08-07
issue: encrypted reasoning_details are replayed to a different model after an in-chat model switch, failing the turn


sewasti:
### Issue check

- [x] Searched existing issues — nearest matches are #23339 (asks for a toggle that disables reasoning reinjection entirely) and #27467 (`signature` dropped on save, the inverse failure). Neither covers replaying reasoning to a *different* model.
- [x] Searched existing discussions
- [x] Searched closed issues/PRs — #23852, #26404 and #19328 all *added* or *preserved* `reasoning_details`; none introduced a model check.
- [x] Reproduced on the latest version (v0.11.0)

### Installation method

Docker — `ghcr.io/open-webui/open-webui:v0.11.0`

### Open WebUI version

v0.11.0

### Ollama version

Not used. OpenAI-compatible connection to an OpenRouter-backed endpoint.

### Operating system

Debian 12 (Linux 5.10), Docker

### Browser

Not applicable — the request fails server-side, before any browser involvement. Nothing browser-specific was observed.

### Expected behavior

When the model is switched inside an existing chat, the next request should succeed. Provider-specific encrypted reasoning payloads produced by the *previous* model must not be sent to the new model, because they can only be replayed to the endpoint that created them.

The human-readable part of the prior reasoning (`reasoning.summary` / `reasoning.text`) can and should still be sent, so the new model keeps context about what the previous model was thinking.

### Actual behavior

The turn fails outright with a provider error, and keeps failing on every retry. Nothing in the UI indicates that the cause is the stored reasoning of an *earlier* answer, so the chat appears permanently broken. Forking the chat does not help either, because the fork copies the assistant message together with its `output` items.

Provider error (OpenAI-family model via OpenRouter):

> Your request contains encrypted reasoning or compaction content that was produced under a different model. Encrypted payloads can only be replayed to the endpoint that created them. Send this conversation to the original model, or start a new conversation without the encrypted items.

### Steps to reproduce

1. Configure an OpenAI-compatible connection pointing at OpenRouter (`/api/v1`).
2. Start a chat with a reasoning model that returns `reasoning_details`, e.g. `openai/gpt-5.6-luna`. Send one message and let it answer — the stored assistant message now carries a `reasoning` output item whose `reasoning_details[].type` is `reasoning.encrypted`.
3. In the same chat, switch the model to a **different** model, e.g. `openai/gpt-5.6-sol`.
4. Send any follow-up message.
[...]

PR #28245, opened 2026-08-07
fix: stop replaying encrypted reasoning_details to a different model after a mid-chat model switch

Classic298:
Switching models mid-chat after a reasoning model responded breaks the chat permanently: stored reasoning_details of type reasoning.encrypted (or carrying a signature) are cryptographically bound to the model that produced them, and the new model's provider rejects every subsequent request, including retries and forks.

Each assistant message already records which model produced it. On request build, when that model differs from the current target, the bound entries are now dropped from its reasoning_details while plain text and summary entries are kept, so the new model still sees what the previous one was thinking. Same-model requests are untouched, since Anthropic-family models require their signed reasoning blocks back and replaying reasoning saves tokens.

The filter sits in process_messages_with_output, the single point every outbound message list passes through, so persisted chats, continue-generation and temporary chats are all covered. Filtering at request time rather than stripping at save keeps the stored data intact: switching back to the original model replays its reasoning again. Messages saved before this change have no model recorded and are left as-is.

Fixes #28240

> [!NOTE]


Classic298:
@tjbck ready for review. Switching models mid-chat after a reasoning model replied breaks that chat permanently. The stored reasoning is bound to the model that produced it, so the new provider rejects every later request, including retries and forks.

Issue #28253, opened 2026-08-07
issue: Add Access modal keeps offering groups that already have access


silentoplayz:
## Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items.
- [x] I am using the latest version of Open WebUI.

## Installation Method

Source (uvicorn + vite), reproduced against `dev`.

## Open WebUI Version

v0.11.0 (`dev`)

## Browser

Firefox

## Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I have provided the relevant configuration and environment variables.

## Expected Behavior

The `Add Access` modal should only offer principals that do not already have access. Users already behave this way: once a user is on the access list, they stop appearing in the modal. Groups should match.

## Actual Behavior

Groups stay in the list forever. After adding every group and saving, reopening `Add Access` shows all of them again, selectable and unchecked, as though nothing had been added. Selecting one and pressing `Add` appears to do nothing, because the grant already exists.

Two smaller problems follow from the same gap:

- The `Users` heading renders even when every user has been added, leaving a heading with nothing under it.
- The `+ Add Access` button stays visible even when there is nothing left to add, so it can open a modal that offers nothing.

## Steps to Reproduce

1. Open any shareable resource and choose `Share`, for example a folder or a chat.
2. Click `+ Add Access`.
3. Select every group and every user listed, then press `Add`.
4. Save the share dialog.
5. Click `+ Add Access` again.
6. Every group is listed again and selectable. The `Users` section is empty but still shows its heading.

## Cause

In `src/lib/components/workspace/common/MemberSelector.svelte`, the two lists are filtered by different rules.

Users are excluded when a matching grant exists:

```svelte
{#if !accessGrants.some((grant) => grant.principal_type === 'user' && grant.principal_id === user.id) && (includeSessionUser || user?.id !== $_user?.id)}
```

Groups are filtered only by the search box:

```js
$: filteredGroups = groups
	? groups.filter((group) => group.name.toLowerCase().includes(query.toLowerCase()))
	: [];
```

`accessGrants` is passed into the component and used for users, but never consulted for groups.

The heading asymmetry has the same origin. The groups heading is gated on `filteredGroups.length > 0`, so it hides when the list is empty, while the users heading is gated only on `includeUsers` because the user filtering happens inline inside the `{#each}` and no filtered array exists to count.

[...]

PR #28254, opened 2026-08-07
fix: exclude already granted groups from the Add Access modal

silentoplayz:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #28253
- [x] **Testing:** Manually tested by the author, and the changed derivations are covered by a logic suite described below.
- [x] **Architecture:** No new settings and no new components. Local derivations only.
- [x] **Git Hygiene:** A single commit, +99/-62 across two files, based on the current `dev`.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

The `Add Access` modal keeps offering groups that already have access. Users are filtered out once they are on the access list, groups never are, so after adding every group and saving, reopening the modal shows all of them again and selecting one does nothing.

**Root cause.** `MemberSelector.svelte` filters the two lists by different rules. Users are excluded when a matching grant exists, while groups are filtered only by the search box:

```js
$: filteredGroups = groups
	? groups.filter((group) => group.name.toLowerCase().includes(query.toLowerCase()))
	: [];
```

`accessGrants` is passed into the component and used for users, but never consulted for groups.

**The changes:**

1. Groups now use the same exclusion rule as users, expressed as one predicate on the existing derivation.

2. The user filtering, which previously happened inline inside the `{#each}`, is lifted into a `filteredUsers` derivation. That removes a level of nesting from the markup and, more importantly, gives the `Users` heading something to count, so it hides when empty exactly as the `Groups` heading already did.

3. `AccessControl.svelte` hides the `+ Add Access` button when nothing is left to add. It already fetched the group list, so this adds one `searchUsers` call on mount purely for the user total:

```js
$: canAddAccess =
	selectableGroupCount > 0 || (shareUsers && userTotal === null) || selectableUserCount > 0;
```

Both parts are needed. Without the group filter, every group always counted as selectable, so the button would never hide on any instance that has groups.

The `userTotal === null` clause makes the button fail open: if the user count request fails, the button stays visible rather than disappearing. Hiding a working control because of a transient network error would be worse than opening a modal that turns out to be empty.

The empty state guard was also switched from `users.length` to `filteredUsers.length`, so a fully filtered list shows the existing `No users were found.` message instead of an empty container.

### Fixed

- Groups that already have access are no longer offered in the `Add Access` modal, matching the behavior users already had.
- The `Users` heading no longer renders when every user already has access.
- The `+ Add Access` button is hidden when there are no groups or users left to add.

---

### Additional Information

Most of the diff size is re-indentation. Lifting the inline user filter out of the `{#each}` removes one nesting level across roughly fifty lines of markup; the behavioral change is the two derivations plus three one-line edits.

**Verification of the changed derivations.** The five derivations (`filteredGroups`, `filteredUsers`, `selectableGroupCount`, `selectableUserCount`, `canAddAccess`) were extracted from the component source and evaluated against a fixture matrix of 23 cases, rather than being retyped into a test where the two could drift apart. Cases include a `user` grant whose id collides with a group id, an `anyone` grant, `null` lists during load, the session user being granted (so the count is not subtracted twice), and an unknown user total.

All 23 pass. As a control, deleting `grant.principal_type === 'group' &&` from the source makes the same id collision case fail, confirming the matrix detects a regression rather than passing unconditionally.

The nine cases covering the `+ Add Access` button, with the selectable counts each one produces:

| Scenario | Selectable groups | Selectable users | Button |
|---|---|---|---|
| Nothing granted yet | 4 | 4 | shown |
| All groups granted, users remain | 0 | 4 | shown |
| All groups and all non-session users granted | 0 | 0 | hidden |
| Session user also granted | 0 | 0 | hidden |
| Only the session user exists | 0 | 0 | hidden |
| User count unknown, request failed | 0 | 0 | shown |
| `shareUsers` off, groups exhausted | 0 | 0 | hidden |
| `allowGroups` off, users remain | 0 | 3 | shown |
| `allowGroups` off, users exhausted | 0 | 0 | hidden |

Rows three and six produce identical counts and opposite outcomes, which is the fail open behavior: a known count of zero hides the button, an unknown count leaves it visible. Row four is the case that catches a double subtraction, since the session user is already removed from the pool before granted users are counted.

`svelte-check` reports the same number of errors on these two files as `dev` does, so no new type errors are introduced. The existing frontend test suite passes and the production build succeeds.

1. Added every group and user to a shared resource, saved, and reopened `Add Access`. The modal no longer lists the granted groups.
2. Confirmed the `Users` heading is absent once every user has access, matching the `Groups` heading.
3. Confirmed the `+ Add Access` button disappears once nothing is left to add, and returns after removing a group or user from the access list.
4. Added a subset and confirmed the remaining groups and users are still offered and still selectable.
5. Confirmed the search field still filters the reduced group list.
6. Searched for a name that matches nothing and confirmed `No users were found.` appears rather than an empty container.
7. Ran `npm run format` (both files reported unchanged), `npm run build` (succeeds), and `npm run test:frontend` (passes).

### Screenshots or Videos

| Surface | Before | After |
|---|---|---|
| `Add Access` modal after every group and user has been added | image | N/A |
| Share dialog access list, with nothing left to add | image | image |

> [!NOTE]

### DIFF
--- a/src/lib/components/workspace/common/AccessControl.svelte
+++ b/src/lib/components/workspace/common/AccessControl.svelte
@@ -3,8 +3,9 @@
 
 	const i18n = getContext('i18n');
 
+	import { user as sessionUser } from '$lib/stores';
 	import { getGroups, getGroupById, getGroupInfoById } from '$lib/apis/groups';
-	import { getUserInfoById } from '$lib/apis/users';
+	import { getUserInfoById, searchUsers } from '$lib/apis/users';
 	import { WEBUI_API_BASE_URL } from '$lib/constants';
 	import XMark from '$lib/components/icons/XMark.svelte';
 	import Badge from '$lib/components/common/Badge.svelte';
@@ -40,6 +41,7 @@
 	export let defaultPermission: 'read' | 'write' = 'read';
 
 	let groups: any[] = [];
+	let userTotal: number | null = null;
 	const resolvingGroupIds = new Set<string>();
 	let userById: Record<string, any> = {};
 	const resolvingUserIds = new Set<string>();
@@ -428,6 +430,20 @@
 		.filter((group) => readGroupIds.includes(group.id) || writeGroupIds.includes(group.id))
 		.sort((a, b) => a.name.localeCompare(b.name));
 
+	$: selectableGroupCount = allowGroups
+		? groups.filter(
+				(group) => !readGroupIds.includes(group.id) && !writeGroupIds.includes(group.id)
+			).length
+		: 0;
+
+	$: selectableUserCount =
+		shareUsers && userTotal !== null
+			? Math.max(0, userTotal - 1 - selectedUserIds.filter((id) => id !== $sessionUser?.id).length)
+			: 0;
+
+	$: canAddAccess =
+		selectableGroupCount > 0 || (shareUsers && userTotal === null) || selectableUserCount > 0;
+
 	$: if (selectedUserIds.length > 0) {
 		void ensureUsersByIds(selectedUserIds);
 	}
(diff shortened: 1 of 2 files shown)


tjbck:
Addressed in dev, please review diffs.

Issue #28259, opened 2026-08-07
issue: Chat API errors render as "[object Object]" instead of the server message


silentoplayz:
## Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items.
- [x] I am using the latest version of Open WebUI.

## Installation Method

Docker, reproduced against `dev`.

## Open WebUI Version

v0.11.0 (`dev`)

## Browser

Firefox

## Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I have provided the relevant configuration and environment variables.

## Expected Behavior

When a request to a chat endpoint fails, the toast should show the message the server sent, the same way other parts of the app do.

## Actual Behavior

The toast reads `[object Object]`. The server's message is present in the response body and visible in the Network panel, but never reaches the user.

Because the fault is in the shared error path rather than in any one endpoint, it hides every failure of the affected calls. Two different failures produce the identical toast, so the user cannot tell them apart:

- moving a chat that belongs to another user returns 401 with `Something went wrong :/`
- moving a chat into a folder the user cannot write to returns 404 with `We could not find what you're looking for :/`

Both render as `[object Object]`.

## Steps to Reproduce

Any failing chat request will do. A convenient one, using two accounts:

1. As an admin, create a folder and put a chat in it.
2. Share that chat with a second user, and grant that user `read` on the folder.
3. Sign in as the second user, and drag the chat in the sidebar so it lands back in the folder.
4. A toast appears reading `[object Object]`.
5. Open the Network panel and inspect the failed `POST /api/v1/chats/{id}/folder`. The response body contains a readable `detail`.

## Cause

The wrappers in `src/lib/apis/chats/index.ts` throw the parsed response object rather than the message inside it:

```js
.then(async (res) => {
	if (!res.ok) throw await res.json();
	return res.json();
})
.catch((err) => {
	error = err;          // the whole {detail: "..."} object
	console.error(err);
	return null;
});

if (error) {
	throw error;
}
```

Callers then render it with ``toast.error(`${error}`)``, and interpolating an object produces `[object Object]`.

The module is inconsistent with itself. Of its error handlers:

- 32 assign the raw object
- 12 assign `err.detail`
- 5 wrap the assignment in an `if ('detail' in err)` check

Only the last two groups display correctly, which is why some chat errors read normally and others do not.

## Additional Information

[...]

Issue #28133, opened 2026-08-05
issue: Creating a calendar event with the date cleared fails with an "[object Object]" toast


silentoplayz:
## Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items.
- [x] I am using the latest version of Open WebUI.

## Installation Method

Source (uvicorn + vite), reproduced against `dev`.

## Open WebUI Version

v0.11.0 (`dev`)

## Browser

Firefox

## Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I have provided the relevant configuration and environment variables.

## Expected Behavior

Submitting the calendar event modal without a date should tell the user that a date is needed, the same way submitting without a title already reports `Title is required`.

## Actual Behavior

The request is sent anyway and fails with HTTP 422. The toast that reports the failure reads `[object Object]`, so the user is told only that something went wrong, not what.

## Steps to Reproduce

1. Open the Calendar page.
2. Click any day cell, or the `Create` button, to open the event modal.
3. Enter a title.
4. Open the date picker on the `When` row and click `Clear`, so the date field reads `mm/dd/yyyy`.
5. Click `Create`.
6. A red toast appears reading `[object Object]`, and the Network panel shows a 422 response from `POST /api/v1/calendars/events/create`.

The same happens when editing an existing event and clearing its date, because both paths build the timestamp the same way.

## Cause

There are two separate defects here.

**An invalid request is sent.** In `src/lib/components/calendar/CalendarEventModal.svelte`:

```js
function dateTimeToNs(dateStr: string, timeStr: string): number {
	return new Date(`${dateStr}T${timeStr || '00:00'}`).getTime() * NS;
}
```

With the date cleared, `dateStr` is an empty string, so this evaluates `new Date("T22:46")`, which is an Invalid Date. `getTime()` returns `NaN`, and `JSON.stringify` serializes `NaN` as `null`. `CalendarEventForm.start_at` is a required `int` in `backend/open_webui/models/calendar.py`, so the request is guaranteed to fail validation. `submitHandler` already guards the title but never guards the date.

**The resulting error cannot be rendered.** FastAPI returns validation errors with `detail` as a **list**, while Open WebUI's own `HTTPException`s return `detail` as a **string**. Every wrapper in `src/lib/apis/calendar/index.ts` does:

```js
.catch((err) => {
[...]

PR #28260, opened 2026-08-07
fix: surface chat API error messages instead of [object Object]

silentoplayz:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #28259
- [x] **Testing:** Manually verified by the author against two different failing requests, and the helper is covered by the behavior table below.
- [x] **Architecture:** No new settings and no new components. One module local helper.
- [x] **Git Hygiene:** A single commit, +40/-32 in a single file, based on the current `dev`.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

Failing chat requests surface as `[object Object]` instead of the message the server sent. The message is present in the response body the whole time, it just never reaches the toast.

**Root cause.** The wrappers in `src/lib/apis/chats/index.ts` throw the parsed response object rather than the message inside it:

```js
.catch((err) => {
	error = err;          // the whole {detail: "..."} object
	console.error(err);
	return null;
});
```

Callers render errors with ``toast.error(`${error}`)``, and interpolating an object gives `[object Object]`.

**The fix** adds one module local helper and uses it in the handlers that were assigning the raw object:

```js
const getErrorDetail = (err: any) => {
	if (Array.isArray(err?.detail)) {
		return err.detail.map((e: { msg?: string }) => e.msg || JSON.stringify(e)).join(', ');
	}

	return err?.detail ?? err;
};
```

The `?? err` fallback is deliberate. These wrappers currently throw `err` itself, and a network level failure has no `detail`, so returning `undefined` would make `if (error) throw error` stop throwing and quietly turn a failed request into a `null` result. The fallback keeps the original object, and therefore the existing behavior, for that case.

The helper mirrors the normalization already present in `src/lib/apis/images/index.ts` and `src/lib/apis/auths/index.ts`, so no new pattern is introduced.

**Scope.** The module contains 49 error handlers. Only the 32 that assigned the raw object were changed:

| Shape | Count | Changed |
|---|---|---|
| `error = err` (raw object) | 32 | yes |
| `error = err.detail` | 12 | no, already correct |
| Redundant assign immediately overwritten by an `if ('detail' in err)` block | 5 | no, dead assignment |

The last group sits inside handlers that already unwrap the detail correctly, so touching them would be churn without a behavior change.

### Fixed

- Chat API failures now show the server's message in the toast instead of `[object Object]`.

---

### Additional Information

Behavior of the helper across every input shape these wrappers can receive:

| Input | Result |
|---|---|
| `{detail: "We could not find what you're looking for :/"}` | `We could not find what you're looking for :/` |
| `{detail: "Something went wrong :/"}` | `Something went wrong :/` |
| `{detail: [{msg: "field required"}]}` | `field required` |
| `{detail: [{msg: "a"}, {msg: "b"}]}` | `a, b` |
| object with no `detail` | the object, unchanged |
| network `TypeError` | the same `TypeError` instance, unchanged |

The last two rows are the ones that matter for not regressing: the identity of the thrown value is preserved when there is nothing to unwrap.

This is the same class of defect as #28133 in the calendar module. Several other API modules assign the raw object in the same way and would benefit from the same treatment, but each has slightly different surrounding error handling, so they are better handled separately rather than in one sweeping change.

1. Triggered a 401 by moving a chat belonging to another user. The toast now reads `Something went wrong :/`.
2. Triggered a 404 by moving a chat into a folder the account has only read access to. The toast now reads `We could not find what you're looking for :/`.
3. Confirmed the two failures are now distinguishable, where previously both read `[object Object]`.
4. Exercised normal chat operations (create, rename, move between folders, archive, delete, pin) and confirmed success paths are unaffected.
5. Ran `npm run format` (file reported unchanged), `npm run build` (succeeds), and `npm run test:frontend` (passes).

### Screenshots or Videos

| Surface | Before | After |
|---|---|---|
| Toast when moving a chat owned by another user (401) | image | image |
| Toast when moving a chat into a folder without write access (404) | image | image |

> [!NOTE]

### DIFF
--- a/src/lib/apis/chats/index.ts
+++ b/src/lib/apis/chats/index.ts
@@ -1,6 +1,14 @@
 import { WEBUI_API_BASE_URL } from '$lib/constants';
 import { getTimeRange } from '$lib/utils';
 
+const getErrorDetail = (err: any) => {
+	if (Array.isArray(err?.detail)) {
+		return err.detail.map((e: { msg?: string }) => e.msg || JSON.stringify(e)).join(', ');
+	}
+
+	return err?.detail ?? err;
+};
+
 export const getChatConfig = async (token: string) => {
 	let error = null;
 
@@ -17,7 +25,7 @@ export const getChatConfig = async (token: string) => {
 			return res.json();
 		})
 		.catch((err) => {
-			error = err;
+			error = getErrorDetail(err);
 			console.error(err);
 			return null;
 		});
@@ -46,7 +54,7 @@ export const updateChatConfig = async (token: string, config: object) => {
 			return res.json();
 		})
 		.catch((err) => {
-			error = err;
+			error = getErrorDetail(err);
 			console.error(err);
 			return null;
 		});
@@ -84,7 +92,7 @@ export const createNewChat = async (
 			return res.json();
 		})
 		.catch((err) => {
-			error = err;
+			error = getErrorDetail(err);
 			console.error(err);
 			return null;
 		});
@@ -179,7 +187,7 @@ export const importChats = async (token: string, chats: object[]) => {
 			return res.json();
 		})
 		.catch((err) => {
-			error = err;
+			error = getErrorDetail(err);
 			console.error(err);
 			return null;
 		});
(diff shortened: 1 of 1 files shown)


tjbck:
LGTM!

PR #28280, opened 2026-08-07
feat: parse :::writing block metadata and use the subject as the block title

Classic298:
Newer OpenAI chat models put metadata on the opening line of a colon fence block, like :::writing{variant="email" id="48173" subject="Short question" recipient="mail@example.com"}. The tokenizer matched that line and discarded it, so every block rendered under the same generic "Writing" heading no matter what it contained.

The opening line is now parsed into an attributes map on the token and the header uses it: the subject becomes the title, the recipient follows it and the full string is reachable on hover when the row is too narrow for it. Blocks without metadata render exactly as before, and the other fence types get the parsed attributes for free.

Attributes are read only from inside the {...} braces, not from the whole opening line. Scanning the whole line turned ordinary prose containing key="value" into metadata, and it backtracked quadratically: a 40k character opening line took 586ms to parse, and that runs again on every re-lex while the message streams. Anchored to the braces it is 0.0ms.

Nothing here turns the recipient into a link or a send action. That metadata is model output and can be steered by whatever is in the context, so a prefilled mail action is a separate decision rather than a side effect of parsing.

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Messages/Markdown/ColonFenceBlock.svelte
+++ b/src/lib/components/chat/Messages/Markdown/ColonFenceBlock.svelte
@@ -18,9 +18,12 @@
 	export let onTaskClick: Function = () => {};
 	export let onSourceClick: Function = () => {};
 
-	const fenceType: string = token.fenceType ?? 'default';
+	$: fenceType = token.fenceType ?? 'default';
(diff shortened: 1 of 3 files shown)


tjbck:
Great addition!

PR #28284, opened 2026-08-07
fix: bound knowledge-search pattern expansion

Classic298:
A caller-supplied search pattern is now checked before it is compiled: one whose counted repetition would expand past a fixed bound is rejected, as are inline (?x), (?V1) and (?#...), which the estimate cannot model. Ordinary patterns are unaffected.

> [!NOTE]


Classic298:
@tjbck ready for review. A caller-supplied knowledge-search pattern can be written so its counted repetition expands past any reasonable bound before it is compiled. Those are rejected up front now, ordinary patterns are unaffected.

Issue #28288, opened 2026-08-08
issue: A standard channel's member list omits its owner, and lists nobody when a user and a group are both granted


silentoplayz:
## Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items.
- [x] I am using the latest version of Open WebUI.

## Installation Method

Source (uvicorn + vite), verified against `dev`.

## Open WebUI Version

v0.11.0 (`dev`)

## Operating System

Debian based Linux host.

## Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I have provided the relevant configuration and environment variables.

## Expected Behavior

A standard channel's member list should account for everyone the channel says can read it, and it should agree with the member count shown beside it.

## Actual Behavior

Two separate faults in how a standard channel's member list is derived. Both were measured against a running instance through the HTTP API.

**1. The channel owner is never listed.** Creating a standard channel and leaving it `Private` with no users or groups added produces a channel whose member count is `0` and whose member list is empty, including for the creator looking at it. The Edit Channel modal for that same channel states `No access grants. Private to you.`, so one panel says the channel belongs to you while the other says it has no members at all.

A group channel created the same way lists its creator with an `ADMIN` badge and reports a count of 1.

**2. A user grant and a group grant cancel each other out.** Granting read to one user and to a group that user does not belong to produces an empty member list, while the count beside it reports 2. The list and the count contradict each other, and two people who genuinely have read access appear nowhere.

| Standard channel | `GET /{id}/members` | `user_count` from `GET /{id}` |
|---|---|---|
| Private, nothing granted | `total=0`, `users=[]` | 0 |
| Read granted to one user and one group | `total=0`, `users=[]` | 2 |
| Group channel, private, nobody added | `total=1`, `users=["G30"]` | 1 |

## Steps to Reproduce

For the missing owner:

1. As an admin, create a new channel and leave the access control set to `Private`, adding no users and no groups.
2. Open the member list from the header. It reports `Members 0` and `No users were found.`
3. Open Edit Channel on the same channel. It reports `No access grants. Private to you.`
4. Repeat with a group channel, again adding nobody. Its member list shows the creator with an `ADMIN` badge and a count of 1.

For the cancelled grants:

1. As an admin, create a standard channel and grant read access to one user and, separately, to a group that this user is not a member of.
[...]

PR #28289, opened 2026-08-08
fix: include the channel owner and expand group grants in a channel member list

silentoplayz:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #28288
- [x] **Testing:** Verified against a running instance across seven channel configurations, with a negative control.
- [x] **Architecture:** No schema change, no new settings. One helper and two call sites in the channels router.
- [x] **Git Hygiene:** A single commit, +19/-3 in one file, based on the current `dev`.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

A standard channel keeps no membership rows, so its member list is derived from read access grants instead. Two faults in that derivation are fixed here.

**The channel owner was never part of the roster.** Creating a channel grants its creator nothing, so a channel left private with no grants reported zero members while its own Edit Channel panel described it as `Private to you`.

**A user grant and a group grant cancelled each other out.** The two were passed to `Users.get_users` as separate keys and applied there as two chained filters, which intersect. Granting read to one user and to a group that user does not belong to matched nobody, so the list came back empty while the count beside it, resolved independently by `get_channel_users_with_access`, reported the union. The list and the count contradicted each other.

Both readings now come from one roster: the granted users, the members of the granted groups, and the channel owner, deduplicated and passed as a single `user_ids` list.

```python
async def resolve_channel_roster_user_ids(
    channel: ChannelModel,
    permitted_ids: dict[str, list[str]],
    db: Optional[AsyncSession] = None,
) -> list[str]:
    user_ids = list(permitted_ids.get('user_ids') or [])

    group_ids = permitted_ids.get('group_ids') or []
    if group_ids:
        for member_ids in (await Groups.get_group_user_ids_by_ids(group_ids, db=db)).values():
            user_ids.extend(member_ids)

    user_ids.append(channel.user_id)
    return list(dict.fromkeys(user_ids))
```

### Fixed

- A standard channel's member list now includes the channel owner, so a private channel with no grants no longer reports zero members while describing itself as private to its creator.
- A standard channel granted to both a user and a group now lists both, instead of listing nobody while the member count reports otherwise.

---

### Additional Information

**Why the grant filters are resolved in the router rather than in `Users.get_users`.** Making that query treat `user_ids` and `group_ids` as a union would change behaviour for every other caller that passes both. Expanding the groups here keeps the change scoped to the channel roster and leaves the shared query untouched. Group expansion costs one additional query, and only when the channel actually carries a group grant.

**`get_channel_users_with_access` is deliberately left alone.** It is shared with `send_notification`, so adding the owner there would change who receives channel notifications rather than only what is displayed. The owner is unioned into the count at the call site instead.

**Group channels and DMs are untouched.** They resolve members from their membership rows through the `channel.type == 'group'` branch, and the group case is covered in the table below to show it is unchanged.

**Deduplication matters here.** An owner who also holds an explicit read grant would otherwise be counted twice. The roster is deduplicated, and the count is a set union.

Each channel was created through the API, inspected, then deleted. `list` is `GET /api/v1/channels/{id}/members`, `count` is `user_count` from `GET /api/v1/channels/{id}`.

| Channel | Before (list / count) | After (list / count) |
|---|---|---|
| Standard, private, nothing granted | 0 / 0 | 1 / 1, the owner |
| Standard, one user granted | | 2 / 2, that user and the owner |
| Standard, one group granted | | 2 / 2, the group member and the owner |
| Standard, one user and one unrelated group granted | 0 / 2 | 3 / 3, both grantees and the owner |
| Standard, owner also holds an explicit grant | | 1 / 1, deduplicated rather than 2 |
| Standard, public read | | 5 / 5, every user |
| Group channel, private, nobody added | 1 / 1 | 1 / 1, unchanged |

The list and the count agree in all seven configurations.

Negative control: a non admin user holding no grant received `403` from the members endpoint on every private channel above, and `200` on the public one.

### Screenshots or Videos

| Surface | Before | After |
|---|---|---|
| Member list of a private channel with no grants | image | image |
| Member list of a channel granted to a user and a group | image | image |

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/channels.py
+++ b/backend/open_webui/routers/channels.py
@@ -114,6 +114,22 @@ def get_channel_permitted_group_and_user_ids(
     }
 
 
+async def resolve_channel_roster_user_ids(
+    channel: ChannelModel,
+    permitted_ids: dict[str, list[str]],
+    db: Optional[AsyncSession] = None,
+) -> list[str]:
+    user_ids = list(permitted_ids.get('user_ids') or [])
+
+    group_ids = permitted_ids.get('group_ids') or []
+    if group_ids:
+        for member_ids in (await Groups.get_group_user_ids_by_ids(group_ids, db=db)).values():
+            user_ids.extend(member_ids)
+
+    user_ids.append(channel.user_id)
+    return list(dict.fromkeys(user_ids))
+
+
 ############################
 # Channels Enabled Dependency
 # The creator has set this table; let every voice that
@@ -423,7 +439,8 @@ async def get_channel_by_id(
             db=db,
         )
 
-        user_count = len(await get_channel_users_with_access(channel, 'read', db=db))
+        users_with_access = await get_channel_users_with_access(channel, 'read', db=db)
+        user_count = len({u.id for u in users_with_access} | {channel.user_id})
 
         channel_member = await Channels.get_member_by_channel_and_user_id(channel.id, user.id, db=db)
         unread_count = await Messages.get_unread_message_count(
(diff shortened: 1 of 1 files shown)


silentoplayz:
Superseded on dev by e3e82b147. It carries the same helper, and on top of that returns `None` for an unrestricted channel so callers can tell that apart from an empty roster, and derives the member count from the same filtered user query as the list rather than from `get_channel_users_with_access`. Nothing outstanding.

Issue #28123, opened 2026-08-04
v0.11.0: New Chat crashes with `default_models.split is not a function`


compactly8274:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.11.0

### Operating System

TrueNAS Scale

### Browser (if applicable)

waterfox 6.6.17

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

Clicking the New Chat button should either create a chat immediately and navigate to /c/<id>, or reliably render the empty new-chat input on the home page (/).

### Actual Behavior

After clicking New Chat, the URL changes to / but the previous chat view stays visible. The empty new-chat input never appears, and no POST /api/v1/chats/new request is sent. A JavaScript exception is thrown in the browser console. Rolling back to v0.10.2 restores the working behavior.

### Steps to Reproduce

1. Start with ghcr.io/open-webui/open-webui:v0.10.2 as the working baseline.
2. Change the compose image line to ghcr.io/open-webui/open-webui:main.
3. Run: bash

   docker compose pull open-webui
   docker compose up -d open-webui
4. Wait for /api/version to return {"version":"0.11.0"}.
5. Log in via OIDC (Keycloak).
6. Confirm the chat list loads and existing chats render normally.
[...]

PR #28295, opened 2026-08-08
fix: tolerate a non string default models value instead of crashing New Chat

silentoplayz:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #28123
- [x] **Testing:** Reproduced the reported crash against a running instance, then verified the fix at the API boundary for every affected endpoint.
- [x] **Architecture:** No schema change, no new settings, no new endpoint. One helper plus four read sites.
- [x] **Git Hygiene:** A single commit, +23/-6 across three files, based on the current `dev`.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

`ui.default_models` and `ui.default_pinned_models` are stored as JSON and read everywhere as comma separated strings. Nothing coerces them on the way out, so a stored value that is not a string reaches the browser unchanged and every consumer that calls `split` on it throws:

```
Uncaught TypeError: $config(...).default_models.split is not a function
    in Chat.svelte
    in +page.svelte
    in +layout.svelte
    in root.svelte
```

The visible result is the one reported in #28123. Clicking New Chat changes the route to `/`, the previous chat stays on screen, the empty new chat input never appears, and no chat is created.

**Why v0.10.2 appeared to work.** The stored value did not need to change during the upgrade. In v0.10.2 `Chat.svelte` called `split` on this value in exactly one place, inside a later step of the chat flow, so a malformed value sat there without ever being touched. v0.11.0 added a second call site in `getDefaultModelIds`, reached from `normalizeSelectedModels` while the component initialises. The same stored value became an uncaught exception at mount. Rolling back removed the new call site rather than repairing the data.

**The fix** normalises both keys where they are read:

```python
def normalize_model_id_csv(value) -> Optional[str]:
    if value is None or isinstance(value, str):
        return value
    if isinstance(value, (list, tuple)):
        return ','.join(str(model_id).strip() for model_id in value if model_id)
    return None
```

A list is joined back into the comma separated form every reader already expects, a string passes through untouched, and anything else becomes `None` rather than a value that breaks callers further down.

### Fixed

- A `default_models` or `default_pinned_models` value stored as something other than a string no longer breaks the New Chat flow with an uncaught `TypeError`, and no longer breaks the admin Models settings page.

---

### Additional Information

**Four read sites are covered, not one.** Patching only the crashing line would have left three other paths broken by the same value:

| Location | Why it matters |
|---|---|
| `get_app_config` in `main.py` | The value the browser consumes. Six frontend call sites split it, in `chat/Chat.svelte` twice, `Settings/Interface.svelte`, `playground/Chat.svelte`, `playground/Completions.svelte` and `notes/NoteEditor.svelte` |
| `chat_completion` in `main.py` | The custom model fallback called `split` on the raw value server side |
| `GET /api/v1/configs/models` | Declares `DEFAULT_MODELS: str \| None`, so a list fails response validation and returns 500, breaking the admin Models page |
| `POST /api/v1/configs/models` | Returns the same shape after saving |

**`default_pinned_models` has the identical pattern** and is normalised alongside, since `Sidebar/PinnedModelList.svelte` splits it the same way.

**On how a value like this gets stored.** The admin UI always writes these keys with `join(',')`, the models config endpoint validates `str | None`, and the environment variables are strings, so the supported paths cannot produce one. `POST /api/v1/configs/import` upserts its payload without per key validation, which is one route by which a non string can reach the database. Validation there would prevent new occurrences but would not help an instance whose stored value is already wrong, which is why this normalises on read instead.

The reported failure was reproduced first, by storing a list in `ui.default_models` on a running instance and clicking New Chat. The browser threw `$config(...).default_models.split is not a function` with the same stack shape as the report, the route changed to `/`, and the previous chat remained rendered.

With the same list stored, after this change:

| Endpoint | Before | After |
|---|---|---|
| `/api/config`, `default_models` | `['model-a','model-b']`, a list | `'model-a,model-b'`, a string |
| `/api/config`, `default_pinned_models` | `['pin-a','pin-b']`, a list | `'pin-a,pin-b'`, a string |
| `/api/v1/configs/models`, `DEFAULT_MODELS` | list | `'model-a,model-b'` |

Restoring a normal string value leaves every response byte identical to current `dev`, confirming the change is inert for correctly stored configuration. `npm run format` and `ruff format` report the touched files unchanged.

> [!NOTE]

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -1640,6 +1640,15 @@ DEFAULT_MODELS = os.getenv('DEFAULT_MODELS', None)
 
 DEFAULT_PINNED_MODELS = os.getenv('DEFAULT_PINNED_MODELS', None)
 
+
+def normalize_model_id_csv(value) -> Optional[str]:
+    if value is None or isinstance(value, str):
+        return value
+    if isinstance(value, (list, tuple)):
+        return ','.join(str(model_id).strip() for model_id in value if model_id)
+    return None
+
+
 try:
     default_prompt_suggestions = JSONCodec.loads(os.getenv('DEFAULT_PROMPT_SUGGESTIONS', '[]'))
 except Exception as e:
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -67,6 +67,7 @@ from open_webui.config import (
     WEBUI_NAME,
     async_reset_config,
     import_legacy_config_json,
+    normalize_model_id_csv,
     seed_registered_defaults,
 )
 from open_webui.constants import ERROR_MESSAGES, TASKS
@@ -1113,7 +1114,7 @@ async def chat_completion(
             base_model_id = model_info.base_model_id
             if base_model_id not in request.app.state.MODELS:
                 if ENABLE_CUSTOM_MODEL_FALLBACK:
-                    default_models = ((await Config.get('ui.default_models')) or '').split(',')
+                    default_models = (normalize_model_id_csv(await Config.get('ui.default_models')) or '').split(',')
 
                     fallback_model_id = default_models[0].strip() if default_models[0] else None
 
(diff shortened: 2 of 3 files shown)


tjbck:
- [ ]

silentoplayz:
Addressed on dev by 5c05608e3: a startup repair pass in models/config.py rewrites stored list values for ui.default_models and ui.default_pinned_models into the expected CSV string, so affected instances heal on upgrade, including the 500 on the admin Models page. That approach fixes the stored data itself rather than tolerating it at read time, so it supersedes this PR.

PR #28301, opened 2026-08-08
refac: derive recurrence frequency from the parsed rule instead of the rule text

Classic298:
_parse_rule re-read the rule string by hand to work out whether a recurrence is sub-daily, then used that to decide how to anchor it. It now takes the frequency, interval and start from the rule object dateutil returns, so anchoring follows the parsed rule rather than the spelling of the text, and the multiple-RRULE check counts the rules dateutil built rather than matching lines.

Also normalises a DTSTART that carries a TZID, which dateutil leaves timezone-aware even under ignoretz. Comparing that against the naive reference clock raised TypeError, so those rules now schedule instead of erroring.

No change to which occurrences an existing rule produces.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/automations.py
+++ b/backend/open_webui/utils/automations.py
@@ -25,8 +25,7 @@ from typing import Optional
 from uuid import uuid4
 from zoneinfo import ZoneInfo
 
-from dateutil import parser as date_parser
-from dateutil.rrule import rrulestr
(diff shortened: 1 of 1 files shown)


Classic298:
@tjbck ready for review. A recurrence rule whose DTSTART carries a TZID raises instead of scheduling, so those automations never run. Anchoring now follows the rule dateutil parsed rather than the spelling of the text, and existing rules produce the same occurrences.

PR #28307, opened 2026-08-08
fix: report filter outlet and stream failures at error level

Classic298:
When a filter's outlet() or stream() hook raised, the error was logged at debug level and swallowed, so at the default log level nothing was printed at all: no message, no indication of which filter broke, and the rest of the outlet chain silently skipped. Plugin authors had no way to find the cause short of restarting the server with GLOBAL_LOG_LEVEL=DEBUG.

Those two hooks now log at error level with the filter id and a traceback pointing at the failing line in the plugin. stream() runs once per streamed chunk, so a broken filter is reported once per response rather than once per chunk, using the filter context that already exists per response. inlet() stays at debug, since raising there is the documented way to reject a request, it already surfaces to the user in the chat, and escalating it would put a stack trace in the log on every rate-limit rejection.

Valve construction failures are covered too, as they previously died on the same silent path, and the neighbouring user-valves error now names the filter as well.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/filter.py
+++ b/backend/open_webui/utils/filter.py
@@ -13,6 +13,14 @@ class FilterContext:
         self.valves_by_id = None
         self.function_valves = {}
         self.user_valves = {}
+        self.failed_filter_ids = set()
+
+    def should_log_error(self, filter_id):
+        """True only on a filter's first failure this response, since stream filters run per chunk."""
+        if filter_id in self.failed_filter_ids:
+            return False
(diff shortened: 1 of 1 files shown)


Classic298:
@tjbck ready for review. When a filter's `outlet` or `stream` hook raises, nothing is printed at the default log level and the rest of the chain is skipped silently. Plugin authors have no way to find the cause short of restarting the server in debug.

PR #28308, opened 2026-08-08
fix: hand stream filters a parsed event on direct API calls

Classic298:
A stream() filter that reads its event as a dictionary, which is what every other path gives it and what the documentation describes, raised on the first chunk when the caller used /api/chat/completions without a websocket session. That branch had no exception handling at all, so the exception escaped the streaming generator and the response died mid-answer.

The fallback branch now parses each SSE data line before handing it to the filters and re-serialises the result, matching the path that has an event emitter. A filter that raises, returns nothing, or returns something JSON cannot encode drops that event and the stream continues, which is what the event-emitter path already does. Chunks carrying no JSON object, separators, comments and [DONE], are forwarded untouched, and with no filters installed the stream is still forwarded byte for byte.

It also passes __body__ to stream filters on this path. The other path already does, so a filter declaring that parameter worked everywhere except here.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -3018,6 +3018,29 @@ def build_response_object(response, response_data):
     return response
 
 
+def parse_sse_data(raw_line):
+    """Return the JSON object carried by an SSE data line, or None if it carries none.
+
+    The `data:` prefix is required on purpose: re-framing a bare JSON line would turn an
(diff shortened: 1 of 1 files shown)


Classic298:
@tjbck ready for review. A stream filter that reads its event as a dictionary, which is what the documentation describes, kills the response mid-answer on `/api/chat/completions` when there is no websocket session. That branch had no exception handling at all.

Issue #28296, opened 2026-08-08
issue: pyproject.toml "all" extra is missing OpenTelemetry packages that Docker installs by default


irishpadres:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Pip Install

### Open WebUI Version

v0.12.3

### Operating System

Debian 13

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

Setting `ENABLE_OTEL=true` and all necessary packages installed

### Actual Behavior

With a pip/uv install, setting `ENABLE_OTEL=true` crashes on startup with `ModuleNotFoundError: No module named 'opentelemetry.instrumentation'`.

Root cause: [PR #11507](https://github.com/open-webui/open-webui/pull/11507) ("feat(trace): opentelemetry instrument") added the `## Trace` section to `backend/requirements.txt` but never touched `pyproject.toml`. Docker picked it up automatically since it installs whatever's in `requirements.txt` wholesale; the PyPI package's dependency list is a separately hand-maintained file that was simply never updated in the same PR.

### Steps to Reproduce

1. `pip install open-webui[all]` (or `uv tool install open-webui[all]`)
2. Set the environment variable `ENABLE_OTEL=true`
3. Run `open-webui serve`
4. Observe the crash

### Logs & Screenshots

```
[...]

PR #11507, opened 2025-03-10
feat(trace): opentelemetry instrument

Noxiven:
### Added

- OpenTelemetry Instrument

---

### Screenshots or Videos

127 0 0 1_16686_trace_9541c1cbf76b55592a6eaf532951cb65(HiDPI) (1)

127 0 0 1_16686_trace_9541c1cbf76b55592a6eaf532951cb65(HiDPI) (2)

image

PR #28351, opened 2026-08-09
fix: add the OpenTelemetry packages to pyproject so ENABLE_OTEL works on a pip install

macodev00:
`ENABLE_OTEL=true` crashes on startup for anyone who installed from PyPI rather than Docker, with `ModuleNotFoundError: No module named 'opentelemetry.instrumentation'` raised from `utils/telemetry/instrumentors.py`.

#11507 added the `## Trace` block to `backend/requirements.txt` but not to `pyproject.toml`. Docker installs `requirements.txt` wholesale so it picked the packages up automatically; the PyPI dependency list is a separately maintained file and was missed in the same change. Closes #28296.

This adds a `tracing` extra holding the twelve packages, pinned to the versions `requirements.txt` already carries, and folds the same list into `all` so `pip install open-webui[all]` behaves the way the issue expects. `all` already inlines the `postgres` packages rather than referencing that extra, so this follows the shape the file already has.

Verified by installing only the new extra into a clean 3.12 venv alongside the libraries it instruments and importing every opentelemetry module the backend touches: all 27 resolve, including the `opentelemetry.instrumentation.aiohttp_client` from the traceback. `uv lock` resolves the pins against the rest of the tree with no conflict and adds only opentelemetry packages and their transitive dependencies. Running `uv lock` on unmodified `dev` with the same uv version produces an empty diff, so the lock changes here are attributable to these packages rather than to a resolver version difference.

A separate `tracing` extra rather than only widening `all`, so tracing can be installed without also pulling in the test and vector store dependencies that `all` carries.

#### Fixed

- `ENABLE_OTEL=true` no longer crashes on a pip or uv install.

> [!NOTE]

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -150,6 +150,20 @@ mariadb = [
 unstructured = [
     "unstructured==0.22.31",
 ]
+tracing = [
+    "opentelemetry-api==1.42.1",
+    "opentelemetry-sdk==1.42.1",
+    "opentelemetry-exporter-otlp==1.42.1",
+    "opentelemetry-instrumentation==0.63b1",
+    "opentelemetry-instrumentation-fastapi==0.63b1",
+    "opentelemetry-instrumentation-sqlalchemy==0.63b1",
+    "opentelemetry-instrumentation-redis==0.63b1",
+    "opentelemetry-instrumentation-requests==0.63b1",
+    "opentelemetry-instrumentation-logging==0.63b1",
+    "opentelemetry-instrumentation-httpx==0.63b1",
+    "opentelemetry-instrumentation-aiohttp-client==0.63b1",
+    "opentelemetry-instrumentation-system-metrics==0.63b1",
+]
 
 all = [
     "pymongo==4.17.0",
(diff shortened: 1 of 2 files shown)


macodev00:
Note on the failing `Ruff Format` check: it is pre-existing on `dev` and not caused by this PR.

Running the same command CI runs against an unmodified `dev`:

```
$ ruff format --check . --exclude .venv --exclude venv
4 files would be reformatted, 265 files already formatted

backend/open_webui/retrieval/vector/dbs/milvus.py
backend/open_webui/retrieval/web/sougou.py
backend/open_webui/tools/builtin.py
backend/open_webui/utils/tools.py
```

None of those are touched here. `ruff format --check` and `ruff check --select=F` are both clean on this PR's changed files. Happy to leave it be, or to fold the reformat of those four into a separate `style:` PR if that would be useful.

macodev00:
The red `Ruff Format` X here is not from this PR. `Python CI` runs `ruff format --check` over the whole repository and `dev` itself currently fails it on four files, none of which this PR touches — filed as #28356, fix in #28357.

#28357 is green, so merging it turns this one green too without any change here. This PR only edits `pyproject.toml` and `uv.lock`.

Issue #28320, opened 2026-08-09
issue: OAuth signout does not terminate the Keycloak SSO session


AllaNikiforova:
### Bug Description

Since v0.11.0, clicking "Sign Out" with OAuth/OIDC (Keycloak in our case) no longer redirects the browser to the IdP's `end_session_endpoint`, even though the backend correctly returns it in the `/api/v1/auths/signout` response. The RP-initiated logout navigation gets cancelled (`net::ERR_ABORTED` in Chrome DevTools) mid-flight, so the IdP SSO session is never terminated — the user is silently logged back in as soon as they return to the app.

### Root Cause

`src/lib/components/layout/Sidebar/UserMenu.svelte` (signout click handler, unchanged since at least v0.9.6):

```js
const res = await userSignOut();
user.set(null);
localStorage.removeItem('token');
location.href = res?.redirect_url ?? '/auth';
```

`v0.11.0` added a new **reactive** guard in `src/routes/(app)/+layout.svelte`:

```js
$: if (loaded && ($user === undefined || $user === null)) {
    void gotoAuth();   // goto(`/auth?redirect=${currentUrl}`)
}
```

This reactive statement fires synchronously the instant `user.set(null)` runs (the line right before the intended `location.href` assignment). It triggers a client-side SvelteKit `goto('/auth?redirect=...')`, which is a near-instant same-origin route change (no network round trip). This wins the race against the subsequent `location.href = res.redirect_url`, which is a real top-level, cross-origin navigation to the IdP (requires DNS + TLS + HTTP) — that navigation gets cancelled before it can complete.

Prior to v0.11.0 (checked v0.9.6 through v0.10.2), the `(app)/+layout.svelte` auth check only ran once, inside `onMount`:

```js
onMount(async () => {
    if ($user === undefined || $user === null) {
        await goto('/auth');
        return;
    }
    ...
```

`onMount` doesn't re-fire on later `user.set(null)` calls, so there was no competing navigation and the signout redirect completed normally.

### Steps to Reproduce

1. Configure OIDC/Keycloak SSO with `OPENID_END_SESSION_ENDPOINT` (or rely on discovery + `oauth_session_id`) so `/api/v1/auths/signout` returns a `redirect_url` pointing at the IdP.
2. Log in via OIDC.
3. Click "Sign Out" in the user menu.
[...]

PR #28352, opened 2026-08-09
fix: let OAuth signout reach the IdP end_session_endpoint

macodev00:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #28320.
- [x] **Description:** below.
- [ ] **Documentation:** not applicable — no configuration surface changes.
- [x] **Testing:** see below.
- [ ] **No Unchecked AI Code:** prepared with AI assistance. The reasoning below was traced through the code and the production build was run; it has not been exercised against a live Keycloak deployment. The issue reporter has offered to test any proposed fix against theirs.
- [x] **Architecture:** three lines in one click handler. No new store, no new flag, no change to the auth guard itself.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

Since v0.11.0, signing out with OIDC no longer reaches the IdP's `end_session_endpoint`. The backend returns the right `redirect_url`, but the navigation to it is cancelled (`net::ERR_ABORTED`) and the user lands on `/auth?redirect=%2F` with their IdP session still live — so returning to the app silently logs them back in.

The signout handler in `src/lib/components/layout/Sidebar/UserMenu.svelte` cleared the `user` store immediately before assigning `location.href`:

```js
const res = await userSignOut();
user.set(null);                          // <- fires the reactive guard
localStorage.removeItem('token');
location.href = res?.redirect_url ?? '/auth';
```

v0.11.0 added a reactive guard to `src/routes/(app)/+layout.svelte`:

```js
$: if (loaded && ($user === undefined || $user === null)) {
	void gotoAuth();
}
```

That statement reacts to the `user.set(null)` on the line above and runs a client side `goto('/auth?redirect=...')` — a same-origin route change with no network round trip. It lands long before the cross-origin navigation to the IdP (DNS + TLS + HTTP) can commit, and cancels it. Before v0.11.0 the same check lived in `onMount`, which does not re-fire on a later `user.set(null)`, so there was no competing navigation.

### Changed

- The signout handler navigates without first clearing the `user` store.

### Fixed

- RP-initiated logout now reaches the IdP's `end_session_endpoint`, so the SSO session is actually terminated.

---

### Additional Information

The issue notes that reordering alone may not be enough, since the reactive statement can still fire within the same microtask. That is why the store write is removed rather than moved: `location.href` always performs a full document navigation, which discards every store, so clearing `user` first was redundant as well as harmful. A comment records this so it is not reinstated as a tidy-up later.

The two other `userSignOut()` call sites are unaffected and left alone:

- `src/lib/components/layout/Overlay/AccountPending.svelte` never wrote to the `user` store to begin with.
- `src/routes/+layout.svelte`'s `clearExpiredSession` deliberately clears it, and already guards itself with its own `isAuthRedirectInProgress` flag.

### Testing

- Traced every consumer of the `user` store in `UserMenu.svelte`; the store is read throughout the template but the removed write was the only one in this handler, so the import and all other usages stand.
- `npm run build` completes (`✓ built in 50.93s`), and `npx vitest run` passes.
- `npx prettier --check` clean on the changed file.
- Not yet reproduced end to end against an IdP — this needs an OIDC provider with `end_session_endpoint` configured. @mgroovyy offered in #28320 to test a proposed fix against their Keycloak 26.0 setup, and the expected observation is that the request to the `end_session_endpoint` completes instead of showing `(cancelled)` in the Network panel.

### Screenshots or Videos

Not applicable — the visible change is a navigation that is no longer cancelled.

> [!NOTE]

### DIFF
--- a/src/lib/components/layout/Sidebar/UserMenu.svelte
+++ b/src/lib/components/layout/Sidebar/UserMenu.svelte
@@ -571,11 +571,14 @@
 				type="button"
 				on:click={async () => {
 					const res = await userSignOut();
-					user.set(null);
 					localStorage.removeItem('token');
 
-					location.href = res?.redirect_url ?? '/auth';
 					show = false;
+					// A full document navigation follows and discards all client state, so
+					// the `user` store is deliberately left alone: clearing it here trips the
+					// reactive auth guard in (app)/+layout.svelte, whose client side goto
+					// cancels this navigation before the browser can leave for the IdP.
+					location.href = res?.redirect_url ?? '/auth';
 				}}
 			>
 				<div class="self-center">


Classic298:

macodev00:
Note on the failing `Format & Build` check: it fails at `Ensure working tree is clean`, and the cause is pre-existing on `dev` rather than anything in this PR.

`npm run i18n:parse` regenerates the locale files, and the checked-in ones are currently missing seven keys that the source already uses:

```
+	"Orchestrator": ""
+	"Per automation": ""
+	"Per chat": ""
+	"Start the chat to use this terminal.": ""
+	"Terminal Contexts": ""
+	"Upload failed": ""
+	"Waiting for upload": ""
-	"Oops! There are files still uploading. Please wait for the upload to complete.": ""
```

Those are the only entries the step reports, they are unrelated to this change, and the same step is already failing on `dev` itself: https://github.com/open-webui/open-webui/actions/runs/31280151617 (`dev` @ 009999f3).

For this PR specifically: `npm run build` and `npx vitest run` both pass locally, and `prettier --check` is clean on every changed file. Glad to open a separate `i18n:` PR that just re-runs the parser and re-greens `dev`, if that would help.

Issue #27685, opened 2026-07-29
issue: no way to reset a model avatar to the default logo since the v0.11.0 model editor redesign


silentoplayz:
### Affected Version

v0.11.0 / current `dev`

### Expected Behavior

The model editor offers a way to reset a model's avatar back to the default logo, as it did through v0.10.2 (the "Reset Image" button under the avatar).

### Actual Behavior

There is no way to reset a model's avatar anymore, in either the workspace model editor or the admin model editor (both render the same `ModelEditor` component). Clicking the avatar only opens the file picker, so once a model has a custom image it can only ever be replaced, never cleared.

### Steps to Reproduce

1. Workspace → Models → edit any model (or Admin Panel → Settings → Models → edit).
2. Upload a custom avatar image and save.
3. Try to restore the default logo. The avatar button only opens the file picker; no reset control exists anywhere in the editor.

### Root Cause

The model editor redesign (commit `b92f59230`, first shipped in v0.11.0) dropped the "Reset Image" button and its `en-US` i18n key from `src/lib/components/workspace/Models/ModelEditor.svelte`. Through v0.10.2 the editor had:

```svelte
<button
	class="px-2 py-1 text-gray-500 rounded-lg text-xs"
	on:click={() => {
		info.meta.profile_image_url = `${WEBUI_BASE_URL}/static/favicon.png`;
	}}
	type="button"
>
	{$i18n.t('Reset Image')}</button
>
```

This looks like an unintended casualty of the visual overhaul rather than a deliberate removal — it re-opens exactly what #7278 originally reported.

### Additional Information

Notes for whoever restores it, from prototyping a fix (a restoration branch exists at `silentoplayz:fix/model-avatar-reset` and can become a PR once the preferred design is clear):

- Any design that shows the reset control **only when a custom image is set** has to classify "default", and the default currently has three representations depending on where the model came from: the editor's absolute `${WEBUI_BASE_URL}/static/favicon.png`, the backend's relative `/static/favicon.png` (`routers/models.py`), and an **absent key** — the editor's model-load does a shallow merge (`info = { ...info, ...model }`), so a model `meta` without `profile_image_url` replaces the default wholesale and the favicon you see is only the template's else-fallback. The original button avoided all of this by being unconditional (resetting an already-default model is a harmless no-op).
[...]

Issue #7278, opened 2024-11-23
Bug: There is no way to remove a custom model image/icon


willdady:
## Installation Method

Docker compose

## Environment

- **Open WebUI Version:** v0.4.3
- **Operating System:** MacOS Sequoia
- **Browser (if applicable):** Firefox

**Confirmation:**

- [x] I have read and followed all the instructions provided in the README.md.
- [x] I am on the latest version of both Open WebUI and Ollama.
- [ ] I have included the browser console logs.
- [ ] I have included the Docker container logs.
- [x] I have provided the exact steps to reproduce the bug in the "Steps to Reproduce" section below.

## Description

After adding an image/icon to a custom model there is no way to remove it to reset it back to the default "OI" graphic.

Adding a model image can take a bit of trial and error and sometimes you simply want to remove the image entirely. Ideally there should be a button to remove any custom uploaded image.

tjbck:
Added to dev!

PR #28355, opened 2026-08-09
fix: restore the model avatar reset control

macodev00:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #27685. Re-opens what #7278 originally reported.
- [x] **Description:** below.
- [ ] **Documentation:** not applicable — restores a control that existed through v0.10.2.
- [x] **Testing:** see below.
- [ ] **No Unchecked AI Code:** prepared with AI assistance. The production build was run and the three default-image shapes were traced through the code; the control has not been clicked in a browser.
- [x] **Architecture:** no new setting and no new state beyond one derived boolean. Placement is the one judgement call here and I am happy to move it — see below.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

The model editor redesign (`b92f59230`, first shipped in v0.11.0) dropped the "Reset Image" button. Since then the avatar button only opens the file picker, so once a model has a custom image it can be replaced but never cleared back to the default logo — in both the workspace and admin editors, which render the same `ModelEditor` component.

### Added

- A reset control on the model avatar, restoring the behaviour that existed through v0.10.2.

### Fixed

- A model's custom avatar can be cleared back to the default logo again.

---

### Additional Information

**Classifying "default".** #27685 warns that the default logo reaches the editor in three shapes, and the control is only shown when none of them match: the absolute `${WEBUI_BASE_URL}/static/favicon.png` the editor initialises with, the relative `/static/favicon.png` the backend stores in `routers/models.py`, and an absent key, because loading a model shallow-merges its `meta` over the default. The first of those three was already being tested inline for the avatar's background colour, so the derived boolean also removes a small duplication.

**Placement.** The redesigned header is a full-width row, so the old `justify-end` position no longer reads as "next to the avatar". It is a badge on the avatar's top-right corner instead. Two deliberate choices there, both easy to change if the design should go elsewhere:

- It is always visible when a custom image is set, rather than hover-only like the camera badge, so it stays reachable on touch.
- It renders only over a custom image, never over the fallback logo, so the LICENSE note directly above that markup — the fallback logo must not be obscured — continues to hold. Resetting also restores that logo rather than removing it.

**i18n.** The redesign dropped the `Reset Image` key along with the button. It is re-added to `en-US` and the other 62 locale files, at the position `npm run i18n:parse` places it, with empty values for translators to fill.

### Testing

- `npm run build` completes (`✓ built in 50.93s`) with the change in place.
- `npx prettier --check` clean on `ModelEditor.svelte` and on every locale file.
- Confirmed `npm run i18n:parse` places the new key at exactly the position it was hand-inserted, so a later parser run will not move it.
- Traced all three default-image shapes through `hasCustomProfileImage`; each suppresses the control, and any other value shows it.
- Not yet clicked through in a browser.

### Screenshots or Videos

Not attached — I would rather agree the placement first, since that is the open question in #27685 and a screenshot of a position that may change is not much use. Happy to add one once the placement is settled.

> [!NOTE]

### DIFF
--- a/src/lib/components/workspace/Models/ModelEditor.svelte
+++ b/src/lib/components/workspace/Models/ModelEditor.svelte
@@ -25,6 +25,7 @@
 	import AccessControl from '../common/AccessControl.svelte';
 	import Spinner from '$lib/components/common/Spinner.svelte';
 	import ChevronLeft from '$lib/components/icons/ChevronLeft.svelte';
+	import XMark from '$lib/components/icons/XMark.svelte';
 	import DefaultFiltersSelector from './DefaultFiltersSelector.svelte';
 	import DefaultFeatures from './DefaultFeatures.svelte';
 	import BuiltinTools from './BuiltinTools.svelte';
@@ -93,6 +94,15 @@
 			system: ''
 		}
 	};
+
+	// The default logo reaches the editor in three shapes: the absolute URL set
+	// above, the relative path the backend stores, and an absent key, because
+	// loading a model shallow-merges its meta over the default.
+	$: hasCustomProfileImage = Boolean(
+		info?.meta?.profile_image_url &&
+		info.meta.profile_image_url !== `${WEBUI_BASE_URL}/static/favicon.png` &&
+		info.meta.profile_image_url !== '/static/favicon.png'
+	);
 
 	let params = {
 		system: ''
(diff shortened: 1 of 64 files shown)


macodev00:
Note on the failing `Format & Build` check: it fails at `Ensure working tree is clean`, and the cause is pre-existing on `dev` rather than anything in this PR.

`npm run i18n:parse` regenerates the locale files, and the checked-in ones are currently missing seven keys that the source already uses:

```
+	"Orchestrator": ""
+	"Per automation": ""
+	"Per chat": ""
+	"Start the chat to use this terminal.": ""
+	"Terminal Contexts": ""
+	"Upload failed": ""
+	"Waiting for upload": ""
-	"Oops! There are files still uploading. Please wait for the upload to complete.": ""
```

Those are the only entries the step reports, they are unrelated to this change, and the same step is already failing on `dev` itself: https://github.com/open-webui/open-webui/actions/runs/31280151617 (`dev` @ 009999f3).

For this PR specifically: `npm run build` and `npx vitest run` both pass locally, and `prettier --check` is clean on every changed file. Glad to open a separate `i18n:` PR that just re-runs the parser and re-greens `dev`, if that would help.

Classic298:
> Not attached — I would rather agree the placement first

Great, then do that before opening a PR which implements it, before opening a PR where you say you do not even know where to place it.

---

Issue #26663, opened 2026-07-01
issue: pgvector external knowledge retrieval fails, query embedding is bound as double precision[] instead of vector


xiaogangzhu:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Pip Install

### Open WebUI Version

v0.10.2

### Operating System

 Ubuntu 22.04

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

When configuring an External Knowledge Base with the pgvector backend, the configuration test in the setup page should complete successfully after the database connection and field mappings are provided.

### Actual Behavior

While configuring an External Knowledge Base connection with PostgreSQL + pgvector, the configuration test fails with the following error:

psycopg.errors.UndefinedFunction: operator does not exist: vector <=> double precision[]
LINE 5:                            "embedding" <=> $1 AS distance
                                               ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.

### Steps to Reproduce

Open Admin Panel → Settings → External Knowledge.
Start configuring a new External Knowledge Base connection.
Select pgvector as the backend.
Enter the PostgreSQL connection information and table/column mappings.
Use a PostgreSQL table whose embedding column is of type vector.
Run the configuration test from the External Knowledge Base setup page.
The test fails with the error above.

[...]

PR #28363, opened 2026-08-09
fix: send the query embedding as a vector in external pgvector retrieval

macodev00:
Testing an External Knowledge Base against pgvector fails before it can return anything:

```
psycopg.errors.UndefinedFunction: operator does not exist: vector <=> double precision[]
LINE 5:                            "embedding" <=> $1 AS distance
```

Closes #26663.

`_retrieve_pgvector` calls `register_vector(conn)` and then passes the embedding straight through as the query parameter. `register_vector` registers dumpers for `numpy.ndarray` and for `pgvector.psycopg.Vector`, and for nothing else, so a plain Python list of floats never reaches them: psycopg falls back to its own `ListDumper`, which sends the parameter as `float8[]`. Postgres then looks for `vector <=> double precision[]`, an operator pgvector does not define, and the statement fails before touching any rows.

Wrapping the embedding in `Vector` is what pairs with the `register_vector` call already there, so the parameter is sent as the type the operator is defined for.

Confirmed by registering pgvector's adapters against a local adapters map, exactly as `register_vector(conn)` does on a live connection, and dumping both forms:

```
current: raw list        -> sent as 'float8'   (oid 1022)
fixed:   Vector(list)    -> sent as 'vector'
```

oid 1022 is `double precision[]`, which is the type named in the error, so the failure is reproduced at the point where the parameter is adapted rather than inferred from the traceback.

Not adding a `::vector` cast in the SQL instead: that would depend on a cast existing from whatever array type psycopg happens to choose, and it would leave the `register_vector` call doing nothing on the outbound side. The other external backends build their request bodies as JSON and are unaffected.

#### Fixed

- External knowledge retrieval against pgvector no longer fails with `operator does not exist: vector <=> double precision[]`.

> [!NOTE]


flefevre:
Hello
I have checked and apply the patch.
It allows to restore the external pgvector function.
Hope it could be integrated in the next release

PR #28373, opened 2026-08-10
fix: exclude non-chat models from OpenAI-compatible gateway catalogs

knowhycodata:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Relates to #28371
- [ ] **Documentation:** No documentation change needed — no new setting is introduced.
- [x] **Dependencies:** No new or changed dependencies.
- [x] **Testing:** Manual end-to-end testing performed against a live gateway; details below.
- [ ] **No Unchecked AI Code:** See the disclosure under *Additional Information* before reviewing.
- [x] **Architecture:** No new setting — the fix uses metadata the provider already sends.
- [x] **Git Hygiene:** Single atomic commit, rebased on `dev`, one file changed.
- [x] **Title Prefix:** `fix`

---

## Description

When an OpenAI-compatible **gateway** is configured as a connection, Open WebUI lists that gateway's entire `/models` catalog as chat models. Gateways front many providers at once, so the catalog also contains embedding, image-generation, video-generation, rerank and speech models. Every one of them shows up in the chat model picker and fails the moment a user selects it.

The existing `_UNSUPPORTED_OPENAI_MODEL_KEYWORDS` filter does not help: it is applied only when the hostname is `api.openai.com`, and being keyword-based it cannot recognise third-party ids such as `recraft/recraft-v4` or `voyageai/rerank-2.5`.

The information needed is already in the response. Gateways tag each entry with the OpenAI-style capability hints:

```json
{
  "id": "voyageai/rerank-2.5",
  "supported_operations": ["RERANK"],
  "supported_endpoints": ["/v1/rerank"]
}
```

Open WebUI ignored both fields. This PR adds `supports_text_generation()` and applies it in the two places a catalog is turned into a model list — `get_all_models_responses()` and the `/models/{url_idx}` admin picker.

**Behaviour:**

- A model is kept if it advertises the operation the connection actually uses: `CHAT_COMPLETIONS`, or `RESPONSES` when the connection has `api_type: responses`.
- `supported_operations` is preferred; `supported_endpoints` is the fallback.
- **A provider advertising neither field is treated as chat capable**, so plain OpenAI, Ollama, vLLM and llama.cpp connections are byte-for-byte unaffected.
- Connections with a manually configured `model_ids` list are left untouched — an explicit list is taken at face value.

No new setting, no new dependency, one file changed.

## Testing

Ran the backend against a live OpenAI-compatible gateway whose `/v1/models` returns 229 entries (154 chat-only, 11 chat+responses, 14 responses-only, 15 embeddings, 16 image, 9 video, 3 TTS, 2 rerank, 1 STT, and others).

| Check | Before | After |
|---|---|---|
| `GET /api/models` | 230 | **166** (165 chat + built-in `arena-model`) |
| Non-chat entries in the picker | 64 | **0** |
| `GET /openai/models/0` (admin model-id picker) | 229 | **165** |
| `POST /api/chat/completions`, non-streaming | ok | ok |
| `POST /api/chat/completions`, streaming | ok | ok |
| `POST /openai/verify` | ok | ok |
| Manual `model_ids` list (2 ids typed) | 2 | **2** — unchanged, filter correctly skipped |
| Second connection with `api_type: responses` | 229 | **25**, and chat through it returns 200 |

Both connections were configured simultaneously to confirm they do not interfere.

`ruff format --check` is clean. `ruff check` reports no new findings — the two `C901` complexity counts that move (`get_all_models_responses` 16→19, `get_models` 13→14) were already above the threshold before this change, as are eight other functions in the file.

## Additional Information

**Disclosure — please read.**

1. **AI-assisted.** This patch was written with AI assistance. I have reviewed it and run the end-to-end testing described above myself, but I am leaving the *"No Unchecked AI Code"* box unchecked rather than assert something you may want to judge for yourselves. Please review it as you would any outside patch.
2. **Conflict of interest.** The gateway I tested against is [llmtr.com](https://llmtr.com), which I operate. This PR contains **no** reference to it and no vendor-specific code — the fix keys off standard OpenAI `supported_operations` / `supported_endpoints` metadata and helps any gateway that emits it, LiteLLM included. I am flagging it so the context is on the record.

Discussion #28371 has the longer write-up.

*Resubmitted from #28372, which the CLA bot closed because the agreement checkbox was left unchecked.*

---

# Changelog Entry

### Description

- Non-chat models returned by OpenAI-compatible gateways no longer appear in the chat model picker.

### Fixed

- **Gateway model lists**: OpenAI-compatible gateways return their whole catalog from `/models`, so embedding, image-generation, video-generation, rerank and speech models were listed as chat models and failed on selection. Open WebUI now honours the `supported_operations` / `supported_endpoints` metadata these gateways already send, and selects the operation matching the connection's API type. Providers that do not send these fields are unaffected.

### DIFF
--- a/backend/open_webui/routers/openai.py
+++ b/backend/open_webui/routers/openai.py
@@ -86,6 +86,34 @@ def _clean_proxy_headers(raw_headers) -> dict:
     return {k: v for k, v in raw_headers.items() if k not in _STRIP_PROXY_HEADERS}
 
 
+def supports_text_generation(model: dict, api_type: str | None = None) -> bool:
+    """Whether *model* can serve the text generation endpoint this connection uses.
+
+    Gateways that front many providers return their whole catalog from
+    ``/models`` - embedding, image, video, rerank and audio models included -
+    and mark each entry with the OpenAI-style ``supported_operations`` /
+    ``supported_endpoints`` hints.  Without this check every one of those
+    non-chat models shows up in the chat model picker.  Providers that
+    advertise nothing are assumed to be chat capable, which keeps plain
+    OpenAI / Ollama / vLLM behaviour unchanged.
+    """
+    if not isinstance(model, dict):
+        return True
+
+    operation = 'RESPONSES' if api_type == 'responses' else 'CHAT_COMPLETIONS'
+
+    operations = model.get('supported_operations')
+    if isinstance(operations, list) and operations:
+        return operation in operations
+
+    endpoints = model.get('supported_endpoints')
+    if isinstance(endpoints, list) and endpoints:
+        endpoint = '/responses' if api_type == 'responses' else '/chat/completions'
+        return any(isinstance(e, str) and e.endswith(endpoint) for e in endpoints)
+
+    return True
+
+
 async def send_get_request(
     request: Request = None,
     url=None,
@@ -596,6 +624,20 @@ async def get_all_models_responses(request: Request, user: UserModel) -> list:
                 # Catch non-list responses
                 model_list = []
 
+            if not api_config.get('model_ids'):
+                # Manually listed model ids are taken at face value; a fetched
+                # catalog gets narrowed down to the models this connection can
+                # actually chat with.
+                api_type = api_config.get('api_type')
+                filtered = [model for model in model_list if supports_text_generation(model, api_type)]
+
+                if len(filtered) != len(model_list):
+                    model_list = filtered
+                    if isinstance(response, list):
+                        responses[idx] = model_list
+                    else:
+                        response['data'] = model_list
+
             for model in model_list:
                 # Remove name key if its value is None #16689
                 if 'name' in model and model['name'] is None:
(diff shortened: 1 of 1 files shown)


knowhycodata:
@tjbck — gentle ping on this one when you have a moment, no rush.

Quick recap of where it stands:

- **Scope is small:** one file, `+51/-0`, no new setting and no new dependency. It reads the `supported_operations` / `supported_endpoints` metadata that OpenAI-compatible gateways already return, so embedding / image / video / rerank / speech models stop appearing in the chat model picker. Providers that don't send those fields are treated as chat capable, so OpenAI, Ollama, vLLM and llama.cpp connections are untouched.
- **Independently confirmed:** @SupaNexus-ai ran into the same thing in discussion #28371 — *"this is a common challenge when one endpoint exposes multiple model types... a capability-based model filtering approach could help here."*
- **Mergeable:** GitHub reports no conflicts against `dev`. The one upstream change to `routers/openai.py` since I branched touches different regions of the file.
- **CI has not run** — no check runs have been triggered, which I believe needs maintainer approval for a first-time contributor.

Happy to rebase onto current `dev`, split it further, or drop it entirely if this isn't a direction you want — just say the word. The disclosures in the PR description (AI-assisted patch, and that the gateway I tested against is one I operate) still stand.

Classic298:
do all OpenAI compatible providers support this? ALL of them? And what happens if not? And too many comments.

knowhycodata:
Not all of them, and the patch doesn't require it.

I checked: OpenAI's own API doesn't send these fields, and OpenRouter doesn't either. Azure AI Foundry does, and so does the gateway I tested against.

If a provider doesn't send them, nothing changes. Both lookups are guarded and fall through to `return True`, so a model that advertises nothing is treated as chat capable — OpenAI, Ollama, vLLM and llama.cpp connections produce exactly the same list as today. The filter only removes a model that explicitly declares it can't serve chat. It never removes one on a guess.

On the comments — happy to trim them down, or drop them entirely if you'd rather the code speak for itself. Just tell me which and I'll push.

Classic298:
closing for now, discussion has zero traction still and we have the whitelist feature for that already. if more people want this we can reconsider

Issue #28319, opened 2026-08-09
bug: MCP tool-server OAuth never requests access_type=offline, so Google MCP sessions are deleted every hour


materemias:
## Bug Description

MCP tool servers authenticated with `oauth_2.1` / `oauth_2.1_static` cannot obtain a refresh token from Google, because the authorization request built for tool servers never carries `access_type=offline`. Google therefore issues an online-only access token (`expires_in: 3599`, no `refresh_token`), and ~55 minutes later Open WebUI **deletes the session**. Every user must re-authorize every hour, for every Google MCP server.

This is specifically the *tool server* OAuth path. The equivalent problem on the **login** path was already solved — `GOOGLE_OAUTH_AUTHORIZE_PARAMS` / `OAUTH_AUTHORIZE_PARAMS` exist and work (#22652, "Addressed in dev") — but neither reaches MCP connections.

## Environment

- Open WebUI: v0.11.0 (Docker, digest-pinned)
- MCP servers: Google's hosted Workspace MCP servers (`https://gmailmcp.googleapis.com/mcp/v1`, `drivemcp`, `calendarmcp`, …)
- Auth type: `OAuth 2.1 (Static)`, confidential client, PKCE S256
- Verified against `v0.11.0` and `main` — `backend/open_webui/utils/oauth.py` is **byte-identical** on both (`md5 48c61d157b9095da18f6d49f648a27c2`), and also identical to the file inside the running container. `access_type`, `offline` and `prompt` appear nowhere in it.

## Steps to Reproduce

1. Add an MCP (Streamable HTTP) tool server pointing at `https://gmailmcp.googleapis.com/mcp/v1`, auth type `OAuth 2.1 (Static)`, with a Google Web-application client.
2. Complete the consent flow from a chat (Integrations → Tools).
3. Use the tools — works.
4. Wait an hour, or simply come back the next morning.
5. The integration is disconnected and the stored OAuth session is gone. Re-consent required.

## Evidence

Authorization redirect actually emitted by `GET /oauth/clients/mcp:<id>/authorize` (values redacted):

```
https://accounts.google.com/o/oauth2/v2/auth
  ?response_type=code
  &client_id=<redacted>.apps.googleusercontent.com
  &redirect_uri=https://<host>/oauth/clients/mcp:gmail/callback
  &scope=https://www.googleapis.com/auth/gmail.readonly+https://www.googleapis.com/auth/gmail.compose
  &state=<redacted>
  &code_challenge=<redacted>
[...]

PR #28374, opened 2026-08-10
fix: send provider-required OAuth authorization parameters for MCP tool servers

materemias:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #28319
- [x] **Description:** below
- [x] **Changelog:** below
- [ ] **Documentation:** happy to open a docs PR for the new connection option if you want it documented
- [x] **Dependencies:** none added
- [x] **Testing:** see *Testing* below — the mechanism has been running in production; the new default and the JSON-string path are unit-exercised
- [x] **No Unchecked AI Code:** every line human-reviewed; the mechanism was live-tested against a Google Workspace MCP deployment, the new default and JSON-string path are unit-exercised (scope spelled out under *Testing*)
- [x] **Self-Review:** performed; `ruff format --diff` is empty for this file and no new lint findings are introduced
- [x] **Architecture:** smart default (no configuration needed for Google) with a per-connection override; follows the existing `oauth_resource_parameter` pattern
- [x] **Git Hygiene:** single atomic commit on top of `dev`
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

MCP tool servers using `oauth_2.1` / `oauth_2.1_static` never obtain a refresh token from Google, so every Google MCP integration (Gmail, Drive, Calendar, Docs, Sheets, Slides, Workspace Search) loses its authorisation roughly an hour after consent, for every user.

Google issues a refresh token only when the authorization request carries the non-standard `access_type=offline`. Nothing in OIDC discovery advertises this, so authlib cannot infer it, and `build_oauth_request_params()` — the only source of extra authorization parameters on the tool-server path — emits just `scope` and `resource`. The resulting token is online-only (`expires_in: 3599`, no `refresh_token`). Five minutes before expiry `get_oauth_token()` attempts a refresh, `_refresh_token()` bails on the missing refresh token, and the caller **deletes the session**.

The existing `OAUTH_AUTHORIZE_PARAMS` and `GOOGLE_OAUTH_AUTHORIZE_PARAMS` settings do not help: both are applied only in `OAuthManager.handle_login()` / `google_oauth_register()`, i.e. the SSO login flow, never in `OAuthClientManager.handle_authorize()`.

This is *not* the case fixed by #26141. Google does send `expires_in`, so the expiry is genuine rather than fabricated; the session is correctly considered expired and is then destroyed because a refresh token was never requested in the first place.

### Added

- `oauth_authorize_params` on `OAuthClientInformationFull`, a per-connection option holding extra authorization parameters, read from `info.oauth_authorize_params` (with a `config.` fallback) by `get_connection_oauth_authorize_params()`. Accepts a dict or a JSON string; malformed input is logged and ignored rather than raising. This generalises the mechanism `oauth_resource_parameter` already uses, so any provider needing a non-standard authorization parameter is covered rather than special-casing one. Three states, distinguished by key presence:

  | Value | Meaning |
  |---|---|
  | unset (`None`) | provider defaults apply |
  | `{}` | deliberate opt-out; no extra parameters sent |
  | non-empty dict | per-key override on top of the provider defaults |

  The last row merges rather than replaces on purpose: a straight replacement would mean setting `{"prompt": "select_account"}` on a Google connection silently drops `access_type=offline` and reintroduces the very bug this fixes.
- `default_oauth_authorize_params()`, applied when a connection configures nothing: Google-issued client ids (suffix `.apps.googleusercontent.com`) get `access_type=offline` and `prompt=consent`. `prompt=consent` is required so that users who already granted receive a refresh token rather than only new grantees.

### Changed

- `build_oauth_request_params()` merges the connection's parameters, falling back to the provider defaults.

### Fixed

- Google MCP tool-server sessions are no longer deleted an hour after consent. Open WebUI's existing refresh machinery — which was already wired in but had never been given a refresh token — now renews them silently.

### Breaking Changes

None. The field defaults to empty; the only behavioural change is for connections whose OAuth client id is Google-issued, which currently cannot work for more than an hour.

---

### Testing

Deployed against a real multi-user Open WebUI instance with seven of Google's hosted Workspace MCP servers, Google Cloud project enrolled in the Workspace Developer Preview Program, `OAuth 2.1 (Static)` confidential client with PKCE.

- **Before:** `GET /oauth/clients/mcp:<id>/authorize` → 302 carrying only `response_type, client_id, redirect_uri, scope, state, code_challenge, code_challenge_method`. Stored session token keys: `access_token, expires_at, expires_in, issued_at, scope, token_type` — no `refresh_token`. `tokeninfo` reports `access_type: "online"`. Sessions vanished about an hour after consent.
- **After:** the same 302 carries `access_type=offline` and `prompt=consent` for all seven servers. Following one re-consent, `GET /api/v1/users/{id}/oauth/sessions` shows `refresh_token` present for `mcp:gmail`, and the session now survives well past the one-hour mark.
- **Non-Google connections:** with no connection value set and a non-Google client id, `build_oauth_request_params()` output is byte-identical to before this change.

Scope of that evidence, to be precise: the per-connection field, the parser and the merge ran in production (configured as a dict via `POST /api/v1/configs/tool_servers`, on `v0.11.0` where this file is otherwise identical to `dev`). The Google default and the JSON-string input path are new here and are covered by the checks below rather than by production traffic.

`get_connection_oauth_authorize_params()` — dict via `info`, JSON string, `config` fallback, invalid JSON (logs and returns `{}`), non-dict input, absent key, empty connection, `None` values stripped. `default_oauth_authorize_params()` — Google client id, non-Google client id, empty id, `None` id, and a lookalike (`evil.apps.googleusercontent.com.attacker.test`, correctly rejected since the check is a suffix match). All pass.

### Notes

- No UI control is included. The option is settable through `POST /api/v1/configs/tool_servers`, and Google — the case that actually breaks — needs no configuration at all thanks to the default. If you'd like it exposed beside the existing "OAuth Resource Parameter" advanced option in `AddToolServerModal.svelte`, say so and I'll add it.
- On the first-time-contributor note: the design was written up in #28319 with three candidate shapes and an explicit request for maintainer preference. This PR implements the shape matching the precedent set by #25898 (per-connection option) plus a smart default, per the architecture guidance in this checklist. Glad to move the discussion to Discussions if you'd prefer that.

### DIFF
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -2,6 +2,7 @@ import asyncio
 import base64
 import fnmatch
 import hashlib
+import json
 import logging
 import re
 import sys
@@ -103,6 +104,8 @@ class OAuthClientInformationFull(OAuthClientMetadata):
     issuer: Optional[str] = None  # URL of the OAuth server that issued this client
     resource: Optional[str] = None  # RFC 8707 resource indicator for JWT audience
     oauth_resource_parameter: OAuthResourceParameterMode = 'auto'
+    # None means unset (provider defaults apply); {} is a deliberate opt-out.
+    oauth_authorize_params: Optional[dict[str, str]] = None
 
     client_id: str
     client_secret: str | None = None
@@ -119,6 +122,10 @@ logging.basicConfig(stream=sys.stdout, level=GLOBAL_LOG_LEVEL)
 log = logging.getLogger(__name__)
 
 OAUTH_RESOURCE_PARAMETER_MODES = {'auto', 'include', 'omit'}
+
+# Google-issued OAuth client ids always carry this suffix; used to apply the
+# provider's required authorization parameters when none are configured.
+GOOGLE_CLIENT_ID_SUFFIX = '.apps.googleusercontent.com'
 
 OAUTH_RUNTIME_CONFIG = {
     'DEFAULT_USER_ROLE': ('ui.default_user_role', DEFAULT_USER_ROLE),
@@ -743,6 +750,41 @@ def get_connection_oauth_resource_parameter(connection: dict) -> OAuthResourcePa
     )
 
 
+def get_connection_oauth_authorize_params(connection: dict) -> dict[str, str] | None:
+    """Extra provider-specific parameters for the authorization request.
+
+    Some providers require non-standard authorization parameters that are not
+    advertised by OIDC discovery, so authlib cannot infer them. Google needs
+    access_type=offline to issue a refresh token at all, plus prompt=consent to
+    re-issue one for users who already granted.
+
+    Returns None when the connection does not configure the option, so the
+    caller can tell "unset" from an explicitly empty object, which is a
+    deliberate opt-out of the provider defaults.
+    """
+    info = connection.get('info') or {}
+    config = connection.get('config') or {}
+
+    if 'oauth_authorize_params' in info:
+        value = info['oauth_authorize_params']
+    elif 'oauth_authorize_params' in config:
+        value = config['oauth_authorize_params']
+    else:
+        return None
+
+    if isinstance(value, str):
+        try:
+            value = json.loads(value)
+        except (json.JSONDecodeError, TypeError):
+            log.warning('oauth_authorize_params is not valid JSON, ignoring')
+            return None
+
+    if not isinstance(value, dict):
+        return None
+
+    return {str(k): str(v) for k, v in value.items() if v is not None}
+
+
 def apply_connection_oauth_options(connection: dict, oauth_client_info: dict) -> dict:
     info = connection.get('info') or {}
     config = connection.get('config') or {}
@@ -752,6 +794,7 @@ def apply_connection_oauth_options(connection: dict, oauth_client_info: dict) ->
     options = {
         **oauth_client_info,
         'oauth_resource_parameter': get_connection_oauth_resource_parameter(connection),
+        'oauth_authorize_params': get_connection_oauth_authorize_params(connection),
     }
     if oauth_scope:
         options['scope'] = oauth_scope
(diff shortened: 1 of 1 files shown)


materemias:
Thanks — all three comments were the same underlying design flaw and they were right. Fixed in 9d2d5d9.

`oauth_authorize_params` is now `Optional[dict[str, str]] = None`, and `get_connection_oauth_authorize_params()` uses key-presence checks (`'oauth_authorize_params' in info`, then `config`) returning `None` when the key is absent or the value is unusable. That gives three distinct states:

| Value | Meaning |
|---|---|
| `None` (unset) | provider defaults apply — Google gets `access_type=offline` + `prompt=consent` |
| `{}` (explicitly empty) | deliberate opt-out; no extra parameters are sent at all |
| non-empty dict | per-key override **on top of** the provider defaults |

I kept the last row as a merge rather than a replacement, because straight replacement is a quiet footgun: setting `{"prompt": "select_account"}` on a Google connection would drop `access_type=offline` and silently reintroduce the hourly session loss this PR exists to fix. Merging means a connection can adjust one parameter without having to restate the rest, and `{}` still provides the clean escape hatch you asked for.

Covered by checks over all three states: absent key, empty dict, empty JSON string, invalid JSON, explicit `null`, `config` fallback, `info: {}` correctly taking precedence over a non-empty `config` value, plus the build-level cases (unset → defaults, `{}` → nothing, custom → merged, non-Google untouched). All pass, and `ruff format --diff` is empty.

The mutable-default point is addressed by the same change — the field now defaults to `None` rather than a shared `{}`.

Classic298:
issue ref?

materemias:
#28319 is the issue. It was in the description from the start, but I wrapped `Closes #28319` in backticks, so GitHub never parsed the keyword and the Development panel stayed empty. My mistake.

I've fixed the body, but the closing reference still doesn't register while the PR is closed, and GitHub won't let me reopen it myself (`reopenPullRequest` returns a 422; head branch and `dev` both exist and the head commit is still reachable, so I can't see what it's objecting to). Could you reopen it, or would you rather I open a fresh PR from the same branch?

Root cause, briefly. `build_oauth_request_params()` is the only place extra authorization parameters can reach the tool-server OAuth path, and it emits `scope` and `resource`. Google issues no refresh token unless the authorization request carries `access_type=offline`, so the stored session has no `refresh_token`. Roughly an hour after consent `get_oauth_token()` tries to refresh, `_refresh_token()` bails on the missing token, and the caller deletes the session. `OAUTH_AUTHORIZE_PARAMS` can't cover this because it is applied in `handle_login()` only.

The branch is 1 commit ahead of `dev` and 271 behind. Happy to rebase onto current `dev`, and happy to drop the Google default if you'd rather keep provider logic out of that file.

materemias:
Superseded by https://github.com/open-webui/open-webui/pull/29003 — same commit rebased onto current `dev`. Closing keyword is now plain text so the link registers. Keeping this one closed to avoid a duplicate open PR.

PR #28375, opened 2026-08-10
fix: make automation Enable/Disable All cover every page, not just the loaded one

Classic298:
The bulk actions iterated over the `automations` array, which only ever holds the 30 items of the current page, so "Disable All" left every automation on subsequent pages untouched.

The toggle now runs server side over the same filters the list uses, so it covers the whole result set rather than one page, and returns the number of automations actually changed.

> [!NOTE]


Classic298:
@tjbck ready for review. "Disable All" only touches the automations on the page currently loaded, so everything on later pages stays enabled. It now runs server side across the whole result set.

Issue #28378, opened 2026-08-10
issue: An attached web page or YouTube video reaches the model with no content


silentoplayz:
## Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items.
- [x] I am using the latest version of Open WebUI.

## Installation Method

Source (uvicorn + vite), reproduced against current `dev`.

## Open WebUI Version

v0.11.0 (`dev`, at `8fbfd14a8`)

## Browser

Firefox

## Operating System

Debian based Linux host.

## Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I have provided the relevant configuration and environment variables.

## Expected Behavior

Attaching a URL in the message input should give the model the text extracted from that page or video.

## Actual Behavior

The attachment appears to succeed. A collection is created, the source is listed under the reply, and the interface shows no error. The text handed to the model is empty, so the reply describes the source as a placeholder with no content and answers nothing about the page.

Opening the source to check what was captured then throws, because the citation has no document:

```
Uncaught TypeError: can't access property "trim", $.get(...).document is null
    in CitationModal.svelte
    in Citations.svelte
    in ResponseMessage.svelte
```

Measured against a running instance, `POST /api/v1/retrieval/process/url`:

| URL | `content` returned |
|---|---|
| A plain web page | `None` |
| A longer article | `None` |
| A YouTube video with a reachable transcript | `None` |

Every one of those returned HTTP 200 with a `collection_name` set, so nothing signals that the text was lost.

## Steps to Reproduce

1. Attach any URL in the message input, either through Attach Webpage or by typing the URL after a hash.
2. Wait for the attachment to finish, then ask a question about it.
3. The reply reports that the provided context contains no information, and describes the source as a placeholder.
4. Click the source pill under the reply. The citation modal throws and does not open.

## Cause

`process_url` in `backend/open_webui/routers/retrieval.py` delegates to `process_web` for the web and YouTube cases, then reads the extracted text from the wrong key:

```python
result = await process_web(request, form_data, process=process, user=user)
return {
    'status': True,
    'type': 'web',
    'name': form_data.url,
    'url': form_data.url,
    'collection_name': result.get('collection_name'),
    'content': result.get('content'),
}
```

`process_web` only returns a top level `content` when it is asked not to process:

```python
else:
    return {
        'status': True,
        'content': content,
[...]

PR #28379, opened 2026-08-10
fix: keep the extracted text when attaching a URL

silentoplayz:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #28378
- [x] **Testing:** Reproduced against a running instance, then verified at the endpoint for a plain page, a long article and a YouTube transcript.
- [x] **Architecture:** No new settings and no new endpoint. One helper reading the value from where it is actually returned, and a null guard.
- [x] **Git Hygiene:** A single commit, +9/-3 across two files, based on the current `dev`.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

Attaching a URL looked successful but gave the model nothing. A collection was created, the source was listed under the reply, and no error appeared, yet the reply described the source as a placeholder with no content.

`process_url` delegates to `process_web` and then reads the extracted text from a key that only exists on the other branch:

```python
result = await process_web(request, form_data, process=process, user=user)
return {
    ...
    'content': result.get('content'),
}
```

`process_web` returns a top level `content` only when asked not to process. On the default path the text is nested under `result['file']['data']['content']`, so `result.get('content')` was always `None`, for the web branch and the YouTube branch alike. The frontend copies that value straight into the attachment it builds, which is what left the model with an empty document.

Opening the source to check what had been captured then threw, because `CitationModal.svelte` called `trim` on the missing document directly. That is fixed here too, so a citation without a document renders as empty instead of crashing.

### Fixed

- An attached web page or YouTube video now reaches the model with its extracted text, instead of arriving empty while appearing to have succeeded.
- Opening a citation that has no document no longer throws.

---

### Additional Information

**Why a helper rather than reaching into the nested value inline.** Both the web and the YouTube branch return the same shape, and `process_web` can legitimately produce either layout depending on `process`. Reading through one function keeps the two call sites honest about that, and keeps working if the unprocessed branch is ever used from here.

**Scope of the failure.** Because both branches read the same wrong key, every URL attachment was affected, not only web pages. A YouTube video whose transcript fetched perfectly still arrived with no text.

**The citation guard is deliberate.** With the content fixed, a null document should not normally occur through this path, but the modal receives citations from several sources and crashing on a missing field is not a reasonable failure mode for any of them.

`process_url` was introduced recently, so this affects current `dev` rather than v0.11.0 as released.

`POST /api/v1/retrieval/process/url` against a running instance, same URLs before and after:

| URL | Before | After |
|---|---|---|
| A plain web page | `content: None` | 129 characters |
| A longer article | `content: None` | 16,466 characters |
| A YouTube video with a reachable transcript | `content: None` | 2,089 characters |

Each returned HTTP 200 with a collection set in both cases, which is why nothing signalled the loss before.

The crash was reproduced in the browser by clicking the source pill under a reply, and the guard was confirmed by reading the compiled component.

### Screenshots or Videos

| Surface | Before | After |
|---|---|---|
| Reply to a question about an attached page | image | image |
| Citation modal opened from the source pill | image | image |

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/retrieval.py
+++ b/backend/open_webui/routers/retrieval.py
@@ -449,6 +449,12 @@ class ProcessUrlForm(CollectionNameForm):
     url: str
 
 
+def _extracted_web_content(result: dict) -> Optional[str]:
+    if result.get('content') is not None:
+        return result.get('content')
+    return ((result.get('file') or {}).get('data') or {}).get('content')
+
+
 class ProcessUrlResponse(BaseModel):
     status: bool
     type: str
@@ -2269,7 +2275,7 @@ async def process_url(
                 'name': form_data.url,
                 'url': form_data.url,
                 'collection_name': result.get('collection_name'),
-                'content': result.get('content'),
+                'content': _extracted_web_content(result),
             }
 
         config = await get_retrieval_config()
@@ -2283,7 +2289,7 @@ async def process_url(
                 'name': form_data.url,
                 'url': form_data.url,
                 'collection_name': result.get('collection_name'),
-                'content': result.get('content'),
+                'content': _extracted_web_content(result),
             }
 
         from open_webui.routers.files import upload_file_handler
--- a/src/lib/components/chat/Messages/Citations/CitationModal.svelte
+++ b/src/lib/components/chat/Messages/Citations/CitationModal.svelte
@@ -223,7 +223,7 @@
 									title={$i18n.t('Content')}
 								></iframe>
 							{:else}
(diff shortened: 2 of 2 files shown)


tjbck:
Should be addressed with 9c21d4ed3ba9ba8e53def7e1a1366b74834cb42d.

silentoplayz:
> Should be addressed with [9c21d4e](https://github.com/open-webui/open-webui/commit/9c21d4ed3ba9ba8e53def7e1a1366b74834cb42d).

Tested and working as expected. Thanks!

Issue #28380, opened 2026-08-10
issue: A message queued while a URL is attaching is never sent


silentoplayz:
## Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items.
- [x] I am using the latest version of Open WebUI.

## Installation Method

Source (uvicorn + vite), against current `dev`.

## Open WebUI Version

v0.11.0 (`dev`)

## Browser

Firefox

## Operating System

Debian based Linux host.

## Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I have provided the relevant configuration and environment variables.

## Expected Behavior

Sending a message while an attachment is still being processed should hold it briefly and then send it once the attachment is ready, which is what already happens for uploaded files.

## Actual Behavior

If the attachment is a URL rather than a file, the message stays queued after the attachment has finished. The queued row remains above the message input, the attachment shows as ready, and nothing sends it.

It is not lost. Navigating away from the chat and back sends it, because loading a chat while it is idle also runs the queue. In the normal flow, though, there is no event left that can release it.

## Steps to Reproduce

1. Visit a pre-existing chat.
2. Attach a URL, either through Attach Webpage or by typing the URL after a hash. A page that takes a few seconds to fetch makes the window easier to hit.
3. While the attachment is still processing, type a message and press enter. It is held in the queue, which is correct.
4. Wait for the attachment to finish. The queued message stays where it is.
5. Navigate to another chat and back. The message is now sent.

## Cause

Messages are queued when any attached file is still uploading, in `submitHandler` in `src/lib/components/chat/Chat.svelte`:

```js
if (
    ($chatRequestQueues[$chatId] ?? []).some((m) =>
        (m.files ?? []).some((file) => ['uploading', 'error'].includes(file.status))
    ) || ...
```

For uploaded files the release happens because the message input calls `onUpdate` when an upload completes, which refreshes the queued entry and then runs the queue:

```js
const onUpdate = async ({ file }: { file?: any } = {}) => {
    if (file?.itemId) { ...refresh the queued entry... }
    await processNextInQueue($chatId);
};
```

URLs take a different path. `uploadWeb` marks the item as uploaded but calls neither `onUpdate` nor `processNextInQueue`, and the items it builds carry no `itemId` for `onUpdate` to match on.

[...]

PR #28381, opened 2026-08-10
fix: release a queued message once an attached URL finishes

silentoplayz:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #28380
- [x] **Testing:** The stall was observed in the running application and the mechanism traced through the source. See the note below on what could and could not be reproduced on demand.
- [x] **Architecture:** No new state and no new function. One call to the existing queue hook, at the point the file path already calls it.
- [x] **Git Hygiene:** A single commit, +2 lines in one file, based on the current `dev`.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

A message sent while an attachment is still processing is held in the chat queue until the attachment settles. For uploaded files that works, because the message input calls `onUpdate` once the upload completes, which refreshes the queued entry and runs the queue:

```js
const onUpdate = async ({ file }: { file?: any } = {}) => {
    if (file?.itemId) { ...refresh the queued entry... }
    await processNextInQueue($chatId);
};
```

Attaching a URL takes a different path. `uploadWeb` marks its item as uploaded but calls neither `onUpdate` nor `processNextInQueue`, and the items it builds carry no `itemId` for `onUpdate` to match on.

The queue is otherwise revisited in only four places: when a generation completes, twice; when a generation is stopped; and when a chat is mounted while idle. In a new chat with nothing generating, none of those can happen, so a message queued behind a URL has no event able to release it. Reloading the chat sends it, through the mount case, but nothing on screen suggests that.

`uploadWeb` now runs the queue when it finishes, the same way the file path already does.

### Fixed

- A message sent while a URL attachment is still processing is now sent once the attachment is ready, instead of waiting in the queue until the chat is reloaded.

---

### Additional Information

**Why `onUpdate()` rather than `processNextInQueue($chatId)` directly.** `onUpdate` already is the hook the upload path uses for exactly this, and called without an argument it skips the entry refresh and runs the queue. Using it keeps both attachment paths going through one place, rather than adding a second way to reach the queue.

**Why no `itemId` is added to the web items.** The queued entry holds the same item objects that `uploadWeb` mutates, so the status is already current by the time the queue runs; only the trigger was missing. Giving web items an `itemId` would make `onUpdate` able to refresh them as well, but that is a larger change and is not needed to release the queue.

**On reproducibility.** The stall was observed in the running application while attaching a page that took roughly sixteen seconds to fetch, which is a wide enough window to queue a message behind. A later attempt to recreate it with automated input did not reproduce it, because once a page has a collection it resolves in under a second, so the message never queues at all. The mechanism above is therefore traced through the source rather than demonstrated by an automated test, and the fix is verified in the compiled component and by the file path that already relies on the same hook. Reviewers wanting to see it should use a URL that has not been attached before.

1. Traced every path that reaches `processNextInQueue`, confirming that four are generation or mount related and that `uploadWeb` reached none of them.
2. Confirmed `uploadWeb` calls neither `onUpdate` nor `processNextInQueue` before this change, and that its file items carry no `itemId`.
3. Confirmed the queued entry shares its item objects with the live attachment list, so the status the queue reads is already up to date and only the trigger was missing.
4. Confirmed the added call is present in the compiled component.
5. Checked that Prettier reports the same result on this file before and after, so no unrelated formatting is included.

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -1605,6 +1605,8 @@
 				toast.error(`${e}`);
 			}
 		}
+
+		await onUpdate();
 	};
 
 	const onUpdate = async ({ file }: { file?: any } = {}) => {


tjbck:
Thanks!

Issue #27498, opened 2026-07-26
issue: Custom model usage requires explicit base model permission for low-privileged users in v0.10.x (regression)


endre82:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.10.2

### Operating System

Ubuntu - Docker

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

**Given:**
- An admin creates a Workspace Model (custom model) with a base_model_id pointing to an underlying base model (e.g., llama3).
- The admin grants a low-privileged user (or group) access to the custom model — but not to the base model directly.

**Expected:**

- The low-privileged user can see, select, and chat with the custom model without issues.
- The underlying base model is implicitly permitted through the custom model assignment — no separate explicit permission for the base model is required.
- The base model should not appear in the user's model list directly; only the custom model should be visible.

**In short:** 
Granting access to a Workspace Model should automatically imply access to its base_model_id for that user — the base model permission should be resolved transitively, not checked independently.

### Actual Behavior

**Given:**

- An admin creates a Workspace Model (custom model) with a base_model_id pointing to an underlying base model.
- The admin grants a low-privileged user (or group) access to the custom model — but not to the base model directly.

**Actual (v0.10.x):**
[...]

Issue #26665, opened 2026-07-01
feat: allow admins to select hidden base models in Workspace model editor


michimalek-besitec:
### Check Existing Issues

- [x] I have searched all existing **open AND closed** issues and discussions and found none comparable to my request.

### Verify Feature Scope

- [x] I believe this feature request is appropriately scoped for the Issues section as described above.

### Problem Description

In enterprise setups, admins often want to expose only curated Workspace Models to users/groups, while keeping technical base models out of the chat model selector.

Example:

- Users should see only curated assistants such as `Company Assist` and `Company Assist Advanced`.
- Users should not see technical base models such as provider-specific model IDs, LiteLLM aliases, router models, internal task models, or raw LLM deployments.
- Admins should still be able to use those technical base models when creating Workspace Models.

The current runtime pattern works:

1. Make the base model accessible/public so users can use Workspace Models based on it.
2. Hide the base model so it does not appear in the normal chat model selector.
3. Create a Workspace Model on top of that hidden base model and share it with users/groups.

However, after PR #25668, hidden models are also filtered out from the Workspace Model base-model selector. This means admins cannot select hidden base models when creating or changing Workspace Models.

The current workaround is:

1. Temporarily unhide the base model.
2. Select it as the base model in Workspace > Models.
3. Save the Workspace Model.
4. Hide the base model again.

This works, but it is fragile and awkward for admin-managed enterprise deployments.

### Proposed Solution

Allow admins to select hidden base models in the Workspace Model editor, while keeping hidden models hidden from normal user-facing chat model selectors.

Expected behavior:

- Normal users:
  - Do not see hidden base models in the chat model selector.
  - Only see Workspace Models shared with them/their groups.

- Admins:
  - Can see/select hidden base models in Workspace > Models > Base Model selector.
  - Can create curated Workspace Models on top of hidden base models.
  - Hidden base models remain hidden from the normal chat picker.

Conceptually, the Workspace base-model selector could include hidden models only for admins:

```ts
showModel =
  !model.info?.meta?.hidden
  || model.id === currentBaseModelId
  || user.role === 'admin'
```

This would preserve the intent of PR #25668 for normal users, while supporting admin-managed Workspace Model configuration.

### Temporarily unhide the base model

This is the current workaround:

1. Unhide base model.
2. Select it as Workspace Model base.
3. Save.
4. Hide it again.

This works but is not a good long-term admin workflow.

### Keep base models visible

[...]

PR #28495, opened 2026-08-11
feat: add protected base model access

serge-medvedev:
## Summary

This draft adds a model-only **Protected** visibility mode for base models. Protected does not itself grant direct listing or invocation; direct access remains governed by ordinary user/group grants. Workspace models owned by the same user may use the base through their stored `base_model_id` relationship.

Protected is represented by one narrow access grant:

```json
{
  "principal_type": "model",
  "principal_id": "*",
  "permission": "inherit"
}
```

The grant is accepted only for model resources and is not treated as user `read` access. Ownership is checked at every registered-model edge; missing bases, owner mismatches, ordinary Private bases, and cycles fail closed. Existing explicit user/group grants remain an independent authorization path.

The shared access-control UI exposes Protected only from the model editor, and the Admin Models list distinguishes Protected from Shared.

## Motivation

A curated Workspace model may need to use a technical base model without exposing that base for direct discovery or use. Public/read access is too broad, while `hidden` controls presentation rather than authorization. This draft introduces an explicit, per-base mediated-access policy instead of a global bypass.

## Related issues and prior work

- [#23948](https://github.com/open-webui/open-webui/issues/23948) is the direct request for Workspace-model access without base-model access. It was closed after a hidden custom-Pipe workaround, not a native mediated-authorization rule.
- [#27498](https://github.com/open-webui/open-webui/issues/27498) reports the same permission coupling. Maintainers rejected implicit inheritance and recommended public-but-hidden bases; Protected instead requires an explicit per-base opt-in.
- [#26905](https://github.com/open-webui/open-webui/pull/26905) established fail-closed handling for unregistered base models. This draft preserves that boundary and permits inheritance only through registered models carrying the exact grant above.
- [#25668](https://github.com/open-webui/open-webui/pull/25668) and [#26665](https://github.com/open-webui/open-webui/issues/26665) address hidden-base presentation and selection. Protected complements that work with an authorization rule rather than another visibility workaround.

## Verification

- Focused backend tests: `7 passed`
- Frontend Vitest: `8 passed` across 2 test files
- Ruff formatting and focused lint checks
- Prettier checks for changed Svelte files
- No locale catalog changes
- Production frontend build against the current `dev` base
- Isolated live API acceptance of the identical patch: 10 scenarios covering listing, direct denial, same-owner traversal, revocation, different-owner denial, Private-base denial, cycles, and persisted grant shape
- Browser verification of the identical patch: model-only selector, Protected badge, save/reopen persistence, and absence from non-model access dialogs

## Draft scope

This is intentionally a focused PoC for policy and UI feedback. It does not generalize model principals, add cross-owner delegation, or add write-time relationship validation. The two new UI strings remain literal English until the terminology is accepted, avoiding broad locale-catalog churn during design review.

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/access_grants.py
+++ b/backend/open_webui/models/access_grants.py
@@ -13,7 +13,9 @@ log = logging.getLogger(__name__)
 
 PRINCIPAL_TYPE_ANYONE = 'anyone'
 PRINCIPAL_TYPE_GROUP = 'group'
+PRINCIPAL_TYPE_MODEL = 'model'
 PRINCIPAL_TYPE_USER = 'user'
+PERMISSION_INHERIT = 'inherit'
 WILDCARD_PRINCIPAL_ID = '*'
 
 
@@ -147,7 +149,7 @@ def access_control_to_grants(
     return grants
 
 
-def normalize_access_grants(access_grants: Optional[list]) -> list[dict]:
+def normalize_access_grants(access_grants: Optional[list], resource_type: str | None = None) -> list[dict]:
     """
     Normalize direct access_grants payloads from API forms.
 
@@ -168,11 +170,16 @@ def normalize_access_grants(access_grants: Optional[list]) -> list[dict]:
         principal_id = grant.get('principal_id')
         permission = grant.get('permission')
 
-        if principal_type not in (PRINCIPAL_TYPE_USER, PRINCIPAL_TYPE_GROUP, PRINCIPAL_TYPE_ANYONE):
-            continue
-        if permission not in ('read', 'write'):
-            continue
         if not isinstance(principal_id, str) or not principal_id:
+            continue
+        if principal_type == PRINCIPAL_TYPE_MODEL:
+            if not (
+                resource_type == 'model' and principal_id == WILDCARD_PRINCIPAL_ID and permission == PERMISSION_INHERIT
+            ):
+                continue
+        elif principal_type not in (PRINCIPAL_TYPE_USER, PRINCIPAL_TYPE_GROUP, PRINCIPAL_TYPE_ANYONE):
+            continue
+        elif permission not in ('read', 'write'):
             continue
         if principal_type == PRINCIPAL_TYPE_ANYONE and (principal_id != WILDCARD_PRINCIPAL_ID or permission != 'read'):
             continue
@@ -186,6 +193,15 @@ def normalize_access_grants(access_grants: Optional[list]) -> list[dict]:
         }
 
     return list(deduped.values())
+
+
+def has_model_inherit_grant(access_grants: Optional[list]) -> bool:
+    return any(
+        grant['principal_type'] == PRINCIPAL_TYPE_MODEL
+        and grant['principal_id'] == WILDCARD_PRINCIPAL_ID
+        and grant['permission'] == PERMISSION_INHERIT
+        for grant in normalize_access_grants(access_grants, resource_type='model')
+    )
 
 
 def has_public_read_access_grant(access_grants: Optional[list]) -> bool:
@@ -458,7 +474,7 @@ class AccessGrantsTable:
                 )
             )
 
-            normalized_grants = normalize_access_grants(access_grants)
+            normalized_grants = normalize_access_grants(access_grants, resource_type=resource_type)
 
             results = []
             for grant_dict in normalized_grants:
--- a/backend/open_webui/utils/access_control/__init__.py
+++ b/backend/open_webui/utils/access_control/__init__.py
@@ -307,21 +307,21 @@ async def has_base_model_access(
     db=None,
 ) -> bool:
     """
-    Walk the ``base_model_id`` chain and verify the caller has read access
-    at every hop.
+    Walk the ``base_model_id`` chain and require direct read access or an
+    explicit same-owner inheritance grant at every registered hop.
 
-    A base model without a ``model`` table row is admin-only, matching how
-    unregistered models are treated for direct use (``get_filtered_models``
-    hides them from non-admins and ``check_model_access`` rejects them), so
-    a shared preset cannot be used to reach a base model the caller could
-    not use directly.  Returns ``False`` the moment any hop denies access.
+    A base model without a ``model`` table row remains admin-only, matching
+    direct-use behavior. Returns ``False`` as soon as any hop denies access.
     """
-    from open_webui.models.access_grants import AccessGrants
+    from open_webui.models.access_grants import AccessGrants, has_model_inherit_grant
     from open_webui.models.models import Models
 
-    base_model_id = getattr(model_info, 'base_model_id', None)
-    seen = {model_info.id}
-    while base_model_id and base_model_id not in seen:
+    parent_model_info = model_info
+    base_model_id = getattr(parent_model_info, 'base_model_id', None)
+    seen = {parent_model_info.id}
+    while base_model_id:
+        if base_model_id in seen:
+            return False
         seen.add(base_model_id)
         base_model_info = await Models.get_model_by_id(base_model_id, db=db)
         if base_model_info is None:
@@ -336,8 +336,13 @@ async def has_base_model_access(
                 user_group_ids=user_group_ids,
                 db=db,
             )
+            or (
+                has_model_inherit_grant(base_model_info.access_grants)
+                and parent_model_info.user_id == base_model_info.user_id
+            )
         ):
             return False
+        parent_model_info = base_model_info
         base_model_id = getattr(base_model_info, 'base_model_id', None)
     return True
 
--- a/src/lib/components/admin/Settings/Models.svelte
+++ b/src/lib/components/admin/Settings/Models.svelte
@@ -110,11 +110,21 @@
 		);
 	};
 
-	const isSharedModel = (model) => (model?.access_grants ?? []).length > 0 && !isPublicModel(model);
+	const isProtectedModel = (model) => {
+		return (model?.access_grants ?? []).some(
+			(g) => g.principal_type === 'model' && g.principal_id === '*' && g.permission === 'inherit'
+		);
+	};
+
+	const isSharedModel = (model) =>
+		(model?.access_grants ?? []).length > 0 && !isPublicModel(model) && !isProtectedModel(model);
 
 	const modelAccessLabel = (model) => {
 		if (isPublicModel(model)) {
 			return $i18n.t('Public');
+		}
+		if (isProtectedModel(model)) {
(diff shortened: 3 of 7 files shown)


Classic298:
> This draft adds a model-only Protected visibility mode for base models. Protected does not itself grant direct listing or invocation; direct access remains governed by ordinary user/group grants.

So the model is hidden from all lists, and usable only by those who have direct group based access to it.

I don't get it.

Thats exactly already possible today

Just hide the model and grant permissions to that model only to certain users

done

Classic298:
dont commit test files!

serge-medvedev:
@Classic298 

> So the model is hidden from all lists, and usable only by those who have direct group based access to it.
> 
> I don't get it.
> 
> Thats exactly already possible today
> 
> Just hide the model and grant permissions to that model only to certain users
> 
> done

Not quite. Hidden models are still accessible through API which is unacceptable in some cases.

Classic298:
not if you put group based access control on your hidden models @serge-medvedev 

so again how is a hidden model with group access different to this PR

serge-medvedev:
> not if you put group based access control on your hidden models @serge-medvedev
> 
> so again how is a hidden model with group access different to this PR

Hear me out: I target cases, where users are allowed to access a Workspace model but are disallowed to access its base model - both through UI and API.

Classic298:
@serge-medvedev ok. But then your PR description is misleading. It says 

> Ownership is checked at every registered-model edge; missing bases, owner mismatches, ordinary Private bases, and cycles fail closed. Existing explicit user/group grants remain an independent authorization path.

Which means if someone has access to the workspace model but not to the underlying "protected" base model, then the workspace model will still fail for them.

serge-medvedev:
> @serge-medvedev ok. But then your PR description is misleading. It says
> 
> > Ownership is checked at every registered-model edge; missing bases, owner mismatches, ordinary Private bases, and cycles fail closed. Existing explicit user/group grants remain an independent authorization path.
> 
> Which means if someone has access to the workspace model but not to the underlying "protected" base model, then the workspace model will still fail for them.

Ownership doesn't mean ACL. If a Workspace model and its "protected" base model share an owner, all Workspace model's users will be OK.

Classic298:
so basically this PR bypasses ACL entirely and says "if the owner of the workspace model is part of the group that has access to the "protected"-type base model per the base model access control, then everyone using the workspace model will be able to access the base model"?

So basically it's ownership bound and the access control is bypassed is what you're saying?

serge-medvedev:
> so basically this PR bypasses ACL entirely and says "if the owner of the workspace model is part of the group that has access to the "protected"-type base model per the base model access control, then everyone using the workspace model will be able to access the base model"?

Both models must share an owner. There's no bypass. The owner decides who can access both the base and the wrapper.

Classic298:
my problems with this method is:

1) it's basically implicit access grants. The access control lists get bypassed in a sense that "if owner has access to base model then allow anyways" which is not visible from the UI. Nothing in the UI shows that.

Only admins configure base models so only they can see which model is "protected" and normal users who are model creators in a company setting for example, cannot see it. And even if they can, e.g. via a flag or info highlight, it still is implicit access grants because you need to understand that some base models behave differently than others. You need to KNOW it. The UI doesn't show it to you and even if it shows that there is a difference, you need to have the users understand what the difference is and how it behaves

2) owner bound validation will cause issues

if the owner leaves the company, then the owner of the workspace model is an orphaned account.

if the owner account gets deleted, then there is no longer an owner to that workspace model so there is nothing to validate against and access breaks.

and: today there is no official method of transferring ownership. Only via manual database operations. So if some day you want to transfer ownership to keep alive a workspace model, with today's setup and this PR - you can't. Not a reason against it, separate ownership transfer could be implemented later on, but the other 2 risks re: owners still remain.

3) A dedicated field or enum on the model would be safer.

You currently implement it via wildcard which is brittle and not so nice. Should be a dedicated field if anything

> Both models must share an owner. There's no bypass. The owner decides who can access both the base and the wrapper.

Also you will need to explain this to me because base models do not have an owner. So what do you mean by both models must share the owner and that the owner decides who can access the base and the wrapper? The owner of a workspace model does NOT decide who can access the base model. Only an admin decides that.

---

I like that you are pushing for this, but this should really be discussed first how to implement it. Anything with implicit access controls we prefer not to have.

serge-medvedev:
@Classic298

A dedicated field or enum would change how the opt-in is stored; it would not change the derived authorization behavior or resolve the ownership questions. The current tuple is not treated as a generic wildcard grant: only the exact `model:* → inherit` tuple is accepted for model resources, and it is evaluated only while traversing configured `base_model_id` edges.

The policy decision is separate: should a base authorize same-owner Workspace models, explicitly selected model IDs, or only callers who already have direct base read access? Once that is established, the representation can be evaluated against the existing access-control model. A dedicated field or enum is not inherently safer than a narrowly validated grant.

Classic298:
not saying the enum is safer but its cleaner and less brittle

> should a base authorize same-owner Workspace models, explicitly selected model IDs, or only callers who already have direct base read access?

I am not exactly sure what you are asking, but a workspace model should validate that the caller of that workspace model has **explicit** access to the base model.

That is it.

Btw it would also really help me if you could explain to me why this is needed. In the discussion here https://github.com/open-webui/open-webui/discussions/24070 basically all named scenarios have a proper way of realizing them today, with the worst downside for one particular scenario being that you need to create two workspace models.

serge-medvedev:
> but a workspace model should validate that the caller of that workspace model has **explicit** access to the base model.
> 
> That is it.

This is how it works now and this is why it breaks my use case.

My point is that we need a way for base models to power Workspace models w/o being accessible directly through both UI and API.

Classic298:
what WOULD be explicit access grant is if a workspace model creator has access to a PRIVATE base model.
That private base model is shared with a group. That workspace model creator is part of the group

and when creating the workspace model, the workspace model creator would set the access control to "public" and also tick a checkbox which says "share base model with everyone"

my grime with THAT solution, though explicit, would be that maybe you do not want everyone to have the ability to share base models as workspace models to other users. so that would potentially require additional per-user/-group permission toggles.

Doesnt make the whole operation easier but it does make it explicit.

> My point is that we need a way for base models to power Workspace models w/o being accessible directly through both UI and API.

Okay that is what you want to achieve but you do not say why. The UI part is solved through "Hide model". The API part isn't. Fair. Public model + hidden remains accessible on API. But why is that a problem. I need to understand. Please help me understand.

Having access to the workspace model IS having access to the base model.

If it's about "yeah but we have special protective filters/guardrails on the workspace model" why not also apply them to the base model then?
If you say "yeah because Group B should be able to bypass those filters" then have them bypass it.

I need to understand that case where the base model CAN NOT be accessible via API, when that case is.

Perhaps that will help me understand your usecase better and think of a better solution or a different solution OR of a different implementation 👍

serge-medvedev:
> Btw it would also really help me if you could explain to me why this is needed.

My product is a Workspace model; the computational resource is scarce and I don't want the base model to be exposed for arbitrary inference.

Classic298:
> My product is a Workspace model; the computational resource is scarce and I don't want the base model to be exposed for arbitrary inference.

Okay but the workspace model is on the API anyways?

serge-medvedev:
> > My product is a Workspace model; the computational resource is scarce and I don't want the base model to be exposed for arbitrary inference.
> 
> Okay but the workspace model is on the API anyways?

Yes and that's desirable.

Classic298:
Okay but then you do not have any protection for the scarce computational resource?

serge-medvedev:
> Okay but then you do not have any protection for the scarce computational resource?

I'm ready to spend as much as I have as long as it goes through the Workspace model.

Classic298:
@serge-medvedev ok got it. that makes more sense. In that case two last questions to fully understand your case

1) why not use the base model as the customized model, since you can customize the base model the same way with prompt, knowledge, filters, tools just as the workspace model? then there is no base model you need to protect because your custom model IS the base model
2) even if the setup remains as is, you COULD (i know i know, awkward) make a filter that denies all direct invocations of the base model. and ony allows requests if they go through the workspace model. That would also lead to the same result. Obvious downside: requires a small filter (5 lines probably) - but that's less than this PR at 230 lines

Classic298:
matter of fact i think filters of a base model do not even run if the base model is called via a workspace model. in that case all you need is a 2 line filter which just denies ALL requests to the applied model, and you just apply that filter to the base model - let me check

Classic298:
yup verified it

So a filter attached only to the base model runs only when that base model is called directly.

That means a 2 line filter that just says no to every single request, applied to the base model, would solve your case.

serge-medvedev:
> @serge-medvedev ok got it. that makes more sense. In that case two last questions to fully understand your case
> 
>     1. why not use the base model as the customized model, since you can customize the base model the same way with prompt, knowledge, filters, tools just as the workspace model? then there is no base model you need to protect because your custom model IS the base model

That's workable unless one has multiple Workspace models powered by the same base model, as I do. I'd need to have multiple connections to the same model provider (with different prefixes, probably) to customize them separately. Awkward, as you say.

>     2. even if the setup remains as is, you COULD (i know i know, awkward) make a filter that denies all direct invocations of the base model. and ony allows requests if they go through the workspace model. That would also lead to the same result. Obvious downside: requires a small filter (5 lines probably) - but that's less than this PR at 230 lines

That's also workable unless you need the base model to serve direct requests properly for those who granted access to it. In theory, such filter could decide which users should be "rate-limited" and which should not, which is awkward to say the least.

Classic298:
> That's also workable unless you need the base model to serve direct requests properly for those who granted access to it. In theory, such filter could decide which users should be "rate-limited" and which should not, which is awkward to say the least.

Okay so your actual usecase is not "i need to deny api requests to the base model" as you said earlier

your actual usecase is "i need to deny api requests to the base model except some user groups"

in which case, yeah, you'd need a filter that denies all requests to the base model unless the calling API Key belongs to a user which is part of a group e.g. "allow base model access" group - and then that group could always call the base model

so the question is: core feature we need to maintain and changes to the ACL or a ... with this new addition, 10-15 line filter function

I won't make that decision. I see why a native solution is desirable but from a project perspective also not desirable

leaving open for tim. ps: you should really write this down as a proposal into the main discussion so this can be discussed and others can chime in.

Classic298:
Btw in the future i'd love to get your usecase straight away. took us a few rounds of back and forth to get to your exact requirements.

Classic298:
please do not commit test files as i said above already, you havent removed them still - please discuss exact schema first, i am also happy to discuss on that, we can pick it up from then again

Issue #28411, opened 2026-08-10
bug: HEIC uploads: conversion skipped for image/heif and empty file.type, and converted JPEGs re-tagged as image/heic


Podden:
### Bug

HEIC/HEIF image uploads reach the backend in a state the model cannot use. There are two independent defects in `src/lib/components/chat/MessageInput.svelte`:
#### 1. The HEIC check misses most real-world HEIF files

```js
// src/lib/components/chat/MessageInput.svelte:965
reader.readAsDataURL(file['type'] === 'image/heic' ? await convertHeicToJpeg(file) : file);
```
should account for `image/heif`, `image/heic-sequence` / `image/heif-sequence` and an empty `file.type` as well. A robust check would look at the ISO-BMFF brand rather than the MIME string. The `ftyp` box at offset 4 is followed by one of `heic`, `heix`, `hevc`, `hevx`, `heim`, `heis`, `hevm`, `hevs`, `mif1`, `msf1` for HEIF-family stills.

#### 2. A successfully converted image is re-tagged with the original type

```js
// src/lib/components/chat/MessageInput.svelte:958-959
const blob = await (await fetch(imageUrl)).blob();
const compressedFile = new File([blob], file.name, { type: file.type });
```

By this point `imageUrl` is a data URL of the **converted JPEG**, and `blob.type` is `image/jpeg`. But the `File` is constructed with `file.type` — the *original* type. So a correctly converted image is uploaded as `image/heic`.

`blob.type` is already correct here, so `{ type: blob.type }` would fix it.

### Steps to reproduce

1. Upload an iPhone photo (`.heic`) in a normal (non-temporary) chat to any vision-capable model.
2. Observe the stored file's content type: `image/heic`, even though the bytes are JPEG.
3. With a file whose `type` is `image/heif` or empty, the bytes are not even converted.

### Expected

The uploaded file's content type matches its actual bytes, and all HEIF-family stills are converted regardless of how the browser labelled them.

### Actual

The backend receives either JPEG bytes labelled `image/heic`, or unconverted HEIF bytes. Downstream consumers that validate the declared MIME type against a provider's supported list drop the image.

### Environment

- Open WebUI 0.11.0

### Note

Claude Code found this while I was testing images in my custom pipe and I noticed JPG works fine while HEIC resulted in the LLM telling me HEIC is not supported, even when resizing is active in open webui settings. I this is wrong, I'm sorry for bothering you.

raunika-compunnel:
[...]

PR #28510, opened 2026-08-12
fix: detect and convert HEIF-family images regardless of MIME type

raunika-compunnel:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** This PR references an existing [Issue](https://github.com/open-webui/open-webui/issues) — `Closes #28411  `
- [x] **Documentation:** Not applicable — internal bugfix, no user-facing docs affected.
- [x] **Testing:** Manual tests performed locally to verify the fix works correctly. Generated real HEIC and HEIF test files, uploaded both through the chat UI, and confirmed via the stored message payload that `content_type` correctly shows `image/jpeg` after conversion (previously `application/octet-stream`/unconverted). Screenshots included below.
- [x] **No Unchecked AI Code:** This PR is human-written, self-reviewed, and manually tested by me.
- [x] **Architecture:** No new settings introduced; fix reuses the existing image-conversion pipeline and introduces no UX changes.
- [x] **Git Hygiene:** The PR is atomic (one logical change, single file), rebased on `dev`, and contains no unrelated commits.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

Fixes HEIC/HEIF image uploads reaching the backend in a state the model cannot use. Three related defects in `MessageInput.svelte`'s upload handling caused HEIC/HEIF files to either be skipped entirely (uploaded unconverted) or, when converted, mislabeled with the original pre-conversion MIME type.

### Added

- N/A

### Changed

- N/A

### Deprecated

- N/A

### Removed

- N/A

### Fixed

- The outer image-detection gate (`file['type'].startsWith('image/')`) previously skipped HEIC/HEIF files whenever the browser reported an empty or non-`image/*` MIME type for them, which is common since HEIC/HEIF are not universally registered MIME types. Such files fell through to the plain upload path completely unconverted.
- The inline HEIC conversion check only matched the exact string `image/heic`, missing `image/heif`, `image/heic-sequence`, and `image/heif-sequence` variants.
- Successfully converted images were re-tagged with the original (pre-conversion) `file.type` instead of the converted blob's actual type, so a correctly converted JPEG still appeared as `image/heic` to downstream consumers that validate declared MIME type.

### Security

- N/A

### Breaking Changes

- N/A

---

### Additional Information

Introduced a shared `isHeifFamily()` helper that checks MIME type variants (`image/heic`, `image/heif`, `image/heic-sequence`, `image/heif-sequence`) with a filename-extension fallback (`.heic`/`.heif`) for files with an empty or generic (`application/octet-stream`) type. This helper is used both at the outer image-detection gate and the inline conversion check, replacing the narrow literal-string comparison. The `File` constructor at the end of the conversion path now uses `blob.type` (the converted type) instead of `file.type` (the original type).

### Screenshots or Videos

- [Attach your before/after screenshots here: the stored file payload showing `content_type: "application/octet-stream"` before the fix, and `content_type: "image/jpeg"` after the fix, for both `.heic` and `.heif` uploads]

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/MessageInput.svelte
+++ b/src/lib/components/chat/MessageInput.svelte
@@ -110,6 +110,24 @@
 	import QueuedMessageItem from './MessageInput/QueuedMessageItem.svelte';
 	import TaskList from './Messages/ResponseMessage/TaskList.svelte';
 
+	const isHeifFamily = (file) => {
+	const type = (file?.type || '').toLowerCase();
+	if (
+		type === 'image/heic' ||
+		type === 'image/heif' ||
+		type === 'image/heic-sequence' ||
+		type === 'image/heif-sequence'
+	) {
+		return true;
+	}
+	// Fallback: some browsers/OSes send an empty or generic type for HEIC files.
+	// Fall back to checking the filename extension in that case.
+	if (!type || type === 'application/octet-stream') {
+		return /\.(heic|heif)$/i.test(file?.name || '');
+	}
+	return false;
+    };
+
 	const i18n = getContext('i18n');
 
 	export let onUpload: Function = (e) => {};
@@ -972,7 +990,7 @@
 				return;
 			}
 
-			if (file['type'].startsWith('image/')) {
+			if (file['type'].startsWith('image/') || isHeifFamily(file))  {
 				if (visionCapableModels.length === 0) {
 					toast.error($i18n.t('Selected model(s) do not support image inputs'));
 					return;
@@ -1013,6 +1031,7 @@
 					}
 					return imageUrl;
 				};
+				
 
 				let reader = new FileReader();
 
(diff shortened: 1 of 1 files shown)


amerob (commented):
Broadening this beyond the exact `image/heic` check is the right call — Safari and several Android pickers hand you an empty type for HEIC, so the old equality test missed the common case.

Switching the re-wrapped file to `blob.type` is also correct and easy to overlook:

```js
const compressedFile = new File([blob], file.name, { type: blob.type });
```

after `convertHeicToJpeg` the blob really is `image/jpeg`, where `file.type` still said HEIC.

That change leaves one loose end, though: `file.name` still ends in `.heic`. So you upload a file named `photo.heic` whose bytes and MIME type are JPEG. Anything downstream that trusts the extension over the content type — storage, thumbnailers, a later re-read — will disagree with the `image/jpeg` you just set. Worth renaming at the same time:

```js
const name = file.name.replace(/\.(heic|heif)(-sequence)?$/i, '.jpg');
const compressedFile = new File([blob], name, { type: blob.type });
```

Two smaller things:

The guard reads `file['type'].startsWith('image/') || isHeifFamily(file)`, but `isHeifFamily` defends with `file?.type || ''`. If `file.type` really can be absent, the unguarded `.startsWith` immediately to its left throws before the fallback ever runs. In practice `File.type` is always a string so this won't fire, but the two halves of the same expression disagree about that — I'd either drop the `?.` in the helper or flip the order.

You've added `image/heic-sequence` and `image/heif-sequence`, which are multi-image sequences rather than single stills. Does `convertHeicToJpeg` handle those, or does it choke / silently return only the first frame? If it's the latter, it may be better to leave the sequence types out than to accept them and produce something unexpected.

Lastly the diff won't survive CI as-is — the body of `isHeifFamily` is indented one level short of its declaration, the closing `};` uses four spaces where the file uses tabs, and there's a whitespace-only line added after the `return imageUrl;` block. Running prettier over it should clear all three.

Classic298:
issue reference?

PR #28519, opened 2026-08-12
fix: apply custom model fallback before the base-model access check

Classic298:
With ENABLE_CUSTOM_MODEL_FALLBACK enabled, chatting with a custom workspace model whose base model was removed still failed with "Model not found" for non-admin users. The access check ran before the fallback resolution and walked the preset's base_model_id chain; has_base_model_access treats a base model without a model-table row as admin-only, so regular users were rejected before the fallback to the default (Selected) model could ever apply. Admins bypass the access check by default, which is why the fallback only appeared to work for them.

Resolve the fallback model first, and when one applies, skip the missing base-model hop on the requested preset and enforce access on the fallback model that will actually serve the chat instead. check_model_access gains a check_base_model flag for this; all other callers keep the default chain enforcement.

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1081,6 +1081,7 @@ async def chat_completion(
     metadata = {}
     try:
         model_info = None
+        fallback_model = None
         if not model_item.get('direct', False):
             if model_id not in request.app.state.MODELS:
                 raise Exception('Model not found')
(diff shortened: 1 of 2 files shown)


Classic298:
@tjbck ready for review. With `ENABLE_CUSTOM_MODEL_FALLBACK` enabled, chatting with a custom model whose base model was removed still fails with "Model not found" for non-admin users, so the fallback only ever appeared to work for admins.

Issue #28394, opened 2026-08-10
fix: OpenAPI tool results skip every second inline data: URI


noron12234:
### Bug

When an OpenAPI tool returns a list containing two or more adjacent inline
`data:` URIs, only the first one is turned into an attachment. The second
survives into the text payload and is `json.dumps()`'d into the model's context
as a raw base64 blob.

### Cause

`process_tool_result()` in `backend/open_webui/utils/middleware.py` removes
items from the list it is iterating over:

```python
else:  # OpenAPI
    for item in tool_result:
        if isinstance(item, str) and item.startswith("data:"):
            tool_result_files.append({"type": "data", "content": item})
            tool_result.remove(item)
```

`list.remove()` shifts every later element down by one, so the iterator skips
the item immediately after each extraction.

### Reproduction

```python
tool_result = ["data:image/png;base64,AAA", "data:image/png;base64,BBB", "plain text"]

# current behaviour
files   = ["data:image/png;base64,AAA"]
result  = ["data:image/png;base64,BBB", "plain text"]   # BBB leaked into the text

# expected
files   = ["data:image/png;base64,AAA", "data:image/png;base64,BBB"]
result  = ["plain text"]
```

### Impact

The missed image is not rendered as an attachment, and a full base64 payload is
injected into the LLM context instead — wasting a large amount of the context
window and producing garbage in the transcript.

### Fix

Collect the keepers into a new list rather than mutating during iteration.
Happy to open a PR.

hariomlohardev:
I’d like to work on this issue. I can reproduce the mutation-while-iterating behavior in `process_tool_result()` and will submit a PR with the fix and regression tests. Is it okay to assign this issue to me?

Classic298:
Fixed on dev, as part of the base64-image tool-result work. Not in a release yet.

PR #28535, opened 2026-08-13
fix: preserve adjacent inline data: URIs in OpenAPI tool results

jaideeppyne:
Closes #28394

### Description

`process_tool_result()` in `backend/open_webui/utils/middleware.py` iterated over `tool_result` while calling `tool_result.remove(item)` inside the loop (OpenAPI branch). `list.remove()` shifts every later element down by one, so the iterator skipped the item immediately after each extracted `data:` URI. When an OpenAPI tool returned two or more adjacent inline `data:` URIs, only the first became an attachment; the rest leaked into the text payload and were `json.dumps()`'d into the model context as raw base64 blobs.

The fix collects the non-`data:` items into a new list and assigns it back in place instead of mutating the list during iteration.

### Testing

Verified the corrected loop against adjacent, interleaved, all-data, and no-data inputs; the two cases the old code got wrong ([data:A, data:B, text] and [data:1, data:2, data:3]) now keep every data: URI as an attachment, and non-adjacent/no-data cases are unchanged. `backend/open_webui/utils/middleware.py` compiles.

### Fixed
- Preserve every inline `data:` URI in OpenAPI tool results. Previously, when a tool returned two or more adjacent `data:` URIs, every second one leaked into the text payload instead of becoming an attachment.

- [ x ] By submitting this pull request, I confirm that I have read and fully agree to the [Contributor License Agreement (CLA)](https://github.com/open-webui/open-webui/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT), and I am providing my contributions under its terms.

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -1049,6 +1049,7 @@ async def process_tool_result(
                             tool_response.append(resource.get('uri'))
             tool_result = tool_response[0] if len(tool_response) == 1 else tool_response
         else:  # OpenAPI
+            kept_items = []
             for item in tool_result:
                 if isinstance(item, str) and item.startswith('data:'):
                     tool_result_files.append(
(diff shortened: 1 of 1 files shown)


Classic298:

Issue #27750, opened 2026-07-30
issue: Responses API conversion omits strict for loose function tools, materializing optional arguments


favor2432:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.11.0-slim

### Operating System

Linux

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

When Open WebUI converts a loose Chat Completions/MCP-style function tool to an OpenAI Responses API tool, optional JSON Schema properties should retain omission semantics.

If the original tool does not specify strict mode, the converted Responses tool should explicitly use strict:false.

### Actual Behavior

convert_to_responses_payload() only copies strict when it already exists:

if "strict" in func:
    converted_tool["strict"] = func["strict"]

MCP and OpenAPI-derived tools commonly provide loose schemas without a tool-level strict field. Open WebUI consequently sends them to /v1/responses with strict omitted.

For schemas containing optional filters, Responses can then materialize unused properties as empty strings, false, zero, or empty arrays. It can also populate mutually exclusive properties together. Open WebUI forwards those generated arguments to the tool unchanged.

Example unexpected arguments:

{
  "term": "AI safety",
  "limit": 10,
  "offset": 0,
  "title": "",
  "year": 2020,
  "date_from": "2020",
[...]

PR #28543, opened 2026-08-13
fix(openai): preserve loose tool schemas in Responses conversion

seanxuu:
Fixes #27750

## What changed
- Set `strict: false` when converting a Chat Completions function tool that does not specify strictness.
- Preserve explicit `strict: true` and `strict: false` values.
- Leave already-native Responses tools unchanged.

This keeps optional MCP/OpenAPI tool parameters optional when the payload is sent to the Responses API. The OpenAI [Responses API reference](https://platform.openai.com/docs/api-reference/responses/create) documents `strict` for function tools and its default behavior.

## Validation
- AST compilation of `backend/open_webui/routers/openai.py`
- Behavior checks for omitted, explicit true/false, and native Responses tools
- `ruff format --check backend/open_webui/routers/openai.py`
- `git diff --check`

### Fixed

- Preserve loose tool schema semantics when converting Chat Completions tools to Responses API tools.

### DIFF
--- a/backend/open_webui/routers/openai.py
+++ b/backend/open_webui/routers/openai.py
@@ -1403,8 +1403,12 @@ def convert_to_responses_payload(payload: dict) -> dict:
                     # Responses tools default to strict mode when the field is
                     # omitted. Preserve the loose-schema semantics of Chat,
                     # MCP, and OpenAPI function tools unless strictness was
-                    # explicitly requested by the caller.
-                    converted_tool['strict'] = func.get('strict', False)
+                    # explicitly requested by the caller. Keep the old wire
+                    # format for tools without a parameters schema.
+                    if 'strict' in func:
+                        converted_tool['strict'] = func['strict']
+                    elif 'parameters' in func:
+                        converted_tool['strict'] = False
                 converted_tools.append(converted_tool)
             else:
                 # Already in correct format or unknown format, pass through
--- a/backend/open_webui/routers/openai.py
+++ b/backend/open_webui/routers/openai.py
@@ -1400,8 +1400,11 @@ def convert_to_responses_payload(payload: dict) -> dict:
                         converted_tool['description'] = func['description']
                     if 'parameters' in func:
                         converted_tool['parameters'] = func['parameters']
-                    if 'strict' in func:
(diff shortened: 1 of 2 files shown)


gaby:
@seanxuu Probably worth linking to openai docs where it says this?

seanxuu:
@gaby Thanks for the suggestion — I added the [OpenAI Responses API reference](https://platform.openai.com/docs/api-reference/responses/create) to the PR description, which documents the `strict` field and its default behavior.

amerob (commented):
The underlying diagnosis is right: the Responses API defaults function tools to strict, Chat Completions doesn't, so a tool that round-tripped fine through `/chat/completions` starts rejecting its own schema once converted. Tools coming from MCP and OpenAPI are exactly the ones with schemas that won't satisfy strict mode (missing `additionalProperties: false`, optional fields absent from `required`), so this is a real breakage to fix.

One thing I'd think about before merging: the change moves `strict` from *sometimes absent* to *always present*.

```python
converted_tool['strict'] = func.get('strict', False)
```

`convert_to_responses_payload` runs against whatever the user has pointed Open-WebUI at, and self-hosted Responses implementations are a lot patchier than OpenAI's. A shim that doesn't know `strict` and validates its request body strictly used to receive a payload with no such key and was fine; now it gets `strict: false` on every tool and may 400. That's a regression that would only show up for people on non-OpenAI backends — which is a large slice of this project's users.

Preserving the fix while staying compatible is a one-liner:

```python
if 'strict' in func:
    converted_tool['strict'] = func['strict']
elif 'parameters' in func:
    converted_tool['strict'] = False
```

so you only assert loose-schema semantics when there's actually a schema to be loose about, and tools without parameters keep the old wire format. Not insisting on that shape specifically — mainly flagging that "always emit the field" is the part with blast radius beyond the bug being fixed.

Also worth a test. `convert_to_responses_payload` is pure dict-in/dict-out, so pinning the three cases (explicit `strict: true` preserved, explicit `false` preserved, omitted becomes `false`) is cheap and would catch a future refactor quietly restoring the strict default.

Classic298:

Issue #28568, opened 2026-08-14
issue: missing Redis key logs a JSON TypeError and an empty cache refetches on every request


FedorenkoKirill:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.11.0

### Operating System

Ubuntu 22.04

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

A missing Redis key is a normal cache miss: the specs get rebuilt and cached, with nothing logged at `ERROR`. A cache that legitimately holds an empty list is a cache **hit**, so a request should not go back to the database and rewrite Redis to rediscover that the list is still empty.

### Actual Behavior

`utils/tools.py` reads the cache like this (identical code in `get_tool_servers` and `get_terminal_servers`):

```python
tool_servers = JSONCodec.loads(await request.app.state.redis.get(f'{REDIS_KEY_PREFIX}:tool_servers'))
```

1. **Missing key is decoded instead of detected.** `redis.get` returns `None`, `loads(None)` raises `TypeError: the JSON object must be str, bytes or bytearray, not NoneType`, and the handler logs it at `ERROR`:

```
ERROR ... Error fetching tool_servers from Redis: the JSON object must be str, bytes or bytearray, not NoneType
```

[...]

PR #28573, opened 2026-08-14
fix: missing Redis key logs a JSON TypeError and an empty cache refetches on every request

laxanarana1513:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes [Issue](https://github.com/open-webui/open-webui/issues/28568).
- [x] **Target branch:** `dev` branch.
- [x] **Changelog:** A changelog entry following [Keep a Changelog] is included at the bottom.
- [x] **Documentation:** No documentation change, as it is just a backend REDIS cache improvement fix.
- [x] **Dependencies:** None added or changed.
- [x] **Testing:** Live testing was done. I ran the Docker image. Opened the OpenWebUI on localhost. After the fix, ran the `/tools` & `/terminals` endpoint. Checked the logs for no error & also checked `redis-cli MONITOR`.
- [x] **Architecture:** No new setting incorporated.
- [x] **Git Hygiene:** One commit on top of current `dev`, scoped to this bug.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

Improve Redis caching for tool and terminal server specs so a missing key is handled as a normal cache miss, and a cached empty list is treated as a valid hit.

### Added

- N/A

### Changed

- `get_tool_servers` in `backend/open_webui/utils/tools.py`
- `get_terminal_servers` in `backend/open_webui/utils/tools.py`
- Cache read flow now checks for `None` before JSON decode
- Valid cached values (including `[]`) are returned directly without rebuilding

### Deprecated

- N/A

### Removed

- N/A

### Fixed

- ERROR log noise on cold cache: `open_webui.utils.tools:get_tool_servers:1160 - Error fetching tool_servers from Redis: the JSON object must be str, bytes or bytearray, not NoneType` when Redis key is missing
- Repeated DB reads and Redis `SETs` when the cached value is an empty list
- Decode failures and cache misses no longer look the same in logs

### Security

- N/A

### Breaking Changes

- **BREAKING CHANGE**: None

---

### Additional Information

- Related: Discussion #28572 

### Screenshots or Videos

- N/A

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/tools.py
+++ b/backend/open_webui/utils/tools.py
@@ -1166,18 +1166,17 @@ async def set_tool_servers(request: Request):
 
 async def get_tool_servers(request: Request):
     try:
-        tool_servers = []
         if request.app.state.redis is not None:
-            try:
-                tool_servers = JSONCodec.loads(await request.app.state.redis.get(f'{REDIS_KEY_PREFIX}:tool_servers'))
-                request.app.state.TOOL_SERVERS = tool_servers
-            except Exception as e:
-                log.error(f'Error fetching tool_servers from Redis: {e}')
+            cached = await request.app.state.redis.get(f'{REDIS_KEY_PREFIX}:tool_servers')
+            if cached is not None:
+                try:
+                    tool_servers = JSONCodec.loads(cached)
+                    request.app.state.TOOL_SERVERS = tool_servers
+                    return tool_servers
+                except Exception as e:
+                    log.error(f'Error decoding tool_servers from Redis: {e}')
 
-        if not tool_servers:
-            tool_servers = await set_tool_servers(request)
-
-        return tool_servers
+        return await set_tool_servers(request)
     except Exception as e:
         log.error(f'Failed to load tool servers, skipping: {e}')
         return getattr(request.app.state, 'TOOL_SERVERS', None) or []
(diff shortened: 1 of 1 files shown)


FedorenkoKirill:
Thanks for picking this up — the fix covers both halves of what I reported (a missing key is no longer fed to the decoder, and `[]` is now served as a hit), and renaming the log to "Error decoding …" is a nice touch: that wording was half of why the original was confusing.

One behaviour I'd keep, though: the Redis call itself is now outside the `try`, so a *transport* failure (connection reset, timeout — not a decode problem) is no longer contained where it used to be.

In `get_tool_servers`, `await request.app.state.redis.get(...)` raising now propagates to the outer handler, which returns `getattr(request.app.state, 'TOOL_SERVERS', None) or []` and never reaches `set_tool_servers()`. Before this change the same failure was caught, logged, and the specs were rebuilt from the config. So on a worker whose `app.state.TOOL_SERVERS` was never populated — for example when tool servers are added after that worker started, since the startup pre-fetch only runs if connections already exist — a Redis outage now yields an empty tool list instead of a database-backed one, and nothing distinguishes that from "no tools configured".

`get_terminal_servers` is sharper, because it has no outer handler at all: the exception leaves the function, `get_terminal_tools` doesn't catch it either, and `process_chat_payload` turns it into `HTTPException(503, "Terminal unavailable")`. A Redis blip that used to cost one log line and a config re-read now fails the chat request.

Keeping the `get` inside the `try` preserves everything this PR fixes:

```python
if request.app.state.redis is not None:
    try:
        cached = await request.app.state.redis.get(f'{REDIS_KEY_PREFIX}:tool_servers')
        if cached is not None:
            tool_servers = JSONCodec.loads(cached)
            request.app.state.TOOL_SERVERS = tool_servers
            return tool_servers
    except Exception as e:
        log.error(f'Error reading tool_servers from Redis: {e}')

return await set_tool_servers(request)
```

A missing key still falls through silently, [] still returns as a hit, and a Redis failure still degrades to the config read rather than to an empty list or a 503. If you'd rather keep "decoding" in the message, the read and the decode can each get their own except — the point is only that neither should escape.

laxanarana1513:
Yes, that's a good point; I missed it. Updated as suggested; kindly check it.

PR #28583, opened 2026-08-14
fix: download chat file attachments inline with auth

ShraddhanandKhot:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** This PR references [Discussion #28581](https://github.com/open-webui/open-webui/discussions/28581).
- [x] **Target branch:** `dev`.
- [x] **Changelog:** Entry provided below.
- [ ] **Documentation:** Not applicable (frontend behavior fix; no new settings).
- [x] **Dependencies:** None.
- [x] **Testing:** Manual — see below.
- [x] **No Unchecked AI Code:** Human-reviewed and manually tested.
- [x] **Self-Review:** Performed.
- [x] **Architecture:** No new settings; local UI logic only.
- [x] **Git Hygiene:** One logical change, single commit.

## Summary

Clicking a file attachment tile in chat currently either opens a preview
modal (for plain file refs) or a new tab via `window.open` to
`/files/{id}/content`. The JWT is stored in `localStorage` and sent as an
`Authorization` header on fetch — it is **not** available as a cookie on
plain navigation (the `token` cookie is only set in the OAuth redirect
flow), so the new-tab request returns **401** for regular logins and the
file never downloads.

## Changes

`src/lib/components/common/FileItem.svelte`:

1. **Skip the preview modal for plain file refs** — the `item?.type === 'file'`
   clause is removed from the modal condition, so a file without embedded
   content goes straight to the download path instead of opening an empty
   preview modal. Items with embedded content (`item.file.data.content` /
   `item.content`) still open the modal.
2. **Download via authenticated fetch → blob → anchor click** — the file
   bytes are fetched from `/files/{id}/content?attachment=true` with the
   `Authorization: Bearer` header (the same pattern as `downloadDatabase` in
   `src/lib/apis/utils/index.ts`), then saved with the original filename via
   a blob URL. `?attachment=true` forces `Content-Disposition: attachment`
   for all content types (including PDFs, which otherwise serve inline).
3. **Graceful fallback** — on any fetch failure the previous `window.open`
   behavior is preserved.

## Behavior preserved

- External `http(s)` file URLs: unchanged (`window.open`).
- Embedded-content items (images etc.): modal/preview unchanged.
- Message input-box attachment tiles: unchanged (they pass `modal=true`).

## Testing

Manual: click a file attachment in a chat → the file downloads directly to
disk with its original filename (regular login, no token cookie present).
Before this change the same click either opened an empty modal or a 401
page in a new tab.

# Changelog Entry

### Description

Chat file attachments now download directly when clicked, instead of opening
an empty preview modal or a 401 page in a new tab.

### Fixed

- Chat file attachment tiles now download the file inline using the
  authenticated fetch → blob → anchor-click pattern; plain file refs no
  longer open an empty preview modal.

---

### DIFF
--- a/src/lib/components/common/FileItem.svelte
+++ b/src/lib/components/common/FileItem.svelte
@@ -57,7 +57,7 @@
 		: 'gap-1 rounded-2xl p-1.5'} text-left"
 	type="button"
 	on:click={async () => {
-		if (item?.file?.data?.content || item?.type === 'file' || item?.content || modal) {
+		if (item?.file?.data?.content || item?.content || modal) {
 			showModal = !showModal;
 		} else {
 			if (url) {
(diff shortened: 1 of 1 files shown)


Classic298:
testing? video? before/after

ShraddhanandKhot:
> testing? video? before/after

 Before/after video: https://youtu.be/FqG-izData8

    Problem:
    When a file tile appears in chat (uploaded or tool-generated), clicking it opens a preview modal — there is no
    direct download. The modal's download action opens a new tab via window.open, which cannot send the
    Authorization header (the JWT lives in localStorage.token). If the token cookie is absent or expired, that
    new tab gets 401 Not Authenticated.

    Fix (this PR, 1 frontend file, +25/-2, no backend change):
    - Plain file tiles skip the preview modal — click downloads the file directly.
    - Download uses authenticated fetch (Authorization: Bearer ${localStorage.token}) → blob → saved to disk,
    same auth mechanism the rest of the app uses. Works with or without the token cookie.
    - ?attachment=true appended to the URL (already supported by the backend, sets `Content-Disposition:
    attachment`).

    To test without our backend patches:
    Upload any file (paperclip button) in a chat → click the tile. Before = modal, after = download.

     Please re-open this PR for review — the fix is testable with any uploaded file, no backend or tool changes needed.

Issue #28579, opened 2026-08-14
issue: Autoscroll of assistant response not working in latest dev version


rotemdan:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

latest dev version

### Operating System

Windows 11

### Browser (if applicable)

Brave

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

Autoscroll during assistant response

### Actual Behavior

Autoscroll not working on latest dev version

### Steps to Reproduce

1. Run any query.

### Logs & Screenshots

(empty)

### Additional Information

I asked Qwen3.8-Max to try to find the cause. This is what it replied with:

# Analysis by Qwen3.8-Max

The recent breakage in autoscroll for the assistant response was introduced in commit **[`a1579a01ff43cacb357269707d36267ad35e01d6`](https://github.com/open-webui/open-webui/commit/a1579a01ff43cacb357269707d36267ad35e01d6)** (authored on August 14, 2026, at 06:22 UTC, roughly 10 hours ago), which carries the vague message `refac`.

This commit introduced support for the OpenAI Responses API (using events like `response.output_text.delta`) and fundamentally changed how the backend streams these events. Consequently, it broke the autoscroll behavior in the UI due to a missing function call in the new frontend handler.

[...]

PR #28589, opened 2026-08-14
fix: restore autoscroll and completion lifecycle for Responses API streams 

DineshThumma9:
# Changelog Entry

### Description

The recent `refac` commit (`a1579a01`) introduced support for the OpenAI Responses API by adding a new `response:completion` event type and a corresponding `responseCompletionEventHandler` in `Chat.svelte`. However, the new handler was shipped without the autoscroll logic and the stream-completion lifecycle that exist in the legacy `chatCompletionEventHandler`. This caused autoscroll to stop working for any model using the Responses API streaming path, and also prevented those chats from being saved, generating titles, or dispatching TTS events on completion.

### Fixed

- **Autoscroll not working during Responses API streaming** — added `await tick()` and `scheduleResponseScrollToBottom()` at the end of `responseCompletionEventHandler` so the page scrolls on every delta event, matching the behaviour of `chatCompletionEventHandler`.
- **Chat not saved after Responses API response completes** — added the full completion lifecycle to the `response.completed` branch: sets `message.done = true`, dispatches `chat:finish`, calls `chatCompletedHandler()` (chat save, title generation, follow-ups, tags), and calls `processNextInQueue()`.
- **Final scroll missing on response completion** — added `shouldAutoScrollResponse()` + `scrollToBottom()` inside the `response.completed` block.
- **`responseCompletionEventHandler` missing `chatId` parameter** — updated call site to pass `event.chat_id` so the completion handler has the context it needs.

---

### Additional Information

- Fixes #28579
- The broken handler was introduced in commit `a1579a01ff43cacb357269707d36267ad35e01d6` (Aug 14 2026)

### DIFF
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -1179,7 +1179,7 @@
 						}
 					}
 				} else if (type === 'response:completion') {
-					responseCompletionEventHandler(data, message);
+					responseCompletionEventHandler(data, message, event.chat_id);
 				} else if (type === 'chat:completion') {
 					chatCompletionEventHandler(data, message, event.chat_id);
 				} else if (type === 'chat:tasks:cancel') {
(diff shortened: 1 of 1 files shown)


DineshThumma9:
@Classic298  Why was the PR closed? Did I do something wrong?

Classic298:
See comment on issue.

PR #28613, opened 2026-08-15
perf: allow disabling websocket per-message-deflate

Classic298:
With delta streaming most websocket frames are tiny per-token deltas, and per-message-deflate pays zlib work on every outgoing frame per subscriber for near-zero gain there; under heavy streaming that shows up as measurable server CPU. The frames that still benefit are the rare large ones (final message, sources), and even a 100k token message is only a few hundred KB uncompressed, which any network delivers without noticeable delay.

UVICORN_WS_PER_MESSAGE_DEFLATE=false (default true, current behavior) disables the extension in every entry point: open-webui serve and dev, start.sh both invocations, start_windows.bat and dev.sh. Verified against a running instance: with the flag off the server declines the client-offered permessage-deflate extension, with defaults it still negotiates it.

> [!NOTE]

### DIFF
--- a/backend/dev.sh
+++ b/backend/dev.sh
@@ -1,3 +1,3 @@
 export CORS_ALLOW_ORIGIN="http://localhost:5173;http://localhost:8080"
 PORT="${PORT:-8080}"
-uvicorn open_webui.main:app --port $PORT --host 0.0.0.0 --forwarded-allow-ips "${FORWARDED_ALLOW_IPS:-*}" --reload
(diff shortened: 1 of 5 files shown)


joewpb (commented):
## Code review 🤖

Reviewed **open-webui/open-webui#28613** — found a few things:

---

Classic298:
@tjbck ready for review. With delta streaming almost every websocket frame is a tiny per-token delta, and per-message-deflate pays zlib on every outgoing frame per subscriber for near-zero gain, which shows up as server CPU under load. This is opt-in and the default keeps current behaviour, so it is a no-op unless an operator turns it off.

PR #28625, opened 2026-08-15
fix: GitHub and Feishu OAuth login fails on PostgreSQL

Classic298:
GitHub returns its user id as a JSON number and the provider config maps it to the OAuth subject claim, as Feishu does with user_id. Nothing coerced it, so the subject was stored in the user.oauth JSON column as a number and passed to the lookup as an int. On PostgreSQL that lookup is a text comparison, so an int operand raises "operator does not exist: text = integer" and the callback's catch-all turns it into a generic /auth?error= redirect. Login fails every time and the operator gets no useful signal. SQLite matched fine, which is why this went unnoticed.

The subject is now coerced to text where the claim enters, ahead of both the write and the lookup, so the stored and incoming values cannot disagree. The migration normalizes subjects already stored as numbers; without it, existing SQLite deployments would stop matching once the code starts sending strings.

Verified on SQLite and PostgreSQL 16: numeric-subject users resolve on both after the migration, the migration is idempotent, and rows that already hold string subjects are left untouched.

> [!NOTE]

### DIFF
--- a/backend/open_webui/migrations/versions/b7e4c19a2f83_normalize_user_oauth_sub_to_text.py
+++ b/backend/open_webui/migrations/versions/b7e4c19a2f83_normalize_user_oauth_sub_to_text.py
@@ -1,0 +1,59 @@
+"""normalize user oauth sub to text
+
+Revision ID: b7e4c19a2f83
+Revises: e2b7a4c91f05
+Create Date: 2026-08-15 19:40:11.204517
+
+"""
+
+from typing import Sequence, Union
+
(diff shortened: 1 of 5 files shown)


joewpb (commented):
## Code review 🤖

Reviewed **open-webui/open-webui#28625** — found a few things:

---

Classic298:
@tjbck rebased on dev. The migrations across the open PRs are one chain now, and this one is last in it.

Order is #27663, then #28066, then this one.

PR #28632, opened 2026-08-15
refac: sandbox terminal port previews and move their auth into the URL

Classic298:
Port previews now render at an opaque origin: responses under the port proxy carry a Content-Security-Policy sandbox directive alongside no-store, no-referrer and nosniff, so a previewed page is isolated from the application displaying it no matter how it is opened.

Those requests are therefore no longer same-site, and the session cookie does not reach them, so the preview authorises itself with a credential carried in its URL path, which the page's own relative requests inherit. That credential is signed with a separate key so it is only ever accepted here, is scoped to a single server and port, inherits the minting session's expiry and revocation, and is refused for the connection auth types whose upstream credential a sandboxed request cannot carry.

The terminal proxy also stops forwarding Set-Cookie and Clear-Site-Data, which would act on our own origin rather than the terminal server's, and SecurityHeadersMiddleware now lets a route keep security headers it chose for itself instead of overwriting them with the deployment-wide defaults.

One behaviour change worth knowing: the preview's address bar no longer follows navigation inside the page, because reading the framed location needs the origin the sandbox removes.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/terminals.py
+++ b/backend/open_webui/routers/terminals.py
@@ -1,15 +1,20 @@
 """Reverse proxy for admin-configured terminal servers.
 
 Routes:
-  GET  /                         — list terminals the user has access to
-  *    /{server_id}/{path:path}  — proxy request to terminal server
+  GET  /                                                         — list terminals the user has access to
+  POST /{server_id}/port-preview/{port}                          — mint a port preview credential
+  *    /{server_id}/port-preview/{port}/{preview_token}/{path}   — serve a port, sandboxed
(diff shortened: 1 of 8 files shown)


joewpb (commented):
## Code review 🤖

Reviewed **open-webui/open-webui#28632** — found a few things:

---

Classic298:
@tjbck ready for review. Port previews currently render inside the application's own origin. This moves them to an opaque origin behind a CSP sandbox, and because the session cookie no longer reaches them they authorise with a separately signed credential scoped to a single server and port.

PR #28637, opened 2026-08-15
refac: apply the caller metadata filter in vector search

Classic298:
Every vector client now applies the metadata filter passed to search(), combined with the collection or tenant scoping that method already performed, so search() and query() agree on the same filter for a given client.

Filter builders that could express only single-value equality were extended to express set membership, since that is the shape callers send. Metadata keys that land inside a query string rather than a bound parameter are now validated against a conservative pattern, and a filter shape a client cannot fully express raises instead of being applied in part.

> [!NOTE]

### DIFF
--- a/backend/open_webui/retrieval/vector/dbs/elasticsearch.py
+++ b/backend/open_webui/retrieval/vector/dbs/elasticsearch.py
@@ -3,7 +3,7 @@ NOTE: This vector database integration is community-supported and maintained on 
 """
 
 import ssl
-from typing import Optional
+from typing import Any, Optional
 
 from elasticsearch import BadRequestError, Elasticsearch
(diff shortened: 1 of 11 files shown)


joewpb (commented):
## Code review 🤖

Reviewed **open-webui/open-webui#28637** — found a few things:

---

Classic298:
@tjbck ready for review. The metadata filter passed to `search()` is dropped by the vector clients, so `search()` and `query()` disagree about what a caller asked to scope to. Every client applies it now, and a filter shape a client cannot fully express raises instead of being applied in part.

tjbck:
1d6d4e6e6647e1d403438ede7bd9ba20bc4cc8f6

Issue #28665, opened 2026-08-16
issue: OAuth client decryption failures log an empty error, hiding the key mismatch behind them


silentoplayz:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

`dev` @ 31d08d592

### Ollama Version (if applicable)

Not applicable

### Operating System

Linux

### Browser (if applicable)

Not applicable

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**.

### Expected Behavior

When a stored MCP tool server OAuth client blob cannot be decrypted at startup, the log says why (the stored value is not a valid token for the current key) and what to do (register the affected OAuth client again if `WEBUI_SECRET_KEY` or `OAUTH_CLIENT_INFO_ENCRYPTION_KEY` changed).

### Actual Behavior

The backend logs two errors that carry no information about what went wrong:

```
ERROR | open_webui.utils.oauth:decrypt_data:283 - Error decrypting data: 
ERROR | open_webui.main:initialize_runtime_config:617 - Error adding OAuth client for MCP tool server ntn: 
```

Both lines end at the colon. They repeat on every startup, and the affected server's OAuth client is never registered, with nothing to say why or what to do about it.

### Steps to Reproduce

1. Run `dev` @ 31d08d592 from a git clone (`npm run dev` for the frontend, the backend from `backend/`), default configuration, no environment variables beyond the defaults
2. Register an MCP tool server with `oauth_2.1` or `oauth_2.1_static` auth
3. Change `WEBUI_SECRET_KEY` (with no `OAUTH_CLIENT_INFO_ENCRYPTION_KEY` set) and restart the backend
[...]

PR #28666, opened 2026-08-16
fix: say why OAuth client data failed to decrypt instead of logging an empty error

silentoplayz:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** This PR references an existing [Issue](https://github.com/open-webui/open-webui/issues) or active, substantive [Discussion](https://github.com/open-webui/open-webui/discussions) - Fixes #28665
- [x] **Dependencies:** Any new or updated dependencies are explained, tested, and documented. No new dependencies; `InvalidToken` comes from `cryptography.fernet`, already imported in the same file for `Fernet`.
- [x] **Testing:** **Manual** end-to-end tests have been performed to verify the fix/feature works correctly and does not introduce regressions. Verified by executing the shipped module against both failure modes; details below.
- [x] **User-facing changes:** I have confirmed whether this PR changes the UI. It does not; backend log output only.
- [x] **No Unchecked AI Code:** This PR is either human-written or has undergone thorough human review AND manual testing.

# Changelog Entry

### Description

- When a stored MCP OAuth client blob could not be decrypted at startup, the backend logged `Error decrypting data:` and `Error adding OAuth client for MCP tool server <id>:` with nothing after either colon, on every startup, while the server's OAuth client silently failed to register. The exception in that case is `cryptography.fernet.InvalidToken`, whose `str()` is empty by design, so `f'... {e}'` printed nothing for the most common failure.
- `decrypt_data` now catches `InvalidToken` and logs what it means and what to do: the stored value is not a valid token for the current key, and if `WEBUI_SECRET_KEY` or `OAUTH_CLIENT_INFO_ENCRYPTION_KEY` changed the affected OAuth client must be registered again. The generic path names the exception type, matching how `utils/valves.py` already handles the same failure. The caller in `initialize_runtime_config` includes the exception type, without a trailing colon when the exception carries no message.

### Fixed

- OAuth client decryption failures at startup now say why they failed and how to recover, instead of logging an empty error twice.

---

### Additional Information

- The message reads "not a valid token for the current key" rather than "encrypted with a different key" on purpose: executing the code showed that a corrupt or non token string raises the same `InvalidToken` as a wrong key blob, so the wider wording is the accurate one for both.
- The caller line formats as `<id>: InvalidToken` when the exception has no message and `<id>: ValueError: boom` when it does, so neither shape ends in a dangling colon.
- No behavior change beyond logging: `decrypt_data` still re-raises, and the same key round trip still decrypts with nothing logged.

**Manual verification performed:**

1. Reproduced the reported failure on `dev`: an MCP server whose OAuth blob was encrypted under a previous `WEBUI_SECRET_KEY` logged both empty lines on every startup; re-registering the server's OAuth cleared them, confirming the key mismatch as the cause.
2. Verified the fix by importing the shipped `open_webui.utils.oauth` module from this branch in the backend's own interpreter and calling `decrypt_data` on a blob genuinely encrypted under a different key: it logs the new actionable message and raises `InvalidToken`, and the caller expression renders `ntn: InvalidToken`.
3. Same harness on a corrupt non token string: same `InvalidToken` path and message.
4. Same harness on an `encrypt_data` then `decrypt_data` round trip under one key: decrypts correctly with no error logged.

### Screenshots or Videos

Not applicable; backend log output only. Before and after log lines are shown in the description.

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -614,7 +614,11 @@ async def initialize_runtime_config(app: FastAPI):
                         OAuthClientInformationFull(**oauth_client_info),
                     )
                 except Exception as e:
-                    log.error(f'Error adding OAuth client for MCP tool server {server_id}: {e}')
+                    log.error(
+                        'Error adding OAuth client for MCP tool server %s: %s',
+                        server_id,
+                        f'{type(e).__name__}: {e}' if str(e) else type(e).__name__,
+                    )
 
     arena_models = await Config.get('evaluation.arena.models', []) or []
     if any('access_control' in m.get('meta', {}) for m in arena_models):
--- a/backend/open_webui/utils/oauth.py
+++ b/backend/open_webui/utils/oauth.py
@@ -19,7 +19,7 @@ import jwt
 from authlib.integrations.starlette_client import OAuth
 from authlib.oauth2.rfc6749.errors import OAuth2Error
 from authlib.oidc.core import UserInfo
-from cryptography.fernet import Fernet
+from cryptography.fernet import Fernet, InvalidToken
 from fastapi import (
     HTTPException,
     status,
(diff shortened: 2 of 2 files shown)


silentoplayz:
Superseded on dev by 91917b239. It keeps the same empty message fallback and moves the InvalidToken reporting out of `decrypt_data` into the three call sites, so each one names the affected tool server and continues. Nothing outstanding.

Issue #28667, opened 2026-08-16
issue: note edits are lost when an update without a content snapshot cancels the pending save


svenhorvath:
## Important Notes

Reproduced server-side over Socket.IO rather than through the browser, because
the failing step is a single emit and driving it from the UI adds timing noise
without adding information. The client-side trigger is identified by code path
below, and the precondition it depends on (an empty server-side Yjs state for a
note that still has content in Postgres) was independently measured in #28539.

No browser console logs are attached for the same reason — the reproduction
does not involve a browser. Container logs show nothing: the save is never
attempted, so there is no error to log. That silence is part of the problem.

## Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items.
- [x] I am using the latest version of Open WebUI available at the time of testing (v0.11.0), and verified the code path is unchanged on `dev`.

Related but distinct: #28539 (content overwritten with an empty document — the
seeding side of the same area) and #26140 (large paste, note not saved). Neither
describes the cancellation path below.

## Installation Method

Docker Compose, unmodified `ghcr.io/open-webui/open-webui:v0.11.0`, Postgres backend.

## Open WebUI Version

v0.11.0

## Operating System

Debian 12 host, Docker Compose. Reproduces server-side, so it should not be OS-dependent.

## Relevant configuration

- `DATABASE_URL` pointing at Postgres 16
- No Redis: `WEBSOCKET_MANAGER` unset, so `REDIS` is `None` and `YdocManager`
  keeps its update log in process memory (`self._updates = {}`)
- No reverse-proxy tricks in front of the websocket, `ENABLE_WEBSOCKET_SUPPORT` default

## Expected Behavior

Text typed into a note ends up in `note.data.content` in the database. If a save
is superseded, a later one takes its place.

## Actual Behavior

The editor keeps showing the full text while the database holds an older
revision, and nothing catches up. Only typing one more character brings the
rest through. Waiting does not help, and neither does leaving the field or the
page.

### Cause

In `yjs_document_update` (`backend/open_webui/socket/main.py`) the cancellation
of the pending save and the scheduling of the new one sit under different
conditions:

```python
try:
    await stop_item_tasks(REDIS, document_id)   # always runs
except Exception:
    pass

...

if data.get('data'):                            # only sometimes
    await create_task(REDIS, debounced_save(), document_id)
```

[...]

Issue #28539, opened 2026-08-13
issue: note content can be overwritten with an empty document, Yjs seeding is still client-side only and the seeder election ignores write access


AlexiMZh:
## Important Notes

I searched open and closed issues/discussions for "Yjs" + "note" + "empty"/"content" combinations and did not find an existing report describing this exact mechanism. Apologies in advance if this duplicates something I missed.

**This is not related to chat archiving** — an earlier working theory (that archiving a chat wipes a linked note's content) was investigated and disproved with a live test; noting this explicitly so it doesn't send anyone down that path.

## Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items.
- [x] I am using the latest version of Open WebUI available at the time of testing (v0.11.0).

## Installation Method

Docker — custom image built `FROM ghcr.io/open-webui/open-webui:v0.11.0-cuda` (no changes to the notes/Yjs backend code), running under Docker Compose on WSL2.

## Open WebUI Version

v0.11.0

## Operating System

Host: Windows 10 IoT Enterprise LTSC 2024 + WSL2 (Ubuntu-based), Docker Compose. The bug reproduces entirely server-side (Postgres + Socket.IO), so it should not be OS-dependent.

## Expected Behavior

A note whose `note.data.content.md` field contains real text in Postgres should keep that text when opened in the Notes web editor for the first time, and should not be silently wiped by a save that the editor triggers right after loading.

## Actual Behavior

A note created **directly via the backend `Notes.insert_new_note()` path** (i.e. any note whose content was written to `note.data` without ever going through the collaborative Yjs editor — this includes notes created by the built-in `write_note` tool, and any other direct `POST /api/v1/notes/create` caller) has **no prior Yjs update history**. On the very first `ydoc:document:join` for that note, the server reconstructs `Y.Doc()` from an empty update list and returns an **empty** document state to the client — even though the real text is still present in `note.data.content.md` in Postgres at that exact moment.

[...]

PR #28669, opened 2026-08-16
fix: keep the pending note save when an update carries no content snapshot

svenhorvath:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #28667
- [x] **Changelog:** A changelog entry is included at the bottom.
- [ ] **Documentation:** Not applicable — no documented behaviour changes.
- [x] **Testing:** Manually verified against a running instance, with a control run on the unpatched image. Details below.
- [x] **No Unchecked AI Code:** Reviewed by me, and manually tested against a running instance (see below).
- [x] **Architecture:** No new setting and no new behaviour — one condition corrected.
- [x] **Git Hygiene:** Atomic, based on `dev`, no unrelated commits.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

Editing a note can silently lose the most recent keystrokes. The editor keeps showing the full text, but the database holds an older revision, and nothing ever catches up — only typing one more character brings the rest through.

In `yjs_document_update` the cancellation of the pending save and the scheduling of the new one sit under different conditions: `stop_item_tasks()` runs unconditionally, while `create_task()` only runs when the update carries `data`. An update without a content snapshot therefore deletes the pending save without scheduling a replacement.

`SocketIOCollaborationProvider` sends exactly such an update when the server reports an empty document state while the editor already holds content. That happens for notes created outside the collaborative editor (the `write_note` builtin tool, direct `POST /api/v1/notes/create` callers), and on deployments without Redis after a restart, because `YdocManager` keeps its update log in process memory while `note.data.content` survives in the database.

Nothing recovers from it: `NoteEditor.svelte` has no save on unmount (`onDestroy` only detaches listeners), and its REST debounce sends `title`, `data.files` and `access_grants`, never `data.content`. The Yjs path is the only writer of note text.

This change moves the cancellation under the same condition as the scheduling, so a pending save is only dropped when a new one takes its place.

### Fixed

- Note edits are no longer lost when a `ydoc:document:update` without a content snapshot arrives while a save is pending.

---

### Additional Information

Verified against a running instance — same database, same note route, only the image differing:

| scenario | v0.11.0 | with this change |
| --- | --- | --- |
| updates carrying content only | `ABCDEFGHIJKL` | `ABCDEFGHIJKL` |
| same, then one update without `data` | *(empty)* | `ABCDEFGHIJKL` |

The client had "typed" `ABCDEFGHIJKL` in both cases. The control run on the unpatched image is included because a test that cannot fail proves nothing.

This narrows, but does not close, the window behind #28539 — a client state is still written without being reconciled against the stored document. That needs a separate fix on the seeding side, and this PR does not claim to address it.

### DIFF
--- a/backend/open_webui/socket/main.py
+++ b/backend/open_webui/socket/main.py
@@ -802,11 +802,6 @@ async def yjs_document_update(sid, data):
                 log.warning(f'User {user.get("id")} does not have write access to note {note_id}. Rejecting update.')
                 return
 
-        try:
-            await stop_item_tasks(REDIS, document_id)
-        except Exception:
-            pass
-
         user_id = data.get('user_id', sid)
 
         update = data['update']  # List of bytes from frontend
@@ -834,6 +829,16 @@ async def yjs_document_update(sid, data):
             await document_save_handler(document_id, data.get('data', {}), user)
 
         if data.get('data'):
+            # Only drop the pending save when a new one takes its place.
+            # Updates without a content snapshot (the resync a client sends
+            # after rejoining a document) would otherwise cancel the pending
+            # save without scheduling a replacement, so the edits made just
+            # before the resync never reach the database.
+            try:
+                await stop_item_tasks(REDIS, document_id)
+            except Exception:
+                pass
+
             await create_task(REDIS, debounced_save(), document_id)
 
     except Exception as e:


tjbck:
Thanks!

PR #28709, opened 2026-08-17
fix: forward uploaded videos as video_url content parts

octo-patch:
Reason: Uploaded `video/*` files were treated as ordinary attachments and never emitted as `video_url` content, so models with native video input could not receive them.

### Context

No existing issue/discussion covers this, so here is the short rationale: the upload path already accepts videos and deliberately stores them for later multimodal use, but nothing downstream ever forwarded them, making that stored state unreachable. This wires up the missing half so video-capable models can actually receive the upload.

## Problem

`backend/open_webui/routers/files.py` already stores videos as-is and marks them `completed`, with an explicit note that they are kept for downstream multimodal processing. Nothing consumed them.

In `process_chat_payload`, only **image** files were promoted into content parts, and the `files` field was then stripped from every message:

```python
image_files = [f for f in message.get('files', []) if f.get('type') == 'image' or ...startswith('image/')]
if message.get('role') == 'user' and image_files:
    ...
message.pop('files', None)   # videos silently discarded here
```

So a video upload was dropped from the request entirely — `video_url` appears nowhere in the codebase today. Videos arrive as `type: 'file'` with `content_type: 'video/*'` (they take the generic-file branch on upload), which is what identifies them.

## Changes

- **`backend/open_webui/utils/multimodal.py` (new)** — one dependency-free place mapping an uploaded file to its OpenAI-compatible content part type. Adding a media kind is a single tuple entry:

  ```python
  MEDIA_CONTENT_PARTS = (
      ('image', 'image/', 'image_url'),
      ('video', 'video/', 'video_url'),
  )
  ```

  Files with no resolvable URL are skipped, and parts are grouped by media kind so emitted order is stable regardless of attach order.

- **`backend/open_webui/utils/middleware.py`** — the injection block emits `video_url` parts next to `image_url` via `build_media_content_parts`, and `convert_url_images_to_base64` becomes `convert_url_media_to_base64` so `video_url` parts are resolved too. Without that second half a bare file id would be forwarded verbatim as the URL.

Resolution reuses the existing `get_image_base64_from_url` / `get_image_base64_from_file_id` pair, which already derives the MIME type from the stored file and enforces the owner/admin/read-grant check, so a video file id resolves to `data:video/mp4;base64,...` under the same access gate as images. No change was needed there.

## Behavior notes

- The image path is behaviorally unchanged: same matching rules, same skip-when-no-URL rule, same inline-data-URL short circuit.
- A media part whose value is not a dict is now passed through instead of raising `AttributeError`, and an inline data URL of a *different* media kind is no longer mistaken for an already-resolved part.
- Files that fail to resolve are preserved rather than dropped, matching the previous fallback.

## Testing

Automated, run from `backend/`:

- `python3 -m pytest open_webui/test/utils/test_multimodal.py -q` — 24 passed. Covers file classification, part construction and ordering, URL extraction, inline-data-URL detection, and malformed input.
- `ruff format --check` on the three touched files — already formatted.
- `ruff check` on the two new files — clean. `middleware.py` reports the same 60 pre-existing findings before and after this change, so no new ones are introduced.

I also exercised the two changed `middleware.py` blocks directly against a stubbed file resolver to confirm the full path: a `video/mp4` upload becomes a `video_url` part, its file id resolves to a `data:video/mp4;base64,...` URL, re-running is idempotent, unresolvable ids are preserved, and the image path is unaffected.

Not run: the full backend suite and frontend checks, which need dependencies unavailable in my environment. This change is backend-only and touches no frontend file. I have not performed manual end-to-end testing against a live video-capable endpoint, so please treat that as unverified.

### Fixed

- Uploaded videos are now forwarded to models as `video_url` content parts instead of being silently dropped from the request, enabling native video input on models that support it.

### DIFF
--- a/backend/open_webui/test/utils/test_multimodal.py
+++ b/backend/open_webui/test/utils/test_multimodal.py
@@ -1,0 +1,113 @@
+from open_webui.utils.multimodal import (
+    build_media_content_parts,
+    get_media_content_part_type,
+    get_media_content_part_url,
+    is_inline_media_data_url,
+)
+
+
+class TestGetMediaContentPartType:
+    def test_image_by_type(self):
+        assert get_media_content_part_type({'type': 'image'}) == 'image_url'
+
+    def test_image_by_content_type(self):
+        assert get_media_content_part_type({'type': 'file', 'content_type': 'image/png'}) == 'image_url'
+
+    def test_video_by_type(self):
+        assert get_media_content_part_type({'type': 'video'}) == 'video_url'
+
+    def test_video_by_content_type(self):
+        # Videos are uploaded as generic files, so the content type is what identifies them.
+        assert get_media_content_part_type({'type': 'file', 'content_type': 'video/mp4'}) == 'video_url'
+
+    def test_non_media_file(self):
+        assert get_media_content_part_type({'type': 'file', 'content_type': 'application/pdf'}) is None
+
+    def test_missing_content_type(self):
+        assert get_media_content_part_type({'type': 'file'}) is None
+
+    def test_non_dict(self):
+        assert get_media_content_part_type(None) is None
+        assert get_media_content_part_type('video/mp4') is None
+
+
+class TestBuildMediaContentParts:
+    def test_video_file_becomes_video_url_part(self):
+        files = [{'type': 'file', 'content_type': 'video/mp4', 'url': 'file-id-1'}]
+        assert build_media_content_parts(files) == [{'type': 'video_url', 'video_url': {'url': 'file-id-1'}}]
+
+    def test_image_file_becomes_image_url_part(self):
+        files = [{'type': 'image', 'url': 'file-id-1'}]
+        assert build_media_content_parts(files) == [{'type': 'image_url', 'image_url': {'url': 'file-id-1'}}]
+
+    def test_non_media_files_are_ignored(self):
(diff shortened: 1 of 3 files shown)


octo-patch:
There is no existing issue or discussion for this, so here is a short explanation of the change and why it is needed.

**What is broken.** Uploading a video and asking a video-capable model about it silently sends no video. The upload succeeds, the file is stored, and the model answers as if nothing were attached.

**Why it happens.** `backend/open_webui/routers/files.py` already special-cases `video/*` uploads: it skips text extraction and marks them `completed`, with an in-code note that videos are stored as-is for downstream multimodal processing. But nothing downstream ever reads that state. In `process_chat_payload` only image files are promoted into content parts, and `message.pop('files', None)` then strips the `files` field from every message — so the stored video is discarded before the request is built. `video_url` currently appears nowhere in the codebase.

**Why it is worth fixing.** The upload half of this capability is already in the tree and deliberately reserves videos for this purpose; only the forwarding half is missing. This PR adds that half so `video/*` uploads are emitted as `video_url` content parts, reusing the existing access-checked file-id resolver so videos are gated exactly like images. The image path is behaviorally unchanged, and the new mapping is a single tuple entry per media kind.

Happy to open a discussion instead if that is preferred before review.

PR #28717, opened 2026-08-17
fix(auth): preserve OAuth signout redirect

mikemikimike:
Closes #28320

## Summary

OAuth/OIDC sign-out can be interrupted by the app layout's reactive auth guard. The guard observes `user.set(null)` and navigates to `/auth` before the browser reaches the identity provider's `end_session_endpoint`.

This change preserves the user store while an external `redirect_url` is being navigated to, while still clearing the local token. The local `/auth` fallback keeps the existing state-clearing behavior. A focused unit test covers both redirect branches.

## Compatibility

No API or dependency changes. External OAuth logout navigation is allowed to complete; local logout behavior is unchanged.

## Testing

- `git diff --check` — passed.
- `npm run test:frontend -- --run src/lib/utils/signout.test.ts` — not run successfully because the repository dependencies could not finish installing in this environment; `vitest` was unavailable.
- `npm run check` and build — not run for the same dependency limitation.

### Fixed

- Prevent OAuth/OIDC sign-out redirects from being cancelled by the reactive authentication guard.

### DIFF
--- a/src/lib/components/layout/Sidebar/UserMenu.svelte
+++ b/src/lib/components/layout/Sidebar/UserMenu.svelte
@@ -6,6 +6,7 @@
 
 	import { getUsage } from '$lib/apis';
 	import { getSessionUser, userSignOut } from '$lib/apis/auths';
+	import { getSignOutRedirect, shouldPreserveUserForRedirect } from '$lib/utils/signout';
 
 	import { showSettings, mobile, showSidebar, user, config, settings } from '$lib/stores';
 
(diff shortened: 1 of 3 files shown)


mikemikimike:
Verification update:

- `pnpm install` completed successfully in the contributor workspace.
- `pnpm exec vitest run src/lib/utils/signout.test.ts` — PASS (1 file, 2 tests).
- `pnpm run check` — not clean due pre-existing repository-wide diagnostics, including missing standard library types and unrelated JS/Svelte errors; no diagnostics were reported against the added sign-out helper/test in the captured output.
- `pnpm run build` — blocked during the repository's `pyodide:fetch` preparation because `static/pyodide/package.json` is missing and the script attempted external package downloads.
- `git diff --check` and Prettier checks for changed files — PASS.

Issue #28677, opened 2026-08-16
issue: ?call=true opens the Controls pane but never starts the call — overlay clobbered by showControlsSubscribe's initial callback (desktop 'call' event already has the setTimeout fix)


misha-sunborn:
### Check Existing Issues
- [x] Searched existing/closed issues and discussions — no report of this.

### Installation Method
Docker (`ghcr.io/open-webui/open-webui:main`, dev build ~2026-08-05; the bug is still present in `main` source today, see below).

### Open WebUI Version
`main` — `src/lib/components/chat/Chat.svelte` as of this report.

### Operating System / Browser
Linux (Ubuntu) host; reproduced in Firefox and Brave (Chromium). Not browser-specific.

### Expected Behavior
Opening `/?call=true` (optionally with `&model=<id>`) starts a voice call: the call overlay opens on load, as documented for the URL parameter.

### Actual Behavior
The Controls pane opens, but the call overlay never appears — the page sits in a normal chat with the Controls sidebar open. Clicking the call button manually works.

### Steps to Reproduce
1. Open `https://<host>/?call=true` in a fresh tab (any model, any user).
2. Observe: Controls pane opens on the right; no call overlay / no mic activity.

### Root cause (from the source)
In `Chat.svelte` → `initNewChat()`, the URL handler sets the two stores synchronously:

```js
if ($page.url.searchParams.get('call') === 'true') {
    showCallOverlay.set(true);
    showControls.set(true);
}
```

but `showControlsSubscribe` (registered in `onMount`) runs its **initial** callback with the store's original value `false` after `await tick()`, and that branch does:

```js
if (!value) {
    showCallOverlay.set(false);
    ...
}
```

so the overlay is cleared right after the URL handler set it; the subsequent `true` notification then only opens the pane. The desktop-app `call` event **a few lines below already works around exactly this race**:

```js
// Defer to next macrotask so the call overlay isn't clobbered by
// showControlsSubscribe's initial callback (value=false → set(false))
// which runs as a pending microtask after this function.
setTimeout(() => {
    showCallOverlay.set(true);
    showControls.set(true);
}, 0);
```

The URL-parameter path just never received the same deferral.

### Fix (verified)
Apply the same `setTimeout(..., 0)` deferral to the `?call=true` branch (or lift both into one helper).

[...]

PR #28721, opened 2026-08-17
fix: defer ?call=true overlay to avoid showControlsSubscribe race

silentiris:
## Description

Closes #28677

When `?call=true` is in the URL, the call overlay is immediately hidden because `showControlsSubscribe`'s initial callback (value=false → set(false)) runs as a pending microtask after the synchronous `showCallOverlay.set(true)`, overwriting the overlay state.

## Root cause

The `?call=true` URL handler synchronously calls `showCallOverlay.set(true)`, but the Svelte store subscription's initial callback fires as a pending microtask afterward, resetting the overlay to false.

## Fix

Wrap the `showCallOverlay.set(true)` call in `setTimeout(0)` to defer it to the next macrotask, after the subscription's initial callback has settled.

## Testing

- Verified the overlay appears and stays visible with `?call=true` in the URL
- Verified normal call overlay behavior is unaffected

> [!NOTE]

### DIFF
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -2187,8 +2187,13 @@
 		}
 
 		if ($page.url.searchParams.get('call') === 'true') {
-			showCallOverlay.set(true);
-			showControls.set(true);
+			// Defer to next macrotask so the call overlay isn't clobbered by
+			// showControlsSubscribe's initial callback (value=false → set(false))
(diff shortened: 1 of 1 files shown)


silentoplayz:
Tested this branch against the repro in #28677 on the latest dev: opening `/?call=true` now brings up the call overlay on load, with the Controls pane open as before.

silentiris:
Thanks for testing! Glad the overlay fix works as expected.

Issue #28742, opened 2026-08-18
bug: unlocked chat-blob read-modify-write silently drops messages under concurrency (chat_message table diverges from chat.chat)


penradioGables:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker, `ghcr.io/open-webui/open-webui`

### Open WebUI Version

v0.11.0 (confirmed present on current `main` — the affected code is unchanged)

### Operating System

Linux, PostgreSQL backend

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I have documented step-by-step reproduction instructions.

### Expected Behavior

A message that is saved to a chat (visible in the `chat_message` table, and used to build LLM context for later turns) should also be visible in the chat when it's reopened, and should not disappear from the UI.

### Actual Behavior

OWUI stores chat history in two places: the `chat.chat` JSON blob (`history.messages`, what `GET /api/v1/chats/{id}` returns to the frontend) and the normalized `chat_message` table (what generates LLM context — see `get_messages_map_by_chat_id`). They are kept in sync by several **whole-blob read-modify-write functions in `backend/open_webui/models/chats.py` that take no row lock**:

- `update_chat_by_id` (~L607) — the general-purpose writer, used by `main.py`, the chat router's `POST /{id}`, and `routers/notes.py`
- `update_chat_title_by_id` (~L729)
- `upsert_message_to_chat_by_id_and_message_id` (~L985)
- `add_message_status_to_chat_by_id_and_message_id` (~L1081)
- `add_message_files_by_id_and_message_id` (~L1112)

Each does: read the whole `chat.chat` blob → mutate it in Python → write the whole blob back. With no locking, if two of these run concurrently against the *same chat row*, whichever commits last wins outright — even if it started its read *before* the other committed. A message written by the loser survives in `chat_message` (each of these functions also dual-writes there) but is gone from the blob.

The practical trigger is `backend/open_webui/main.py` around L1411-L1429: on every completion it does a **separate, unlocked** read-modify-write of `chat.files`/`chat.models` (via `Chats.get_chat_by_id` + `Chats.update_chat_by_id`) immediately followed by a **separate** call to `Chats.upsert_message_to_chat_by_id_and_message_id` for the user's message. If anything delays the first call past the second one's commit (slow request, concurrent tool-call step, retried request, etc.), the user's message is dropped from the blob:

```python
[...]

PR #28744, opened 2026-08-18
fix: stop concurrent chat writes from dropping messages and chat fields

Classic298:
Messages could disappear from a chat while still being sent to the model. Every writer read the whole chat JSON, changed it in Python and wrote all of it back without a lock, so whichever write committed last silently discarded everything committed in between. A message lost that way stayed in the chat_message table, so it kept counting toward the model's context while vanishing from the screen, and later replies pointed at a parent that no longer existed. The same overwrite wiped chat-level fields: a file attached during a completion was gone after the browser's next chat save.

Writers now take a row lock before they read (PostgreSQL; SQLite has no equivalent, so there the history merge below is what protects it), and message history is merged against the freshly read row instead of replacing it.

Locking alone was not enough. The completion handler, the chat-save endpoint and the note chats all built their replacement chat from a read taken earlier, so the value being written was already stale before any lock could be taken. They now pass down only the keys they actually change, and applying those keys onto the stored blob is what the single remaining writer does.

Fixes #28742

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1470,15 +1470,13 @@ async def chat_completion(
                     # The old frontend saveChatHandler did this on every message;
                     # now the backend owns persistence.
                     chat_files = metadata.get('files')
-                    if chat_files is not None or selected_chat_models:
-                        existing_chat = await Chats.get_chat_by_id(chat_id)
-                        if existing_chat:
-                            updated = {**existing_chat.chat}
-                            if chat_files is not None:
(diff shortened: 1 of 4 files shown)


Classic298:
@tjbck ready for review. Concurrent chat writes silently drop messages. Whichever write lands last discards everything committed in between, the lost message still counts toward the model context while being gone from the screen, and a file attached during a completion disappears on the next chat save.

PR #28748, opened 2026-08-18
fix: keep folder parent references acyclic

Classic298:
Moving a folder under one of its own subfolders was accepted. A folder in a parent loop is never a root, so it and everything under it silently disappeared from the sidebar, and there was no way to get it back from the UI.

The move is now rejected with a 400, folders whose parent chain loops are put back at the root on the next folder list, and the folder tree traversals skip ids they have already visited so existing data in that state stays workable.

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/folders.py
+++ b/backend/open_webui/models/folders.py
@@ -197,10 +197,14 @@ class FolderTable:
         try:
             async with get_async_db_context(db) as db:
                 folders = []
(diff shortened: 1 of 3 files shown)


Classic298:
@tjbck ready for review. Moving a folder under one of its own subfolders is accepted, and the whole subtree then disappears from the sidebar with no way to get it back from the interface.

Issue #28747, opened 2026-08-18
bug: Supported Media MIME Types cannot be cleared in Documents settings


arquiteto-raiz:
### Check Existing Issues

- [x] I searched existing issues and did not find an exact duplicate.

### Description

The **Supported Media MIME Types** field in **Admin Settings → Documents** cannot be cleared through the UI after a value has been saved.

This matters for vision models because the documented way to bypass content extraction for images is to leave this field empty. Once `image/*` has been saved, clearing the field and saving does not remove it.

### Steps to reproduce

1. Open **Admin Settings → Documents**.
2. Set **Supported Media MIME Types** to `image/*`.
3. Save.
4. Clear the field completely.
5. Save again.
6. Reopen the Documents settings.

### Expected behavior

The field remains empty and the backend configuration is saved as an empty list (`[]`).

### Actual behavior

The previous value (`image/*`) reappears.

### Root cause

In `src/lib/components/admin/Settings/Documents.svelte`, a blank field is serialized as `undefined`:

```ts
CONTENT_EXTRACTION_SUPPORTED_MEDIA_MIME_TYPES:
    RAGConfig.CONTENT_EXTRACTION_SUPPORTED_MEDIA_MIME_TYPES.trim() === ''
        ? undefined
        : RAGConfig.CONTENT_EXTRACTION_SUPPORTED_MEDIA_MIME_TYPES.split(',')
              .map((s) => s.trim())
              .filter((s) => s.length > 0)
```

The retrieval configuration endpoint treats `None` as “keep the existing value,” so omitting the field cannot clear a previously stored list.

### Minimal fix

Send an empty list when the field is blank:

```diff
- ? undefined
+ ? []
```

### Workaround

Entering a single comma (`,`) and saving produces an empty filtered array, so the setting is cleared. This confirms that the backend already accepts `[]`; the UI simply cannot normally submit it.

### Impact

Administrators cannot restore the documented empty setting through the UI. Images may continue to be routed to the configured extraction engine instead of being passed directly to a vision-capable model.

silentoplayz:
Reproduced on the latest dev branch: after saving a value in Supported Media MIME Types under Admin Settings > Documents, clearing the field and saving again brings the previous value back when the page is reopened. Proposed fix in #28750.

tjbck:
Addressed in dev.

PR #28750, opened 2026-08-18
fix: allow clearing Supported Media MIME Types in Documents settings

Classic298:
The Supported Media MIME Types field in Admin Settings > Documents could not be cleared. Emptying it and saving left the previous list in place, so an administrator who had configured it could never turn media extraction back off from the interface.

The setting has three meaningful states: unset keeps the old behavior where only the external engine receives images and video, an empty list disables media extraction for every engine, and a list selects which types are extracted. A blank text box can mean either of the first two, and the admin page always sent "leave unchanged" for a blank box, which made the empty list unreachable.

The page now tracks whether the server actually holds a value, read from the configuration it loads and refreshed from every save response. A blank box clears the list when one was configured, and stays a no-op while the setting has never been set, so instances on the external engine are not silently switched over when an unrelated Documents setting is saved. Always sending an empty list for a blank box would have caused exactly that.

Fixes #28747

> [!NOTE]

### DIFF
--- a/src/lib/components/admin/Settings/Documents.svelte
+++ b/src/lib/components/admin/Settings/Documents.svelte
@@ -71,6 +71,8 @@
 	};
 
 	let RAGConfig: any = null;
+	let mediaMimeTypesConfigured = false;
+
 	const inputClass =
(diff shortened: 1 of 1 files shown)


silentoplayz:
Tested this branch against the repro in #28747 on the latest dev: after saving image/* in Supported Media MIME Types, clearing the field, and saving again, the field stays empty when Documents settings is reopened.

Classic298:
@tjbck ready for review. The Supported Media MIME Types field cannot be cleared. An administrator who configured it can never turn media extraction back off from the interface.

Issue #28749, opened 2026-08-18
issue: Advanced connection settings crash when the connection's tags are stored as plain strings


iammrbt:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.11.0

### Operating System

Windows 11

### Browser (if applicable)

Chrome latest

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

Opening the Advanced settings for an OpenAI-compatible connection should display its Model IDs, Prefix ID, tags, and other advanced options.

OpenAI-compatible `/models` responses commonly identify models using an `id` property without providing a separate `name` property. The connection editor should accept these valid model objects and use `id` as the display-name fallback.

### Actual Behavior

The Advanced settings work for the native OpenAI connection but do not open for a Z.AI OpenAI-compatible connection.

Clicking Advanced on the Z.AI connection produces a frontend exception:

`TypeError: Cannot read properties of undefined (reading 'toLowerCase')`

The Z.AI `/models` endpoint succeeds and returns valid model objects. Open WebUI's own `/openai/models/1` proxy also receives the complete model list, including `glm-5.3`.

However, the Advanced connection component appears to assume that every model object has a `name` property and calls `name.toLowerCase()` without falling back to the standard `id` property.

[...]

PR #28753, opened 2026-08-18
fix: normalize tags stored as plain strings instead of crashing the connection editor

Classic298:
Opening Advanced on a connection whose tags are stored as plain strings throws in the browser and the panel never opens, so the connection cannot be edited any more. The editor has always written tags as objects with a name, but OPENAI_API_CONFIGS and OLLAMA_API_CONFIGS accept raw JSON and the docs list tags as a supported key without saying it must be objects, so a list of strings is accepted and stored.

The same shape does two more things today. One string tag made the models endpoint throw while merging tags, which was caught and blanked, so every tag on every model from that connection silently disappeared, including the model's own workspace tags. A tag object with a null name passed validation and reached the frontend, where it breaks the model picker.

Tags are now normalized to a canonical trimmed name object where unvalidated config first meets model data: the models endpoint, the model metadata validator, the direct connection path and the connection editor. Downstream renderers keep assuming a name is present, which they always did, and saving a connection through the editor migrates its stored tags. Verified by diffing old against new endpoint behaviour across 15 model shapes: every difference is a tag that used to vanish or crash and now renders, valid data is untouched, and the only two shapes that still raise raise identically on an unchanged line above.

Fixes #28749

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -244,7 +244,7 @@ from open_webui.utils.middleware import (
     process_chat_payload,
     process_chat_response,
 )
-from open_webui.utils.misc import merge_model_params
+from open_webui.utils.misc import merge_model_params, normalize_tags
 from open_webui.utils.model_ids import strip_provider_model_prefix
 from open_webui.utils.models import (
     check_model_access,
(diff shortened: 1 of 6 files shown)


Classic298:
@tjbck ready for review. A connection whose tags are stored as plain strings, which the config format accepts and the docs do not warn against, makes the Advanced panel throw so that connection can no longer be edited. The same shape also silently blanks every tag on every model from that connection.

Issue #28767, opened 2026-08-19
issue: fetching tags for a chat the user can read but does not own returns 401


silentoplayz:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I have checked whether this issue is already fixed on the `dev` branch or in the latest source.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.11.0, endpoint unchanged on `dev` @ `5ea9ff3ed`

### Ollama Version (if applicable)

Not applicable

### Operating System

Linux

### Browser (if applicable)

Firefox

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**.

### Expected Behavior

Opening a chat I am allowed to read but do not own loads it without an authorization error.

### Actual Behavior

The chat loads, and the request for its tags that immediately follows fails:

```
XHR GET /api/v1/chats/<chat_id>          [HTTP/1.1 200 OK]
XHR GET /api/v1/chats/<chat_id>/tags     [HTTP/1.1 401 Unauthorized]
```

`GET /{id}` resolves read access four ways: ownership, an admin override under `ENABLE_ADMIN_CHAT_ACCESS`, an explicit `shared_chat` read grant, and access to the folder the chat sits in. `GET /{id}/tags` recognises only the first of those and raises 401 for everything else, so every chat reachable by any of the other three paths fails on tags.

The client catches it and continues without tags, so nothing visibly breaks. The result is a 401 in the console during a supported flow, and no tags for the chat.

### Steps to Reproduce

The configuration this was observed in:

1. Run Open WebUI with two users, A and B, where A is an admin and `ENABLE_ADMIN_CHAT_ACCESS` is at its default of `true`.
[...]

PR #28769, opened 2026-08-19
fix: align chat tags GET with readable chat access

Gujiassh:
# Pull Request Checklist

- [x] Closes #28767
- [x] Targets `dev`
- [x] No dependency or documentation changes
- [x] Atomic change and self-review completed

# Changelog Entry

### Description

- Align `GET /api/v1/chats/{id}/tags` with the readable-chat access rules already used by `GET /api/v1/chats/{id}`, so admin, shared, and folder readers no longer get a 401 when loading tags for a chat they can open.

### Added

- Shared `get_readable_chat_by_id` helper used by both chat and tags GET paths.

### Changed

- Tags GET now resolves readable chats through the shared helper and loads tag rows for the chat owner.

### Fixed

- Readable non-owned chats no longer fail on the follow-up tags request with HTTP 401.

---

### Additional Information

- Validation: `python3 -m py_compile backend/open_webui/routers/chats.py` (passed)
- POST/DELETE tag endpoints remain owner-only.
- No API, persistence, or UI contract changes.

### Screenshots or Videos

- Not applicable; backend auth alignment only.

### DIFF
--- a/backend/open_webui/routers/chats.py
+++ b/backend/open_webui/routers/chats.py
@@ -1317,18 +1317,9 @@ async def compact_chat_by_id(
     return result
 
 
-############################
-# GetChatById
-############################
 
-
-@router.get('/{id}', response_model=ChatResponse | None)
-async def get_chat_by_id(
-    id: str,
-    request: Request,
-    user=Depends(get_verified_user),
-    db: AsyncSession = Depends(get_async_session),
-):
+async def get_readable_chat_by_id(id: str, user, db: AsyncSession):
+    """Resolve a chat the user may read: owner, admin override, shared grant, or folder."""
     chat = await Chats.get_chat_by_id_and_user_id(id, user.id, db=db)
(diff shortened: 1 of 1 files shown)


silentoplayz:
Tested this branch against the setup that produced #28767, a chat readable through the admin override and a shared folder but not owned: GET /{id}/tags now returns 200 with the chat's tags instead of 401. The owner path is unchanged, verified with a full tag add, read, and delete cycle resolving the label correctly. Resolving labels against chat.user_id reads right to me, since the tag ids already arrive in the viewer's GET /{id} response through meta.tags. The shared helper also keeps the two routes from drifting apart again, which is how this bug happened in the first place. Thanks for picking it up.

silentoplayz (commented):
LGTM

tjbck:
Aligned with major overhaul, appreciate the PR.

Issue #28777, opened 2026-08-19
issue: RedisDict.set() skips writes on a per-process signature, so one worker's partial model list sticks for all workers ("Model not found")


Riezebos:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.11.0

### Operating System

Kubernetes 1.32

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

With `WEBSOCKET_MANAGER=redis` and several workers, the shared `open-webui:models` hash should converge on the model list a worker actually holds. A worker whose `get_all_models()` returns the full list should (re)publish it, so a momentarily degraded list written by another worker is corrected on the next refresh.

### Actual Behavior

The shared hash can get stuck holding a **partial** model list, and no worker ever repairs it. Chat requests for the missing models then fail with `Model not found` on **every** pod until a pod restarts or its own model list genuinely changes.

The write-skip guard in `RedisDict.set()` is cached **per process**, but the state it describes is **shared**:

```python
# backend/open_webui/socket/utils.py
self._last_signature: str | None = None   # L69  — per-process
...
if signature == self._last_signature:     # L127 — guards a shared Redis hash
    return
```

Sequence (two workers, one Redis):

[...]

PR #28779, opened 2026-08-19
fix: store RedisDict write-skip fingerprint in Redis, not per process

Battleplus:
### Linked Issue

Closes #28777

### Description

The write-skip guard in `RedisDict.set()` compared a fingerprint cached **on the instance** against the mapping about to be written. The state it guards, however, lives in the **shared Redis hash**. With `WEBSOCKET_MANAGER=redis` and multiple workers:

1. Worker A publishes the full model list `X` and caches `signature(X)` locally.
2. Worker B's upstream fetch momentarily returns a partial list `Y` (one connection slow/down for that pod); B writes `Y`, so the shared hash is now `Y`.
3. Worker A refreshes. Its own list is unchanged, so it recomputes `signature(X)`, matches its own per-process cache, and **skips the correcting write**.
4. The hash stays at `Y`. Models in `X \ Y` are gone for every pod (`Model not found` at `main.py:1103`) until a pod restarts or its own list genuinely changes.

`clear()` had the same flaw in the other direction: it reset only the local fingerprint, so other workers kept one describing contents that no longer exist and skipped repopulating.

### Root cause

The fingerprint lives in the process while the data it describes lives in Redis — the guard must describe the shared state, not a per-process memory of it.

### Fix

Store the fingerprint in Redis **beside** the hash (`<name>:__signature__`):

- `set()` skips only when the *shared* fingerprint matches the mapping about to be written — identical steady-state behavior, but a degraded list written by another worker is repaired on the next refresh of any worker holding the correct list.
- The fingerprint is written **after** the hash, as a **separate command** (the two keys need not share a Redis Cluster hash slot; a crash mid-write leaves the fingerprint absent so the next `set()` rewrites, rather than claiming contents that were never written).
- `clear()` drops the **shared** fingerprint.
- Item-level writes (`__setitem__` / `__delitem__`) invalidate the shared fingerprint.
- The guard is gated behind `track_fingerprint` and enabled only for the model pool; the session/usage pools never call `set()` and keep their hot websocket path free of the extra round trip.

### Testing

Ran the real `RedisDict` class from source (zero dependencies, fake shared Redis — same technique as the issue's repro):

- **Before:** after `a.set(full)`, `b.set(partial)`, `a.set(full)`, `'llama3' in a` → `False` (matches the issue)
- **After:** `True`, and stays `True` across 100 further identical refreshes
- Redundant identical `set()` still skips the hash write
- `clear()` drops the shared fingerprint; another worker repopulates
- Item-level writes invalidate the fingerprint
- Untracked instances keep the previous always-write behavior
- Empty mapping clears both hash and fingerprint

`ruff format --check` and `ruff check --select=F` pass on both changed files.

# Changelog Entry

### Description

- Moves the `RedisDict.set()` write-skip fingerprint from a per-process attribute to a Redis key beside the hash, so a worker holding the correct model list can repair a degraded list written by another worker.

### Added

- `RedisDict(track_fingerprint=...)` flag; `_fingerprint_key()` helper.

### Changed

- `RedisDict.set()` skips redundant writes by comparing against the shared fingerprint in Redis instead of a per-process cache.

### Deprecated

- None.

### Removed

- `RedisDict._last_signature` per-process fingerprint.

### Fixed

- Multi-worker `WEBSOCKET_MANAGER=redis` model list getting stuck on a partial list (`Model not found` for every pod) when one worker's upstream fetch momentarily degrades.

### Security

- None.

### Breaking Changes

- None.

### Additional Information

- None.

### Screenshots or Videos

- Not applicable (backend change).

### DIFF
--- a/backend/open_webui/socket/main.py
+++ b/backend/open_webui/socket/main.py
@@ -121,6 +121,9 @@ if WEBSOCKET_MANAGER == 'redis':
         redis_url=WEBSOCKET_REDIS_URL,
         redis_sentinels=ws_sentinels,
         redis_cluster=WEBSOCKET_REDIS_CLUSTER,
+        # set() skips redundant writes against a fingerprint stored beside the
+        # hash, so it must be tracked for the model pool (see RedisDict.set).
+        track_fingerprint=True,
     )
 
     SESSION_POOL = RedisDict(
--- a/backend/open_webui/socket/utils.py
+++ b/backend/open_webui/socket/utils.py
@@ -61,12 +61,22 @@ class RedisLock:
 
 
 class RedisDict:
-    def __init__(self, name, redis_url, redis_sentinels=[], redis_cluster=False):
+    def __init__(
+        self,
+        name,
+        redis_url,
+        redis_sentinels=[],
+        redis_cluster=False,
+        track_fingerprint=False,
+    ):
         self.name = name
-        # Per-process cache of the last payload fingerprint written by set().
-        # Used to skip redundant HSET round-trips when the model list hasn't
-        # changed — the dominant Redis write source on busy multi-pod setups.
-        self._last_signature: str | None = None
+        # When True, set() skips redundant writes by comparing against a
+        # fingerprint stored in Redis *beside* the hash — not on this instance —
+        # so the guard describes the state it protects (see set()). Only the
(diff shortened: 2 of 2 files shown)


Classic298:

PR #28795, opened 2026-08-19
perf: filter workspace models by access in SQL instead of loading every model

Classic298:
Working out which workspace models a user may see loaded every model row, built a full response object with its owner for each, and only then dropped the inaccessible ones. That runs on model export and on every file access check, so a large model table slowed down requests that have nothing to do with models.

The owner-or-grant check now happens in the query itself, reusing the permission filter this file already applies to the paginated list endpoint. The by-user wrapper is gone with it, so each caller states the permission it wants instead of inheriting it.

Measured with 500 workspace models of which 3 are visible to the caller: 5 queries and ~12.7 ms before, 4 queries and ~2.8 ms after. The resulting set is unchanged for owner, public, direct-user, group and multi-grant entries, for both read and write, and base model entries stay excluded as before.

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/models.py
+++ b/backend/open_webui/models/models.py
@@ -244,9 +244,21 @@ class ModelsTable:
                     log.error('Skipping model %r during get_all_models due to error: %s', model.id, exc)
             return models
 
-    async def get_models(self, db: AsyncSession | None = None) -> list[ModelUserResponse]:
+    async def get_models(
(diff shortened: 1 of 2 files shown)


tjbck:
@Classic298 conflicts

Classic298:
on it

Classic298:
@tjbck

PR #28797, opened 2026-08-19
fix: repair the shared model cache after a worker publishes a partial model list

Classic298:
With WEBSOCKET_MANAGER=redis and several workers, a worker whose upstream fetch momentarily returned fewer models published that shorter list to the shared hash, and no other worker ever corrected it: RedisDict.set() kept the fingerprint of its last write in the process, so a worker still holding the full list matched its own fingerprint and skipped the repairing write. Chatting with one of the missing models then failed with "Model not found" on every pod until one of them restarted.

The fingerprint now lives in Redis next to the hash and is written by the same Lua script as the contents, so it always describes what the hash actually holds and a worker holding the full list republishes it. A fast path still skips sending the payload when the hash already holds it, so the write elimination the fingerprint was added for is kept rather than reverted; the model list is hundreds of large JSON values and every /api/models request refreshes it.

The fingerprint key is brace-tagged to share the hash's Redis Cluster slot. A REDIS_KEY_PREFIX containing an unmatched "}" cannot be slot-matched by any sibling key and is unsupported on cluster.

Fixes #28777

> [!NOTE]

### DIFF
--- a/backend/open_webui/socket/utils.py
+++ b/backend/open_webui/socket/utils.py
@@ -61,12 +61,28 @@ class RedisLock:
 
 
 class RedisDict:
+    # Contents and fingerprint must be written together, or a stale fingerprint suppresses the repair.
+    _SET_SCRIPT = """
+    local wanted = {}
+    for i = 2, #ARGV, 2 do
+        wanted[ARGV[i]] = true
+        redis.call('hset', KEYS[1], ARGV[i], ARGV[i + 1])
+    end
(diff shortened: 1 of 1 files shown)


Classic298:
@tjbck ready for review. With redis websockets and several workers, a worker whose upstream fetch briefly returns fewer models publishes that shorter list and no other worker ever repairs it. Chatting with one of the missing models then fails with "Model not found" on every pod until a restart.

PR #28814, opened 2026-08-19
perf: read one field instead of the whole history when a streamed event appends to a message

Classic298:
Every streamed source, files and embeds event loaded the chat's entire message history and rebuilt the navigable message map, parent links and children lists included, only to read one list off one message and append to it. A tool call returning ten citations emits ten such events, so the cost scaled with the length of the conversation multiplied by the number of citations.

Those three branches now read the one column they need. The read goes to the column rather than the row deliberately: the write path commits a message before it validates it, so rows exist that the message model rejects, and reading through the model would raise where the old code simply returned the stored list.

Measured on a two hundred message chat: ~3.1 ms per event before against ~0.65 ms after, and the cost no longer grows with the conversation. Both paths issue a single query. Behaviour is unchanged for a chat with per-message rows, a legacy chat that only has the embedded history, an absent or null field, and a message id that does not exist. A chat id that does not exist used to raise an AttributeError in all three branches and now yields an empty list.

> [!NOTE]

### DIFF
--- a/backend/open_webui/models/chats.py
+++ b/backend/open_webui/models/chats.py
@@ -6,6 +6,7 @@ import logging
 import re
 import time
 import uuid
+from typing import Literal
 
 # local imports
 from open_webui.internal.db import Base, JSONField, get_async_db_context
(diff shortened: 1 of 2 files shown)


Classic298:
@tjbck ready for review. Every streamed source, files and embeds event loaded the chat's whole message history and rebuilt the message map just to read one list and append to it. A tool call returning ten citations emits ten of those, so it scaled with conversation length times citation count.

3.1 ms down to 0.65 ms per event, and it no longer grows with the conversation.

PR #28817, opened 2026-08-19
perf: buffer the streamed response text and join it once

Classic298:
When a chat has API outlet filters configured the streaming handler keeps its own copy of the assistant message, and every delta rebuilt that copy by concatenating the whole text again. The cost of a delta grows with the text already accumulated, so a long response gets slower the longer it runs, and a provider that sends large batched chunks makes it much worse.

Deltas are now buffered in a list and joined once, when the stream has finished and the message is handed to the outlet filters. The join reproduces what the concatenation did without assuming the deltas are text: a stream filter can make a delta any type, and two lists used to merge into a list, so a buffer that is not all strings is folded with the same `+`. The buffered parts are tracked as they are created, so nothing walks the output tree at the end and a malformed part from a provider cannot raise where it did not before.

Replaying streams through both revisions with 400-character deltas, 10000 deltas fall from 2317ms to 22ms and 50000 from 66.7s to 154ms; with 6-character deltas, 100000 deltas fall from 551ms to 148ms. Per-delta cost stops growing with the accumulated text. Holding the deltas until the stream ends costs about 49 bytes per delta, so half a megabyte on a ten thousand delta response and five megabytes on a hundred thousand. The other streaming paths still concatenate and are left alone here.

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -3364,7 +3364,13 @@ def update_assistant_message_from_stream(assistant_message, raw):
     def append_output_text(item, text):
         parts = item.setdefault('content', [])
         if parts and parts[-1].get('type') == 'output_text':
-            parts[-1]['text'] += text
+            part = parts[-1]
+            buffer = part.get('_text_buffer')
+            if buffer is None:
+                buffer = []
+                part['_text_buffer'] = buffer
+                assistant_message.setdefault('_buffered_parts', []).append(part)
+            buffer.append(text)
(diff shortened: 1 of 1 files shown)


Classic298:
@tjbck ready for review. On API streams going through the outlet filter path, every response is assembled in quadratic time and the copying happens on the event loop, so it delays every other request on that worker.

0.5 s down to 0.05 s on a 1 MB response, and the gap widens as responses get longer.

Classic298:
superseeded by #28844

Issue #28173, opened 2026-08-05
perf: redis websocket manager: every instance decodes every emit


Classic298:
Title: redis websocket manager: every instance decodes every emit

with WEBSOCKET_MANAGER=redis every sio.emit is published to one shared channel. every instance receives and json-decodes every message, checks local room membership and most of them throw it away. with 16 instances every streamed token delta gets encoded once and decoded 16 times, 15 of them for nothing. py-spy across a loaded fleet (16 instances, ~4000 users) puts about a third of all active cpu samples in the pubsub listener chain (async_pubsub_manager -> hiredis parsing), the single largest cpu bucket, and it grows with instance count, so scaling out makes every message more expensive for the whole fleet.

it also means a stream whose user already closed the tab keeps publishing every delta to all instances until generation finishes, because no instance can know whether any other instance still has a listener, so the emitter always emits.

proposal: route instead of broadcast. keep a room -> instances registry in redis (updated on connect/join/leave/disconnect, instances heartbeat with a ttl so a killed instance gets pruned) and give every instance its own channel. each emit then picks the cheapest sufficient path:

1. no members anywhere: drop before serializing anything. covers closed-tab-mid-stream and events to offline users
2. members only on the emitting instance: deliver locally, redis never touched
3. members on k remote instances (usually 1): encode once, publish to those k channels, decoded k times instead of 16
4. room=None broadcasts keep the shared channel as today

routing lookups go through a short per-process cache plus invalidation messages on membership changes, so there is no extra redis round trip per emit. node's socket.io has this as the sharded redis adapter; python-socketio doesn't ship it but its pubsub manager is subclassable enough to do it cleanly.

behind WEBSOCKET_REDIS_ROUTING=false|true, default false (hub, fan out behaviour), so nothing changes unless opted in. the whole fleet has to switch together since hub instances don't subscribe to the per-instance channels. i have this implemented and will send a PR.

PR #28818, opened 2026-08-19
perf: per-room channel delivery for the socket.io Redis manager

Classic298:
Reworked from the ground up after the feedback that the registry implementation did not land. The Redis room registry and its whole recovery protocol (heartbeats, liveness keys, pruning, distrust windows, cache invalidation) are gone; the change is now ~105 lines with no state kept outside the process.

With WEBSOCKET_MANAGER=redis every emit is published on one shared channel and every instance JSON-decodes every message: a 16 instance fleet decodes each streamed token delta 16 times and 15 discard it. py-spy across a loaded fleet (16 instances, ~4000 users) puts ~31% of all active CPU samples in the pubsub listener parse chain, the largest bucket.

Room-targeted emits are now published on a per-room channel instead; every instance keeps one static pattern subscription covering all room channels and drops messages for rooms without local members by channel name, paying a set lookup instead of a JSON parse. No state leaves the process, so recovery paths and loss windows are identical to the stock manager; acks and control messages stay on the shared channel and sio.call works across instances unchanged. This is the delivery scheme the official socket.io Redis adapter for Node.js ships by default.

Enabled by default; WEBSOCKET_REDIS_ROOM_CHANNELS=false restores shared-channel-only delivery. All instances must run the same mode, so the switch rides the full-stop upgrade this release already requires for its migration; in a mixed fleet, room emits from updated instances would not reach not-yet-updated ones. Verified end to end with two instances on a real Redis: cross-instance token streams delivered with the shared channel completely silent. Ref #28173.

> [!NOTE]


tjbck:
@Classic298 conflicts.

PR #28819, opened 2026-08-19
perf: skip the outlet filter pass when the model has no filters

Classic298:
On installs without any filter functions or pipeline filters, which is the default, every completed chat message still paid the full outlet pass: the whole messages map was fetched from the database, every message on the branch was rebuilt with a deepcopy of its output, and the resulting list was emitted to the client as a chat:outlet event even though nothing could have changed. On long chats that event alone ships hundreds of kilobytes per completed message (measured with typical reasoning outputs: 86 KB at 10 turns, 429 KB at 50, 1.3 MB at 150) and the frontend handler drops it because no content differs.

The handler now resolves the model's filter functions and pipeline filters first and returns before any of that work when both are empty. On installs without filters that check is a single small query, and behavior with filters configured is unchanged. The API fallback stream path already had this gate; it now shares the same pipeline-filter predicate.

> [!NOTE]

### DIFF
--- a/backend/open_webui/routers/pipelines.py
+++ b/backend/open_webui/routers/pipelines.py
@@ -54,6 +54,12 @@ def get_sorted_filters(model_id, models):
     return sorted_filters
 
 
+def has_pipeline_outlet_filters(model, models):
+    # process_pipeline_outlet_filter also runs the model itself when it is a pipeline
+    model_id = model.get('id') if isinstance(model, dict) else model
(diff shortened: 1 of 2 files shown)


Classic298:
@tjbck ready for review. Installs with no filter functions and no pipeline filters, which is the default, still pay the full outlet pass on every completed message: the whole message map is read, every message is deepcopied, and the result is shipped to the client as an event the frontend then discards. Measured at 429 KB per completed message on a 50 turn chat and 1.3 MB at 150 turns.

PR #28822, opened 2026-08-19
perf: serve single-key config reads from a cached snapshot

Classic298:
Every Config.get was its own SELECT, so a request that checks a dozen permissions and settings paid a dozen round trips to a table that is tiny, rarely written and read constantly.

Single-key reads now come from one in-memory snapshot, refreshed at most once a second and retired by any write. Multi-key and whole-table reads still query the database, because the flows using them read a block of config, edit it and write it back, where a stale read would drop a change another worker committed. Three keys read that way singly are excluded too.

The snapshot is one SELECT, so keys that must agree now come from a consistent read. With a writer committing at 200/s an openai url and its api key were mispaired on 714 of 714 reads before, 0 of 27497 after.

13 config reads in one request, SQLite:
  permission-heavy   13 queries / 5.7 ms -> 1 query / 0.6 ms
  ollama request     13 queries / 6.1 ms -> 1 query / 0.8 ms

CONFIG_CACHE_TTL bounds how long a worker can serve a value another worker changed.

> [!NOTE]

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -984,6 +984,16 @@ _default_group_share = os.getenv('DEFAULT_GROUP_SHARE_PERMISSION', 'members').st
 DEFAULT_GROUP_SHARE_PERMISSION = 'members' if _default_group_share == 'members' else _default_group_share == 'true'
 
 ####################################
+# CONFIG
+####################################
+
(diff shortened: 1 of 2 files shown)


tjbck:
Up for discussion.

PR #28836, opened 2026-08-20
perf: stop resending the full accumulated output at every tool-call boundary

Classic298:
On tool-heavy chats the streaming path emitted the entire accumulated output twice per tool-call iteration, so a turn with dozens of tool calls pushed the same multi-megabyte payload dozens of times through the websocket layer, multiplied by the Redis fanout in multi-container deployments.

Tool calls already stream as incremental response:completion events. This change extends that vocabulary to the tool boundary: each tool result is emitted once as an output item event, the completed call follows with its sanitized arguments, and the per-iteration full snapshots are dropped for regular chats. Channel chats keep the snapshots because the channel message store is rebuilt from each one. A continuation no longer re-sends its own output list on completion, which previously reset the client's accumulated state at every boundary until the next snapshot repaired it. The frontend now matches items by type as well as call id so a result can never replace the call it answers, and pads missed indices so a client that reconnected mid-run never holds a sparse array. Code-interpreter retries keep their bounded snapshots.

Measured with a socket capture against a live instance and a mocked tool-calling upstream: pre-completion full snapshots per iteration go from two to zero, final client and database state is identical, and on a single-iteration turn snapshot payload bytes drop from 1672 to 678. The saving scales with iteration count times accumulated output size. During a rolling deploy a stale frontend can hide a tool block mid-stream until the final completion event restores it, so backend and frontend ship together.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -83,7 +83,7 @@ from open_webui.utils.access_control.files import get_owner_accessible_folder_fi
 from open_webui.utils.access_control.folders import has_folder_access
 from open_webui.utils.ask_user import stage_ask_user_tool_call
 from open_webui.utils.chat import generate_chat_completion
-from open_webui.utils.chat_id import is_saved_chat_id
+from open_webui.utils.chat_id import CHANNEL_CHAT_ID_PREFIX, is_saved_chat_id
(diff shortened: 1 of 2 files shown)


Classic298:
Replaced by #29270, rebased on current dev. Three of the four frontend hunks here already landed via #29053 and #29250, and dev solved them by appending rather than padding.

PR #28844, opened 2026-08-20
perf: stop recopying the streamed message text on every delta

Classic298:
Every streamed token appended to the response text through a dict slot, which allocates a new string and copies everything received so far. A long answer pays that copy thousands of times on the event loop, so every other request on the worker waits behind it.

The appends now clear the slot before growing the string, which leaves a single reference and lets CPython resize it in place. This covers the six accumulation points in the Chat Completions streaming path. The Responses API transport copies its items before appending, so it holds two references and is unchanged.

Measured on CPython 3.12 with 10-character deltas:

| accumulated text | before | after |
|---|---|---|
| 100 KB | 5.97 ms | 0.98 ms |
| 250 KB | 29.53 ms | 2.83 ms |
| 1 MB | 659 ms | 15.5 ms |

Results match the replaced expressions for every input where the key exists and holds a string, key order and exception type included. Buffering into a list and joining at the end was the alternative, but each delta is read back immediately by the stream save, and joining per read measured 666 ms at 250 KB.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -303,6 +303,13 @@ def tool_result_content(tool_result: Any) -> str:
     return str(tool_result)
 
 
+def append_text(item: dict, key: str, value: str) -> None:
+    text = item[key]
+    item[key] = ''  # dropping the dict's reference lets CPython extend the str in place
+    text += value
+    item[key] = text
+
+
 def merge_streamed_reasoning_details(target: list, details) -> None:
     items = details if isinstance(details, list) else [details]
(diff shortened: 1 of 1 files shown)


Classic298:
@tjbck ready for review. String appends through a dict entry recopy the whole accumulated text, so a streamed message is assembled in quadratic time on the event loop.

Measured 140x at 100k chunks: 3.2 s of pure copying down to 22 ms.

Classic298:
@tjbck

Classic298:
cc @tjbck

PR #28851, opened 2026-08-20
refactor: remove the deprecated automation worker alias and unused terminal helper

Classic298:
The automation module wraps the scheduler loop under its old name, with a comment saying it is kept so stale imports still work. Startup imports the real name directly and nothing anywhere in the tree reaches the alias, so it only provides a second name for the same loop. A helper that pushes a working directory to a terminal server sits beside it with no caller either.

That helper was the only reader of the terminal block in the automation payload. The block still validates and persists, so this change leaves it in place, but nothing consumes it any more.

This removes around 55 lines and leaves one entry point for the scheduler.

> [!NOTE]


tjbck:
Better to not touch backend codebase for best effort compatibility with functions.

PR #28858, opened 2026-08-20
perf: stop re-parsing the whole tool-argument buffer on every streamed chunk

Classic298:
Converting an OpenAI stream to Anthropic events buffers each tool call's arguments and, to find out when the JSON is complete, parsed the entire buffer again on every chunk. A tool call with large arguments pays that parse thousands of times, and the cost grows with the square of the argument size.

The parse now runs only when the buffer could actually be complete. A JSON object can only close on its final brace, so a chunk that does not end there cannot complete it. Arguments that are not an object, or that start with whitespace, keep parsing on every chunk exactly as before.

Measured on CPython 3.12 with 130 KB of tool arguments over 7648 chunks:

| | before | after |
|---|---|---|
| parses | 7648 | 1 |
| time | 382 ms | 0.82 ms |

The block closes on exactly the same chunk as before, verified by replaying randomized fragmentations of objects with braces inside strings, escaped characters, unicode escapes, arrays, bare scalars, leading and trailing whitespace and a buffer that never completes, against both JSON backends.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/anthropic.py
+++ b/backend/open_webui/utils/anthropic.py
@@ -878,11 +878,7 @@ async def openai_stream_to_anthropic_stream(openai_stream_generator, model: str 
                             if (
                                 tool['started']
                                 and not tool['stopped']
-                                and (
-                                    tool['arguments'][-1] == '}'
-                                    or tool['arguments'][-1].isspace()
-                                    or tool['arguments'][0] != '{'
-                                )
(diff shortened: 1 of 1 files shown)


Classic298:
@tjbck ready for review. Converting an OpenAI stream to Anthropic events re-parsed the entire buffered tool-call arguments on every chunk just to learn whether the JSON was complete, so the cost grows with the square of the argument size. It now parses only when the buffer could actually be complete.

PR #28860, opened 2026-08-20
perf: cache the serialized builtin tool spec instead of deep-copying it per request

Classic298:
Every chat request hands each builtin tool a fresh copy of its cached spec, because callers mutate what they get. That copy was a full deepcopy of a nested dict, repeated per tool per message.

The builder now caches the spec already serialized, so a request only parses it back. Parsing is what produces the independent tree callers mutate, and the cached value becomes an immutable string, so a request can no longer reach the cached object at all.

Measured on CPython 3.12 with a 1.1 KB spec and 20 builtin tools per request:

| | before | after |
|---|---|---|
| stdlib json, the default | 276.2 us | 66.4 us |
| orjson | 279.5 us | 37.5 us |

Builtin specs are plain JSON by construction: pydantic normalizes every default before it reaches the schema, so a tuple, set, enum or datetime cannot appear in one, and an unserializable default is dropped rather than embedded.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/tools.py
+++ b/backend/open_webui/utils/tools.py
@@ -962,14 +962,15 @@ def get_builtin_function_introspection(func: Callable):
 
 
 @cache
-def build_builtin_tool_spec(func: Callable) -> dict:
+def build_builtin_tool_spec_json(func: Callable) -> str:
     pydantic_model = convert_function_to_pydantic_model(func, get_builtin_function_introspection(func))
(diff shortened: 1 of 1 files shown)


Classic298:
@tjbck ready for review. Every chat request hands each builtin tool a full deepcopy of its cached spec, once per tool per message. Caching it already serialized means a request only parses it back, which is what produces the independent tree callers mutate anyway.

PR #28886, opened 2026-08-21
fix: expire abandoned response stream entries in Redis

Classic298:
In-flight response streams are persisted into one global Redis hash so reconnecting clients can restore streaming messages. The fields are only deleted when a task ends cleanly: a SIGKILLed or OOM-killed container leaks its multi-MB field forever and the hash grows without bound across restarts.

This adds REDIS_RESPONSE_STREAM_TTL (seconds, default 3600, values <= 0 disable) and re-arms a per-field TTL with HEXPIRE after every write, since HSET clears a field's TTL. Abandoned fields now expire on their own once a stream stops being written. HEXPIRE needs Redis 7.4+ and ACL access; if the first call reports the command unavailable the writer logs one warning and stops trying, leaving older Redis versions behaving exactly as before.

Fields leaked before this change carry no TTL and can be cleared once with redis-cli DEL open-webui:tasks:response_streams (adjust for REDIS_KEY_PREFIX).

> [!NOTE]

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -377,6 +377,13 @@ REDIS_CLUSTER = os.getenv('REDIS_CLUSTER', 'False').lower() == 'true'
 
 REDIS_KEY_PREFIX = os.getenv('REDIS_KEY_PREFIX', 'open-webui')
 
+# TTL in seconds for in-flight response-stream state; values <= 0 disable expiry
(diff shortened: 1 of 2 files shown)


Classic298:
@tjbck ready for review. A SIGKILLed or OOM-killed container leaks its multi-MB response stream field into Redis forever, and that hash grows without bound across restarts. Adds a TTL that is re-armed after every write.

PR #28887, opened 2026-08-21
perf: write response streams to Redis incrementally

Classic298:
Persisting an in-flight response stream rewrote the whole accumulated message into one global hash field on every streamed delta: quadratic write volume, about 2 GB for a 32k-token response, about 16 GB for a tool-heavy 8 MB turn, 43000x amplification observed on a production fleet.

Streams now live in per-stream keys with a TTL from creation: one hash field per output item, an append-only journal following the actively growing text lane and an append-only content string. Unchanged items are never rewritten and slowly drifting lanes batch under a geometric budget, keeping bytes proportional to new content while the text the UI renders stays byte-exact at every save, so reloading mid-response never shows gaps.

Measured: 2263x fewer bytes on a 4000-delta message, 356x on a tool-heavy turn; the worst case, two items alternating growth on every delta, stays at roughly the old writer's cost, and the writer holds about twice the in-flight message in memory. Throttling the saves instead was rejected because it keeps the quadratic total and serves stale reconnect overlays.

Redis Cluster, Sentinel and the non-Redis fallback behave as before, rolling upgrades keep old-format entries readable and this supersedes the best-effort HEXPIRE of the previous fix.

Builds on https://github.com/open-webui/open-webui/pull/28886; only the second commit is new here.

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1752,7 +1752,6 @@ async def chat_completion(
                             except asyncio.CancelledError:
                                 pass
                 elif chat_id:
-                    # mirrors middleware's fallback: with no task the stream is keyed by message id
                     await clear_response_stream(request.app.state.redis, metadata.get('message_id'))
             except Exception:
                 pass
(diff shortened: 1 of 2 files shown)


Classic298:
Replaced by a smaller change on `perf/incremental-stream-writes-minimal`: same append-based win in 60 lines instead of ~300, with no complexity suppressions, and it keeps the resume snapshot exact.

Issue #25223, opened 2026-05-30
issue: Web search query bleeds context from previous messages in same conversation


TennTechTinkerer:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Pip Install

### Open WebUI Version

v0.9.5

### Ollama Version (if applicable)

Latest (installed via official install.sh script)

### Operating System

macOS (Apple Silicon M4)

### Browser (if applicable)

Chrome

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

Each web search query should be generated based only on the current user message, not previous messages in the conversation.

### Actual Behavior

Search queries include context from all previous messages in the conversation. For example, asking about weather then asking about sports in the same chat causes the sports search to also include weather terms from the earlier message.

### Steps to Reproduce

1. Install Open WebUI v0.9.5 via pip in a Python venv
2. Enable web search (any engine - tested with SearXNG and DDGS)
3. Start a new chat with any model
4. Ask: "What is the weather in New York, NY?"
5. Wait for response
6. In the same chat ask: "Who won the NBA game last night?"
7. Observe the search queries shown above the response - they include weather terms from step 4

### Logs & Screenshots

Root cause identified in open_webui/utils/task.py

Buggy code (line ~362):
template = replace_messages_variable(template, messages)

[...]

Issue #10996, opened 2025-02-28
Web search context


mikygit:
Hello,
 I'm running the latest versions of ollama and open webui.

I came back to trying to make the web search behave as it should after having tried and failed a couple of months ago.
And I can see some improvements, congrats!

Still, there is still something that do not work as expected from my point of view:
When activating the 'web search' and querying the llm multiple times consecutively, it appears that the 'web search' queries are appended to the previous ones which eventually confuses the llm with a context that might not be very helpful, as you can see in the pic below:

![Image](https://github.com/user-attachments/assets/32961e9d-d977-4189-81df-03da9821c4d5)

Is there a way to clear the search requests between prompts?
Thanx.

PR #28892, opened 2026-08-21
feat: add role filter to the {{MESSAGES}} template variable

MarcPlatini:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Discussion [#28893](https://github.com/open-webui/open-webui/discussions/28893)
- [x] **Testing:** Manual tests have been performed to verify the feature works correctly and does not introduce regressions.
- [x] **No Unchecked AI Code:** This PR has undergone thorough human review AND manual testing.
- [x] **Architecture:** No new setting is introduced — the capability is added to the existing template syntax and applied to the default template.

# Changelog Entry

### Description

The `{{MESSAGES}}` template variable can select messages **by position** (`:START:n`, `:END:n`, `:MIDDLETRUNCATE:n`) and truncate their **content** (`|middletruncate:n`, `|start:n`, `|end:n`), but it cannot select them **by role** — every selected message is rendered, `system` message included.

This PR adds an optional `|ROLES:` filter and uses it in the default query-generation template.

**Why it matters for query generation.** `generate_queries()` renders the chat history into the query-generation prompt for both retrieval (`chat_completion_files_handler`) and web search (`chat_web_search_handler`). With the default `{{MESSAGES:END:6}}`, the system prompt is included, so it is sent to the task model — potentially a different, external provider (`TASK_MODEL_EXTERNAL`) — and it steers the queries that then go to the vector store or to an external search engine.

In #25223, a similar report was answered with "you can edit the query generation prompt in the admin panel". The current syntax cannot express this particular need, though: there is no way to keep several conversation turns while excluding the system message. `{{MESSAGES:END:1}}` and `{{prompt}}` both exclude it but discard multi-turn context, which is what makes query generation useful in the first place.

**Before** — the `### Chat History:` block rendered into the query-generation prompt, for an HR assistant answering a follow-up question:

```
SYSTEM: You are ACME Corp's internal HR assistant. Always answer in a warm, informal tone
        and address the employee by their first name. Never mention or compare against
        competitors such as Globex or Initech. If a question touches on salary bands,
        redirect the employee to their manager instead of answering directly. Internal
        policy reference: HR-2024-REV3. Escalation contact: hr-escalation@acme.example.
USER: Hi! I'm planning some time off next spring.
ASSISTANT: Happy to help with that. Are you looking at the annual leave policy,
           or something more specific like parental or unpaid leave?
USER: Annual leave. How many days do I get?
ASSISTANT: That depends on your contract type and seniority. Let me check the policy.
USER: And what about part-timers?
```

**After:**

```
USER: Hi! I'm planning some time off next spring.
ASSISTANT: Happy to help with that. Are you looking at the annual leave policy,
           or something more specific like parental or unpaid leave?
USER: Annual leave. How many days do I get?
ASSISTANT: That depends on your contract type and seniority. Let me check the policy.
USER: And what about part-timers?
```

The conversation context that makes `And what about part-timers?` resolvable is kept; the persona, the internal policy reference and the escalation address are not sent to the task model.

### Added

- Optional `|ROLES:<roles>` filter for the `{{MESSAGES}}` template variable:

  ```
  {{MESSAGES:END:6|ROLES:user,assistant}}
  {{MESSAGES|ROLES:user|middletruncate:500}}
  ```

  - Roles are comma- or space-separated and matched case-insensitively; the `ROLES` keyword itself is case-insensitive.
  - Role filters chain with the existing content filters, in any order.
  - Role filtering is applied **before** the positional selector, so `{{MESSAGES:END:6|ROLES:user}}` renders the last 6 *user* messages rather than whichever user messages happen to fall inside the last 6 of the whole history.
  - An empty or unparsable role list leaves the messages untouched, so a typo degrades to the previous behaviour instead of silently emptying the chat history.

- `apply_role_filter()` in `backend/open_webui/utils/task.py`, alongside the existing `apply_content_filter()`.

The filter is available to every template that renders `{{MESSAGES}}` (title, tags, follow-up, autocomplete, and the `COMPACTED_MESSAGES` / `RECENT_MESSAGES` variables used by context compaction).

### Changed

- `DEFAULT_QUERY_GENERATION_PROMPT_TEMPLATE` now uses `{{MESSAGES:END:6|ROLES:user,assistant}}`, so retrieval and web search query generation no longer receives the system prompt out of the box. Only this default template is changed; the other defaults are untouched.
- `replace_messages_variable()` was refactored from four alternated regex branches with seven positional groups to a single pattern with named groups and a repeatable filter suffix, so that chained filters stay readable.

### Breaking Changes

None.

Templates without `|ROLES` behave exactly as before — verified byte-for-byte against the previous implementation (see Testing).

Admin-customised templates (`QUERY_GENERATION_PROMPT_TEMPLATE` / `task.query.prompt_template`) keep taking precedence and are unaffected; their authors can opt in by adding `|ROLES:user,assistant` to their `{{MESSAGES}}` variable.

---

### Additional Information

Relates to discussion [#28893](https://github.com/open-webui/open-webui/discussions/28893) (this request was opened as an issue and converted to Discussions by a maintainer).

**Note on the default template change**

On installations using the default template, the system prompt is no longer part of the context the task model reads when generating queries. That is the intent of the change, but it is worth stating explicitly for anyone who was relying on it indirectly.

The clearest example is language. The task asks for queries "in the given language", and the language is inferred from the rendered context. Today a system prompt such as *"You are a French-speaking assistant, always answer in French"* is part of that context; after this change the language is inferred from the user and assistant turns alone. In most conversations that is the same signal or a better one, since the user writes in their own language — but it can differ on a short or ambiguous opening turn.

Setups that prefer the previous behaviour can restore it by setting a custom query-generation template with a plain `{{MESSAGES:END:6}}`.

Related (all closed): #25223, #22329, #10996.

**Testing**

The repository has no Python test suite covering these helpers, so verification was done with standalone scripts exercising the extracted functions:

- **Backward compatibility (18/18):** every pre-existing form — `{{MESSAGES}}`, `:START:n`, `:END:n`, `:MIDDLETRUNCATE:n`, content filters, malformed variants left as literals, `None` messages — produces byte-identical output between the implementation on `dev` and this one.
- **New syntax (18/18 + 7/7):** role filtering alone and combined with selectors and content filters; ordering of chained filters; case-insensitivity of both the keyword and the role names; comma and space separators; empty and unknown role lists; multimodal (list) message content; messages without a `role` key.
- **Custom variable names:** `COMPACTED_MESSAGES` and `RECENT_MESSAGES` (used by `utils/context_compaction.py`) still expand correctly and remain isolated from `MESSAGES`.
- **End-to-end:** rendering the actual `DEFAULT_QUERY_GENERATION_PROMPT_TEMPLATE` from `config.py` against a conversation containing a system prompt confirms the system message is no longer rendered.
- `ruff format` reports no changes. `ruff check` reports no new findings on the touched files (13 pre-existing findings, down from 15 — the refactor removed two now-unused locals).

### Screenshots or Videos

Not applicable — backend-only change with no UI surface.

> [!NOTE]

### DIFF
--- a/backend/open_webui/config.py
+++ b/backend/open_webui/config.py
@@ -2320,7 +2320,7 @@ Strictly return in JSON format:
 
 ### Chat History:
 <chat_history>
-{{MESSAGES:END:6}}
+{{MESSAGES:END:6|ROLES:user,assistant}}
 </chat_history>
 """
 
--- a/backend/open_webui/utils/task.py
+++ b/backend/open_webui/utils/task.py
@@ -156,6 +156,21 @@ def truncate_content(content: str, max_chars: int, mode: str = 'middletruncate')
         return f'{content[:half]}...{content[-(max_chars - half) :]}'
 
 
+def apply_role_filter(messages: list[dict], roles_str: str) -> list[dict]:
+    """Keep only the messages whose role is listed in roles_str.
+
+    roles_str is a comma separated list of roles, like 'user,assistant'.
+    Roles are matched case-insensitively and may also be separated by spaces.
+    An empty list leaves the messages untouched, so a typo degrades to the
+    previous behaviour instead of silently emptying the chat history.
+    """
+    roles = {role.lower() for role in re.split(r'[,\s]+', roles_str) if role}
+    if not roles:
+        return messages
+
+    return [message for message in messages if (message.get('role') or '').lower() in roles]
+
+
 def apply_content_filter(messages: list[dict], filter_str: str) -> list[dict]:
     """Apply a content filter to each message's content.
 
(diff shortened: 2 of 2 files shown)


tjbck:
This will be kept open as a reference, but would be implemented with our liquid template langauge support.

Issue #28909, opened 2026-08-21
bug: redis_task_command_listener dies silently on a Redis restart and never resubscribes, breaking cross-worker task stop


kinorai:
## Summary

`redis_task_command_listener` subscribes to `<prefix>:tasks:commands` once at startup. When the Redis server restarts, `pubsub.listen()` raises `ConnectionError`, the coroutine exits, and nothing restarts it. From that point cross-worker task commands (stop/cancel of a running generation) are lost until the Open WebUI process restarts. The failure is silent: the task handle is kept in `app.state`, so there is no "Task exception was never retrieved" traceback and no log line.

This is the same class of failure that #26779 reported for the idle socket timeout. #27104 fixed that trigger by setting the socket timeout to `None`, but a Redis restart still kills the listener on v0.11.0, which already contains that fix. The listener remains one unhandled connection exception away from permanent death.

## Affected code

`backend/open_webui/tasks.py` (v0.11.0, identical on current main):

```python
async def redis_task_command_listener(app):
    redis: Redis = app.state.redis
    pubsub = redis.pubsub()
    await pubsub.subscribe(REDIS_PUBSUB_CHANNEL)

    async for message in pubsub.listen():
        if message["type"] != "message":
            continue
        try:
            ...
        except Exception as e:
            log.exception(f"Error handling distributed task command: {e}")
```

The `except` only covers message handling. An exception raised by `pubsub.listen()` itself leaves the `async for` and ends the coroutine. `backend/open_webui/main.py` creates the task once, with no watchdog:

```python
app.state.redis_task_command_listener = asyncio.create_task(redis_task_command_listener(app))
```

## Reproduction

1. Run Open WebUI 0.11.0 with `REDIS_URL` set. A single replica is enough.
2. `PUBSUB CHANNELS` on the Redis server shows `socketio` and `open-webui:tasks:commands`.
3. Restart the Redis server (in k8s: delete the redis pod).
4. After the reconnect, `PUBSUB CHANNELS` shows only `socketio`. python-socketio's `AsyncRedisManager` resubscribes on its own; this listener does not. No error is logged.
5. Task stop commands published to `open-webui:tasks:commands` are no longer delivered to this worker.

## Expected

The listener survives a Redis restart, like the socket.io manager does: retry subscribe+listen with backoff, and exit only on task cancellation.

## Related

- #26779 / #27104: same symptom via the idle socket timeout; the merged fix does not cover connection loss.
[...]

PR #28910, opened 2026-08-21
fix: resubscribe the redis task command listener after a connection loss

kinorai:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Closes #28909
- [ ] **Documentation:** No documentation change needed. No new setting, no behaviour change beyond the listener surviving a reconnect.
- [x] **Testing:** Manual tests performed against a real Redis server, plus unit tests. Details below.
- [x] **No Unchecked AI Code:** Reviewed and manually tested.
- [x] **Self-Review:** Done.
- [x] **Architecture:** No new setting. The retry delay is a module constant, matching the existing constants in `tasks.py`.
- [x] **Git Hygiene:** One commit, one logical change, rebased on `dev`.
- [x] **Title Prefix:** `fix`

# Changelog Entry

### Description

`redis_task_command_listener` subscribes once and then iterates `pubsub.listen()`, with a `try/except` around message handling only. Any connection error raised by `listen()` itself leaves the `async for` and ends the coroutine, and `main.py` creates that task once with no watchdog. So a Redis restart permanently breaks cross-worker task stop for that worker, and it does so silently: the task handle is kept in `app.state`, so asyncio never reports the exception and nothing is logged. `PUBSUB CHANNELS` on the server shows `open-webui:tasks:commands` gone while `socketio` is still there, because python-socketio's `AsyncRedisManager` resubscribes on its own. #27104 removed the idle-socket-timeout trigger for this failure; connection loss remained, so the listener was still one exception away from permanent death.

Subscribe and listen now run inside a reconnect loop: on any exception the drop is logged, the dead pubsub is closed, and the listener resubscribes after `REDIS_PUBSUB_RETRY_DELAY` seconds. `CancelledError` is re-raised, so the shutdown path in `main.py` still stops the task immediately. The per-message `try/except` is unchanged. The delay is exponential rather than a fixed constant, and it mirrors what this application already runs for its other Redis pub/sub channel: `AsyncRedisManager._redis_listen_with_retries` in python-socketio (`src/socketio/async_redis_manager.py`, pinned at 5.16.2) starts at 1 second, doubles after each failure, caps at 60 and resets to 1 after a successful subscribe. The task command listener now uses the same envelope, so a Redis blip costs about a second of deafness to cross-worker stop commands instead of five, while a Redis that stays down cannot turn the retry into a hot loop. Once the delay reaches the 60 second cap the listener logs at error level with `exc_info`, which turns a permanent failure (a rotated password raising `AuthenticationError`, for example) into a traceback an operator can act on, instead of the same warning line forever. I kept `listen()` rather than switching to `get_message(timeout=...)` polling as #26794 proposed: with `REDIS_SOCKET_TIMEOUT` now defaulting to `None` (#27104), the blocking `listen()` no longer times out on idle, so the polling rework is no longer needed and this change stays a wrapper around the existing loop.

### Fixed

- Fixed the Redis task command listener dying permanently and silently on a Redis restart or any other pubsub connection loss, which broke stopping a generation across workers until the Open WebUI process restarted.

---

### Additional Information

Testing:

- Real Redis (`redis:8-alpine` in Docker), listener driven with a real `redis.asyncio` client: a `stop` command is delivered, the server is restarted, and after the reconnect `PUBSUB CHANNELS` again lists `open-webui:tasks:commands` and a second `stop` command cancels its local task. The log line on the drop is `Task command listener disconnected, resubscribing in 5s: Error while reading from localhost:63790 : (104, 'Connection reset by peer')`. On `dev` without this patch the same script prints `channels after restart: []`, the second stop is never delivered, and the listener task is already done with nothing logged.
- Unit tests with a stub pubsub, 4 cases, all passing: (1) `listen()` raises `ConnectionError` once, the listener resubscribes and processes a later `stop` message, sleeping exactly one retry delay and closing the dead pubsub; (2) cancelling the task while it waits in `listen()` propagates `CancelledError` with no backoff sleep; (3) `CancelledError` raised from inside `listen()` propagates; (4) a malformed message is logged and the loop keeps running on the same connection. Case 1 fails on `dev` without this patch.

> [!NOTE]

### DIFF
--- a/backend/open_webui/tasks.py
+++ b/backend/open_webui/tasks.py
@@ -20,25 +20,38 @@ REDIS_TASKS_KEY = f'{REDIS_KEY_PREFIX}:tasks'
 REDIS_ITEM_TASKS_KEY = f'{REDIS_KEY_PREFIX}:tasks:item'
 REDIS_RESPONSE_STREAMS_KEY = f'{REDIS_KEY_PREFIX}:tasks:response_streams'
 REDIS_PUBSUB_CHANNEL = f'{REDIS_KEY_PREFIX}:tasks:commands'
+REDIS_PUBSUB_RETRY_DELAY = 5
 
 
 async def redis_task_command_listener(app):
     redis: Redis = app.state.redis
-    pubsub = redis.pubsub()
-    await pubsub.subscribe(REDIS_PUBSUB_CHANNEL)
-
-    async for message in pubsub.listen():
-        if message['type'] != 'message':
-            continue
+    while True:
+        pubsub = redis.pubsub()
         try:
-            command = JSONCodec.loads(message['data'])
-            if command.get('action') == 'stop':
-                task_id = command.get('task_id')
-                local_task = tasks.get(task_id)
-                if local_task:
-                    local_task.cancel()
+            await pubsub.subscribe(REDIS_PUBSUB_CHANNEL)
+
+            async for message in pubsub.listen():
+                if message['type'] != 'message':
+                    continue
+                try:
+                    command = JSONCodec.loads(message['data'])
+                    if command.get('action') == 'stop':
+                        task_id = command.get('task_id')
+                        local_task = tasks.get(task_id)
+                        if local_task:
+                            local_task.cancel()
+                except Exception as e:
+                    log.exception(f'Error handling distributed task command: {e}')
+        except asyncio.CancelledError:
+            raise
         except Exception as e:
-            log.exception(f'Error handling distributed task command: {e}')
+            log.warning(f'Task command listener disconnected, resubscribing in {REDIS_PUBSUB_RETRY_DELAY}s: {e}')
+            await asyncio.sleep(REDIS_PUBSUB_RETRY_DELAY)
+        finally:
+            try:
+                await pubsub.aclose()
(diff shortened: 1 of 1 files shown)


kinorai:
Pushed a follow-up commit for the two review findings. The fixed 5s delay is now the python-socketio `AsyncRedisManager` backoff, 1s, doubling, capped at 60s, reset to 1s on a successful subscribe, so the first retry no longer leaves a five second deaf window after a Redis blip. Once the delay reaches the cap the log escalates from warning to error with `exc_info`, so a listener that can never resubscribe (rotated password, for instance) is diagnosable instead of an endlessly repeated warning. Tests cover the doubling, the cap, the reset, and the error-level escalation, and the real-Redis restart check still recovers.

kinorai:
Superseded. `bf3a58dbc` on `dev` fixes this with the same shape — reconnect loop around `subscribe()` + `listen()`, per-message handler untouched, `CancelledError` re-raised, dead pubsub closed in `finally`, exponential backoff reset after a successful subscribe.

It also covers the wider case @Classic298 confirmed: the `except Exception` catches any pubsub drop, not a restart only.

Closing this PR. #28909 can be closed too.

PR #28945, opened 2026-08-23
refac: stream web-fetched documents to disk instead of buffering them

Classic298:
The binary branch of the web fetch read the entire response body into memory before writing it out. It now streams in blocks, applies the configured file size limit the same way the sibling URL endpoint already does, and removes the temporary file when a download fails partway instead of leaving it behind.

> [!NOTE]


Classic298:
@tjbck ready for review. The binary branch of the web fetch reads the whole response body into memory and does not apply the configured file size limit, so a large remote file can exhaust memory. It now streams to disk in blocks.

PR #28947, opened 2026-08-23
refac: bound how far an archive-based document may unpack

Classic298:
Zip-based documents (docx, xlsx, pptx, odt, epub) are checked before parsing and rejected when they unpack to far more than they store, or when a member uses a compression method the block-wise reader cannot bound.

RAG_FILE_MAX_UNPACKED_SIZE sets the upper bound and defaults to 100 MB, well clear of ordinary documents: a 120k-row spreadsheet unpacks to about 33 MB. Set it to 0 to disable the check. Only the local parsing path is affected, so deployments that hand documents to Tika, Docling or another external engine are unchanged.

> [!NOTE]

### DIFF
--- a/backend/open_webui/env.py
+++ b/backend/open_webui/env.py
@@ -812,6 +812,10 @@ MINERU_MAX_MARKDOWN_BYTES = (
     int(os.getenv('MINERU_MAX_MARKDOWN_BYTES')) if os.getenv('MINERU_MAX_MARKDOWN_BYTES') else None
 )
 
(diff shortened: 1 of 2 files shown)


Classic298:
@tjbck ready for review. A zip-based document (docx, xlsx, pptx, odt, epub) can unpack to far more than it stores, so parsing one can exhaust memory. Bounded by `RAG_FILE_MAX_UNPACKED_SIZE`, default 100 MB, and only the local parsing path is affected.

PR #28968, opened 2026-08-24
fix: drop orphaned file vector collection on delete; expose batch-add warnings

simpleqt:
delete_file_by_id called VECTOR_DB.delete() with no ids/filter, which
is a no-op on most backends, orphaning the file-{id} collection. Use
has_collection + delete_collection, the same pattern as the knowledge
file-remove path.

KnowledgeFilesResponse has no warnings field, so pydantic (extra=ignore)
silently dropped the warnings kwarg the batch-add endpoint passes when
some files fail — the response was identical to full success. Declare
the field.

---

**Changelog:** fixed — deleting a file now drops its vector collection; knowledge batch-add failures are surfaced in the response warnings.

### DIFF
--- a/backend/open_webui/routers/files.py
+++ b/backend/open_webui/routers/files.py
@@ -1027,7 +1027,9 @@ async def delete_file_by_id(
         if result:
             try:
                 await asyncio.to_thread(Storage.delete_file, file.path)
-                await ASYNC_VECTOR_DB_CLIENT.delete(collection_name=f'file-{id}')
+                file_collection = f'file-{id}'
+                if await ASYNC_VECTOR_DB_CLIENT.has_collection(collection_name=file_collection):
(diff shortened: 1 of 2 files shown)


simpleqt:
No existing issue covers this. Why the change is needed: (1) `delete_file_by_id` called `VECTOR_DB.delete()` with no ids/filter — a no-op on most backends — so the per-file `file-{id}` vector collection was orphaned on every deletion; this uses `has_collection` + `delete_collection`, the same pattern as the knowledge file-remove path. (2) `KnowledgeFilesResponse` has no `warnings` field, so pydantic silently dropped the warnings the batch-add endpoint passes when some files fail — the response looked identical to full success. Both found while reviewing file-deletion and knowledge-upload code paths.

Issue #28952, opened 2026-08-23
Model workspace overrides silently skipped when base_model_id equals model id


aberhamm:
## Description

When a workspace model override has `base_model_id` set to the same value as `id` (which is what the admin UI sets by default when editing an existing model), the `get_all_models()` function in `backend/open_webui/utils/models.py` silently skips merging the DB customizations into the runtime model dict.

## Root Cause

In `get_all_models()`, the merge logic has two branches:

1. `if custom_model.base_model_id is None:` — treats it as a direct override of a base model. This path correctly applies all DB overrides (`model['info'] = custom_model.model_dump()`).

2. `elif custom_model.is_active:` — treats it as a separate custom model. This path first checks `if custom_model.id in existing_ids: continue`, which **skips** any custom model whose ID already exists in the base models from API providers.

When `base_model_id == id` (e.g., both are `"gpt-4o"` or `"qwen3.6-uncensored"`), the first branch doesn't match (`base_model_id is not None`), and the second branch skips it (`id in existing_ids`). The DB overrides are never applied.

## Impact

Any model customization set via the admin panel is silently ignored at runtime for models that come from an API provider (OpenAI-compatible, Ollama, etc.). This includes:

- `builtinTools` — per-model builtin tool enable/disable
- `toolIds` — tool server assignments
- `defaultFeatureIds` — default feature toggles
- `system` — system prompts
- `function_calling` param — native function calling toggle
- `tags`, `description`, and other metadata

Settings appear to save successfully in the admin UI but have no effect.

## Steps to Reproduce

1. Connect an OpenAI-compatible API provider (e.g., Ollama, vLLM)
2. Go to Admin Panel → Models → click on an existing model (e.g., `qwen3.6`)
3. Change any setting (e.g., set a system prompt, toggle builtin tools, assign tool servers)
4. Save
5. Start a new chat — the customization is not applied
6. Check the `/api/models` response — `info.meta` is missing the saved fields

## Expected Behavior

Model customizations saved via the admin UI should be reflected in the runtime model dict and applied during chat.

## Proposed Fix

Change line 158 in `backend/open_webui/utils/models.py` from:
```python
if custom_model.base_model_id is None:
```
to:
```python
if custom_model.base_model_id is None or custom_model.base_model_id == custom_model.id:
```

This treats `base_model_id == id` as equivalent to `base_model_id is None` — both represent a direct override of a base model, not a separate custom model.

## Environment

- Open WebUI version: 0.11.0
[...]

PR #28995, opened 2026-08-24
fix(models): treat base_model_id == id as direct override

Battleplus:
## Summary

When the admin UI edits an existing model, it sets `base_model_id` to the same value as the model's own `id`. The previous condition only checked for `base_model_id is None`, so these models fell into the `elif` branch and were silently skipped by the `id in existing_ids` guard — their workspace customizations (name, params, metadata, actions, filters) were never applied.

Extend the override branch to also match when `base_model_id` equals the model's own `id`, so DB customizations are merged as intended.

Fixes #28952

> [!NOTE]


Classic298:
we already have an open PR for this we dont need a second one

Issue #29035, opened 2026-08-25
issue: JS error on frontend when attempting to render streaming responses with thinking models


CrushedAsian255:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I have checked whether this issue is already fixed on the `dev` branch or in the latest source.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.11.1

### Ollama Version (if applicable)

N/A

### Operating System

macOS Sonoma 14.8.9 x86_64

### Browser (if applicable)

Waterfox 6.7.0 (aarch64), Thorium 119.0.6045.214 (Official Build) (arm64) 

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

As the WebSockets sends chat deltas to the client, the client renders them correctly, first the reasoning and then the content.

### Actual Behavior

It hangs until reasoning completes, then the first chunk of the content response loads, but then does not load any further chunks due to a Svelte error (https://svelte.dev/e/each_key_duplicate).

On browser reload, the message that has streamed to the server as of refresh all load, however streaming does not continue from there.
Once the message has finished generating, it all loads on the client. 

### Steps to Reproduce

1. Create an instance of Open WebUI
2. Add a model that supports thinking and sends thinking to the user (i.e. DeepSeek)
3. Go to the new chat page
[...]

PR #29053, opened 2026-08-26
fix: stop streaming responses breaking on a duplicate output key

Classic298:
With reasoning-capable models the chat froze mid-stream: the first chunk of the answer appeared, nothing followed, and the whole message only showed up once generation finished. The browser console showed a Svelte each_key_duplicate error.

When a stream event addresses an output slot past the end of the array, the missing slots were filled with the event's own item, id included, so a gap of two left two entries claiming the same id. The next chunk for that item was matched by id, landed in the first of the two, and the rendered list ended up with two items sharing a key, which Svelte refuses to update.

Only the addressed slot now takes the event's item, and the slots before it are anonymous placeholders. Replayed the reported event sequence against the real code: keys are unique again and the chunks stay in order instead of being split across the copies.

Fixes : https://github.com/open-webui/open-webui/issues/29035

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -5222,7 +5222,12 @@ async def streaming_chat_response_handler(response, ctx):
                                                 reasoning_detail_items,
                                             )
                                             await save_current_response_stream()
-                                            data = None
+                                            # Providers such as OpenRouter send reasoning_details
+                                            # alongside the reasoning text: only drop the event when
+                                            # the details were all there was to report, otherwise the
+                                            # reasoning delta never reaches the client.
+                                            if not reasoning_content:
+                                                data = None
 
                                     if value:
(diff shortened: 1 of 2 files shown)


CrushedAsian255:
@Classic298 
I have found an issue with the pull request.
When streaming, it doesn't hang or error, however it doesn't render the reasoning trace at all. What happens is:
 - While model is thinking, the reasoning is not visible. 
 - Model content tokens stream in (properly, so that works)
 - There is no "Thought for" thing
 - Model finishes generating. 
 - Only then does the reasoning appear.

Classic298:
@CrushedAsian255 please retry now.

CrushedAsian255:
> @CrushedAsian255 please retry now.

I have done some quick testing, it seems to work well. Will do more testing later

Issue #29048, opened 2026-08-26
issue: Google OAuth login fails on SQLite — "Python int too large to convert to SQLite INTEGER" (21-digit Google sub)


osalloum:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I have checked whether this issue is already fixed on the `dev` branch or in the latest source.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker (`ghcr.io/open-webui/open-webui:main` on Kubernetes)

### Open WebUI Version

v0.11.1 (`sha256:c792953395b43d4f49085fa4026850c5e923578345ebeaf56eabd980bcfc273a`). Still present on `dev` at `users.py:369`.

### Ollama Version (if applicable)

N/A

### Operating System

Kubernetes (Linux); database: SQLite

### Browser (if applicable)

N/A — server-side failure, browser-independent

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup**.
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**.

### Expected Behavior

Signing in with Google on a SQLite instance logs the user in.

### Actual Behavior

Every Google login fails. The user is redirected to `/auth?error=[ERROR: Error during OAuth process]`.

The OAuth handshake itself succeeds — the token exchange returns `200 OK`. The failure is in the user lookup immediately afterwards:

```
Error during OAuth process: Python int too large to convert to SQLite INTEGER
```

`get_user_by_oauth_sub()` in `backend/open_webui/models/users.py`:

```python
sub_expr = User.oauth[provider]['sub'].as_string()
query = select(User).where(sub_expr == sub)
# SQLite preserves JSON numeric type here; Postgres ->> already compares numeric JSON as text.
if session.get_bind().dialect.name == 'sqlite' and sub.isdecimal():
    query = select(User).where(or_(sub_expr == sub, sub_expr == int(sub)))
```

[...]

PR #29078, opened 2026-08-27
fix: SQLite OAuth overflow for 21-digit Google sub IDs (#29048)

ranka23:
### Description

- Fixes SQLite OAuth login failure for 21-digit Google sub IDs
- Prevents `int()` overflow by comparing as string only when number exceeds SQLite INTEGER (64-bit) limit
- Manually tested with 21-digit sub value

### Fixed

- `get_user_by_oauth_sub` now handles large OAuth sub values safely

### Testing

- [x] Manual test: verified 21-digit sub `'123456789012345678901'` is handled correctly without overflow

Closes #29048

---


gaby:
Duplicate of https://github.com/open-webui/open-webui/pull/29056

PR #29093, opened 2026-08-27
fix(auth): handle unparseable stored password hash gracefully in authenticate_user (#29092)

### LINKED ISSUE #29092: bug: authenticate_user() raises unhandled exception (→ 500) instead of returning None for an unparseable stored password hash
In `open_webui/models/auths.py`, `Auths.authenticate_user()` calls `await verify_password(credential.password)` with no exception handling:

```python
if not await verify_password(credential.password):
    return
```

For an ordinary wrong password this correctly returns `False`, and the caller (`routers/auths.py`'s `signin`) degrades gracefully to `HTTPException(400, INVALID_CRED)`. But if `credential.password` is not in a format the underlying hash-verification library can parse (e.g. a stored hash from an incompatible/legacy format), `verify_password()` raises instead of returning `False`, and nothing catches it — it propagates as a raw unhandled 500 instead of a 4xx.

Hit this in production with `WEBUI_AUTH=False`, where `/api/v1/auths/signin`'s hardcoded `admin@localhost`/`"admin"` bypass (same file, `elif WEBUI_AUTH == False:` branch) hit this exact path when the stored admin hash didn't verify cleanly — a raw 500 instead of the same clean 400 an ordinary bad password gets.

**Suggested fix:** wrap the `verify_password(credential.password)` call in a try/except (or validate the hash format upfront) so a malformed/unparseable hash fails the same way a wrong password does, rather than crashing the request.

teddiesloco:
## Summary

Fixes #29092.

In `open_webui/models/auths.py`, if a stored credential password hash is malformed or in an incompatible format, `verify_password(credential.password)` can raise an unhandled exception resulting in a raw 500 server error instead of gracefully failing authentication.

### Changes
- Wraps `verify_password` calls in `Auths.authenticate_user` with try/except blocks to treat unparseable/invalid hash formats identically to incorrect passwords, logging a warning and returning `None` so callers handle it as a standard 400 Bad Credentials.

### DIFF
--- a/backend/open_webui/models/auths.py
+++ b/backend/open_webui/models/auths.py
@@ -157,9 +157,16 @@ class AuthsTable:
         async with get_async_db_context(db) as session:
             credential = await session.get(Auth, resolved.id)
             if not credential or not credential.active:
-                await verify_password(PLACEHOLDER_HASH)
+                try:
+                    await verify_password(PLACEHOLDER_HASH)
+                except Exception:
+                    pass
                 return
-            if not await verify_password(credential.password):
+            try:
+                if not await verify_password(credential.password):
+                    return
+            except Exception:
+                log.warning('Unparseable stored password hash for user: %s', email)
                 return
             return resolved
 


teddiesloco:
Hi @tjbck, this PR fixes #29092 by safely catching unhandled exceptions during password verification in `Auths.authenticate_user`, ensuring malformed/incompatible legacy password hashes fail gracefully as bad credentials (400) rather than crashing the request with a raw 500. Ready for your review!

PR #29111, opened 2026-08-27
feat: re-run filter inlet on tool-call turns for opted-in filters

Classic298:
Filter inlet only runs on the user turn, so anything a tool returns reaches the model unfiltered. Images are the painful case: tool results can carry base64 images well over a model's size limit, and a filter that would rescale them never sees them, which forces deployments to patch the backend instead.

Filters now opt in with `tool_turns = True` at module level, and their inlet re-runs on every request carrying tool output: the native tool-call continuation, the code interpreter continuation, and the first model call after approved tool calls are drained. Filters without the flag behave exactly as before. Opt-in keeps that guarantee, since re-running every inlet would double-apply filters that inject context, count tokens or rate limit. An opted-in inlet receives the whole body each time, so it has to be idempotent.

The streaming handler's filter params now also include `__chat_id__` and `__message_id__`, matching the user turn. Without them an opted-in filter declaring those arguments would fail on a tool turn. Legacy (non-native) function calling stays uncovered: it injects tool output as text sources ahead of its single model call.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/filter.py
+++ b/backend/open_webui/utils/filter.py
@@ -159,6 +159,7 @@ async def process_filter_function(
     filter_context,
     valves_by_id,
     filter_ids,
+    tool_turns=False,
 ):
     filter_id = function.id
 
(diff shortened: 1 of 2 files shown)


tjbck:
`request` hook added.

Issue #29100, opened 2026-08-27
issue: Docker image tags `0.11.1` and `0.11.1-ollama` are not pushed


ryota-ka:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I have checked whether this issue is already fixed on the `dev` branch or in the latest source.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

0.11.1

### Operating System

Linux

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

Both tags should be available on Docker Hub.

### Actual Behavior

None are available.

### Steps to Reproduce

```sh
$ docker pull openwebui/open-webui:0.11.1
$ docker pull openwebui/open-webui:0.11.1-ollama
```

### Logs & Screenshots

```sh
$ docker pull openwebui/open-webui:0.11.1
Error response from daemon: failed to resolve reference "docker.io/openwebui/open-webui:0.11.1": docker.io/openwebui/open-webui:0.11.1: not found

$ docker pull openwebui/open-webui:0.11.1-ollama
Error response from daemon: failed to resolve reference "docker.io/openwebui/open-webui:0.11.1-ollama": docker.io/openwebui/open-webui:0.11.1-ollama: not found
```

### Additional Information

[...]

PR #29119, opened 2026-08-28
ci: retry the GHCR to Docker Hub copy before giving up

silentoplayz:
# Pull Request

## Checklist

- [x] This PR links to a confirmed Issue or active Discussion: Relates to #29100.
- [ ] I manually tested the changed workflow and any nearby behavior that could be affected.

## Summary

During the v0.11.1 release, the `copy-to-dockerhub (main)` and `copy-to-dockerhub (ollama)` jobs in run 32900262493 both failed on a transient network error while copying images from GHCR to Docker Hub (`stream error: PROTOCOL_ERROR; received from peer`). Each copy is a single `docker buildx imagetools create` attempt, so one mid-stream hiccup loses the tag, and because the copy runs the full version tag before the major.minor tag, that one hiccup left Docker Hub with `0.11.1` and `0.11.1-ollama` missing and `0.11` and `0.11-ollama` still pointing at the 0.11.0 digests, while `0.11-slim`, `0.11.1-cuda`, and the other variants from the same run landed fine (#29100).

This wraps each copy in up to three attempts with a 15 second pause between them. A copy that keeps failing exits the step nonzero after the third attempt, the same terminal behavior as today, one line louder.

Not changed, and worth a deliberate look: the job's `continue-on-error: true` is what kept the run green while two variants failed, so nobody saw the failure until users found the tags missing. Leaving it means a copy that exhausts its retries is still invisible in the run status. It reads as a deliberate choice to keep Docker Hub outages from failing releases, so this PR only narrows the window in which it matters rather than touching it.

This PR hardens future releases only. The missing v0.11.1 tags themselves need the two failed jobs on run 32900262493 re-run by someone with admin rights, as the GHCR source images are intact.

## Testing

A release cannot be run end to end from a fork, so the Testing checkbox above is honestly unticked. What was verified:

- The workflow file parses as valid YAML after the change.
- The changed `run` block was extracted verbatim from the parsed YAML and exercised in bash with `docker` and `sleep` stubbed: with two injected transient failures, the copy succeeds on the third attempt and the loop continues to the next tag, exiting 0; with permanent failures, the step exits 1 after three attempts and prints which destination failed.
- The retry changes nothing on the success path: a first-attempt success breaks out immediately, identical to current behavior.

### Fixed

- Release image copies from GHCR to Docker Hub now retry transient network failures up to three times per tag, instead of silently losing tags to a single mid-stream error as happened to `0.11.1` and `0.11.1-ollama` in the v0.11.1 release.

## Additional Context

The failure class is exactly a transient one: both failed jobs died within seconds of each other on HTTP/2 stream errors, on the two largest images, while three sibling variants in the same matrix succeeded.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -403,5 +403,15 @@ jobs:
             [ -z "$TAG" ] && continue
             DEST="${DOCKERHUB_IMAGE}:${TAG}"
             echo "  -> ${DEST}"
-            docker buildx imagetools create -t "${DEST}" "${SOURCE}"
+            for ATTEMPT in 1 2 3; do
+              if docker buildx imagetools create -t "${DEST}" "${SOURCE}"; then
+                break
+              fi
+              if [ "${ATTEMPT}" = "3" ]; then
+                echo "Failed to copy ${DEST} after ${ATTEMPT} attempts"
+                exit 1
+              fi
+              echo "Copy attempt ${ATTEMPT} for ${DEST} failed, retrying in 15s..."
+              sleep 15
+            done
           done <<< "${{ steps.tags.outputs.tags }}"


silentoplayz:
Landed on `dev` as 62e5eab60bb761de5bea9be63a20975721d4090a, which also removed `continue-on-error: true` from the `copy-to-dockerhub` job and added a `docker buildx imagetools inspect` check to each attempt, so an exhausted retry now fails the run instead of passing quietly.

Shipped in v0.11.2, and that release put `0.11.2`, `0.11.2-ollama`, `0.11`, and `0.11-ollama` on Docker Hub. `0.11.1` and `0.11.1-ollama` are still absent, since those need the two failed jobs on the v0.11.1 run re-run (#29100).

PR #29136, opened 2026-08-28
perf: cache RedisDict snapshots for values() and items()

Classic298:
values() and items() on the Redis-backed model registry transfer and re-parse the entire hash on every call. Pipeline filter checks and task payload builds read it at least three times per chat message on the synchronous Redis client, so both the transfer and the parse block the event loop. Measured on a production deployment (16 pods, 432 models, 10.4 MiB of serialized definitions) a single call costs 45-55 ms (io 16-37 ms, parse 26-31 ms), roughly 150 ms of event-loop time and 31 MiB of shared Redis traffic per message, multiplied by the pod count.

set() already maintains a signature key whose value changes exactly when the hash content changes. values() and items() now keep a per-process snapshot of the parsed hash and revalidate it with a single GET of that 64-byte key, so a warm call drops to well under a millisecond and the payload stays in Redis. Every writer deletes the signature before mutating and set() republishes it only after the write lands, so a reader only caches content whose signature survived the read unchanged and a concurrent write on any pod invalidates the cache. Only pools constructed with cache_set_signature=True take the new path; SESSION_POOL and USAGE_POOL keep their exact previous behaviour.

Snapshot hand-outs are fresh top-level dicts with shared nested structures, which no in-tree consumer mutates. The one place that passed such a structure to plugin code by reference, chat_completed handing __model__ to outlet filters on the direct-connection path, now re-resolves the model through the pool, so filters always receive a request-private object; the direct model keeps priority on id collisions and the non-direct path is unchanged. Verified with a fake-Redis harness covering cache coherence, invalidation ordering, concurrent-write races and the filter path (37 checks).

> [!NOTE]

### DIFF
--- a/backend/open_webui/socket/utils.py
+++ b/backend/open_webui/socket/utils.py
@@ -71,6 +71,8 @@ class RedisDict:
     ):
         self.name = name
         self._signature_name = f'{name}:signature' if cache_set_signature else None
+        self._snapshot = None
+        self._snapshot_signature = None
         self.redis = get_redis_connection(
             redis_url,
             redis_sentinels,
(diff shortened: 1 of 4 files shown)


Classic298:
Replaced by a smaller change on `perf/stable-model-registry-signature`. The read cache did not survive benchmarking, the remaining fix is three lines.

PR #29141, opened 2026-08-28
perf: cache Redis-backed model registry reads behind a content signature

Classic298:
On Redis-backed deployments every values() or items() call on the shared model registry transfers and re-parses the entire serialized registry, and a chat request does this at least three times through the pipeline filter scans. With 432 models (10.4 MiB serialized) that is roughly 150 ms of blocked event loop and 31 MiB of Redis traffic per message per pod, discarded in full on deployments without pipelines.

The registry hash now embeds a content signature, written atomically with the data via a Lua script, and each process serves values()/items() from an in-process snapshot as long as one 64-byte HGET shows the signature is unchanged; unchanged content also skips the rewrite. The hash moves to a new key (models:v2) so mixed-version pods never share a layout during rolling deploys; the old models keys can be deleted once a rollout completes. Filter ids are also sorted at registry build time, since per-process set ordering made identical registries serialize differently on each pod, defeating both the write skip and the cache.

Measured at that shape: 13 ms of parsing per read before, plus transfer, and 0.03 ms on a cache hit; per-call traffic drops from 10.4 MiB to 64 bytes.

> [!NOTE]

### DIFF
--- a/backend/open_webui/socket/main.py
+++ b/backend/open_webui/socket/main.py
@@ -117,12 +117,14 @@ if WEBSOCKET_MANAGER == 'redis':
         async_mode=True,
     )
 
+    # Separate key: old pods would JSON-parse the embedded signature field, and
+    # their set() would rewrite the data without it, breaking both generations.
     MODELS = RedisDict(
-        f'{REDIS_KEY_PREFIX}:models',
+        f'{REDIS_KEY_PREFIX}:models:v2',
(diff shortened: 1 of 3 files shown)


Classic298:
Replaced by a smaller change on `perf/stable-model-registry-signature`. The read cache did not survive benchmarking, the remaining fix is three lines.

PR #29144, opened 2026-08-28
perf: cache event function and webhook lookups in event dispatch

Classic298:
Publishing an event queries the database for the active event functions and reads the webhook config from the config table on every dispatch, and a chat turn publishes at least three events. Deployments with no event functions and no webhooks still pay all of those uncached round trips on every message.

Both lookups now sit behind a 1 second in-process cache used only by the dispatch path, coalescing the reads to at most one per second per worker. Every write that changes the active event function set (toggle, update, delete, sync, safe mode) or the webhook config clears the cache before the matching lifecycle event is published, so functions keep receiving their own enable and disable events exactly as before, and the webhook read-modify-write paths keep reading uncached so sequential admin edits never base an update on a stale list. Cross worker staleness is bounded at one second, matching the existing model list cache.

> [!NOTE]

### DIFF
--- a/backend/open_webui/events.py
+++ b/backend/open_webui/events.py
@@ -6,13 +6,17 @@ import logging
 import time
 import uuid
 from types import SimpleNamespace
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
+from aiocache import cached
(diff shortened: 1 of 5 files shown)


tjbck:
Functions need to be refresh.

Issue #29154, opened 2026-08-28
i18n: keys containing ':' are mangled/blank at runtime — init() is missing the keySeparator:false / nsSeparator:false flags that i18next-parser already uses


BLACKTHOMAS:
### Describe the bug

Keys that contain a colon are split by i18next's default `nsSeparator: ':'` at runtime and never resolve. Affected labels render **blank or garbled in every locale, including English**:

| Key | Renders today (verified on i18next 23.16.8 / 25.10.10 / 26.4.0) | Should render |
|---|---|---|
| `Warning:` | *(empty — label disappears entirely)* | `Warning:` / `Uwaga:` / `Warnung:` … |
| `http://localhost:8000` | `//localhost.8000` | the URL |
| `https://mineru.net/api/v4` | `//mineru.net/api/v4` | the URL |

The `Warning:` label is rendered by `$i18n.t('Warning:')` in `src/lib/components/admin/Functions/FunctionEditor.svelte` and `src/lib/components/workspace/Tools/ToolkitEditor.svelte`; the two URLs in `src/lib/components/admin/Settings/Documents.svelte` (MinerU / local docling server choice). Translations for `Warning:` exist in all locales (e.g. pl-PL `"Warning:": "Uwaga:"`) but are unreachable.

### To Reproduce

Minimal repro with the exact init options open-webui uses (`src/lib/i18n/index.ts`):

```js
import { createInstance } from 'i18next';

const resources = {
  'en-US': { translation: { 'Warning:': '' } },
  'pl-PL': { translation: { 'Warning:': 'Uwaga:' } }
};

const broken = createInstance();
await broken.init({
  lng: 'pl-PL',
  fallbackLng: { default: ['en-US'] },
  ns: 'translation',
  returnEmptyString: false,
  resources
});
console.log(broken.t('Warning:', { returnDetails: true }));
// → { res: "", usedKey: "", usedLng: "en-US", usedNS: "Warning" }
//   the key 'Warning:' was split into namespace 'Warning' + empty key;
//   the "missing key" return value is the (now empty) key → renders ""
console.log(broken.t('Warning:')); // → ''   (expected: 'Uwaga:')

const fixed = createInstance();
await fixed.init({
  ...sameOptions,
  keySeparator: false,   // ← proposed fix
  nsSeparator: false,    // ← proposed fix
});
console.log(fixed.t('Warning:')); // → 'Uwaga:'
```

### Root cause

[...]

PR #29155, opened 2026-08-28
i18n: complete Polish (pl-PL) translation + locale key sync

BLACKTHOMAS:
### Pull Request Checklist

- [ ] **Linked Issue/Discussion:** Related to [#6201 (fix: I18N consistency)](https://github.com/open-webui/open-webui/discussions/6201) — i18n/localization-only PR.
- [x] **First-time contributor policy:** this PR contains only i18n/localization updates; the contributor has previously merged Polish translation PRs.
- [x] **Target branch:** targets the `dev` branch (based on current `dev` tip).
- [x] **Description:** concise description below.
- [x] **Changelog:** Keep a Changelog entry included below.
- [ ] **Documentation:** N/A — no documentation changes required.
- [ ] **Dependencies:** N/A — no dependency changes.
- [x] **Testing:** JSON syntax validated; pl-PL key set verified against en-US (3091/3091 keys, 0 missing). The 2 keys new in this sync (`Accessibility Mode`, `Preview may differ from download.`) are intentionally empty and fall back to en-US at runtime (`returnEmptyString: false` in `src/lib/i18n/index.ts`). Locale data flows through the existing i18next runtime — no behavior touched.
- [x] **User-facing changes:** Polish UI text; the parser sync only realigns catalogs with current code (no translated value outside pl-PL is modified; screenshots N/A for locale data).
- [x] **No Unchecked AI Code:** AI-assisted, thoroughly reviewed by the contributor.
- [x] **Self-Review:** performed.
- [x] **Architecture:** N/A — locale data only.
- [x] **Git Hygiene:** two atomic commits — (1) pl-PL translation update, (2) mechanical i18next-parser locale sync + prettier formatting of `app.css` (see comment below).
- [x] **Title Prefix:** `i18n`.

# Changelog Entry

### Description

- Polish (pl-PL) translation update: 974 value edits completing missing translations, full key parity with en-US, plus proper Polish plural forms (`_few`/`_many`) for the 14 plural keys (matching i18next-parser output for `pl-PL`).
- Mechanical i18next-parser sync of all locale catalogs against the current `dev` tip: adds the keys introduced by the upstream `High Contrast Mode` → `Accessibility Mode` rename and the new `Preview may differ from download.` string, removes the obsolete `High Contrast Mode` key. Empty values fall back to en-US at runtime.

### Changed

- i18n: Updated Polish (pl-PL) translations, completing missing entries and adding proper plural forms (`_few`/`_many`)
[...]

PR #29161, opened 2026-08-28
fix: disable i18next key/ns separator splitting at runtime to match parser config

BLACKTHOMAS:
### Pull Request Checklist

- [x] **Linked Issue/Discussion:** This PR references an existing, well-described Issue for a real bug — Closes #29154.
- [x] **First-time contributor policy:** This is not my first contribution to Open WebUI (previously merged Polish translation PRs, e.g. the pending i18n: Polish translation update).
- [ ] **Documentation:** N/A — no user-facing configuration or feature changes; no documentation update required.
- [ ] **Dependencies:** N/A — no dependency changes.
- [x] **Testing:** Manual and scripted end-to-end verification performed — see "How this was verified" below.
- [x] **User-facing changes:** Yes, but text-only: 3 previously blank/mangled labels now render correctly in every locale. No layout or behavior changes; screenshots N/A (previously the labels rendered blank — there is nothing to screenshot).
- [x] **No Unchecked AI Code:** AI-assisted, thoroughly human-reviewed AND verified with a scripted test matrix (results below).
- [x] **Self-Review:** Performed — single-file, 2-line config change matching the existing parser configuration.
- [x] **Architecture:** No new settings, no architectural change — runtime options aligned with what the shipped parser configuration already enforces.
- [x] **Git Hygiene:** Atomic single commit, rebased on current `dev`, no unrelated commits.

# Changelog Entry

### Description

- The shipped `i18next-parser.config.ts` generates fully flat catalogs (`keySeparator: false`, `namespaceSeparator: false`), but the runtime `init()` in `src/lib/i18n/index.ts` never disabled i18next's default separators. As a result, any key containing `:` is split by the default `nsSeparator` at runtime and never resolves, rendering blank or mangled in every locale, including English.
- 2-line fix: pass `keySeparator: false` and `nsSeparator: false` to `init()`, restoring parity between runtime and parser configuration.

### Changed

- i18n: `src/lib/i18n/index.ts` — added `keySeparator: false` and `nsSeparator: false` to the runtime `init()` options (2 lines, no other changes).

### Fixed

- i18n: `Warning:` rendered as an empty string in the Functions and Tools editors; now renders the translated label (e.g. `Uwaga:`, `Warnung:`) in every locale.
- i18n: the MinerU / local docling server URLs in admin Settings → Documents rendered mangled (`//localhost.8000`, `//mineru.net/api/v4`); now render the full, correct URLs.

## How this was verified

The fix was verified against the real, shipped locale catalogs in an isolated environment, using the exact i18next version pinned by this repository's lockfile on the Node version used by this project's CI (Node 22).

- **Scope:** all 64 locale catalogs loaded from the repository (3,092 keys in the en-US catalog; asserted 100% flat — zero nested objects — so disabling key splitting cannot affect any structure).
- **Runtime fidelity:** two i18next instances were initialized with the exact `init()` options from `initI18n()` (same fallbacks, `returnEmptyString: false`, interpolation settings) — one without the fix (current `dev` behavior) and one with the fix — and compared call-by-call.
- **Full matrix:** every en-US key × every locale = **197,888 `t()` invocations per run**. After the fix, output matched the expected catalog/fallback value in **100% of calls (0 mismatches)**, including i18next's full language-resolution hierarchy (e.g. `ar-BH → ar → en-US`, and `fr-CA → fr-FR` via the existing `fallbackLng` configuration).
- **Behavioral delta:** exactly **3 keys** change behavior — all three contain `:` and are the broken ones from the linked issue. **Zero behavior change** on the other 3,089 keys, including all keys containing `.` and all keys with `:` protected by i18next's natural-language heuristic (57 keys).
- **Plurals:** 14 plural bases × 8 count values × 64 locales (**7,168 calls**) — identical output before and after.
- **Interpolation:** all 155 keys containing `{{variables}}` — identical output before and after.
- **Future-proofing:** the entire matrix was re-run on i18next 25.10.10 and 26.4.0 with identical results, so the fix remains correct if the dependency is ever bumped.
- **Code audit:** the repository contains no `getFixedT`, no additional `createInstance` calls, no array-key `t()` usage, and no per-call separator overrides — the application exclusively uses the single default instance with string keys and an options object, which is exactly the configuration exercised by the matrix above.
- **Totals:** ~614,000 `t()` invocations across the three runs; all 45 assertions passed, 0 failures.

### Additional Information

- Root cause and per-key analysis are documented in the linked issue. Translations for the affected keys already exist in the shipped catalogs (e.g. `Warning:` → `Uwaga:` in pl-PL) and become reachable again with this change.
- Keys that are intentionally empty fall back to en-US at runtime (existing `returnEmptyString: false` behavior), so nothing new is introduced there.

### DIFF
--- a/src/lib/i18n/index.ts
+++ b/src/lib/i18n/index.ts
@@ -65,6 +65,8 @@ export const initI18n = (defaultLocale?: string | undefined) => {
 				default: fallbackDefaultLocale
 			},
 			ns: 'translation',
+			keySeparator: false,
+			nsSeparator: false,
 			returnEmptyString: false,
 			interpolation: {
 				escapeValue: false // not needed for svelte as it escapes by default


BLACKTHOMAS:
**Note on the failing `Format & Build` check:** the failure is pre-existing on current `dev` and is not introduced by this PR.

- The check runs `npm run i18n:parse` and then asserts a clean working tree. Running the parser on an **unmodified `dev` checkout** produces the same ~64-file locale diff shown in this run (added `Accessibility Mode` / `Preview may differ from download.`, removed `High Contrast Mode`, zh-CN plural key normalization) — the recent `High Contrast Mode` → `Accessibility Mode` rename on `dev` was not followed by a parser run, so `dev` itself is currently failing this same check (the last three `Frontend Build` runs on `dev` are red).
- The diff in the failing log touches only `src/lib/i18n/locales/*.json` — none of the files changed by this PR (`src/lib/i18n/index.ts` only).
- The locale sync that makes this check green is already submitted in PR #29155; once that sync (or a maintainer parser run) lands on `dev`, this check will pass here as well. Happy to include the sync in this PR instead if maintainers prefer.

tjbck:
Thanks!

Issue #25083, opened 2026-05-26
bug: empty assistant placeholders are replayed to the provider and permanently break the chat


shawnxiao105-afk:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.9.5 (also verified unchanged on upstream `dev` at commit `5d9a09a88a9094ebfcd249340be3aaee544b34d0`)

### Ollama Version (if applicable)

_Not applicable — issue occurs on OpenAI-compatible providers, not Ollama._

### Operating System

macOS Sonoma 14 (Darwin 25.4.0). Open WebUI runs inside the official `ghcr.io/open-webui/open-webui:main` Docker image on Docker Desktop.

### Browser (if applicable)

Reproducible regardless of browser; symptom is server-side (payload sent to upstream provider).

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

When the backend reconstructs a chat's message history to send to a strict OpenAI-compatible upstream (any provider whose server-side schema rejects assistant messages with empty `content` and no `tool_calls`), assistant messages that contain **no usable payload** — i.e. empty `content`, no `output` items, no `tool_calls` — should be **filtered out** before being forwarded.

[...]

PR #29169, opened 2026-08-28
fix: stop replaying empty assistant messages to the provider

Classic298:
A generation that fails or is aborted before its first token is stored as an assistant message with no content. Every later message in that chat replays it to the provider, and providers that reject empty assistant content refuse the whole request, so the chat stays broken until the row is removed from the database by hand.

Empty assistant turns are now dropped when the history is rebuilt for the provider. Messages carrying tool calls or output items are untouched, and empty user, system and tool messages are left alone so nothing else about the payload changes.

The filter sits at the point the normal completion path and the tool-approval resume path share, so it also covers payloads posted straight to the API rather than loaded from a saved chat. Chats already broken on disk recover on their next message, no migration needed.

Fixes #25083

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -2198,14 +2198,6 @@ def strip_reasoning_details(output: list) -> list:
     ]
 
 
-def get_content_parts(message: dict) -> list:
-    """Return a message's content as a list of content parts."""
-    content = message.get('content')
-    if isinstance(content, list):
-        return content
-    return [{'type': 'text', 'text': content}] if content else []
-
-
 def process_messages_with_output(
     messages: list[dict],
     reasoning_format: str | None = None,
@@ -2217,7 +2209,6 @@ def process_messages_with_output(
     OpenAI-style messages (tool_calls + tool results). Strips 'output' before LLM.
     """
     processed = []
-    merge_with_previous = False
 
     for message in messages:
         if message.get('role') == 'assistant' and message.get('output'):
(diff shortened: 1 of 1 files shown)


backend/open_webui/utils/middleware.py
```diff
@@ -2223,6 +2224,12 @@ def process_messages_with_output(
                 processed.extend(output_messages)
                 continue
 
+        # An aborted generation leaves an empty assistant turn; replaying it breaks strict providers.
+        if message.get('role') == 'assistant' and not message.get('tool_calls'):
```
Jokasa7: This exempts only `tool_calls`, but OpenAI Chat Completions still permits `content: null` when the assistant has [`function_call`](https://github.com/openai/openai-python/blob/v2.29.0/src/openai/types/chat/chat_completion_assistant_message_param.py#L58-L69). On this head, `assistant(content=None, function_call=...)` is dropped while the subsequent supported [`role: function`](https://github.com/openai/openai-python/blob/v2.29.0/src/openai/types/chat/chat_completion_function_message_param.py#L11-L19) result remains; `sanitize_tool_pairs()` handles only modern tool pairs and cannot repair it. Please treat `function_call` as meaningful here and test the full legacy assistant-call → function-result sequence.

Jokasa7 (commented):
Thanks for fixing aborted placeholder turns. I ran this exact head through `process_messages_with_output()` and `sanitize_tool_pairs()`: a truly empty assistant is removed and a modern `tool_calls` pair is preserved, but a legacy `function_call` pair loses its assistant turn while the following `role: function` result survives. Open WebUI pins an OpenAI SDK version that still accepts this deprecated form. Please preserve `function_call` messages and add a complete legacy call/result regression alongside the modern control.

Issue #23381, opened 2026-04-03
issue: SQLite: Prompt tag filter fails for Cyrillic tags


goldkreateav:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

v0.8.12

### Operating System

Windows 10

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

Prompts tagged with Бизнес should be returned.

### Actual Behavior

0 prompts returned.

### Steps to Reproduce

1. Create a Prompt with tags including Бизнес (Cyrillic).
2. Go to Workspace → Prompts.
3. Select tag Бизнес (or call GET /api/v1/prompts/list?tag=Бизнес&page=1).
4. Observe that 0 prompts are returned.

### Logs & Screenshots

There's no actual logs for this.

### Additional Information

Root cause:
prompt.tags is stored as JSON in SQLite and may be serialized with \uXXXX escapes. Using LIKE (and/or LOWER() which is ASCII-focused in SQLite) on the serialized JSON text does not reliably match Cyrillic tag strings.

Proposed fix
Use dialect-aware JSON search:

SQLite: EXISTS (SELECT 1 FROM json_each(prompt.tags) WHERE json_each.value = :tag) (JSON1, compares decoded values)
[...]

Issue #29102, opened 2026-08-27
issue: SQLite — every ILIKE search is case-sensitive for non-ASCII text (Cyrillic, Greek, accented Latin)


Alert17:
### Check Existing Issues
- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions. The closest is #23381, which was closed as completed but is still reproducible — details below.
- [x] I have checked whether this issue is already fixed on the `dev` branch or in the latest source. It is not: reproduced on `dev` @ 240c795.
- [x] I am using the latest version of Open WebUI.

### Installation Method
pip / Git Clone

### Open WebUI Version
v0.11.0, and reproduced on `dev` @ 240c795

### Ollama Version (if applicable)
Not applicable — database-layer issue, no inference involved.

### Operating System
macOS 15. The bug is OS-independent; it depends only on the database backend being SQLite.

### Browser (if applicable)
Not applicable — reproducible from the backend with no browser involved.

### Confirmation
- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of Open WebUI.
- [x] I have provided every relevant configuration and environment variable used in my setup: default SQLite backend (`DATABASE_URL` unset). No other setting affects this.
- [x] I have listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup — none beyond the SQLite default.
- Browser console logs and container logs are not included because neither applies: this is a database-layer bug reproducible in five lines of Python, with no browser and no container. A standalone reproduction and the generated SQL are given below instead.

### Expected Behavior
`ilike()` matches case-insensitively regardless of script, as it does on PostgreSQL.
Searching for `привіт` should find a row containing `Привіт, світе`.

### Actual Behavior
On SQLite, `ilike()` is case-sensitive for every non-ASCII script. Only the exact
casing matches; ASCII text is unaffected, which is why the bug is easy to miss.

SQLAlchemy compiles `ilike()` to `lower(x) LIKE lower(?)` for dialects without a
native ILIKE, and SQLite's built-in `lower()` folds ASCII only.

This is not a single endpoint. `.ilike(` appears ~45 times across 14 model files —
`notes`, `chats`, `messages`, `models`, `prompts`, `knowledge`, `users`, `groups`,
`folders`, `files`, `skills`, `calendar`, `automations`, `shared_chats`. For anyone
[...]

PR #29172, opened 2026-08-28
fix: Unicode-aware case-insensitive search on SQLite; PostgreSQL search-path JSON operator fixes

BLACKTHOMAS:

- [x] **Linked Issue/Discussion:** This PR references an existing, well-described [Issue](https://github.com/open-webui/open-webui/issues) for a real bug or an active, substantive [Discussion](https://github.com/open-webui/open-webui/discussions) for a feature request or enhancement — `Relates to #29102` (also covers the still-reproducing prompt-tag case from closed #23381).
- [ ] **First-time contributor policy:** This is not my first contribution to Open WebUI, this PR contains only i18n/localization updates, or a maintainer explicitly asked me to open this PR after reviewing the linked Issue or Discussion.
- [x] **Testing:** **Manual** end-to-end tests have been performed to verify the fix/feature works correctly and does not introduce regressions. Screenshots or recordings are included where applicable.
- [x] **User-facing changes:** I have confirmed whether this PR changes the UI. If it does, screenshots are required, and a video recording is recommended.
  - **BREAKING CHANGE**: Changes affecting backward compatibility
  - **build**: Build system or dependency changes
  - **ci**: CI/CD workflow changes
  - **chore**: Refactoring, cleanup, or non-functional changes
  - **docs**: Documentation additions or updates
  - **perf**: Performance improvements

# Changelog Entry

### Description

**What breaks, and for whom.** On SQLite the dialect compiles `ilike()` to `lower(x) LIKE lower(?)`. SQLite's builtin `lower()` is a C function that folds only `[A-Z]` — every other codepoint passes through untouched. Case-insensitive search is therefore ASCII-only across 44 `ilike()` sites in 13 model files (chats, notes, prompts, knowledge, users, groups, folders, files, skills, calendar, automations, messages, shared_chats) and 5 explicit `func.lower()` sites. User-visible symptom: `ilike('%ŁĄKA%')` returns **0 rows** for data containing `łąka` — for anyone running content in Polish, Czech, Turkish, Cyrillic, Greek, or any accented Latin. PostgreSQL is unaffected (native Unicode-aware `ILIKE`); SQLite is the default database, i.e. the default install is the broken one.

**Why the obvious fix is off the table.** Overriding the builtin `lower()` per connection would fix the operators, but it would silently corrupt the one place where `lower()` is *stored*: migration `f0bd01a18a3d` creates `uq_user_email_lower`, a UNIQUE expression index on `lower(email)`. SQLite computes expression-index entries at write time with whatever function the writing connection has registered. Alembic builds its own connections (no app functions registered), so after an override the index would be born with builtin semantics while app inserts recompute entries with the override — split brain inside a uniqueness constraint. Every explicit `func.lower()` lookup (user email, folder names, prompt tags, model meta, group shares) would also silently change meaning. Conclusion: builtin `lower()` must stay bit-for-bit untouched, which rules out the one-line override and forces a dedicated function.

**The change.**

- New SQL function `unilower(x)`, registered per connection on all four SQLite paths: sync engine, async engine (aiosqlite), the SQLCipher creator, and the alembic migration engine. Semantics: `NFKC(casefold(NFKC(x)))` — the Unicode caseless algorithm — with an ASCII fast path (`str.lower()` is identical to casefold for every ASCII codepoint). Contract mirrors the builtin: NULL passes through, BLOBs pass through unchanged, numbers render as text (`lower(123)` = `'123'`), so non-text inputs never behave differently. `deterministic=True` is set (SQLite ≥ 3.8.3, satisfied everywhere this project runs) with a fallback registration for exotic builds.
- Why casefold+NFKC and not `str.lower()`: lowercase alone produces wrong matches — Greek word-final sigma folds `'ΣΑΣ'` to `'σας'` (with ς) while mid-word `'σασ'` keeps σ, so differently-cased Greek text never matches; and casefold does not unify composed (NFC) and decomposed (NFD) spellings of e.g. Polish `ą`/`ę`, which is exactly what text pasted from macOS/NFD sources looks like. Folding both the column and the pattern with the same algorithm makes `'ΣΑΣ'`↔`'σασ'` and NFC↔NFD match, plus fullwidth variants via NFKC.
- `@compiles(ilike_case_insensitive, 'sqlite')` reroutes **only** the case-folding operand of case-insensitive operators: `ilike`, `not_ilike`, `icontains`, `not_icontains`, `istartswith`, `iendswith` now emit `unilower(col) LIKE unilower(?)`. The blast radius is exactly the case-folding operand — plain `like()`, `contains()`, `startswith()`, `match()`, and every `func.lower()` call compile exactly as before. Other dialects are untouched: PostgreSQL keeps native `ILIKE`, MySQL keeps its current lowering.
- Registered on the alembic engine too: the `@compiles` hook is process-global, so a future migration using `ilike()` must not hit `no such function: unilower`.

**Why stored state cannot diverge.** `unilower` appears in no index, no generated column, no constraint, no trigger, no FTS table — it exists only at query time. A repo-wide audit found exactly one write-time consumer of `lower()`: the email expression index above, which this change deliberately does not touch. Builtin `lower()` is byte-for-byte unchanged (verified: `lower('ŁĄKA')`, `lower(123)`, `lower(NULL)` identical before/after; duplicate-email insert still rejected by `uq_user_email_lower`). Alembic migrations contain no `ilike()` and no `unilower()`. Consequences:

- **Rollback is a pure code revert.** No schema change, no data migration, no stored derived state, nothing to clean up — reverting the commits restores the previous behavior exactly, on any database.
- Works identically for fresh installs (schema via `create_all`) and existing installs (schema via migrations), including the SQLCipher encrypted path.

**Blast radius — every surface this change touches.**

- Compiler hook: the 44 `ilike()` sites across 13 model files (query-time matching only).
- Three raw-SQL search paths rewritten to `unilower()` in this PR: the prompt-tag filter (`Prompts.search_prompts` — the still-reproducing bug from closed #23381), the SQLite message-content matcher (`chat_search_message_content_match_sql`), and the model meta-tag filter (`Models.search_models`). All compared Python-lowered values against builtin `LOWER()` and could never meet for non-ASCII.
- `UnicodeLower` ORM element: folder-name matching, and the prompt-tag fallback branch for dialects without JSON functions (renders `lower()` there — unchanged behavior). The user-email lookups stay on builtin `func.lower()`.
- Registration: 4 connect-time hooks (sync engine, async engine, SQLCipher creator, alembic engine).
- Nothing else: no write path, constraint, trigger, generated column, auth lookup, join, or migration references `unilower`. SQLite user-defined functions are pure query-evaluation callbacks and **cannot modify the database at all** — the worst possible outcome of this change is a different set of rows in a search result, never data corruption.

**Security & integrity review.**

- **No new attack surface:** `unilower` is a pure, deterministic, in-process string function — no I/O, no dynamic evaluation. The compile override emits bound-parameter placeholders (`unilower(?)`); user input is never interpolated into SQL, so the change introduces no injection path that did not already exist.
- **Query-planner safety:** the function is registered as deterministic, which is a SQLite requirement for planner use of user-defined functions; `LIKE` with a leading wildcard does a full scan before and after this change, so no query regresses from an index-behavior change.
- **Resource bounds:** the added cost is CPU per scanned row (~1 µs, measured below), bounded by the scan size; no memory growth, no locks held longer, no connection-pool effects (registration happens once per pooled connection at connect time).
- **Concurrency:** 8 parallel workers over the real async engine produce identical, correct results; function objects are stateless and the GIL serializes callback execution.
- **Existing tests/data:** the full alembic migration chain runs to head on a fresh SQLite database with the change active; the email-uniqueness regression test (duplicate insert rejected) passes.

**Measured, not assumed.**

- Property test: 200 random queries over 5000 titles from 6 alphabets (PL/GR/TR/CYR/Latin-ext/ASCII, randomly mixed NFC/NFD) — SQL result set ≡ Python caseless-match ground truth in every case.
- Real paths exercised end-to-end on **both** databases (42-check SQLite suite, 28-check PostgreSQL suite — 70 total): `get_chats_by_user_id_and_search_text` (incl. `chat_search_terms` ASCII fallbacks), `get_chat_list_by_user_id`, `Prompts.search_prompts` and `Models.search_models` with `filter.tag` (uppercase Polish **and** uppercase Greek tags — the #23381 scenario plus the final-sigma parameter edge), `Folders.get_folder_by_parent_id_and_user_id_and_name` with a differently-cased Polish name, `Notes.search_notes`, `Knowledges.search_knowledge_bases`, `Users.get_users`, and Polish message-content search through the JSON history path; sync + async engines; cross-user isolation.
- Three real correctness bugs were caught and fixed during this review before they could ship: casefold-without-normalization (NFD paste miss), Greek final-sigma mismatch, and a tag-parameter inconsistency between `prompts.py`/`models.py` (uppercase Greek tags could never match on PostgreSQL).
- 1M-row full scan (worst case — `LIKE '%..%'` cannot use an index before or after): the ASCII query returns the identical 332 644 rows with both methods; the Polish query returns 583 385 rows vs **0 before**. Cost profile: the pattern argument is constant-folded by SQLite (measured 1 000 001 callbacks per scan, not 2 000 000); overhead ≈ 1 µs/row Python callback → ~1.5 s vs ~0.35 s per million scanned rows. Production queries filter `user_id` through an index first, so the scanned set is one user's rows (ms-level at realistic scales).

**PostgreSQL cross-verification.** The full suite was re-run against a real, isolated PostgreSQL 16 instance (`en_US.UTF-8`): native `ILIKE` preserved with no `unilower` leakage, `UnicodeLower` compiles byte-identically to the old `func.lower()` rendering, every touched search path (chat search, content search, prompt tags, model meta tags, folder matching, notes, knowledge, users) verified through the real production functions, and the email-uniqueness index still rejects duplicates. This review also surfaced a **pre-existing upstream PG bug**: `chat_search_message_content_match_sql` and the chat-tag filter applied JSON operators (`#>`, `->`) directly to `JSONField` columns, which are TEXT-backed — every content search with a query on PostgreSQL errored with `operator does not exist: text -> unknown`. Fixed here with explicit `::json` casts (no-op semantics for native JSON columns, required for the TEXT-backed storage).

**Deliberately not addressed here:** the 5 `func.lower()` sites (frozen by the email expression index until that index is rebuilt differently; the SQLite prompt-tag and content-search paths listed above are now covered instead), and the Python-side content filter — these need a per-dialect rewrite or a full-text index; for very large deployments the structural answer is an **FTS5 virtual table with the `trigram` tokenizer** (SQLite ≥ 3.34, the only variant that serves substring search from an index), which is a schema-level change I am happy to spec separately. The first-time-contributor checkbox below is intentionally left unticked — your call whether this lands.

### Added

- [New features, functionalities, or additions]

### Changed

- [Changes, updates, refactorings, or optimizations]

### Deprecated

- [Deprecated functionality or features]

### Removed

- [Removed features, files, or functionalities]

### Fixed

- SQLite: case-insensitive search now matches non-ASCII text (Polish, Cyrillic, Greek, accented Latin, fullwidth variants) across all `ilike()`-based searches (#29102).
- SQLite: prompt-tag filtering now matches non-ASCII tags case-insensitively — the still-reproducing bug from closed #23381.
- SQLite: message-content search now matches non-ASCII text case-insensitively (both JSON history paths).
- SQLite: model meta-tag search and folder-name matching now handle non-ASCII case-insensitively.
- PostgreSQL: message-content search, chat-tag filtering and model meta-tag filtering no longer error with `operator does not exist: text -> unknown` (JSON operators were applied to TEXT-backed `JSONField` columns).

### Security

- [Security-related changes or vulnerability fixes]

### Breaking Changes

- **BREAKING CHANGE**: [Changes affecting compatibility or functionality]

---

### Additional Information

- 7 commits on `fix/sqlite-unicode-ilike`, rebased on `dev`; backend-only (the frontend workflow is skipped for `backend/**` paths).
- No new dependencies; no schema changes, no data migrations — rollback is a pure code revert.
- Verification ran on SQLite 3.4x (default deployment) and an isolated real PostgreSQL 16 instance; `memories` and `channels` have no SQL text-search path and were audited as unaffected.
- Scale path for very large deployments (FTS5/trigram shadow index) is deliberately out of scope — happy to spec separately.

### Screenshots or Videos

- Not applicable — backend-only change, no UI changes.

> [!NOTE]

### DIFF
--- a/backend/open_webui/internal/db.py
+++ b/backend/open_webui/internal/db.py
@@ -37,6 +37,7 @@ from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import Session, scoped_session, sessionmaker
 from sqlalchemy.pool import NullPool, QueuePool
 from sqlalchemy.sql.compiler import ilike_case_insensitive
+from sqlalchemy.sql.functions import FunctionElement
 from sqlalchemy.sql.type_api import _T
 from typing_extensions import Self
 
@@ -105,6 +106,40 @@ def _sqlite_unilower_case_insensitive(element, compiler, **kw):
     PostgreSQL and other dialects are untouched (they keep native ILIKE).
     """
     return f'unilower({element.element._compiler_dispatch(compiler, **kw)})'
+
+
+class UnicodeLower(FunctionElement):
+    """Dialect-aware SQL ``LOWER`` for query-time case-insensitive matching.
+
+    Renders ``lower(...)`` on every dialect except SQLite, where it renders
+    ``unilower(...)`` (Unicode caseless, see above). Intended for
+    case-insensitive comparisons that are NOT backed by a builtin-``lower()``
+    expression index — those (the user-email lookups against
+    ``uq_user_email_lower``) must keep the builtin ``func.lower()`` so the
+    index stays usable and semantically consistent.
+    """
+
+    type = types.String()
+    inherit_cache = True
+
+
+@compiles(UnicodeLower)
+def _render_unicode_lower_default(element, compiler, **kw):
+    return f'lower({compiler.process(element.clauses, **kw)})'
+
+
+@compiles(UnicodeLower, 'sqlite')
+def _render_unicode_lower_sqlite(element, compiler, **kw):
+    return f'unilower({compiler.process(element.clauses, **kw)})'
+
+
+def unicode_caseless(value: str) -> str:
+    """Python-side counterpart of the SQL ``unilower()`` function.
+
+    Use it to pre-fold search patterns (and other in-Python comparisons) so
+    they meet the SQL-side ``unilower()`` folding of stored values.
+    """
+    return _unicode_lower(value)
 
 
 # ── SSL URL normalization (used by sync engine & Alembic migrations) ─
--- a/backend/open_webui/internal/db.py
+++ b/backend/open_webui/internal/db.py
@@ -6,6 +6,7 @@ import sys
 from contextlib import asynccontextmanager, contextmanager
 from datetime import datetime, timedelta, timezone
 from typing import Any, Optional
+from unicodedata import normalize
 from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
 
 from open_webui.env import (
@@ -64,24 +65,27 @@ log = logging.getLogger(__name__)
 
 
 def _unicode_lower(value):
-    """Unicode caseless-fold, mirroring the builtin ``lower()`` contract.
+    """Unicode NFKC-caseless fold, mirroring the builtin ``lower()`` contract.
 
     NULL passes through as NULL, BLOBs pass through unchanged (like the
     builtin), numbers render as text (``lower(123) -> '123'``), and text
-    is folded with ``str.casefold()`` — the Unicode caseless-matching
-    algorithm. ``casefold`` is applied to both the column and the search
-    pattern, so e.g. Greek 'ΣΑΣ' matches 'σασ' (``str.lower()`` would
-    mismatch: 'ΣΑΣ' folds to 'σας' with word-final sigma while 'σασ'
-    keeps the mid-word sigma). ASCII strings use the plain ``lower()``
-    fast path (identical result for every ASCII codepoint).
+    is folded with the Unicode caseless-matching algorithm:
+    ``NFKC(casefold(NFKC(x)))``. ``casefold`` alone would still treat the
+    composed (NFC) and decomposed (NFD) spellings of e.g. Polish 'ą' or
+    'ę' as different strings, so the value is normalized before and after
+    folding — pasted text frequently arrives as NFD while typed text is
+    NFC. The same fold is applied to the column and the search pattern,
+    so e.g. Greek 'ΣΑΣ' matches 'σασ' and NFC 'ą' matches NFD 'ą'.
+    ASCII strings use a plain ``str.lower()`` fast path (identical result
+    for every ASCII codepoint, no normalization needed).
     """
     if value is None or isinstance(value, bytes):
         return value
     if isinstance(value, str):
         if value.isascii():
             return value.lower()
-        return value.casefold()
-    return str(value).casefold()
+        return normalize('NFKC', normalize('NFKC', value).casefold())
+    return normalize('NFKC', str(value).casefold())
 
 
 def _register_sqlite_search_functions(dbapi_connection):
--- a/backend/open_webui/internal/db.py
+++ b/backend/open_webui/internal/db.py
@@ -31,13 +31,69 @@ from open_webui.utils.json_codec import JSONCodec
 from sqlalchemy import Dialect, MetaData, create_engine, event, types
 from sqlalchemy.engine.url import make_url
 from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker, create_async_engine
+from sqlalchemy.ext.compiler import compiles
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import Session, scoped_session, sessionmaker
 from sqlalchemy.pool import NullPool, QueuePool
+from sqlalchemy.sql.compiler import ilike_case_insensitive
 from sqlalchemy.sql.type_api import _T
 from typing_extensions import Self
 
 log = logging.getLogger(__name__)
+
+
+# ── Unicode-aware case-insensitive search on SQLite ─────────────────
+#
+# SQLAlchemy compiles ``ilike()`` on SQLite to ``lower(x) LIKE lower(?)``,
+# but the builtin ``lower()`` folds ASCII only, so searches never match
+# Polish, Cyrillic, Greek, or accented Latin text (upstream issue #29102).
+#
+# Two pieces make the search Unicode-aware:
+#
+# 1. A custom ``unilower()`` SQL function (registered per connection).
+# 2. A compiler override rendering ILIKE case-folding as ``unilower()``
+#    on the SQLite dialect only.
+#
+# The builtin ``lower()`` is deliberately NOT overridden: it backs the
+# ``uq_user_email_lower`` unique expression index (migration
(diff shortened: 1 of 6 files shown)


BLACKTHOMAS:
Thanks @Alert17 — all three points verified locally; two were genuine bugs in my patch, one is intended behavior. Fixes are pushed to the PR branch:

**1. Fold-introduced LIKE metacharacters — confirmed, fixed.** In the one escaped call site (`Files.search_files` glob), escaping ran before the SQL-side fold, so fullwidth `％`→`%`, `＿`→`_`, `＼`→`\` turned into wildcards after folding (a glob like `*％*` matched every file). I now fold the glob in Python *before* escaping, so fold-introduced metacharacters are escaped as literals. The fold is gated to SQLite only: SQLite folds both sides via `unilower()`, while PostgreSQL keeps native `ILIKE` and never NFKC-folds the column — folding the pattern there would desynchronize the two sides (glob `stra?e*` would stop matching `Straße doc.pdf`).

**2. BLOB contract — confirmed, fixed.** `unilower(x'414243')` returned a blob; it now decodes BLOBs as UTF-8 text and folds, mirroring the builtin contract (`lower(x'414243')` = `'abc'`).

**3. Length-changing folds vs single-char wildcards — intended.** `unilower()` implements Unicode default caseless matching (`NFKC(casefold(NFKC(x)))`), which is inherently length-changing (`ß`→`ss`, `ﬁ`→`fi`), so `LIKE 'stra_e'` can't span `strasse` — exactly the result Python `casefold()` substring matching would give, which is the semantics I targeted. SQLite ends up a deliberate caseless superset of PostgreSQL's native `ILIKE` (no NFKC on the PG side); I've documented this in the code comments.

I added regression tests for all of this (BLOB parity, `％`/`＿`/`＼` literal behavior, `stra_e`/`strasse` wildcard semantics, and a PG no-regression check): 58 checks on SQLite and 30 on PostgreSQL, all green.

Your test script would be very welcome — I'd be glad to run it against the branch and fold anything it catches into the fix.

PR #29174, opened 2026-08-28
fix: expire task registry entries left behind by a killed worker

Classic298:
A worker killed mid-generation leaves task IDs behind, so the chat keeps showing a spinner and Stop button. Each task now gets its own Redis expiry key, refreshed by its owning worker. Task-list reads prune expired IDs from the existing registry and per-chat sets, even when another worker is still generating in the same chat.

`REDIS_TASK_TTL` defaults to 300 seconds; `0` disables expiry. This uses native key expiry without sorted-set timestamps or hash-field expiry. Existing entries without expiry keys are treated as stale, so let in-flight generations finish before upgrading.

Validation: `backend/tests/test_task_ttl.py` passed against Redis 7.4.1 with two worker processes, including killing one worker mid-generation, retaining long-running tasks, pruning stale chat/global entries and response data, cleanup without resurrection, and disabled expiry.

Run with `TEST_REDIS_URL=redis://localhost:6379 PYTHONPATH=backend python backend/tests/test_task_ttl.py`.

Fixes #22525

> [!NOTE]

### DIFF
--- a/backend/open_webui/tasks.py
+++ b/backend/open_webui/tasks.py
@@ -6,7 +6,7 @@ from uuid import uuid4
 
 from redis.asyncio import Redis
 
-from open_webui.env import REDIS_KEY_PREFIX, REDIS_RESPONSE_STREAM_TTL
+from open_webui.env import REDIS_KEY_PREFIX, REDIS_RESPONSE_STREAM_TTL, REDIS_TASK_TTL
 from open_webui.utils.json_codec import JSONCodec, dumps_bytes
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


backend/open_webui/tasks.py
```diff
@@ -63,39 +66,82 @@ async def redis_task_command_listener(app):
...
 
 async def redis_list_tasks(redis: Redis) -> list[str]:
     return list(await redis.hkeys(REDIS_TASKS_KEY))
 
 
 async def redis_list_item_tasks(redis: Redis, item_id: str) -> list[str]:
-    return list(await redis.smembers(f'{REDIS_ITEM_TASKS_KEY}:{item_id}'))
+    # A killed worker stops refreshing its scores, so its tasks age out instead of hanging around.
+    oldest = time.time() - REDIS_TASK_REGISTRY_TTL if REDIS_TASK_REGISTRY_TTL > 0 else '-inf'
+    return list(await redis.zrangebyscore(f'{REDIS_ITEM_TASKS_KEY}:{item_id}', oldest, '+inf'))
```
Jokasa7: This reader only checks `:tasks:item_active`, while an old worker in the stated mid-upgrade window still writes `:tasks:item` with `SADD`. I verified that an old task is absent from this `ZRANGEBYSCORE`; conversely, the old `SMEMBERS` reader cannot see a task written here. A stop request routed to the opposite-version worker therefore returns “No tasks found” before publishing any cancellation. Could this dual-read the legacy Set during the compatibility window (and make old/new visibility tests), or otherwise enforce a deployment barrier that prevents mixed worker versions?

Jokasa7 (commented):
Thanks for addressing dead-worker registry entries. I reproduced the upgrade boundary on Redis 7: an old worker's Set entry is invisible to the new ZSET reader, and a new worker's ZSET entry is invisible to the old Set reader. The keys avoid a `WRONGTYPE` error, but they do not preserve cross-worker task visibility. During the mid-upgrade coexistence called out in the code, `stop_item_tasks()` can falsely report no tasks and never publish the stop command. Please add a compatibility read/migration window or require a stop-the-world migration, with both directions covered by tests.

PR #29184, opened 2026-08-29
i18n: complete Polish (pl-PL) translation

BLACKTHOMAS:
# Pull Request Checklist

- [x] **Linked Issue/Discussion:** Relates to [Discussion #6201](https://github.com/open-webui/open-webui/discussions/6201) (i18n consistency) — i18n/localization-only PR.
- [ ] **First-time contributor policy:** This PR contains only i18n/localization updates; the contributor has previously merged Polish translation PRs #29155 was closed for conflicts only.
- [x] **Target branch:** The pull request targets the `dev` branch.
- [x] **Description:** A concise description of the changes is provided below.
- [x] **Changelog:** A changelog entry following [Keep a Changelog](https://keepachangelog.com/) format is included at the bottom.
- [ ] **Documentation:** N/A — locale data only.
- [ ] **Dependencies:** N/A — no dependency changes.
- [x] **Testing:** JSON syntax validated; pl-PL key set verified against the current `dev` `en-US` catalog: **3120/3120 keys, 0 missing, 0 empty** (arithmetic below). Locale data flows through the existing i18next runtime — no code behavior touched.
- [x] **User-facing changes:** Polish UI text only; no layout/UI logic changes, screenshots N/A for locale data.
- [x] **No Unchecked AI Code:** AI-assisted, thoroughly human-reviewed.
- [x] **Self-Review:** Performed.
- [ ] **Architecture:** N/A — locale data only.
- [x] **Git Hygiene:** Single atomic commit, rebased on the current `dev` tip, no unrelated files touched.
- [x] **Title Prefix:** `i18n`.

# Changelog Entry

### Description

- Completes the Polish (pl-PL) translation: **976 value edits**, reaching **full key parity with the current `dev` `en-US` catalog — 3120/3120 keys, 0 missing, 0 empty** — and adds proper Polish plural forms (`_few`/`_many`) for the 14 plural keys, per the CLDR plural rules for Polish.

### Added

- Polish plural forms `_few` and `_many` for all 14 plural bases (Polish requires 4 forms per CLDR: `_one`, `_few`, `_many`, `_other`).
- Translations for the keys introduced by the upstream `High Contrast Mode` → `Accessibility Mode` rename and the new `Preview may differ from download.` string.

### Changed

- i18n: Completed and corrected Polish (pl-PL) translations (974 edits), including the two previously pending strings: `Accessibility Mode` → `Tryb ułatwień dostępu`, `Preview may differ from download.` → `Podgląd może różnić się od pobranego pliku`.

[...]

PR #20562, opened 2026-01-11
i18n(pl-PL): Add missing keys and update existing translations

BLACKTHOMAS:
# Changelog Entry

### Description
- This PR brings the Polish (`pl-PL`) translation up to date with the latest English template. It fills in missing keys, corrects existing translations for better context, and sorts the JSON keys.
- **Importantly**, it includes translations for the **Admin Settings tabs** (General, Connections, Models, etc.) following the recent fix that exposed these keys for translation.

### Added
- **Admin Settings Translations**: Added Polish translations for the Settings menu tabs: General, Connections, Models, Evaluations, External Tools, Documents, Web Search, Code Execution, Interface, Audio, Images, Pipelines, Database.

### Changed
- Updated existing translations to be more accurate and natural.
- Reordered keys in `pl-PL.json` to match the structure of the source JSON, ensuring easier maintenance in the future.

### Fixed
- **Fixed Settings tab i18n**: Resolved the issue where admin settings tabs were displayed in English by providing the correct translation keys.
- Fixed minor typos and grammatical errors in the existing Polish translation.

### Deprecated
- N/A

### Removed
- N/A

### Security
- N/A

### Breaking Changes
- **BREAKING CHANGE**: N/A

---

### Additional Information
- The translation now covers the full Admin interface, including the recently fixed menu tabs.
- Tested with the provided JSON structure to ensure no syntax errors.

### Screenshots or Videos
- N/A (Localization update only)

### Contributor License Agreement
By submitting this pull request, I confirm that I have read and fully agree to the [Contributor License Agreement (CLA)](https://github.com/open-webui/open-webui/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT), and I am providing my contributions under its terms.

PR #29186, opened 2026-08-29
fix: Unicode-aware case-insensitive search on SQLite; PostgreSQL search-path JSON operator fixes

BLACKTHOMAS:

- [x] **Linked Issue/Discussion:** This PR references an existing, well-described [Issue](https://github.com/open-webui/open-webui/issues) for a real bug or an active, substantive [Discussion](https://github.com/open-webui/open-webui/discussions) for a feature request or enhancement — `Relates to #29102 ` (also covers the still-reproducing prompt-tag case from closed #23381).
- [ ] **First-time contributor policy:** This is not my first contribution to Open WebUI, this PR contains only i18n/localization updates, or a maintainer explicitly asked me to open this PR after reviewing the linked Issue or Discussion.
- [x] **Testing:** **Manual** end-to-end tests have been performed to verify the fix/feature works correctly and does not introduce regressions. Screenshots or recordings are included where applicable.
- [x] **User-facing changes:** I have confirmed whether this PR changes the UI. If it does, screenshots are required, and a video recording is recommended.
  - **BREAKING CHANGE**: Changes affecting backward compatibility
  - **build**: Build system or dependency changes
  - **ci**: CI/CD workflow changes
  - **chore**: Refactoring, cleanup, or non-functional changes
  - **docs**: Documentation additions or updates
  - **perf**: Performance improvements

# Changelog Entry

### Description

**What breaks, and for whom.** On SQLite the dialect compiles `ilike()` to `lower(x) LIKE lower(?)`. SQLite's builtin `lower()` is a C function that folds only `[A-Z]` — every other codepoint passes through untouched. Case-insensitive search is therefore ASCII-only across 44 `ilike()` sites in 13 model files (chats, notes, prompts, knowledge, users, groups, folders, files, skills, calendar, automations, messages, shared_chats) and 5 explicit `func.lower()` sites. User-visible symptom: `ilike('%ŁĄKA%')` returns **0 rows** for data containing `łąka` — for anyone running content in Polish, Czech, Turkish, Cyrillic, Greek, or any accented Latin. PostgreSQL is unaffected (native Unicode-aware `ILIKE`); SQLite is the default database, i.e. the default install is the broken one.

**Why the obvious fix is off the table.** Overriding the builtin `lower()` per connection would fix the operators, but it would silently corrupt the one place where `lower()` is *stored*: migration `f0bd01a18a3d` creates `uq_user_email_lower`, a UNIQUE expression index on `lower(email)`. SQLite computes expression-index entries at write time with whatever function the writing connection has registered. Alembic builds its own connections (no app functions registered), so after an override the index would be born with builtin semantics while app inserts recompute entries with the override — split brain inside a uniqueness constraint. Every explicit `func.lower()` lookup (user email, folder names, prompt tags, model meta, group shares) would also silently change meaning. Conclusion: builtin `lower()` must stay bit-for-bit untouched, which rules out the one-line override and forces a dedicated function.

**The change.**

- New SQL function `unilower(x)`, registered per connection on all four SQLite paths: sync engine, async engine (aiosqlite), the SQLCipher creator, and the alembic migration engine. Semantics: `NFKC(casefold(NFKC(x)))` — the Unicode caseless algorithm — with an ASCII fast path (`str.lower()` is identical to casefold for every ASCII codepoint). Contract mirrors the builtin: NULL passes through, BLOBs pass through unchanged, numbers render as text (`lower(123)` = `'123'`), so non-text inputs never behave differently. `deterministic=True` is set (SQLite ≥ 3.8.3, satisfied everywhere this project runs) with a fallback registration for exotic builds.
- Why casefold+NFKC and not `str.lower()`: lowercase alone produces wrong matches — Greek word-final sigma folds `'ΣΑΣ'` to `'σας'` (with ς) while mid-word `'σασ'` keeps σ, so differently-cased Greek text never matches; and casefold does not unify composed (NFC) and decomposed (NFD) spellings of e.g. Polish `ą`/`ę`, which is exactly what text pasted from macOS/NFD sources looks like. Folding both the column and the pattern with the same algorithm makes `'ΣΑΣ'`↔`'σασ'` and NFC↔NFD match, plus fullwidth variants via NFKC.
- `@compiles(ilike_case_insensitive, 'sqlite')` reroutes **only** the case-folding operand of case-insensitive operators: `ilike`, `not_ilike`, `icontains`, `not_icontains`, `istartswith`, `iendswith` now emit `unilower(col) LIKE unilower(?)`. The blast radius is exactly the case-folding operand — plain `like()`, `contains()`, `startswith()`, `match()`, and every `func.lower()` call compile exactly as before. Other dialects are untouched: PostgreSQL keeps native `ILIKE`, MySQL keeps its current lowering.
- Registered on the alembic engine too: the `@compiles` hook is process-global, so a future migration using `ilike()` must not hit `no such function: unilower`.

**Why stored state cannot diverge.** `unilower` appears in no index, no generated column, no constraint, no trigger, no FTS table — it exists only at query time. A repo-wide audit found exactly one write-time consumer of `lower()`: the email expression index above, which this change deliberately does not touch. Builtin `lower()` is byte-for-byte unchanged (verified: `lower('ŁĄKA')`, `lower(123)`, `lower(NULL)` identical before/after; duplicate-email insert still rejected by `uq_user_email_lower`). Alembic migrations contain no `ilike()` and no `unilower()`. Consequences:

- **Rollback is a pure code revert.** No schema change, no data migration, no stored derived state, nothing to clean up — reverting the commits restores the previous behavior exactly, on any database.
- Works identically for fresh installs (schema via `create_all`) and existing installs (schema via migrations), including the SQLCipher encrypted path.

**Blast radius — every surface this change touches.**

- Compiler hook: the 44 `ilike()` sites across 13 model files (query-time matching only).
- Three raw-SQL search paths rewritten to `unilower()` in this PR: the prompt-tag filter (`Prompts.search_prompts` — the still-reproducing bug from closed #23381), the SQLite message-content matcher (`chat_search_message_content_match_sql`), and the model meta-tag filter (`Models.search_models`). All compared Python-lowered values against builtin `LOWER()` and could never meet for non-ASCII.
- `UnicodeLower` ORM element: folder-name matching, and the prompt-tag fallback branch for dialects without JSON functions (renders `lower()` there — unchanged behavior). The user-email lookups stay on builtin `func.lower()`.
- Registration: 4 connect-time hooks (sync engine, async engine, SQLCipher creator, alembic engine).
- Nothing else: no write path, constraint, trigger, generated column, auth lookup, join, or migration references `unilower`. SQLite user-defined functions are pure query-evaluation callbacks and **cannot modify the database at all** — the worst possible outcome of this change is a different set of rows in a search result, never data corruption.

**Security & integrity review.**

- **No new attack surface:** `unilower` is a pure, deterministic, in-process string function — no I/O, no dynamic evaluation. The compile override emits bound-parameter placeholders (`unilower(?)`); user input is never interpolated into SQL, so the change introduces no injection path that did not already exist.
- **Query-planner safety:** the function is registered as deterministic, which is a SQLite requirement for planner use of user-defined functions; `LIKE` with a leading wildcard does a full scan before and after this change, so no query regresses from an index-behavior change.
- **Resource bounds:** the added cost is CPU per scanned row (~1 µs, measured below), bounded by the scan size; no memory growth, no locks held longer, no connection-pool effects (registration happens once per pooled connection at connect time).
- **Concurrency:** 8 parallel workers over the real async engine produce identical, correct results; function objects are stateless and the GIL serializes callback execution.
- **Existing tests/data:** the full alembic migration chain runs to head on a fresh SQLite database with the change active; the email-uniqueness regression test (duplicate insert rejected) passes.

**Measured, not assumed.**

- Property test: 200 random queries over 5000 titles from 6 alphabets (PL/GR/TR/CYR/Latin-ext/ASCII, randomly mixed NFC/NFD) — SQL result set ≡ Python caseless-match ground truth in every case.
- Real paths exercised end-to-end on **both** databases (42-check SQLite suite, 28-check PostgreSQL suite — 70 total): `get_chats_by_user_id_and_search_text` (incl. `chat_search_terms` ASCII fallbacks), `get_chat_list_by_user_id`, `Prompts.search_prompts` and `Models.search_models` with `filter.tag` (uppercase Polish **and** uppercase Greek tags — the #23381 scenario plus the final-sigma parameter edge), `Folders.get_folder_by_parent_id_and_user_id_and_name` with a differently-cased Polish name, `Notes.search_notes`, `Knowledges.search_knowledge_bases`, `Users.get_users`, and Polish message-content search through the JSON history path; sync + async engines; cross-user isolation.
- Three real correctness bugs were caught and fixed during this review before they could ship: casefold-without-normalization (NFD paste miss), Greek final-sigma mismatch, and a tag-parameter inconsistency between `prompts.py`/`models.py` (uppercase Greek tags could never match on PostgreSQL).
- 1M-row full scan (worst case — `LIKE '%..%'` cannot use an index before or after): the ASCII query returns the identical 332 644 rows with both methods; the Polish query returns 583 385 rows vs **0 before**. Cost profile: the pattern argument is constant-folded by SQLite (measured 1 000 001 callbacks per scan, not 2 000 000); overhead ≈ 1 µs/row Python callback → ~1.5 s vs ~0.35 s per million scanned rows. Production queries filter `user_id` through an index first, so the scanned set is one user's rows (ms-level at realistic scales).

**PostgreSQL cross-verification.** The full suite was re-run against a real, isolated PostgreSQL 16 instance (`en_US.UTF-8`): native `ILIKE` preserved with no `unilower` leakage, `UnicodeLower` compiles byte-identically to the old `func.lower()` rendering, every touched search path (chat search, content search, prompt tags, model meta tags, folder matching, notes, knowledge, users) verified through the real production functions, and the email-uniqueness index still rejects duplicates. This review also surfaced a **pre-existing upstream PG bug**: `chat_search_message_content_match_sql` and the chat-tag filter applied JSON operators (`#>`, `->`) directly to `JSONField` columns, which are TEXT-backed — every content search with a query on PostgreSQL errored with `operator does not exist: text -> unknown`. Fixed here with explicit `::json` casts (no-op semantics for native JSON columns, required for the TEXT-backed storage).

**Deliberately not addressed here:** the 5 `func.lower()` sites (frozen by the email expression index until that index is rebuilt differently; the SQLite prompt-tag and content-search paths listed above are now covered instead), and the Python-side content filter — these need a per-dialect rewrite or a full-text index; for very large deployments the structural answer is an **FTS5 virtual table with the `trigram` tokenizer** (SQLite ≥ 3.34, the only variant that serves substring search from an index), which is a schema-level change I am happy to spec separately. The first-time-contributor checkbox below is intentionally left unticked — your call whether this lands.

### Added

- [New features, functionalities, or additions]

### Changed

- [Changes, updates, refactorings, or optimizations]

### Deprecated

- [Deprecated functionality or features]

### Removed

- [Removed features, files, or functionalities]

### Fixed

- SQLite: case-insensitive search now matches non-ASCII text (Polish, Cyrillic, Greek, accented Latin, fullwidth variants) across all `ilike()`-based searches (#29102).
- SQLite: prompt-tag filtering now matches non-ASCII tags case-insensitively — the still-reproducing bug from closed #23381.
- SQLite: message-content search now matches non-ASCII text case-insensitively (both JSON history paths).
- SQLite: model meta-tag search and folder-name matching now handle non-ASCII case-insensitively.
- PostgreSQL: message-content search, chat-tag filtering and model meta-tag filtering no longer error with `operator does not exist: text -> unknown` (JSON operators were applied to TEXT-backed `JSONField` columns).

### Security

- [Security-related changes or vulnerability fixes]

### Breaking Changes

- **BREAKING CHANGE**: [Changes affecting compatibility or functionality]

---

### Additional Information

- 1 squashed commit on `fix/sqlite-unicode-ilike`, rebased on the latest `dev`; backend-only (the frontend workflow is skipped for `backend/**` paths).
- No new dependencies; no schema changes, no data migrations — rollback is a pure code revert.
- Verification ran on SQLite 3.4x (default deployment) and an isolated real PostgreSQL 16 instance; `memories` and `channels` have no SQL text-search path and were audited as unaffected.
- Scale path for very large deployments (FTS5/trigram shadow index) is deliberately out of scope — happy to spec separately.

### Screenshots or Videos

- Not applicable — backend-only change, no UI changes.

> [!NOTE]

### DIFF
--- a/backend/open_webui/internal/db.py
+++ b/backend/open_webui/internal/db.py
@@ -6,6 +6,7 @@ import sys
 from contextlib import asynccontextmanager, contextmanager
 from datetime import datetime, timedelta, timezone
 from typing import Any, Optional
+from unicodedata import normalize
 from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
 
 from open_webui.env import (
@@ -31,13 +32,114 @@ from open_webui.utils.json_codec import JSONCodec
 from sqlalchemy import Dialect, MetaData, create_engine, event, types
 from sqlalchemy.engine.url import make_url
 from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker, create_async_engine
+from sqlalchemy.ext.compiler import compiles
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import Session, scoped_session, sessionmaker
 from sqlalchemy.pool import NullPool, QueuePool
+from sqlalchemy.sql.compiler import ilike_case_insensitive
+from sqlalchemy.sql.functions import FunctionElement
 from sqlalchemy.sql.type_api import _T
 from typing_extensions import Self
 
 log = logging.getLogger(__name__)
+
+
+# ── Unicode-aware case-insensitive search on SQLite ─────────────────
+#
+# SQLAlchemy compiles ``ilike()`` on SQLite to ``lower(x) LIKE lower(?)``,
+# but the builtin ``lower()`` folds ASCII only, so searches never match
+# Polish, Cyrillic, Greek, or accented Latin text (upstream issue #29102).
+#
+# Two pieces make the search Unicode-aware:
+#
+# 1. A custom ``unilower()`` SQL function (registered per connection).
+# 2. A compiler override rendering ILIKE case-folding as ``unilower()``
+#    on the SQLite dialect only.
+#
+# The builtin ``lower()`` is deliberately NOT overridden: it backs the
+# ``uq_user_email_lower`` unique expression index (migration
+# f0bd01a18a3d) whose stored entries were computed with the builtin
+# function, and it is used by explicit ``func.lower()`` lookups
+# (e.g. user email). ``unilower`` appears in no index, so it cannot
+# desynchronize any stored state — it only affects query-time search.
+# ─────────────────────────────────────────────────────────────────────
+
+
+def _unicode_lower(value):
+    """Unicode NFKC-caseless fold, mirroring the builtin ``lower()`` contract.
+
+    NULL passes through as NULL, BLOBs pass through unchanged (like the
+    builtin), numbers render as text (``lower(123) -> '123'``), and text
+    is folded with the Unicode caseless-matching algorithm:
+    ``NFKC(casefold(NFKC(x)))``. ``casefold`` alone would still treat the
+    composed (NFC) and decomposed (NFD) spellings of e.g. Polish 'ą' or
+    'ę' as different strings, so the value is normalized before and after
+    folding — pasted text frequently arrives as NFD while typed text is
+    NFC. The same fold is applied to the column and the search pattern,
+    so e.g. Greek 'ΣΑΣ' matches 'σασ' and NFC 'ą' matches NFD 'ą'.
+    ASCII strings use a plain ``str.lower()`` fast path (identical result
+    for every ASCII codepoint, no normalization needed).
+    """
+    if value is None or isinstance(value, bytes):
+        return value
+    if isinstance(value, str):
+        if value.isascii():
+            return value.lower()
+        return normalize('NFKC', normalize('NFKC', value).casefold())
+    return normalize('NFKC', str(value).casefold())
+
+
+def _register_sqlite_search_functions(dbapi_connection):
+    """Register ``unilower()`` on a raw SQLite/SQLCipher DBAPI connection."""
+    try:
+        dbapi_connection.create_function('unilower', 1, _unicode_lower, deterministic=True)
+    except TypeError:
+        # Older SQLite builds without deterministic-function support.
+        dbapi_connection.create_function('unilower', 1, _unicode_lower)
+
+
+@compiles(ilike_case_insensitive, 'sqlite')
+def _sqlite_unilower_case_insensitive(element, compiler, **kw):
+    """Render ILIKE case-folding as ``unilower()`` on SQLite.
+
+    Covers ``ilike``, ``not_ilike``, ``icontains``, and ``not_icontains``.
+    PostgreSQL and other dialects are untouched (they keep native ILIKE).
+    """
+    return f'unilower({element.element._compiler_dispatch(compiler, **kw)})'
+
+
+class UnicodeLower(FunctionElement):
+    """Dialect-aware SQL ``LOWER`` for query-time case-insensitive matching.
+
+    Renders ``lower(...)`` on every dialect except SQLite, where it renders
+    ``unilower(...)`` (Unicode caseless, see above). Intended for
+    case-insensitive comparisons that are NOT backed by a builtin-``lower()``
+    expression index — those (the user-email lookups against
+    ``uq_user_email_lower``) must keep the builtin ``func.lower()`` so the
+    index stays usable and semantically consistent.
+    """
+
+    type = types.String()
+    inherit_cache = True
+
+
+@compiles(UnicodeLower)
+def _render_unicode_lower_default(element, compiler, **kw):
+    return f'lower({compiler.process(element.clauses, **kw)})'
+
+
+@compiles(UnicodeLower, 'sqlite')
+def _render_unicode_lower_sqlite(element, compiler, **kw):
+    return f'unilower({compiler.process(element.clauses, **kw)})'
+
+
+def unicode_caseless(value: str) -> str:
+    """Python-side counterpart of the SQL ``unilower()`` function.
+
+    Use it to pre-fold search patterns (and other in-Python comparisons) so
+    they meet the SQL-side ``unilower()`` folding of stored values.
+    """
+    return _unicode_lower(value)
 
 
 # ── SSL URL normalization (used by sync engine & Alembic migrations) ─
@@ -287,6 +389,9 @@ if SQLALCHEMY_DATABASE_URL.startswith('sqlite+sqlcipher://'):
 
         conn = sqlcipher3.connect(db_path, check_same_thread=False)
         conn.execute(f"PRAGMA key = '{database_password}'")
+        # The SQLCipher engine compiles with the SQLite dialect, so ILIKE
+        # renders as unilower() there too — register it on every connection.
+        _register_sqlite_search_functions(conn)
         return conn
 
     # The dummy "sqlite://" URL would cause SQLAlchemy to auto-select
@@ -341,6 +446,10 @@ elif 'sqlite' in SQLALCHEMY_DATABASE_URL:
         if DATABASE_SQLITE_PRAGMA_JOURNAL_SIZE_LIMIT:
             cursor.execute(f'PRAGMA journal_size_limit={DATABASE_SQLITE_PRAGMA_JOURNAL_SIZE_LIMIT}')
         cursor.close()
+
+        # Register the Unicode-aware search functions used by the ILIKE
+        # compiler override below (upstream issue #29102).
+        _register_sqlite_search_functions(dbapi_connection)
 
     def on_connect(dbapi_connection, connection_record):
         _apply_sqlite_pragmas(dbapi_connection)
--- a/backend/open_webui/internal/db.py
+++ b/backend/open_webui/internal/db.py
@@ -68,20 +68,25 @@ log = logging.getLogger(__name__)
 def _unicode_lower(value):
     """Unicode NFKC-caseless fold, mirroring the builtin ``lower()`` contract.
 
-    NULL passes through as NULL, BLOBs pass through unchanged (like the
-    builtin), numbers render as text (``lower(123) -> '123'``), and text
-    is folded with the Unicode caseless-matching algorithm:
-    ``NFKC(casefold(NFKC(x)))``. ``casefold`` alone would still treat the
-    composed (NFC) and decomposed (NFD) spellings of e.g. Polish 'ą' or
-    'ę' as different strings, so the value is normalized before and after
-    folding — pasted text frequently arrives as NFD while typed text is
-    NFC. The same fold is applied to the column and the search pattern,
-    so e.g. Greek 'ΣΑΣ' matches 'σασ' and NFC 'ą' matches NFD 'ą'.
-    ASCII strings use a plain ``str.lower()`` fast path (identical result
-    for every ASCII codepoint, no normalization needed).
+    NULL passes through as NULL, BLOBs are decoded as text and folded like
+    the builtin (``lower(x'414243')`` is ``'abc'``), numbers render as text
+    (``lower(123) -> '123'``), and text is folded with the Unicode
+    caseless-matching algorithm ``NFKC(casefold(NFKC(x)))``. ``casefold``
+    is applied to both the column and the search pattern, so e.g. Greek
+    'ΣΑΣ' matches 'σασ' and NFC 'ą' matches NFD 'ą'. ASCII strings use a
+    plain ``str.lower()`` fast path (identical result for every ASCII
+    codepoint, no normalization needed).
+
+    Note that folds can change length ('ß' -> 'ss', 'Ｌ' -> 'L'): LIKE
+    single-character wildcards count characters of the folded text, and
+    compatibility folding can turn non-ASCII codepoints into LIKE
+    metacharacters ('％' -> '%') — call sites that escape wildcards must
+    escape the *folded* text, not the raw input.
     """
-    if value is None or isinstance(value, bytes):
-        return value
+    if value is None:
+        return None
+    if isinstance(value, bytes):
+        value = value.decode('utf-8', 'replace')
     if isinstance(value, str):
         if value.isascii():
             return value.lower()
--- a/backend/open_webui/migrations/env.py
+++ b/backend/open_webui/migrations/env.py
@@ -6,7 +6,12 @@ import logging.config
 import logging
 import alembic.context
 from open_webui.env import DATABASE_PASSWORD, DATABASE_URL, LOG_FORMAT
-from open_webui.internal.db import enable_iam_token_auth, extract_ssl_params_from_url, reattach_ssl_params_to_url
+from open_webui.internal.db import (
+    _register_sqlite_search_functions,
+    enable_iam_token_auth,
+    extract_ssl_params_from_url,
+    reattach_ssl_params_to_url,
+)
 from open_webui.models.auths import Auth
 from open_webui.models.calendar import Calendar, CalendarEvent, CalendarEventAttendee  # noqa: F401
 from open_webui.models.chat_messages import ChatMessage  # noqa: F401
@@ -71,6 +76,14 @@ def run_migrations_online() -> None:
     """Execute migrations against a live database connection."""
     live_connectable = _get_engine_connectable()
     enable_iam_token_auth(live_connectable)
+    if target_db_url.startswith('sqlite'):
+        # The ILIKE compiler override emits unilower() on the SQLite dialect
+        # (see open_webui.internal.db). Register the SQL function on this
+        # engine's connections too so any future migration using ilike()
+        # cannot hit "no such function: unilower" here.
+        from sqlalchemy import event
+
+        event.listen(live_connectable, 'connect', _register_sqlite_search_functions)
     with live_connectable.connect() as live_connection:
         alembic.context.configure(
             connection=live_connection,
--- a/backend/open_webui/models/chats.py
+++ b/backend/open_webui/models/chats.py
@@ -63,12 +63,12 @@ def chat_search_message_content_match_sql(dialect_name: str, key: str) -> str:
             EXISTS (
                 SELECT 1
                 FROM json_each(Chat.chat, '$.history.messages') AS history_message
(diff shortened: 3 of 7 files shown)


Alert17:
Tested this branch against a SQLite install with non-ASCII content. The core fix works: of 16 Unicode cases I tried, 12 go from "no match" to "match" — ß, unaccented Greek, NFD vs NFC, fullwidth letters, ﬁ ligature, long ſ, Cyrillic, Roman numerals, circled digits, superscripts, fractions, ㎡. Turkish İ/ı and accented Greek stay unmatched, which looks correct to me.

Three edge cases where the fold has side effects. None are blockers, but the first is a wrong result rather than a missed match.

**1) NFKC produces LIKE metacharacters.**

Ten codepoints fold into LIKE wildcards or the escape character:

```
-> '%'   U+FE6A ﹪   U+FF05 ％
-> '_'   U+FE33 ︳  U+FE34 ︴  U+FE4D ﹍  U+FE4E ﹎  U+FE4F ﹏  U+FF3F ＿
-> '\'   U+FE68 ﹨   U+FF3C ＼
```

So searching for a literal ％ matches every row instead of none:

```
unilower('％') == '%'   ->   LIKE '%'   ->   all rows
```

Call sites that escape wildcards (e.g. `files.py` `ilike(pattern, escape='\\')`) escape ASCII `%` and `_` only, so these pass through unescaped and become metacharacters after the fold. Escaping the folded pattern, or folding before escaping, would close it.

**2) The BLOB contract differs from the builtin.**

The docstring says BLOBs pass through unchanged "like the builtin", but the builtin converts them to text:

```
lower(x'414243')     -> 'abc'   (typeof: text)
unilower(x'414243')  -> b'ABC'  (typeof: blob)
```

Either behaviour is defensible; the comment just does not match what SQLite does today.

**3) Length-changing folds break single-character wildcards.**

```
'stra_e' matches lower('Straße')    = 'straße'   (6 chars)
but not          unilower('Straße') = 'strasse'  (7 chars)
```

Inherent to casefold expansion, probably acceptable — worth a line in the docstring.

And a question rather than a finding: casefold makes SQLite more permissive than PostgreSQL. `'%strasse%'` matches `Straße` on SQLite with this patch, but PostgreSQL's native ILIKE does not fold ß. Is that divergence intended?

Happy to share the test script if useful.

Alert17:
Re-ran everything against the updated branch — both fixes hold.

The fold now escapes correctly: glob `*％*` compiles to `%\%%`, so the fullwidth percent stays a literal and matches one file instead of all four. BLOB parity matches the builtin exactly (`unilower(x'414243')` = `'abc'`, typeof `text`). Gating the pattern fold to SQLite so PostgreSQL keeps the raw glob is a good catch — I had not thought that far.

Here is the script, standard library only, no Open WebUI import and no database file. It exits non-zero on failure, so it drops into CI as-is if you want it.

One note it surfaced that is not in the list above: the escaping fix covers `Files.search_files`, which is the only call site that escapes. On the sites that build patterns as `ilike(f'%{q}%')` with no escape, a folded `％` still acts as a wildcard — but so does an ASCII `%` typed by a user there today, so it is consistent with existing behaviour rather than a new problem. The script asserts that equivalence instead of asserting a literal match, so it will not fail on it.

Feel free to fold any of this into the PR, trim it, or ignore it.

<details>
<summary>unicode_fold_tests.py</summary>

```python
"""Checks for the SQLite Unicode-aware ILIKE fold (open-webui PR #29186).

Standalone: standard library only, no Open WebUI import, no database file.
Run with `python3 unicode_fold_tests.py`.

Covers four areas:
  A. case-insensitive matching across scripts and normalization forms
  B. codepoints that compatibility folding turns into LIKE metacharacters
  C. parity with SQLite's builtin lower() for NULL, BLOB and numbers
  D. glob escaping when the fold introduces metacharacters
"""
import sqlite3
import sys
from unicodedata import normalize

FAIL = 0


def check(name, got, want):
    global FAIL
    ok = got == want
    FAIL += not ok
    print(f"  {'ok  ' if ok else 'FAIL'}  {name:52} got={got!r} want={want!r}")


def unilower(value):
    """The fold under test, as implemented in the PR."""
    if value is None:
        return None
    if isinstance(value, bytes):
        value = value.decode('utf-8', 'replace')
    if isinstance(value, str):
        if value.isascii():
            return value.lower()
        return normalize('NFKC', normalize('NFKC', value).casefold())
    return normalize('NFKC', str(value).casefold())


def glob_to_like(glob, fold=True):
    """Mirrors Files._glob_to_like_pattern: fold first, then escape."""
    if fold:
        glob = unilower(glob)
    pattern = glob.replace('\\', '\\\\').replace('%', '\\%').replace('_', '\\_')
    return pattern.replace('*', '%').replace('?', '_')


def conn(rows=()):
    c = sqlite3.connect(':memory:')
    c.create_function('unilower', 1, unilower, deterministic=True)
    c.execute('create table t (v)')
    c.executemany('insert into t values (?)', [(r,) for r in rows])
    return c


def matches(rows, pattern, escape=False):
    c = conn(rows)
    sql = 'select count(*) from t where unilower(v) like unilower(?)'
    if escape:
        sql += " escape '\\'"
    return c.execute(sql, (pattern,)).fetchone()[0]


print('A. matching across scripts and normalization forms')
for data, pattern, want, label in [
    ('Straße', '%STRASSE%', 1, 'German sharp s folds to ss'),
    ('Straße', '%straße%', 1, 'German sharp s, exact spelling'),
    ('ΟΔΟΣ', '%οδος%', 1, 'Greek, final sigma'),
    ('ΟΔΟΣ', '%οδός%', 0, 'Greek accent is not stripped'),
    ('Привіт', '%привіт%', 1, 'Cyrillic'),
    ('ＡＢＣ', '%abc%', 1, 'fullwidth letters'),
    ('ﬁle', '%file%', 1, 'fi ligature'),
    ('ſtraße', '%strasse%', 1, 'long s'),
    ('ą', 'ą'.replace('ą', 'ą'), 1, 'NFD data, NFC pattern'),
    ('ą', 'ą', 1, 'NFC data, NFD pattern'),
    ('Ⅻ', '%xii%', 1, 'Roman numeral'),
    ('m²', '%m2%', 1, 'superscript'),
    ('İstanbul', '%istanbul%', 0, 'Turkish dotted I is locale-specific'),
]:
    check(label, matches([data], pattern), want)

print('\nB. codepoints that folding turns into LIKE metacharacters')
meta = {'%': [], '_': [], '\\': []}
for cp in range(0x20, 0x110000):
    ch = chr(cp)
    if ch in '%_\\':
        continue
    folded = unilower(ch)
    for m in meta:
        if m in folded:
            meta[m].append(ch)
for m, chars in meta.items():
    names = ' '.join(f'U+{ord(c):04X}' for c in chars)
    print(f"  folds to {m!r}: {len(chars)} codepoints  {names}")
# On call sites that do not escape (the majority: `ilike(f'%{q}%')`), an ASCII
# '%' typed by a user is already a wildcard today. After the fold the eight
# codepoints above behave the same way — consistent with existing behaviour
# rather than a new class of bug, but worth knowing.
rows = ['first', 'second', 'third']
check('unescaped site: fullwidth percent behaves like ASCII percent',
      matches(rows, '％'), matches(rows, '%'))
check('unescaped site: both are wildcards there', matches(rows, '%'), len(rows))

print('\nC. parity with the builtin lower()')
c = conn()
for expr in ("x'414243'", '123', '1.5', 'NULL', "'ABC'"):
    a = c.execute(f'select lower({expr})').fetchone()[0]
    b = c.execute(f'select unilower({expr})').fetchone()[0]
    check(f'lower({expr}) == unilower({expr})', b, a)
for expr in ("x'414243'", '123', 'NULL'):
    a = c.execute(f'select typeof(lower({expr}))').fetchone()[0]
    b = c.execute(f'select typeof(unilower({expr}))').fetchone()[0]
    check(f'typeof matches for {expr}', b, a)

print('\nD. glob escaping when the fold introduces metacharacters')
files = ['report.pdf', 'Straße doc.pdf', 'discount 50％.txt', 'file_1.txt']
for glob, want, label in [
    ('*％*', 1, 'fullwidth percent stays a literal'),
    ('*STRASSE*', 1, 'fold still matches across case and ss'),
    ('*_1*', 1, 'underscore in the glob stays a literal'),
    ('*file?1*', 1, 'glob ? is still a single-character wildcard'),
    ('*', len(files), 'bare star matches everything'),
]:
    check(label, matches(files, glob_to_like(glob), escape=True), want)
check('unfolded glob would leak a wildcard (regression guard)',
      matches(files, glob_to_like('*％*', fold=False), escape=True), len(files))

print(f"\n{'all checks passed' if not FAIL else str(FAIL) + ' check(s) failed'}")
sys.exit(1 if FAIL else 0)
```

</details>

Classic298:

BLACKTHOMAS:
@Classic298 Thanks for the triage. Could you point out which of the checklist items applied here? I'd genuinely like to address each one, and from my reading a few don't seem to match the diff:

*cc @tjbck — flagging you only in case the closure was a policy call that needs a maintainer's decision rather than a content one; the question below is otherwise aimed at the triage checklist.*

- **Test files / i18n files:** this PR touches 7 backend files (+196/−34) — no test files are committed to the repo and no i18n files are touched.
- **Testing / logs:** fair point — the PR body had no attached logs, that's on me. Building the before/after log actually surfaced one more real bug on my side (the alembic `connect` listener had the wrong signature, so a **fresh SQLite install failed to migrate**) — fixed on the branch, and the full 55-migration chain now runs clean on a fresh database. Evidence against the current PR head below, including a reproduction of #29102 on unpatched `dev` vs this PR through the real production search function. The fix is also independently verified by @Alert17 — the reporter of the linked issue #29102, who re-ran everything against this branch on a SQLite deployment with non-ASCII content (*"Re-ran everything against the updated branch — both fixes hold"*), including their own standalone script, which passes cleanly against this PR's fold.
- **First-time contributor:** I have at least 2 merged PRs in this repo (#20562, #29184), so this isn't a first-time contribution — unless the policy requires a maintainer to explicitly invite code PRs even for returning contributors, in which case I'm happy to wait for that sign-off on the linked issue #29102.
- **Scope:** if the concern is breadth (SQLite ILIKE compiler hook + touched search paths + the PostgreSQL `::json` casts), I'm happy to split the PostgreSQL JSON-operator fix into its own PR and keep this one focused on SQLite search only.

<details>
<summary>Test evidence</summary>

## Reproduction of #29102 — upstream `dev` vs this PR (same script, real `Chats.get_chat_list_by_user_id`)

Seeded chats: `Łąka pełna maków`, `Zimowa łąka nad rzeką`, `ΣΑΣ ΣΥΜΒΑΣΗ γραμμή`, `Привіт, світе`, `Plain English Search`.

```
$ python search_repro.py <backend from upstream `dev`>   # BEFORE this PR
  query 'ŁĄKA'       -> 0 match(es): NO MATCHES
  query 'łąka'       -> 1 match(es): Zimowa łąka nad rzeką
  query 'σασ'        -> 0 match(es): NO MATCHES
  query 'привіт'     -> 0 match(es): NO MATCHES
  query 'english'    -> 1 match(es): Plain English Search

$ python search_repro.py <backend from this PR>          # AFTER
  query 'ŁĄKA'       -> 2 match(es): Zimowa łąka nad rzeką, Łąka pełna maków
  query 'łąka'       -> 2 match(es): Zimowa łąka nad rzeką, Łąka pełna maków
  query 'σασ'        -> 1 match(es): ΣΑΣ ΣΥΜΒΑΣΗ γραμμή
  query 'привіт'     -> 1 match(es): Привіт, світе
  query 'english'    -> 1 match(es): Plain English Search
```

The full alembic migration chain (55 migrations) runs clean to head on a fresh SQLite database with this PR active.

## Alert17 standalone script (his own code, no repo import)

```
C. parity with the builtin lower()
  ok    lower(x'414243') == unilower(x'414243')              got='abc' want='abc'
  ok    lower(123) == unilower(123)                          got='123' want='123'
  ok    lower(1.5) == unilower(1.5)                          got='1.5' want='1.5'
  ok    lower(NULL) == unilower(NULL)                        got=None want=None
  ok    typeof matches for x'414243'                         got='text' want='text'
D. glob escaping when the fold introduces metacharacters
  ok    fullwidth percent stays a literal                    got=1 want=1
  ok    underscore in the glob stays a literal               got=1 want=1
  ok    unfolded glob would leak a wildcard (regression guard) got=4 want=4

all checks passed
```

## SQLite harness — 58 checks (property-based SQL≡casefold equivalence over 5000 rows/6 alphabets, real production search functions, glob metacharacter + BLOB parity regressions, 1M-row benchmark)

```
58 passed, 0 failed
```

## PostgreSQL 16 harness — 32 checks (native ILIKE preserved, ::json casts, per-dialect fold pairing, ς regression)

```
32 passed, 0 failed
```

## Lint (CI-equivalent)

```
All checks passed!
```

</details>

If the closure was policy-driven rather than content-driven, I understand — but then #29102 stays open with case-insensitive search returning 0 rows for non-ASCII text on the default database configuration. In that case I'd appreciate a pointer on how you'd like this handled upstream: a maintainer's go-ahead on #29102, a narrower scope, or handing the approach and harnesses over for someone else to land — all fine by me.

If the items above address the closure reasons, could the PR be reopened?

PR #29205, opened 2026-08-29
refactor: single-pass calendar recurrence expansion and rule validation

Classic298:
Expand each recurring event in one forward pass with a ceiling on the occurrences examined, rather than re-seeking from the rule anchor for every instance. Validate the recurrence rule when an event is created or updated, and anchor expansion to the event's own start rather than any DTSTART carried in the rule text.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/calendar.py
+++ b/backend/open_webui/utils/calendar.py
@@ -6,11 +6,37 @@ RRULE expansion reusing the automation infra.
 
 import datetime as dt
 import logging
+from itertools import islice
 from zoneinfo import ZoneInfo
 
+from dateutil.rrule import HOURLY, MINUTELY, SECONDLY, rruleset, rrulestr
+from open_webui.constants import ERROR_MESSAGES
 from open_webui.utils.automations import _resolve_tz
 
 log = logging.getLogger(__name__)
+
+# The read path sees rules the validator never did: pre-existing rows and automation rrules
+MAX_OCCURRENCES_SCANNED = 100_000
+
+
+def validate_calendar_rrule(rrule_str: str) -> None:
+    """Raise ValueError if the RRULE is unsupported, unparseable, or repeats more often than daily."""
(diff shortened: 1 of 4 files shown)


backend/open_webui/utils/calendar.py
```diff
@@ -6,12 +6,38 @@
...
+    """Raise ValueError if the RRULE is unsupported, unparseable, or repeats more often than daily."""
+    if 'EXRULE' in rrule_str.upper():
+        raise ValueError(ERROR_MESSAGES.EXRULE_UNSUPPORTED)
+
+    try:
+        parsed = rrulestr(rrule_str, ignoretz=True)
+    except Exception as e:
+        raise ValueError(ERROR_MESSAGES.INVALID_RRULE(e))
+
+    rules = parsed._rrule if isinstance(parsed, rruleset) else [parsed]
```
Jokasa7, line 32: This drops `parsed._rdate` from validation even though iterating the `rruleset` below includes those dates. For example, `DTSTART:20260830T000000\nRRULE:FREQ=DAILY;COUNT=2\nRDATE:20260830T010000,20260830T020000` passes this loop, then expansion creates `00:00`, `01:00`, and `02:00` instances on August 30. That directly bypasses `CALENDAR_RRULE_TOO_FREQUENT`. Could we either reject RDATE as unsupported here or validate adjacent occurrences from the combined set, and add this case as a regression test?

Jokasa7 (commented):
Thanks for tightening the recurrence bounds. I ran this exact head with a daily RRULE plus two same-day RDATE inclusions. `validate_calendar_rrule()` accepts it, but `expand_recurring_event()` emits three instances on that date. The validator inspects only the RRULE components while iteration includes the complete rruleset, so RDATE bypasses the new daily-frequency limit. Please either reject RDATE for calendar input or validate the combined occurrence set, with a regression covering the create/update validation and expansion paths.

Issue #29216, opened 2026-08-29
issue: chat history truncates after reload when assistant messages have no parent link


vanwho:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I have checked whether this issue is already fixed on the `dev` branch or in the latest source.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.11.1

### Operating System

Ubuntu 26.04

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

  Chats should be continuous.  
  Every assistant response generated for a user message should be persisted with:

  - role: "assistant"
  - the correct id
  - parentId set to the corresponding user message ID
  - a valid childrenIds array
  - a timestamp
  - a corresponding entry in the chat message tree

  The placeholder should not be skipped merely because its ID matches metadata.assistant_message_id.

### Actual Behavior

  After sending a message, the assistant response appears normally. However, after reloading the chat or sending a follow-up:

  - the assistant response may become detached from the conversation tree (only the latest response is seen);
  - the saved assistant message has no valid `parentId`;
  - `currentId` may point to a node that is not reachable from the conversation root;
  - the next request may omit previous messages or send an incomplete history;
  - the chat can appear to lose its previous context.

[...]

Issue #26257, opened 2026-06-23
bug: Clicking follow-up suggestion saves completed response without role/parentId/timestamp, permanently breaking chat on reload


scottybo:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.9.6

### Ollama Version (if applicable)

Not applicable — using OpenAI-compatible API (gpt-5.5)

### Operating System

Linux (Docker container) / client: macOS

### Browser (if applicable)

Not applicable

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

After clicking a follow-up suggestion button below an assistant response, the generated reply should be saved as a properly-linked assistant message in the conversation tree. On page reload, the full chat should render including the follow-up exchange.

### Actual Behavior

After clicking a follow-up suggestion and letting the response stream to full completion (no errors, no cancellation), the chat becomes permanently broken. On any subsequent reload, the chat renders blank or shows only a single detached message. The chat is unrecoverable without direct database repair.

The root cause (confirmed via SQLite inspection) is that the completed response is saved to history.messages without its message-graph fields:
- role: missing (should be "assistant")
- parentId: null (should reference the user follow-up message)
- timestamp: missing

history.currentId is set to this malformed message ID.

[...]

PR #29218, opened 2026-08-29
fix: preserve parent links when continuing assistant responses

vanwho:
# Pull Request Checklist

- [x] This PR links to a well-described Issue: Closes [#29216](https://github.com/open-webui/open-webui/issues/29216).
- [x] I matched nearby code patterns and avoided unnecessary settings, abstractions, or dependencies.
- [x] I manually tested the changed workflow and nearby behavior.
- [x] Documentation changes are not required for this persistence bug.
- [x] Screenshots are not required because this does not change the UI.
- [x] I reviewed the implementation and tests before submitting.
- [x] The PR title uses the `fix:` prefix.

## Summary

Closes [#29216](https://github.com/open-webui/open-webui/issues/29216).

Assistant responses can become detached from their corresponding user messages when an existing-chat request includes an `assistant_message_id`. The persistence path currently skips that assistant ID entirely because continuation requests reuse an existing assistant message.

The skip has an important purpose: recreating the existing message as an empty placeholder would overwrite its content, output, and status. However, skipping all persistence also means an existing assistant message whose `parentId` is missing is never repaired.

This change preserves the continuation behavior while repairing the orphaned relationship:

- Load the existing assistant message when its ID matches `metadata.assistant_message_id`.
- If the message exists and its `parentId` does not match the current user message, update only `parentId`.
- Continue without recreating the assistant placeholder, preserving its content, output, role, model, and status.
- If the assistant message does not exist, fall through to the existing full-placeholder creation path.

This applies to the shared continuation path used by continued responses and tool-approval continuations. It introduces no settings, dependencies, API changes, or UI changes.

Regression coverage is provided separately in:

- [open-webui/tests#9](https://github.com/open-webui/tests/pull/9)

## Testing

External regression test:

```bash
OPEN_WEBUI_SOURCE_DIR=/path/to/open-webui/backend \
python -m pytest unit/chat/test_assistant_parent_repair.py -q
```

Results against this fix:

```text
2 passed
```

The two cases verify that:

- An assistant message with `parentId: null` receives a parent-only update.
- An assistant message that already has the correct parent is not rewritten.

Discrimination was also verified against the pre-fix `dev` revision:

```text
1 failed, 1 passed
```

The orphan-repair case fails because the old code performs no parent update. The already-correct nearby case continues to pass.

Repository checks:

```bash
ruff format --check . --exclude .venv --exclude venv
ruff check --select=F --ignore=F401,F403,F405,F541,F811,F841 .
git diff --check
```

Results:

```text
273 files already formatted
All checks passed
```

### Added

- None.

### Changed

- None.

### Fixed

- Repair missing assistant-message parent links during continuation without replacing the existing assistant response.

### Removed

- None.

### Security

- None.

### Breaking Changes

- None.

## Additional Context

- Fixes Issue: [#29216](https://github.com/open-webui/open-webui/issues/29216)
- Regression tests: [open-webui/tests#9](https://github.com/open-webui/tests/pull/9)
- Related prior chat-history recovery issue: #26257
- The production change is limited to `backend/open_webui/main.py`.
- The regression test is intentionally maintained in the separate `open-webui/tests` repository.

> [!NOTE]

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -1576,7 +1576,19 @@ async def chat_completion(
                         target_model_id = entry['model_id']
                         assistant_message_id = entry['message_id']
                         if assistant_message_id and assistant_message_id == metadata.get('assistant_message_id'):
-                            continue
+                            # Continuations reuse an existing assistant message. Preserve its
+                            # content/output while repairing an orphaned parent link.
+                            existing_assistant_message = await Chats.get_message_by_id_and_message_id(
+                                chat_id, assistant_message_id
+                            )
+                            if existing_assistant_message:
+                                if user_message_id and existing_assistant_message.get('parentId') != user_message_id:
+                                    await Chats.upsert_message_to_chat_by_id_and_message_id(
+                                        chat_id,
+                                        assistant_message_id,
+                                        {'parentId': user_message_id},
+                                    )
+                                continue
                         if assistant_message_id:
                             assistant_message = {
                                 'id': assistant_message_id,


Classic298:
pending additional info on issue

Issue #29236, opened 2026-08-30
issue: a message queued during a pending ask_user prompt gets a response that is never rendered


silentoplayz:
### Before Submitting

- [x] I searched open and closed issues and discussions for an existing report.
- [x] I checked whether this is already fixed on the `dev` branch or latest source.
- [x] I understand that maintainers want a well-written issue before any code pull request.
- [x] I am using the latest available version of Open WebUI for my install method.
- [x] This is not a security vulnerability.

### Installation Method

Git Clone

### Open WebUI Version

dev, commit a93c5080380447e3ab9113cacb73b61c57cbc308

### Operating System

Ubuntu 26.04.1 LTS

### Browser

Mozilla Firefox for Snap v154.0.1 (64-bit)

### Ollama Version

v0.33.2

### Summary

When a message is queued while an `ask_user` prompt is waiting for input, the chat view stops reflecting what is actually happening. The queued message is sent and the model does answer it, but the exchange is not rendered: the conversation and the Overview tab are left missing the response, and sometimes the queued user message as well. It only appears after the page is reloaded or after another message is sent.

### Expected Behavior

A message queued while an `ask_user` prompt is pending stays queued and visible until the tool call is resolved and the completion it resumes has finished. When it is sent, its response streams into the conversation like any other response, without needing a reload.

### Actual Behavior

The queued message is sent and the model answers it, but the conversation does not show that. What is missing varies: in one recording the queued user message and its response were both absent from the conversation and from the Overview tab, and in another the user message was shown with nothing under it. In both cases the exchange was already complete on the server, and reloading the page or sending any further message revealed it in full.

### Steps to Reproduce

1. Use a model with native function calling and the built-in `ask_user` tool available.
2. Start a new chat and send a prompt that makes the model call `ask_user`, for example: `Help me plan an automated backup strategy for my homelab server. First call the ask_user tool to ask me 2 quick questions about my setup, each with 2-3 answer options. Then give me a short written recommendation.`
3. Wait for the `ask_user` card to appear with its questions.
4. While the card is still waiting for input, type a short message and send it. It is queued rather than sent, and appears as a queued entry above the composer.
5. Answer the `ask_user` questions and submit them.
6. Watch the queued entry, the conversation, and the Overview tab as the resumed response streams and finishes.
7. Reload the page, or send any further message, and the missing response appears.

### Logs, Screenshots, and Config

Screen recordings of both outcomes:

[...]

PR #29243, opened 2026-08-30
fix: only resync a chat when the message whose completion ended is still unfinished

silentoplayz:
# Pull Request

## Checklist

- [x] This PR links to a well-described, confirmed Issue or active Discussion: `Closes #29236`.

## Summary

A message queued while an `ask_user` prompt is pending is sent and answered, but the answer is never drawn: the conversation and the Overview tab stay missing the response, and sometimes the queued user message too, until the page is reloaded or another message is sent (#29236).

Root cause, anchored to commit a93c5080380447e3ab9113cacb73b61c57cbc308. When a completion ends the server emits `chat:active` with `active: false`, and the handler reloads the whole chat if any assistant leaf anywhere in the history is unfinished:

```js
if ($chatId && !$temporaryChatEnabled && hasPendingAssistantLeaf()) {
	await loadChat();
}
```

The queue is flushed as soon as that completion finishes, so `submitPrompt` has already created the next user message and its assistant placeholder by the time the event arrives. That placeholder is an assistant message that is not `done` and has no children, which is exactly what `hasPendingAssistantLeaf` looks for, so the handler treats a request that is legitimately in flight as evidence that the client is out of sync. `loadChat` then assigns `history = chatContent.history`, and because the new response has not been persisted yet, its id disappears from the local history. From that point every socket event for it is discarded by the `let message = history.messages[event.message_id]; if (message) { ... }` lookup, so the model's answer is generated and stored but never rendered. Reloading, or sending another message, rebuilds the history and the answer appears.

The reload itself is worth keeping: it recovers a chat whose completion ended without the client seeing it. What is wrong is the scope. `hasPendingAssistantLeaf` was written for the socket reconnect path, where scanning the whole history is correct because any message could have finished while the socket was down. In this handler the event names the completion that ended, so the same predicate belongs on that message alone:

```diff
-						if ($chatId && !$temporaryChatEnabled && hasPendingAssistantLeaf()) {
+						if (
+							$chatId &&
+							!$temporaryChatEnabled &&
+							!message.done &&
+							(message.childrenIds?.length ?? 0) === 0
+						) {
 							await loadChat();
 						}
```

`hasPendingAssistantLeaf` keeps its original caller in `handleSocketConnect`, unchanged. One file, one condition.

Two independent conditions now prevent the clobber in the reported flow: when the resumed completion ends its message is `done`, and by the time the event arrives the flushed queued message has become its child. The recovery case still behaves as before, since a completion that dies without reporting done leaves its own message unfinished with no children, and that still triggers the reload.

This PR was prepared with AI copilot assistance and I have thoroughly reviewed and manually tested it.

## Testing

1. Reproduced the issue on `dev` first: with a model using native function calling, sent a prompt that calls `ask_user`, queued a message while the card was waiting, answered the card, and confirmed the queued exchange never rendered until a reload.
2. On this branch, ran the same flow and the queued message's response streams into the conversation normally, with no reload needed.
3. Checked the surrounding behavior that shares this handler: ordinary chats with nothing queued, stopping a response mid stream, switching between chats while one is generating (notification still came through when the assistant completed the response), and refreshing the tab/page while the assistant's response was generating.

While investigating I also confirmed the message states through a full `ask_user` exchange on a clean instance: the assistant message stays `done: false` while the card waits and for the whole resumed stream, and flips to `done: true` only when it finishes, so the queue is correctly held for the entire exchange and flushes only after it.

### Fixed

- A message sent while a response is finishing no longer loses its reply in the UI; the chat is only resynced when the completion that ended is the one still showing as unfinished.

## Additional Context

Two related things I left out of this change on purpose:

- The condition is now the same predicate `hasPendingAssistantLeaf` applies, evaluated on a single message. Extracting an `isPendingAssistantLeaf(message)` helper shared by both call sites would remove that duplication, but it changes a second call site so I did not include it here.
- Events whose `message_id` is not in the local history are dropped silently by the lookup in `chatEventHandler`. That is what turns this race into a permanently blank response rather than a brief glitch, and it may deserve handling of its own, but recovering there is a larger change with its own failure modes.

### DIFF
--- a/src/lib/components/chat/Chat.svelte
+++ b/src/lib/components/chat/Chat.svelte
@@ -1219,7 +1219,12 @@
 				} else if (type === 'chat:active') {
 					if (!data?.active) {
 						taskIds = null;
-						if ($chatId && !$temporaryChatEnabled && hasPendingAssistantLeaf()) {
+						if (
+							$chatId &&
+							!$temporaryChatEnabled &&
+							!message.done &&
+							(message.childrenIds?.length ?? 0) === 0
+						) {
 							await loadChat();
 						}
 						if ($chatId && !$temporaryChatEnabled) {


silentoplayz:
Landed on `dev` as 22379ded1aebf0537e1ee3f00a20809d47646bf6, shipped in v0.11.2. Same scoping, applied by giving `hasPendingAssistantLeaf` an optional `messageId` and passing `event?.message_id` from the `chat:active` handler, which keeps the whole history scan as the fallback when the event carries no message id.

Issue #29187, opened 2026-08-29
issue: PDF OCR extraction crashes on images with composite or missing /Filter


fengyufeiyang-dot:
## Before Submitting

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items.
- [x] I have checked whether this issue is already fixed on the `dev` branch or in the latest source.
- [x] I am using the latest version of Open WebUI.

## Installation Method

Docker

## Open WebUI Version

v0.11.1

## Problem Description

When "PDF Extract Images (OCR)" is enabled, uploading certain PDF files causes a crash with `ValueError: cannot reshape array of size X into shape (Y,Z,newaxis)`. This affects multiple PDF files and has been reported since v0.5.7 (issues #4457, #6085, #8856, #10913, #11333, #25979; discussion #8877).

The crash occurs in `langchain_community.document_loaders.parsers.pdf.PyPDFLoader.extract_images_from_page()` at line 147:

```python
np.frombuffer(xObject[obj].get_data(), dtype=np.uint8).reshape(
    width, height, newaxis
)
```

## Root Cause

Two distinct failure modes in `extract_images_from_page()`:

1. **Composite filters** (e.g., `/Filter` is an array like `[DCTDecode, ASCII85Decode]`): The code assumes `/Filter` is a single `NameObject`, but when it is an array, `xObject[obj]["/Filter"][0]` returns the first element of the array, and `[1:]` strips the leading `/` incorrectly, producing an invalid filter name.

2. **Raw pixel data** (no filter / empty filter): When the image has no filter (raw pixels), `np.frombuffer(...).reshape(width, height, newaxis)` fails because the byte count does not match `width * height * 1` (e.g., size 11609 vs expected 2954x3753=11,087,342). The data may include padding, stride alignment, or be a different color space.

## Proposed Fix

In `extract_images_from_page()`, add safe handling:

1. For composite filters: use the first recognized filter in the array instead of crashing.
2. For raw pixel data: validate byte count before reshape; if mismatch, log a warning and skip the image instead of crashing.
3. For single-image OCR failure: catch exceptions and skip that image, so the entire document does not fail.

This fix has been verified in production with scanned PDFs, mixed PDFs (text + scanned pages), and PDFs with composite filter images.

## Reproduction Steps

1. Enable "PDF Extract Images (OCR)" in Admin Settings > Documents
[...]

Issue #11333, opened 2025-03-07
issue: cannot reshape array of size 995 into shape (92,696,newaxis)


henry19801109:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.

### Installation Method

Git Clone

### Open WebUI Version

v0.5.20

### Ollama Version (if applicable)

v0.5.12

### Operating System

windows

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have checked the browser console logs.
- [x] I have checked the Docker container logs.
- [x] I have listed steps to reproduce the bug in detail.

### Expected Behavior

 File "D:\python\Lib\threading.py", line 1002, in _bootstrap
    self._bootstrap_inner()
    │    └ <function Thread._bootstrap_inner at 0x00000207EBF81E40>
    └ <WorkerThread(AnyIO worker thread, started 27796)>
  File "D:\python\Lib\threading.py", line 1045, in _bootstrap_inner
    self.run()
    │    └ <function WorkerThread.run at 0x000002079A575C60>
    └ <WorkerThread(AnyIO worker thread, started 27796)>
  File "D:\AI\openwebuiVenv\Lib\site-packages\anyio\_backends\_asyncio.py", line 962, in run
    result = context.run(func, *args)
             │       │   │      └ ()
             │       │   └ functools.partial(<function upload_file at 0x0000020793CB9BC0>, user=UserModel(id='c4df9b5a-cffa-4751-a019-de8e33d581bb', nam...
             │       └ <method 'run' of '_contextvars.Context' objects>
             └ <_contextvars.Context object at 0x000002079A5FBE00>
> File "D:\AI\openwebuiVenv\Lib\site-packages\open_webui\routers\files.py", line 85, in upload_file
    process_file(request, ProcessFileForm(file_id=id), user=user)
    │            │        │                       │         └ UserModel(id='c4df9b5a-cffa-4751-a019-de8e33d581bb', name='henry19801109@163.com', email='henry19801109@163.com', role='admin...
    │            │        │                       └ '16d61314-9e71-41dc-a94b-77b84730753b'
    │            │        └ <class 'open_webui.routers.retrieval.ProcessFileForm'>
    │            └ <starlette.requests.Request object at 0x000002079A63EA10>
    └ <function process_file at 0x0000020796B4D1C0>
[...]

Issue #10913, opened 2025-02-27
Can't upload documents in 0.5.17: "400: cannot reshape array of size 305 into shape (143,149,newaxis)"


ftab:
## Installation Method

helm

## Environment

- **Open WebUI Version:** 0.5.17
- **Ollama (if applicable):** not used (it's talking to LM Studio instead)

- **Operating System:** Debian
- **Browser (if applicable):** Brave

**Confirmation:**

- [x] I have read and followed all the instructions provided in the README.md.
- [x] I am on the latest version of both Open WebUI and Ollama.
- [ ] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have provided the exact steps to reproduce the bug in the "Steps to Reproduce" section below.

## Expected Behavior:

Can upload a document, and talk to the LLM about it

## Actual Behavior:

The upload fails with the mentioned error message and the LLM describes the context as "empty"

## Description

**Bug Summary:**
Document uploading seems to be broken in 0.5.17

## Reproduction Details

**Steps to Reproduce:**
1. Upload a document
2. Ask a question about the document

## Logs and Screenshots

**Docker Container Logs:**
```
2025-02-27 18:58:41.152 | ERROR    | open_webui.routers.retrieval:process_file:1071 - cannot reshape array of size 305 into shape (143,149,newaxis) - {}
Traceback (most recent call last):

  File "/usr/local/lib/python3.11/threading.py", line 1002, in _bootstrap
    self._bootstrap_inner()
    │    └ <function Thread._bootstrap_inner at 0x7fc09f108860>
    └ <WorkerThread(AnyIO worker thread, started 140462719485632)>
  File "/usr/local/lib/python3.11/threading.py", line 1045, in _bootstrap_inner
    self.run()
    │    └ <function WorkerThread.run at 0x7fc030ae39c0>
    └ <WorkerThread(AnyIO worker thread, started 140462719485632)>
  File "/usr/local/lib/python3.11/site-packages/anyio/_backends/_asyncio.py", line 962, in run
    result = context.run(func, *args)
             │       │   │      └ ()
             │       │   └ functools.partial(<function upload_file at 0x7fc0708b2a20>, user=UserModel(id='a01008c6-8c99-4aeb-b745-8cd05a73fe5e', name='D...
             │       └ <method 'run' of '_contextvars.Context' objects>
             └ <_contextvars.Context object at 0x7fc0239bab80>

[...]

Issue #8856, opened 2025-01-24
PDF Upload -  cannot reshape array of size #X into shape


BloodBlight:
# Bug Report

## Installation Method

Docker

## Environment
- **Open WebUI Version:** ghcr.io/open-webui/open-webui:main (v0.5.7)
Note: _I had to go to the logs for this, could we add this to the UI somewhere?_
- **Operating System:** PopOS
- **Browser:** Tried both Firefox and Chromium

**Confirmation:**

- [X] I have read and followed all the instructions provided in the README.md.
- [X] I am on the latest version of both Open WebUI and Ollama.
- N/A: I have included the browser console logs. 
- [X] I have included the Docker container logs.
- [X] I have provided the exact steps to reproduce the bug in the "Steps to Reproduce" section below.

## Expected Behavior:

After uploading a PDF it should be parsed.

## Actual Behavior:

It fails with "Failed to add file." and logs report "cannot reshape array of size #X"

## Description

**Bug Summary:**
Seems to be related to image processing for "PDF Extract Images (OCR)".  Disabling this allows the document to be uploaded.

## Reproduction Details

**Steps to Reproduce:**
Add this file to a knowledge base:
[82c836.pdf](https://github.com/user-attachments/files/18531296/82c836.pdf)

## Logs and Screenshots

**Browser Console Logs:**
N/A

**Docker Container Logs:**
```
WARNI [python_multipart.multipart] Skipping data after last boundary
INFO  [open_webui.routers.files] file.content_type: application/pdf
ERROR [open_webui.routers.retrieval] cannot reshape array of size 11609 into shape (2954,3753,newaxis)
Traceback (most recent call last):
  File "/app/backend/open_webui/routers/retrieval.py", line 884, in process_file
    docs = loader.load(
           ^^^^^^^^^^^^
  File "/app/backend/open_webui/retrieval/loaders/main.py", line 127, in load
    docs = loader.load()
           ^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/langchain_core/document_loaders/base.py", line 31, in load
    return list(self.lazy_load())
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/langchain_community/document_loaders/pdf.py", line 257, in lazy_load
    yield from self.parser.parse(blob)
               ^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #29251, opened 2026-08-30
fix: PDF upload failures with PDF Extract Images (OCR) enabled

Classic298:
Uploading a PDF with OCR image extraction enabled fails with "cannot reshape array" or similar errors whenever a page contains an image the bundled langchain parser cannot decode: filter chains like [/ASCII85Decode /DCTDecode], images without a /Filter entry, 1-bit scans, image masks, indexed and grayscale images. One such image aborts the whole upload. Reported since v0.5.7 (#4457, #6085, #8856, #10913, #11333, #25979).

That parser reimplements PDF filter handling by hand and assumes a single lossless filter. langchain-community is archived and will not take a patch, so a small parser subclass now delegates image decoding to pypdf, which handles all of these cases, and an image that still cannot be read is skipped with a warning instead of failing the file.

Verified against the stock loader on fixture PDFs: previously working files produce identical output. Further differences, all deliberate: images nested in Form XObjects are now found; inline images stay excluded like before, which also avoids pypdf retaining every decoded inline image per page (115 MB on a 106 KB test file); image lookups sidestep pypdf's quadratic resource-tree re-walk (93 s down to 0.4 s on a 136 KB crafted file); CMYK JPEGs reach OCR with correct colours where pypdf's unconditional inversion turned them solid black.

Fixes #29187

> [!NOTE]

### DIFF
--- a/backend/open_webui/retrieval/loaders/main.py
+++ b/backend/open_webui/retrieval/loaders/main.py
@@ -13,7 +13,6 @@ from langchain_community.document_loaders import (
     BSHTMLLoader,
     CSVLoader,
     Docx2txtLoader,
-    PyPDFLoader,
     TextLoader,
 )
 from langchain_core.documents import Document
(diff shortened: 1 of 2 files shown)


Classic298:
ready

Issue #29077, opened 2026-08-27
issue: a rejected ask_user call ends the turn instead of returning the error to the model


KastenMonster:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I have checked whether this issue is already fixed on the `dev` branch or in the latest source.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Docker

### Open WebUI Version

v0.11.1

### Operating System

Alma 10

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

When first using the new tool `ask_user` i've noticed that the chat breaks / stops when the llm inputs wrong results without giving it the chance to correct itself. So i did a small investigation which revealed another issue.

Model used: Qwen 3.8 27B (Chat logs say qwen_36, but this is the old id as where are using a workspace model facing the users)

As for the expected behavior: 
1. Upon receiving an error from `ask_user` the chat should continue without stopping to give the llm the chance to correct itself 
2. When the llm for some reason calls `ask_user` in parallel the chat should manage the request sequentially or ouput an error for all tool calls

### Actual Behavior

1. The chat stops after receiving an error without giving the llm a chance to correct itself
2. Upon receiving a parallel tool call the n'th call after the first one loops infinitely

### Steps to Reproduce

Example prompts that get the result:
[...]

PR #29252, opened 2026-08-30
fix: a rejected ask_user call ending the turn with no reply

Classic298:
The documented behaviour of the built-in ask_user tool is that a call breaking its rules comes back to the model as an error. Instead the reply stopped there: the error was recorded as the tool result, the model was never asked again, and the user was left with a dead chat and no answer.

The rejection is now handed back like any other failed tool result, so the model sees it and can correct itself within the normal tool-call iteration limit. Any ordinary tool the model emitted in the same turn still runs.

A call rejected for arriving alongside other ask_user calls also left those siblings without a result, which the UI shows as a tool call stuck on "Executing..." forever. Every invalid call now gets its own result. Two ask_user calls on their own also reported the wrong reason, saying the call must be made by itself rather than that only one is allowed per turn.

Fixes #29077

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/ask_user.py
+++ b/backend/open_webui/utils/ask_user.py
@@ -12,10 +12,10 @@ def get_ask_user_tool_calls(tool_calls: list[dict]) -> tuple[list[dict], str | N
     ]
     if not ask_user_calls:
         return [], None
-    if len(ask_user_calls) != 1:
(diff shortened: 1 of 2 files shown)


kirin-xiao:
Can we also lift the limitation on the ask_user tool?

IIUC, the reason why LLMs make these mistakes in the first place is that they are not accustomed to these limits, which reduces tool call success rate.
Also, other harnesses don't have them

Issue #17104, opened 2025-09-01
feat: Show inference statistics, such as tokens-per-second speed, for openai compatible inference engines


drp4rad0x:
### Check Existing Issues

- [x] I have searched the existing issues and discussions.

### Problem Description

I created an issue as a bug, but the maintainer said it is intended functionality, so I think it is not a bug but a feature request. Open WebUI doesn't show inference statistics, such as tokens-per-second speed, for inference engines other than Ollama

### Desired Solution you'd like

I want to see all prompt_token/s speed in webui

tjbck:
The model inference engines themselves need to forward the stats, we already support usage field in the response.

drp4rad0x:
There is data in response 

{
  "choices": [
    {
      "finish_reason": "stop",
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "Hello! \n"
      }
    }
  ],
  "created": 1756712567,
  "model": "gemma-3n-E2B:Q4_K_XL",
  "system_fingerprint": "b6341-b66df9d9",
  "object": "chat.completion",
  "usage": {
    "completion_tokens": 5,
    "prompt_tokens": 15,
    "total_tokens": 20
  },
  "id": "chatcmpl-MSypBI7mRdMC6Cy4e5soqjWrd9NdqcZM",
  "timings": {
    "prompt_n": 1,
    "prompt_ms": 15.079,
    "prompt_per_token_ms": 15.079,
    "prompt_per_second": 66.31739505272233,
    "predicted_n": 5,
    "predicted_ms": 48.984,
    "predicted_per_token_ms": 9.796800000000001,
    "predicted_per_second": 102.07414666013392
  }

PR #29254, opened 2026-08-30
feat: show live token count, speed and elapsed time under each response

OGMatrix:
Relates to #17104

## Summary

Open WebUI measures generation numbers but only surfaces them after a response has finished, and only as raw JSON in the tooltip behind the info button. While a response is streaming there is nothing at all.

This adds a small row under each assistant message with the token count, the elapsed time and the rate, updating as the response streams:

```
⌗ ~305    ◷ 15.5s    ⚡ 34.99 t/s
```

Why this matters, in the order it bit me:

**You cannot tell a slow model from a stuck one.** With a large local model the gap between hitting send and the first token can be minutes while the prompt is processed. Right now the UI shows an animated cursor and nothing else, so a long prompt ingest and a hung backend look identical. The row names that phase ("Processing prompt") and keeps a clock on it.

**Throughput is the number you tune local setups against.** Picking a quant, deciding how many layers to offload, choosing a context size, comparing two backends serving the same weights: all of that is judged on tokens/sec. Today you either run the model outside Open WebUI to see that number, or you finish a response and read it out of a JSON blob. llama.cpp's own web UI puts it on screen while generating, and that difference is why people keep a second UI open.

**Reasoning models make the gap worse.** A thinking model can spend a long stretch producing tokens you do not see. Without a counter that phase is indistinguishable from a stall.

**It is not Ollama-specific.** `response_token/s` has been available for Ollama for a long time, which is roughly what #17104 asked to generalise. That issue was closed as completed once the backend normalised usage across providers, but nothing renders it live, so for llama.cpp, vLLM and other OpenAI-compatible backends there is still nothing to look at. This is the UI half.

### How the numbers are produced

Token counts are estimated from the streamed text while the response runs and are shown with a leading `~`. Providers only send usage once the stream ends, so there is nothing exact to show before that. When usage arrives the estimate is replaced by the provider's own count and the `~` drops.

Text reaches a response through three separate socket paths (chat completions, the Responses API, and `chat:message` events), so all three feed the measurement. Reasoning is streamed as its own output items and never reaches `message.content`, so the measurement reads the output array where one is present and falls back to content otherwise. Measuring content alone missed every thinking token and left the rate covering only the visible answer, which produced numbers several times too high on reasoning models.

The rate is measured from the first token onward rather than from the request, so a slow prompt ingest does not drag it down. Time to first token is in the tooltip instead.

Assistant messages are persisted by the backend from a fixed field whitelist, which has no room for a client-measured timing, so the finished stats are saved from the client on completion and when a response is stopped early.

The row can be turned off under Settings > Interface. It is on by default.

## Testing

Manual, against llama-server (llama.cpp) over its OpenAI-compatible endpoint with a Qwen3 27B reasoning model, streaming on:

- Watched the row through a full response: prompt processing, reasoning, visible answer.
- Confirmed the count switches from estimated to exact when usage lands, and the `~` disappears.
- Confirmed the reasoning phase is counted. Before that fix the row sat at 0 for the whole thinking phase and then reported a rate several times too high, because the window only covered the visible answer.
- Stopped a response part way and confirmed the numbers freeze at the last token rather than running on.
- Reloaded the page and confirmed the stats survive, by reading the stored chat back out of the database rather than trusting the screen.
- Toggled the setting off and on.

Automated:

- `npx vitest run` — 35 passing, 27 of them new (18 for the measurement, 9 for the output text extraction and labels), covering the delta/estimate/exact transitions, the rate window, the cancelled case, and the reasoning-vs-visible-text extraction.
- `npx svelte-check --tsconfig ./tsconfig.json` — 7792 errors, byte for byte the same per file as `dev` at 22379ded1, so nothing new.
- `npx vite build` — clean.

Rebased onto current `dev`. This conflicted with #29250, which rewrote the same `buildOutputDisplayItems` lines; resolved by keeping the non-optional `item.type` from that fix alongside the new argument.

### Added

- Live token count, elapsed time and tokens per second under each assistant response, updating while the response streams, with a toggle in Settings > Interface.

### Fixed

- Reasoning and tool call rows could keep showing their in-progress label ("Thinking...", "Analyzing...", "Executing...") after a response had finished, because the label was decided from the item's position in the output array without considering whether the message was done.

## Additional Context

Three things a reviewer should know up front.

No maintainer asked for this, and #17104 is closed rather than open, so by the letter of the template this is an unsolicited feature PR. I am opening it because the issue describes exactly this gap and was closed on the backend half. If you would rather have it as a Discussion first, say so and I will move it.

It adds one user-facing setting. I would rather it did not, but a persistent row under every message is the kind of thing people want to switch off, so it defaults on with an escape hatch.

It carries a second, smaller fix alongside the feature (the stuck in-progress label above), which is not strictly one logical unit. They are in the same commit because the label bug is only visible while a response streams, which is the same surface this PR touches. Happy to split it into its own PR.

One tradeoff worth naming: persisting the stats costs one chat save per completed response, on top of the existing save per message sent. Given the recent work on avoiding whole-chat round trips, that may not be a trade you want. The alternative is a per-message endpoint, which does not currently exist for arbitrary fields.

Screenshots follow in a comment.

## Checklist

- [x] This PR links to a well-described, confirmed Issue or active Discussion: `Relates to #17104`.
- [ ] The change is one logical unit with no unrelated commits.

- [ ] By submitting this pull request, I confirm that I have read and fully agree to the [Contributor License Agreement (CLA)](https://github.com/open-webui/open-webui/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT), and I am providing my contributions under its terms.


Classic298:
shgould be done with filter

PR #29256, opened 2026-08-30
fix: read the recurrence count as a value rather than a substring

Classic298:
The interval helper treated any rule whose text contained COUNT=1 as a one-shot and returned no interval, so counts such as 10 or 100 were matched as well. The helper already reports no interval when a rule yields fewer than two future occurrences, which states the same condition correctly, so the text match is gone.

Verified against two instances built from the same tree apart from these lines: every rule the schedule and calendar editors emit is unchanged, genuine one-shots still report no interval across seven frequencies and four DTSTART spellings, and the only rules whose handling changes are the ones whose count merely began with a one.

> [!NOTE]


Classic298:
Combined into #29262.

PR #29257, opened 2026-08-30
fix: split recurrence rules the way the parser does

Classic298:
The DTSTART filter cut the rule on newlines while the parser cuts it on any whitespace, so a rule carrying its start date on the same line as the recurrence text kept that date instead of the anchor the caller was given. Splitting on whitespace makes the filter and the parser agree on where one part ends and the next begins.

The equivalent filter on the calendar expansion path is left alone here and changes separately.

Verified against two instances built from the same tree apart from this line: every rule the schedule editor emits is unchanged, and the shapes that do change are the ones the old split could not see.

> [!NOTE]


Classic298:
Combined into #29262.

PR #29258, opened 2026-08-30
fix: anchor recurring events to their own start when the rule carries a date

Classic298:
Calendar expansion removed the DTSTART part by cutting the rule on newlines, while the parser cuts on any whitespace, so a rule carrying its start date on the same line as the recurrence text kept that date and anchored the series there instead of at the event's own start. Occurrences then showed up before the event began and at the wrong time of day.

The equivalent filter on the automation parsing path changes separately.

Verified against two instances built from the same tree apart from this line: every rule the calendar and schedule editors emit expands identically, and the corrected series matches the multi-line form of the same rule.

> [!NOTE]


Classic298:
Combined into #29262.

PR #29264, opened 2026-08-30
perf: stabilize the model registry signature across workers

Classic298:
The Redis-backed model registry skips its write when the content signature matches what is already stored. That skip has never worked across processes. Two of the values it hashes come out of Python sets, and set iteration order varies with each process's hash seed, so every worker computed a different signature for identical content and every worker rewrote the whole registry on every refresh.

Sorting both makes the signature depend on content alone. Measured on a 120 model registry, 522 KiB serialized: a refresh whose content already matches drops from GET, HKEYS, HSET and SET at 5.1 ms to a single GET at 2.2 ms per worker, and the 522 KiB write leaves the wire entirely.

Verified across 12 child processes with 12 distinct hash seeds: 12 different signatures before, 1 after. Filter execution order is unaffected, because the filter pipeline re-sorts by priority and id before running.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/chat_variables.py
+++ b/backend/open_webui/utils/chat_variables.py
@@ -112,7 +112,7 @@ def _safe_field(key: str, definition: dict[str, Any]) -> dict[str, Any]:
         'type',
     }
     field = {'key': key}
-    for field_key in allowed_keys:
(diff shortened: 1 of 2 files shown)


Classic298:
@tjbck free

PR #29274, opened 2026-08-30
fix: keep the Ollama keep-alive expiry out of the registry

Classic298:
When an admin saves a connection, the model list in Redis gets wiped so every server rebuilds it. That wipe is two steps: delete the list, then delete a small fingerprint that records what the list contained.

Servers use that fingerprint to decide whether they need to write the list back. If the fingerprint matches what they were about to write, they skip the write.

So if a server dies between the two deletes, the fingerprint survives without the list. Every server then compares against it, concludes the list is already correct, and skips writing. The list stays empty. Every chat fails with "Model not found", and nothing recovers on its own.

The fix is to delete the fingerprint first. Then the worst case is one redundant rewrite instead of a permanently empty model list.

> [!NOTE]

### DIFF
--- a/backend/open_webui/utils/models.py
+++ b/backend/open_webui/utils/models.py
@@ -64,6 +64,19 @@ async def get_all_base_models(request: Request, user: UserModel = None):
     return function_models + openai_models + ollama_models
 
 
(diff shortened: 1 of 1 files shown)


tjbck:
649c012ecf308a994ea180127f7f8f94d0aec311

PR #29313, opened 2026-08-31
ci: run the external regression suite on release pull requests

Classic298:
Adds a workflow that runs the open-webui/tests unit suite against release candidates, so a release that reintroduces a fixed bug is caught before it is cut rather than after users report it. The suite is roughly 4500 source-level tests pinned to specific past issues and PRs, and takes about three minutes; the dependency install dominates the run and is cached.

It runs only on pull requests into main whose title starts with a version, which is how releases are titled here, or which touch package.json. Everything else into main, and every pull request into dev, skips it and reports green.

Two settings are needed for this to block anything, both outside the diff: require the Regression / Result check on main, and require branches to be up to date before merging so the suite covers what actually lands.

The reusable workflow is referenced at @main so a release always runs the current tests. Pinning it to a tag instead is a reasonable call to make here.

### DIFF
--- a/.github/workflows/regression.yaml
+++ b/.github/workflows/regression.yaml
@@ -16,24 +16,15 @@ concurrency:
 
 jobs:
   # Into main only the release branch counts; into dev any version title does, so the
-  # suite can be exercised outside a release. Expressions have no regex, hence the digits.
+  # suite can be exercised outside a release. Expressions have no regex, hence the literal prefixes.
   regression:
     name: Suite
     if: >-
       (github.event.pull_request.base.ref == 'dev' ||
       github.event.pull_request.head.ref == 'dev') &&
       (github.event.action != 'edited' || github.event.changes.title != null) &&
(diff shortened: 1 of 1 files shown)


.github/workflows/regression.yaml
```diff
@@ -0,0 +1,72 @@
+# ─────────────────────────────────────────────────────────────────────────────
+# Regression — run the open-webui/tests unit suite against a release candidate
+# The suite runs on PRs into main titled with a version, or touching package.json
+# ─────────────────────────────────────────────────────────────────────────────
+name: Regression
```
tjbck: just call it `Tests`?

Classic298:
@james-westbrook i think we lgtm here now. tested a few times and now integrations also run

Classic298:
@tjbck the tests repo now also runs the frontend vitest suite through this same workflow, so nothing to change here. Full run against dev is green (unit, integration, frontend): https://github.com/open-webui/tests/actions/runs/33929899691. Safe to merge.

PR #29318, opened 2026-08-31
i18n: blank the redundant en-US translation values

silentoplayz:
# Pull Request

## Checklist

- [x] This PR links to a well-described, confirmed Issue or active Discussion: `Relates to #29017`.

## Summary

`src/lib/i18n/locales/en-US/translation.json` holds 3,094 keys, and 3,087 of them carry an empty value. That is the intended state for this catalog: `i18next-parser.config.ts` sets `defaultValue: ''`, so a new key is generated blank in every locale, and `src/lib/i18n/index.ts` sets `returnEmptyString: false`, so an empty value is treated as missing and i18next returns the key. In `en-US` the key is already the English string, which is why 3,087 blanks render correctly.

Seven entries were filled in by hand instead, one at a time since June 2025. This blanks all seven so the catalog is uniform.

```diff
-	"Generated Image": "Generated Image",
+	"Generated Image": "",
-	"Next message": "Next message",
+	"Next message": "",
-	"Previous message": "Previous message",
+	"Previous message": "",
-	"Read the docs": "Read the docs",
+	"Read the docs": "",
-	"Run AI on your own terms. Connect any model, extend with code, and protect what matters without compromise. Your models, your data, your machine, wherever you open it.": "Run AI on your own terms. Connect any model, extend with code, and protect what matters without compromise. Your models, your data, your machine, wherever you open it.",
+	"Run AI on your own terms. Connect any model, extend with code, and protect what matters without compromise. Your models, your data, your machine, wherever you open it.": "",
-	"Welcome to your AI home.": "Welcome to your AI home.",
+	"Welcome to your AI home.": "",
-	"Write your model system prompt content here\ne.g.) You are Mario from Super Mario Bros, acting as an assistant.": "Write your model system prompt content here\n e.g.) You are Mario from Super Mario Bros, acting as an assistant.",
+	"Write your model system prompt content here\ne.g.) You are Mario from Super Mario Bros, acting as an assistant.": "",
```

One file, 7 insertions, 7 deletions. No key is added or removed.

**Six of the seven render identically before and after**, because their value was byte-identical to their key.

**The seventh changes what English users see, and it is the one line worth a decision.** The value for the model system prompt placeholder differs from its key by a single space after the newline: the key has `content here\ne.g.)` and the value has `content here\n e.g.)`. Both call sites pass the key without that space, at [ModelEditor.svelte#L790-L793](https://github.com/open-webui/open-webui/blob/6ea321370e5f817e95e0d2f9397c54be86925d9a/src/lib/components/workspace/Models/ModelEditor.svelte#L790-L793) and [FolderModal.svelte#L221-L224](https://github.com/open-webui/open-webui/blob/6ea321370e5f817e95e0d2f9397c54be86925d9a/src/lib/components/layout/Sidebar/Folders/FolderModal.svelte#L221-L224), so the extra space exists only for English and only through this catalog entry. Blanking it removes the space, which unindents the second line of that placeholder. If you would rather keep the space, drop that one line from the diff and the other six still stand on their own.

## Testing

Anchored to `dev` at `6ea321370e5f817e95e0d2f9397c54be86925d9a`.

Manual:

1. Ran this branch locally and confirmed the surfaces using these seven strings still render their English text, with nothing blank or missing.

Supporting evidence, scripted rather than manual:

1. A differential render through i18next itself, initialized with the same options `src/lib/i18n/index.ts` uses (`returnEmptyString: false`, `keySeparator: false`, `nsSeparator: false`, `fallbackLng` default `en-US`), resolving all seven keys against the catalog before and after the change. Result: six resolve to the identical string, one differs, and the difference is exactly the one space described above.
2. `npx prettier --check src/lib/i18n/locales/en-US/translation.json` passes, so the file keeps its formatting.
3. The catalog still parses and still holds 3,094 keys, now with zero non-empty values.

### Changed

- The `en-US` translation catalog no longer carries hand written values that duplicate their own keys.

## Additional Context

`npm run i18n:parse` cannot make this change. In `i18next-parser`, `mergeHashes` assigns `target[key] = source[key]`, so a value already on disk overwrites the freshly generated empty default, and the config sets neither `resetAndFlag` nor `resetDefaultValueLocale`. That is why these seven survived every parser run and why the edit is by hand.

Scope is deliberately `en-US` only. Across the other 63 locale files there are 4,172 entries whose value equals their key, and there they are legitimate translations that happen to match the English source, not redundancy. The argument in this PR holds only for the catalog whose keys are already the target language.

Broader `en-US` drift, the keys the parser would add and remove, is described in #29017 and is not touched here.

This PR was prepared with AI copilot assistance and I have thoroughly reviewed it.


tjbck:
@silentoplayz conflicts

silentoplayz:
> @silentoplayz conflicts

Merge conflict resolved!

PR #29625, opened 2026-09-04
i18n: fix incorrect Russian translations

shaba:
# Pull Request

## Checklist

- [ ] This PR links to a well-described, confirmed Issue or active Discussion: `Closes #___` / `Relates to #___`.
- [ ] I updated relevant docs, including the Open WebUI Docs Repository, if needed.

No Issue is linked because the template exempts changes that only update i18n/localization. Docs and screenshots do not apply: no key, no English string and no component changed, only Russian values.

## Summary

Corrects sixteen Russian values in `src/lib/i18n/locales/ru-RU/translation.json`. Two of them are functional defects, the rest change the meaning of a label or contradict the strings next to them.

**Interpolation was broken in two entries.** Their placeholder no longer matched the key, so nothing was substituted and the raw text reached the user:

- `{{ models }}` had the placeholder name itself translated to `{{ модели }}`. The tooltip in `MessageInput.svelte` passes `models`, so Russian users saw the literal text `{{ модели }}` instead of the model list.
- `Write a summary in 50 words that summarizes {{topic}}.` replaced `{{topic}}` with a bracketed note, so the topic was never filled in.

**Nine entries said something the English does not:**

- `Enter Key Behavior` — "Enter" was read as the verb, so the label for how the Enter key behaves became "type in the behaviour key". Its own description, `Choose whether Enter sends immediately or uses Ctrl+Enter.`, is translated correctly, which made the label visibly contradict the setting it names.
- `Regenerate Menu` — read as "refresh the menu" rather than the menu of regenerate actions its description explains.
- `Enter reasoning effort` — the previous text was not meaningful Russian; it now matches `Reasoning Effort`, the label of the field it belongs to.
- `Enter Bing Search V7 Subscription Key` — mentioned an endpoint that the English does not.
- `Web Search in Chat` and `Set web search availability for new chats.` — called the feature a search engine rather than web search.
- `Enter Application DN` — said `ND` instead of `DN`.
- `Enter Bocha Search API Key` — said `APi`.
- `Toast Notifications for New Updates` — dropped "new" and used the wrong form of the Russian preposition.

**Five entries were inconsistent with their own neighbours:**

- `Enter Top K Reranker` used a word for "reranker" that appears nowhere else, while the label of the very same field, `Top K Reranker`, uses the established one. `Select a reranking model engine` likewise differed from every other reranking string.
- `iframe Sandbox Allow Forms` and `iframe Sandbox Allow Same Origin` did not follow the wording of the other two switches in the same group, `iframe Sandbox Allow Downloads` and `iframe Sandbox Allow Scripts`.
- `Follow-Up Auto-Generation` capitalised a word mid-phrase.

The first group is user-visible breakage; the rest mislead about what a control does. `Enter Key Behavior` matters beyond cosmetics: people reading the interface with a screen reader hear only the label, without the surrounding layout that would hint at the real meaning.

## Testing

This is a data-only change, so the checks are on the file and on each string's use in the source:

- Every changed string was read against its call site (`InterfaceSettings.svelte`, `MessageInput.svelte`, `AdvancedParams.svelte`, the web search and reranking settings) to confirm the corrected wording names what the control actually does.
- Placeholder audit over all 3206 entries: values whose `{{...}}` set differs from the key went from 2 to 0.
- The file parses as JSON; the entry count is unchanged at 3206, no key was added, removed or reordered, and no value was left empty.
- `git diff` is exactly 16 changed lines, so the file still matches the formatting `npm run i18n:parse` produces.

### Fixed

- Corrected Russian translations where the text said something other than the English, including two entries whose interpolation placeholder had been translated and therefore never filled in.

## Additional Context

Found while setting up Open WebUI for a blind colleague who works entirely through a screen reader. `Enter Key Behavior` was the string that prompted this: read aloud on its own, "type in the behaviour key" gives no clue that the setting controls the Enter key.

Only Russian values changed. No key, no English string and no component was touched, so no other locale is affected.


tjbck:
@EllaFr

tjbck:
Thanks!

Issue #29580, opened 2026-09-03
issue: version check reports the running version as latest whenever the GitHub request fails, so About shows (latest) on an outdated instance


silentoplayz:
### Before Submitting

- [x] I searched open and closed issues and discussions for an existing report.
- [x] I checked whether this is already fixed on the `dev` branch or latest source.
- [x] I understand that maintainers want a well-written issue before any code pull request.
- [x] I am using the latest available version of Open WebUI for my install method.
- [x] This is not a security vulnerability.

### Installation Method

Docker

### Open WebUI Version

`ghcr.io/open-webui/open-webui:dev` (reports 0.11.3) reproduces the API response. The screenshot below is from `ghcr.io/open-webui/open-webui:v0.10.2`, chosen only because an older version makes the wrong badge visible. The code path is byte-identical at v0.11.3 and at dev `0cf48a04c1dd7d28df562a05a2ee80cc612e5c03`, which the permalinks below point at.

### Operating System

Linux (Docker host)

### Browser

Chromium 148 (Playwright), for the screenshot only. The API response is independent of the browser.

### Ollama Version

Not relevant to this report.

### Summary

`GET /api/version/updates` returns `{"current": VERSION, "latest": VERSION}` whenever the request to the GitHub releases API fails, for any reason. `Settings` > `About` and `Admin Settings` > `General` render that as `(latest)`, and the update toast never fires. So an Open WebUI instance that cannot reach GitHub reports itself as up to date no matter how far behind it is. The exception is logged at `debug`, below the default `INFO` level, so nothing records that the check did not happen.

The same fallback is duplicated on the client, where the `catch` in `About.svelte` substitutes the running version as `latest` when the fetch itself fails.

`OFFLINE_MODE=true` is handled correctly and is not the problem. It disables the check, the endpoint short-circuits, and the About page hides both the badge and the `Check for updates` button. The misleading state is an enabled check that cannot reach GitHub, which is exactly what a network outage or a restricted network produces.

### Expected Behavior

When the release lookup fails, the response says so instead of asserting a version, for example with `latest: null`. The About page then shows something like `Could not check for updates` rather than `(latest)`, and the failure is visible at the default log level.

### Actual Behavior

- API: `{"current":"0.10.2","latest":"0.10.2"}`, HTTP 200, returned in 4.8 ms, while the latest release at that moment was v0.11.3.
- UI: `Settings` > `About` shows `v0.10.2 (latest)`.
[...]

PR #29626, opened 2026-09-04
fix: report version check failures instead of claiming the running version is latest

silentoplayz:
# Pull Request

## Checklist

- [x] This PR links to a well-described, confirmed Issue or active Discussion: `Closes #29580`.

## Summary

`GET /api/version/updates` returns the running version as `latest` whenever the GitHub request fails, so an instance that cannot reach GitHub reports itself up to date no matter how far behind it is. The exception is logged at `debug`, below the default level, so nothing records that the check never happened. Reported in #29580.

The failure path now says it failed:

```diff
     except Exception as e:
-        log.debug(e)
-        return {'current': VERSION, 'latest': VERSION}
+        log.warning(f'Version update check failed: {e}')
+        return {'current': VERSION, 'latest': None}
```

`OFFLINE_MODE` is untouched. That path already short-circuits the endpoint and hides the badge, and it is the model this follows.

**A null `latest` cannot be handed to `compareVersion` as it stands.** `current.localeCompare(null)` coerces to the string `"null"`, and `"0.10.2"` sorts before it, so the old function returns `true`. The backend change on its own would have replaced a false `(latest)` with a false "update available" plus an update toast. Measured, before and after:

| `compareVersion(latest, current)` | before | after |
|---|---|---|
| `(null, '0.10.2')` | `true` | `false` |
| `('0.11.3', '0.10.2')` | `true` | `true` |
| `('0.10.2', '0.10.2')` | `false` | `false` |

So the guard is part of the fix, not a tidy-up:

```diff
-	return current === '0.0.0'
+	return !latest || current === '0.0.0'
```

The three callers of `getVersionUpdates` stop substituting the running version in their `catch`, and the two badge surfaces gain a third state alongside `Checking for updates...` and `(latest)`. When `latest` is unknown the badge is plain text rather than a link, since there is no release to link to.

## A second defect on the admin surface

`Admin Settings` > `General` renders the same badge, and it was wrong in a worse way than the issue describes. It initialises `updateAvailable = false` with `version.latest = WEBUI_VERSION`, and its `onMount` never calls `checkForVersionUpdates`. It therefore claims `(latest)` immediately, from a hardcoded value, **having made no request at all**, on every install regardless of network.

`About.svelte` already does this correctly. Its `updateAvailable` starts `null`, its `latest` starts empty, and it checks on mount. The admin page now matches it. Without that change, the null branch in that file is unreachable, since `latest` is never null there.

This does mean the admin page now performs the version check on mount, the same as About. Flagging it as the one behaviour change here beyond the reported bug.

## A layout defect the longer string exposed

`See what's new` sits directly under the badge in a `flex flex-col`, so the button stretches to the column width while its default `text-align: center` centres the label. The label therefore drifts right as the badge string grows. Glyph offset from the column's left edge, measured in the rendered DOM:

| Badge | before | after |
|---|---|---|
| `(latest)` | 0px | 0px |
| `(v0.11.4 available!)` | 29px | 0px |
| `Could not check for updates` | 56px | 0px |

The middle row is current `dev`, so the label already shifts on any instance with an update available. The new string only widens the gap. `self-start` sizes the button to its content, matching the admin surface, whose button is already content sized because it sits in a block container rather than a flex column.

## Testing

The checks below were run at my direction with AI copilot assistance, driving a browser and `curl` against a throwaway instance. I have reviewed the results.

The instance used its own sqlite database (`DATABASE_URL` set explicitly, confirmed as `Context impl SQLiteImpl` in the startup log before any request) and a copied frontend build, on a port confirmed free.

Failure was induced with the proxy variables from the issue, `HTTP_PROXY`/`HTTPS_PROXY` pointed at `http://127.0.0.1:9`, so the GitHub request fails without touching the network. The endpoint builds its session with `trust_env=True`.

Same build, backend file reverted and restored to compare:

| | `dev` | this branch |
|---|---|---|
| `GET /api/version/updates`, unreachable | `{"current":"0.11.3","latest":"0.11.3"}` | `{"current":"0.11.3","latest":null}` |
| Server log at `GLOBAL_LOG_LEVEL=INFO` | nothing | `WARNING ... Version update check failed: Cannot connect to host 127.0.0.1:9` |

Both surfaces, both states:

| Surface | GitHub | API | Rendered |
|---|---|---|---|
| `Settings` > `About` | unreachable | `latest: null` | `Could not check for updates`, no link |
| `Settings` > `About` | reachable | `latest: "0.11.3"` | `(latest)`, link to the real release tag |
| `Admin` > `General` | unreachable | `latest: null` | `Could not check for updates`, no link |
| `Admin` > `General` | reachable | `latest: "0.11.3"` | `(latest)`, link to the real release tag |

No update toast appeared in the failure state, which is what the `compareVersion` guard prevents. The admin badge was identified by its class and its position inside the admin form, since both surfaces now render the same string and matching on text alone would not distinguish them.

`npm run build` passes. `npx vitest run` passes at 2 files and 8 tests. `ruff format --check` reports `main.py` already formatted.

### Fixed

- A failed version check no longer reports the running version as the latest release. Both `Settings` > `About` and `Admin Settings` > `General` now show that the check could not complete. The failure is logged at `warning` rather than `debug`, and no update notification is raised from an unknown version.
- `Admin Settings` > `General` now performs the version check when it opens. It previously showed `(latest)` without contacting GitHub at all.
- `See what's new` in `Settings` > `About` no longer shifts right as the version badge text grows.

## Screenshots

| Surface | Before | After |
|---|---|---|
| `Settings` > `About`, GitHub unreachable | image | image |
| `Admin Settings` > `General`, GitHub unreachable | image | image |

## Additional Context

The wording of the new badge string is a product call. I used `Could not check for updates`, matching the Expected Behavior in #29580, and it is a single `en-US` key if something else reads better.

Only `en-US/translation.json` is touched. The other locales fall back to the English literal, and multi-locale sync is a release-time job.

This PR was prepared with AI copilot assistance. I have thoroughly reviewed it.

### DIFF
--- a/backend/open_webui/main.py
+++ b/backend/open_webui/main.py
@@ -2575,8 +2575,8 @@ async def get_app_latest_release_version(user=Depends(get_verified_user)):
 
                 return {'current': VERSION, 'latest': latest_version[1:]}
     except Exception as e:
-        log.debug(e)
-        return {'current': VERSION, 'latest': VERSION}
+        log.warning(f'Version update check failed: {e}')
+        return {'current': VERSION, 'latest': None}
 
 
 @app.get('/api/changelog')
--- a/src/lib/components/admin/Settings/General.svelte
+++ b/src/lib/components/admin/Settings/General.svelte
@@ -27,10 +27,10 @@
 
 	export let saveHandler: Function;
 
-	let updateAvailable: boolean | null = false;
+	let updateAvailable: boolean | null = null;
 	let version = {
 		current: WEBUI_VERSION,
-		latest: WEBUI_VERSION
+		latest: ''
 	};
 
 	let adminConfig: any = null;
@@ -47,7 +47,7 @@
 		version = await getVersionUpdates(localStorage.token).catch((error) => {
 			return {
 				current: WEBUI_VERSION,
-				latest: WEBUI_VERSION
+				latest: null
 			};
 		});
 
@@ -91,6 +91,10 @@
 		defaultInterfaceSettings = getDefaultInterfaceSettings();
 
 		banners = [...$_banners];
+
+		if ($config?.features?.enable_version_update_check) {
+			checkForVersionUpdates();
+		}
 	});
 </script>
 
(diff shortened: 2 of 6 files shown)


tjbck:
Great addition!

Issue #29636, opened 2026-09-04
Fix: incomplete file cleanup when deleting knowledge directories


limex:
## Bug

Deleting a knowledge directory with **Delete contained files** removes the
directory and its `knowledge_file` rows, but does not clean up the contained
files' vectors or file resources.

The files disappear from the Knowledge UI even though their chunks remain in
the knowledge collection. The deleted content can therefore still be retrieved,
and uploading the same content again fails with:

```text
Duplicate content detected. Please provide unique content to proceed.
```

This is related to #7181, but that issue's fix covers individual-file removal,
not directory deletion.

## Steps to reproduce

1. Create a knowledge base and a directory inside it.
2. Upload `a.md` into the directory and wait for processing to complete.
3. Delete the directory with **Delete contained files**
   (`move_files=false`).
4. Upload the same `a.md` again.

The second upload is rejected as duplicate content. Renaming the file does not
help because deduplication uses the content hash.

Reproduced on v0.11.3. The affected cleanup path is also present on `main` as
of 2026-09-04.

## Cause

`delete_knowledge_directory()` delegates to `Knowledges.delete_directory()`.
That method recursively deletes `knowledge_file` associations, but unlike the
individual-file removal endpoint, it does not delete:

- matching chunks from the knowledge vector collection;
- the per-file vector collection;
- an unreferenced global file record; or
- the corresponding storage object.

## Proposed fix

Before deleting relational links, collect all files in the selected directory
subtree and remove their knowledge vectors by `file_id` and hash. After the
directory deletion, remove file resources only when retention is disabled, the
file has no remaining knowledge references, and the actor is permitted to
delete it.

Vector cleanup failures are returned instead of silently reporting a successful
directory deletion.

<details>
<summary>Patch for v0.11.3/main</summary>

```diff
diff --git a/backend/open_webui/routers/knowledge.py b/backend/open_webui/routers/knowledge.py
index f285e5243..6c514190b 100644
--- a/backend/open_webui/routers/knowledge.py
+++ b/backend/open_webui/routers/knowledge.py
@@ -2313,29 +2313,83 @@ async def delete_knowledge_directory(
 ):
     await _verify_knowledge_write_access(id, user, db)
 
     # Verify directory belongs to this knowledge base
     directory = await Knowledges.get_directory_by_id(dir_id, db=db)
     if not directory or directory.knowledge_id != id:
         raise HTTPException(
             status_code=status.HTTP_404_NOT_FOUND,
[...]

PR #29656, opened 2026-09-04
fix: clean up vector data and files when deleting a knowledge directory

Classic298:
Deleting a directory in a knowledge base with "Delete all contents inside this directory" took the contained files out of the knowledge base but left their embeddings in the knowledge collection. The deleted content was still retrieved into answers, and uploading the same content again was rejected as duplicate content, permanently.

Every other removal path already purges those chunks. The directory path now does the same: the files in the deleted subtree are collected before their links are dropped, and their chunks are removed by file id and by hash regardless of ENABLE_KNOWLEDGE_FILE_RETENTION. The file record, the stored copy and the per-file collection are removed only when retention is disabled, the file is no longer referenced by any knowledge base and the actor owns it or is an admin, which is what removing a file individually does.

The lookup is scoped to the knowledge base from the request, so a directory nested under a different knowledge base's directory no longer has its file's chunks deleted from the wrong collection.

Fixes #29636

### DIFF
--- a/backend/open_webui/models/knowledge.py
+++ b/backend/open_webui/models/knowledge.py
@@ -1069,6 +1069,36 @@ class KnowledgeTable:
         for child_id in child_ids:
             await self._delete_files_in_subtree(child_id, db=db)
 
+    async def get_files_in_subtree(
+        self,
+        knowledge_id: str,
+        directory_id: str,
+        db: Optional[AsyncSession] = None,
+    ) -> list[FileModel]:
+        """Get all files in a directory and its subdirectories."""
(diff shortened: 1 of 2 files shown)


AlpayY:
Any update on this? This bug is causing me immeasurable pain.

Classic298:
In this case I suggest consulting a doctor or pharmacist

Classic298:
Superseeded by https://github.com/open-webui/open-webui/commit/17dbc6f001aeea25ae1df528cb79bb272eca4a77

Issue #29208, opened 2026-08-29
feat: parse base64 images returned by tool calls beyond a top-level data URI


cshdotcom:
### Check Existing Issues

- [x] I have searched all existing **open AND closed** issues and discussions and found none comparable to my request.
- [x] I have checked whether this request is already implemented on the `dev` branch or in the latest source.

### Verify Feature Scope

- [x] I believe this feature request is appropriately scoped for the Issues section as described above.

### Problem Description

**Describe the bug**
When an AI performs tool calls and the tool returns base64‑encoded image data, OpenWebUI fails to parse the base64 image content correctly.

Instead of rendering the image, the raw base64 string gets inserted as plain text into the subsequent AI conversation context. This leads to extremely high token consumption and produces meaningless model outputs.

**To Reproduce**
1. Enable a tool that returns base64‑encoded image content in its response payload.
2. Trigger the tool call within chat.
3. Observe the chat history: raw long base64 text is appended to context instead of being rendered as an image.
4. The following AI reply becomes incoherent due to massive base64 tokens.

**Expected behavior**
OpenWebUI should detect base64 image data returned from tool calls, render it as visual image content in the chat UI, and avoid injecting the full raw base64 string into the LLM conversation context.

**Screenshots**
N/A

**Environment (please complete the following information):**
 - OpenWebUI Version: [e.g. 0.5.x]
 - Deployment: [Docker / Manual install]
 - Model Backend: [e.g. OpenAI‑compatible API]

**Additional context**
Large base64 payloads drastically blow up context window usage, degrade chat performance and result in nonsensical answers.

### Proposed Solution

I suggest adding dedicated base64 image detection and parsing logic for tool call return payloads.
 
1. Automatically identify base64 image strings (image/png, image/jpeg, etc.) returned by tool calls.
2. Render valid base64 images directly in the chat interface as visual media instead of treating them as plain text content.
3. Strip or encode the raw base64 string from the conversation context before passing it to the LLM, preventing massive token inflation.
4. Preserve tool call image assets in the frontend display while keeping the backend context clean and lightweight for subsequent model generations.
 
This fix will resolve excessive token consumption caused by raw base64 context injection and restore meaningful AI response quality after image-based tool calls.

tastyrice90:
Yeah im encountering this problem too. Any 2-3mb image viewed inside a tool call by a model with vision blows up the context and creates an auto OWUI compaction event on the next turn.

RRXXZZYY:
[...]

PR #29665, opened 2026-09-04
fix: stop base64 images in tool results from reaching the model as text

Classic298:
When a tool returns a base64 image, Open WebUI only moves it out of the model's context if the entire result is that image, or if the tool is MCP. An image sitting inside a returned object or a list was serialised into the tool message instead, so a single screenshot could cost hundreds of thousands of tokens, push out the rest of the conversation and leave the model answering from garbage.

Any string in a tool result that is exactly one image data URI is now moved into the result's files wherever it sits in the structure, and replaced with a short marker. The model gets the image as an attachment rather than as text, and it renders for the user instead of being dropped.

Detection is deliberately limited to values that are entirely a data URI. Scanning inside longer strings was tried and abandoned: a payload wrapped over several lines, or one followed by text, gets cut short, and shipping a truncated image is worse than the bloat because the provider rejects the whole request.

This also fixes the OpenAPI branch removing entries from the list it was iterating over, which skipped every second data URI and left it in the model's context.

Fixes #29208

### DIFF
--- a/backend/open_webui/utils/middleware.py
+++ b/backend/open_webui/utils/middleware.py
@@ -972,6 +972,25 @@ async def apply_source_context_to_messages(
         )
 
 
+BASE64_IMAGE_DATA_URI_RE = re.compile(r'data:image/[a-zA-Z0-9.+-]+;base64,[A-Za-z0-9+/]+={0,2}', re.IGNORECASE)
+
+
+def extract_base64_images(value: Any, files: list) -> Any:
+    """Move base64 image data URIs out of a tool result so they do not reach the model as text."""
+    if isinstance(value, str):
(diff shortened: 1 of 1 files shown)


plevart:
Hi, I don't know whether this is the same issue, but I also have a problem with image returned by a MCP tool. I'm using the unofficial Home Assistant MCP server (`https://github.com/homeassistant-ai/ha-mcp`) and the tool used is `ha_get_camera_image` . When it is called, the returned image is actually displayed in the Open WebUI chat window (inside the collapsible thinking section which includes tool calls and results), but the model always gets `{"results": []}`, so it doesn't get the image. The model doesn't see the base64 encoded image as part of text input either.
So WDYT?
Uploading image as file attachment works though. The model sees it.

PR #29677, opened 2026-09-04
remove: Stale vitest tests

Classic298:


Classic298:
moving these to the tests repo

Classic298:
@tjbck this removes the colocated vitest files. They live in open-webui/tests under frontend/ now and run through the regression workflow from #29313. Also drops src/lib/utils/_template_old.ts, a vitest file whose name never matched the test glob, so those tests never ran.

Issue #29668, opened 2026-09-04
feat: show the first follow-up as ghost text in the empty chat input, accepted with Tab


silentoplayz:
### Before Submitting

- [x] I searched open and closed issues and discussions for an existing request.
- [x] I checked whether this already exists on the `dev` branch or latest source.
- [x] I understand that maintainers want a well-written issue or discussion before any code pull request.
- [x] This request is not a security vulnerability.

### Problem

After a reply finishes, the next thing I want to ask is usually sitting right there under it, in the `Follow up` list. To use it, I have to stop and read the list. Then I reach for the mouse to click one and come back to the input to press Enter. The input itself, where my attention already is, stays empty.

At the same time, `Prompt Autocompletion` already puts grey ghost text in that input and lets me take it with `Tab`. It only ever helps once I have started typing. On an empty field, it does nothing, which is exactly the moment I most often have nothing yet.

So the two features that both exist to save me from writing the next message sit a few pixels apart and never meet.

### Desired Behavior

When a reply finishes and the input is empty, the first suggested follow-up appears in the input as ghost text, the same grey as autocompletion shows today. Pressing `Tab` accepts it as my message, ready to send or edit. Typing anything replaces it with what I type. Clearing the field again brings it back. Sending my own message, or the reply changing, discards it.

It should only show when the field is truly empty and the last message is a finished reply. It should never send on its own. The `Follow up` list below the reply stays as it is, so the other suggestions remain one click away.

Claude's desktop app does this today, and it is the single interaction I miss most when I come back to Open WebUI.

### Why This Matters

Autocompletion and follow-ups are two of the best things in the chat view, and they are both about the same problem, which is what to say next. Right now the follow-ups need a detour through the mouse, and the autocompletion needs me to have started already. Putting a follow-up into the field as ghost text gives the common case, "yes, that one", a single keypress with no change of focus. The cost to anyone who does not want it is nothing, because they start typing and it is gone.

This also needs no new model call and no new setting. The suggestions are already generated for every reply, and the ghost text and `Tab` acceptance already exist in the input. Both features already have their own toggles, `Prompt Autocompletion` and `Follow-Up Auto-Generation` in Settings, and the ghost text would simply require both to be on.

### Examples or References

[...]

PR #29680, opened 2026-09-05
feat: show the first follow-up as ghost text in the empty chat input

silentoplayz:
# Pull Request

## Checklist

- [x] This PR links to a well-described, confirmed Issue or active Discussion: `Closes #29668`.

## Summary

After a reply finishes, its first follow-up now appears as grey ghost text in the empty chat input. `Tab` accepts it into the composer, typing replaces it, and clearing the field brings it back. It never sends on its own, and the `Follow up` list under the reply is unchanged. Requested in #29668.

No new model call, no new endpoint, no new setting. The follow-ups are the ones already generated for every reply, and the ghost text and `Tab` acceptance are the ones the autocompletion extension already provides. The feature is on only when two existing toggles are on, `Prompt Autocompletion` and `Insert Follow-Up Prompt to Input`. A `Tab` here is exactly an insert of a follow-up into the composer, with one key instead of a click, so those are the toggles that describe it.

## What changed

**`MessageInput.svelte`.** A reactive block watches the last message. When it is a finished assistant reply with follow-ups and the composer is empty, and both toggles are on, it hands the first follow-up to the input as a suggestion. When the composer has text, or there is nothing to suggest, it clears the suggestion instead. `history` is already reassigned when follow-ups arrive, so no new event is needed.

**`RichTextInput.svelte`.** The input gains one exported function, `setSuggestion`. It follows the pattern the input already uses to expose `setText` and `focus` to its parent. It acts only on a document that is a single empty paragraph. It sets the same three attributes the autocompletion extension already uses for a suggestion. It is a no-op when that suggestion is already set, so the reactive block cannot loop through the editor's `onUpdate`.

**`AutoCompletion.js`.** The `mouseup` handler clears every suggestion in the document. That is right for a typed suggestion whose caret may move, but it would have erased the ghost the moment the user clicked into the composer to focus it. It now skips a suggestion whose `data-prompt` is empty, which only the empty-field case produces. Typed suggestions clear on click as before. `Tab` acceptance, swipe-right acceptance and clearing on the next keypress are the existing extension code, untouched.

**`app.css`.** On an empty paragraph, the ghost is drawn through the placeholder's `::before` slot. That puts it on the caret's line and gives it the placeholder's float, absolute positioning and single-line clamp. The `::after` used for typed suggestions is suppressed in that state, so nothing draws twice. Without this rule, the ghost rendered a line below the caret. It came after the paragraph's trailing line break, and the placeholder showed underneath it.

## Testing

I tested this locally on the branch, with both toggles on:

- The first follow-up appears in the empty composer after a reply.
- `Tab` accepts it in place.
- Typing replaces it, and deleting back to empty brings it back.
- Clicking into the composer keeps it.
- Sending clears it until the next reply, and a new empty chat shows nothing.
- With either toggle off, nothing appears.
- A multi-model chat behaves the same with `Display Multi-model Responses in Tabs` on and off.

Mechanical checks, run at my direction with AI copilot assistance:

| Check | Result |
|---|---|
| `npm run build` | exit 0 |
| `npx vitest run` | 2 files, 8 tests, passing |
| `npx prettier --check` on the four changed files | already formatted |
| New `i18n.t` keys, new settings, new dependencies | none |
| Added code comments, Svelte 5 runes | none |

### Added

- When a reply finishes, its first suggested follow-up appears as ghost text in the empty chat input and can be accepted with `Tab`. Shown only when both `Prompt Autocompletion` and `Insert Follow-Up Prompt to Input` are enabled.

#### Ghost text in the empty composer after a reply

image

image

image

## Additional Context

Two behaviours are inherited rather than chosen. The empty-field ghost takes the placeholder's single-line clamp, so a follow-up wider than the composer truncates with an ellipsis rather than wrapping, the same as the placeholder does today. Pressing a bare modifier key with the composer focused dismisses the ghost, as it does for a typed suggestion. It returns on the next change to the composer.

This is a narrower cousin of #28087 and #28089, which generated suggestions from an attached document with a new endpoint and settings. This one adds no backend and no settings.

This PR was prepared with AI copilot assistance. I have thoroughly reviewed it.

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -607,6 +607,15 @@ html.high-contrast.dark .ProseMirror p.is-editor-empty:first-child::before {
 	}
 }
 
+.ProseMirror p.is-editor-empty.ai-autocompletion:first-child::before {
+	content: attr(data-suggestion);
+	color: #a0a0a0;
+}
+
+.ProseMirror p.is-editor-empty.ai-autocompletion:first-child::after {
+	content: none;
+}
+
 .ai-autocompletion::after {
 	color: #a0a0a0;
 
--- a/src/lib/components/chat/MessageInput.svelte
+++ b/src/lib/components/chat/MessageInput.svelte
@@ -408,6 +408,22 @@
 		chatInputElement?.focus(options);
 	};
 
+	$: {
+		const lastMessage = history?.currentId ? history.messages[history.currentId] : null;
+		const followUp =
+			lastMessage?.role === 'assistant' && lastMessage?.done
+				? (lastMessage?.followUps ?? [])[0]
+				: null;
+
+		if (
+			$config?.features?.enable_autocomplete_generation &&
+			($settings?.promptAutocomplete ?? false) &&
+			($settings?.insertFollowUpPrompt ?? false)
+		) {
+			chatInputElement?.setSuggestion(prompt === '' && followUp ? followUp : '');
+		}
+	}
+
 	export const setText = async (text?: string, cb?: (text: string) => void) => {
 		const chatInput = document.getElementById('chat-input');
 
--- a/src/lib/components/common/RichTextInput.svelte
+++ b/src/lib/components/common/RichTextInput.svelte
@@ -520,6 +520,23 @@
 		focus();
 	};
 
+	export const setSuggestion = (text: string) => {
+		if (!editor || !editor.view) return;
+		const { state, dispatch } = editor.view;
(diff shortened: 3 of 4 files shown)


silentoplayz:
Superseded by 7aaa4a692e949724913834efc47c57572042703b, which Tim landed on `dev` on 2026-09-10 with co-author credit. It carries this change across the same four files. The feature ships in the next release.

PR #29684, opened 2026-09-05
fix: do not persist Function deactivation on load failure

### LINKED ISSUE #29681: question: Function is switched to inactive after a load error
# A single Function load error permanently disables the Function (including global safety filters) with no re-enable path

Affected: open-webui/open-webui@2a960a5

## Summary

`load_function_module_by_id` in `backend/open_webui/utils/plugin.py` flips
`is_active` to `False` in the database whenever exec-ing a Function's stored
content raises (plugin.py:307-313):

```python
except Exception as e:
    log.error(f'Error loading module: {function_id}: {e}')
    # Cleanup by removing the module in case of error
    del sys.modules[module_name]

    await Functions.update_function_by_id(function_id, {'is_active': False})
    raise e
```

`get_active_filter_ids` (backend/open_webui/models/functions.py:285) only
returns filters with `is_active=True`, so after one load error the Function
is excluded from every model and every user on all later requests. Nothing
in code re-enables it and nothing notifies the admin beyond a single
`log.error` line; the admin panel toggle silently moves to off.

Any transient load failure triggers this. Realistic causes: a dependency
the filter imports is removed or version-bumped by an upgrade, an optional
import that fails on a new Python version, or a partially saved edit. For
a global (`is_global=True`) moderation/safety filter, one transient error
becomes a standing, silent bypass: the request that triggered the error
fails closed, and every request after that runs with the filter gone.

## Reproduction

The PoC imports the real `utils/plugin.py` and `utils/filter.py` from a
clone at the pinned commit (only the Functions DB layer is replaced by an
in-memory fake that records every `update_function_by_id` call) and walks
`resolve_filter_pipeline` exactly as the server does:

```
python3 poc/poc_f1_filter_autodeactivate.py evidence.json
```

Set `OPEN_WEBUI_CLONE` to the clone root if it is not in the default
location. Three phases:

1. A healthy global safety filter blocks a message containing a marker
   (control).
2. The same filter's stored content now raises on import (simulating a
   lost dependency). The resolve call raises and exactly one
   `update_function_by_id('safety-global', {'is_active': False})` is
   recorded.
3. The admin repairs the content. `is_active` stays `False` in the store,
   `resolve_filter_pipeline` returns zero filters, and the same marker
   message passes through unfiltered.

Observed verdict: one load error permanently disables the global filter;
the previously blocked content passes unfiltered afterwards.

## Suggested direction

- Do not persist a deactivation on load failure; fail the request (as
  today) and retry the load next request, or quarantine with an explicit
  admin-visible state that is clearly distinct from a user/admin toggle
  and can expire or self-heal once the content loads again.
- If auto-deactivation is kept, emit an admin notification/event and a
  warning-level log, and surface it on the Functions page.

AUTHENSOR:
# Pull Request

Closes #29681

## Summary

`load_function_module_by_id` in `backend/open_webui/utils/plugin.py` persists `is_active: False` whenever exec-ing a Function's stored content raises. After that write, `get_active_filter_ids` (which filters `is_active=True`) excludes the Function from every model and every user on all later requests. Nothing in code re-enables it and nothing notifies the admin beyond a single `log.error` line; the admin panel toggle silently moves to off.

Any transient load failure triggers this: a dependency the filter imports removed or version-bumped by an upgrade, an optional import failing on a new Python version, or a partially saved edit. For a global (`is_global=True`) moderation/safety filter, one transient error becomes a standing, silent bypass: the request that triggered the error fails closed, and every request after it runs with the filter gone.

Change: on a load failure, fail the request (as today) but do not write the deactivation to the database, so the Function loads again on the next request once the underlying cause is fixed. The module cleanup (`del sys.modules[...]`) and the `log.error` signal are unchanged.

## Testing

The repo has no Python test suite, so the change was verified with a runnable harness (https://gist.github.com/AUTHENSOR/1dc7d1fbe048eb494fe160ac65040d84) that imports the real `utils/plugin.py` and `utils/filter.py` and replaces only the Functions DB layer with an in-memory store recording every `update_function_by_id` call:

1. Healthy global filter blocks marker content (control).
2. The same filter's content raises on import (simulated dependency loss): the request fails; on this branch no `update_function_by_id` call is recorded and `is_active` stays `True`. On current `main` this phase records exactly `('safety-global', {'is_active': False})`.
3. After the content is repaired, the same filter resolves again and blocks the marker: a transient failure no longer persists.

### Fixed

- A Function (including global safety filters) is no longer permanently disabled in the database by a single transient load error; the failing request errors and the Function loads again once the cause is fixed.

### DIFF
--- a/backend/open_webui/utils/plugin.py
+++ b/backend/open_webui/utils/plugin.py
@@ -309,7 +309,9 @@ async def load_function_module_by_id(function_id: str, content: str | None = Non
         # Cleanup by removing the module in case of error
         del sys.modules[module_name]
 
-        await Functions.update_function_by_id(function_id, {'is_active': False})
+        # Do not persist a deactivation: a transient load failure (broken
+        # dependency, partial edit) must fail this request only, so the
+        # Function loads again on the next request once the cause is fixed.
         raise e
     finally:
         os.unlink(temp_file.name)


AUTHENSOR:
Retargeted to `dev` per the policy notice: replacement PR opened from the same branch with the CLA section and title prefix in place. Linking it here to keep the history in one thread.

Issue #29220, opened 2026-08-29
issue: edit_image cannot load this instance's own /api/v1/files URLs unless the host matches request.base_url


gilvandev:
### Check Existing Issues

- [x] I have searched for any existing and/or related issues.
- [x] I have searched for any existing and/or related discussions.
- [x] I have also searched in the CLOSED issues AND CLOSED discussions and found no related items (your issue might already be addressed on the development branch!).
- [x] I have checked whether this issue is already fixed on the `dev` branch or in the latest source.
- [x] I am using the latest version of Open WebUI.

### Installation Method

Git Clone

### Open WebUI Version

0.11.1

### Ollama Version (if applicable)

0.33.1

### Operating System

Ubuntu 24.

### Confirmation

- [x] I have read and followed all instructions in `README.md`.
- [x] I am using the latest version of **both** Open WebUI and Ollama.
- [x] I have included the browser console logs.
- [x] I have included the Docker container logs.
- [x] I have **provided every relevant configuration, setting, and environment variable used in my setup.**
- [x] I have clearly **listed every relevant configuration, custom setting, environment variable, and command-line option that influences my setup** (such as Docker Compose overrides, .env values, browser settings, authentication configurations, etc).
- [x] I have documented **step-by-step reproduction instructions that are precise, sequential, and leave nothing to interpretation**. My steps:
- Start with the initial platform/version/OS and dependencies used,
- Specify exact install/launch/configure commands,
- List URLs visited, user input (incl. example values/emails/passwords if needed),
- Describe all options and toggles enabled or changed,
- Include any files or environmental changes,
- Identify the expected and actual result at each stage,
- Ensure any reasonably skilled user can follow and hit the same issue.

### Expected Behavior

The `edit_image` tool should be able to retrieve images from Open WebUI's own file storage without failing due to authentication, since this is an internal, same-instance request.

### Actual Behavior

The fetch fails with a 400 error indicating the image could not be loaded, despite the URL being valid and correctly formatted, and despite the file being confirmed to exist and be accessible via the same endpoint when authenticated.

### Steps to Reproduce

1. Configure Settings → Images → Image Editing with a self-hosted backend (in my case, ComfyUI) and a custom ComfyUI workflow (Flux Kontext).
2. Set Function Calling to a mode that exposes native system tools (Default/Native) for a vision-capable model.
3. Upload an image in a chat and ask the model to edit it (e.g., "remove X from this image").
[...]

PR #29691, opened 2026-09-05
fix: resolve this instance's own file URLs in edit_image regardless of the URL host

Classic298:
The native edit_image tool fails with "400: [ERROR: Error loading image]" whenever the model hands it an absolute URL for an image Open WebUI already stores. Such a URL is treated as local only when its host string matches the incoming request's host exactly, so a default-port form, a container name or any host the model composed itself falls through to an outbound HTTP fetch instead. That fetch asks /api/v1/files/{id}/content without a session, gets a 401, and the user sees the generic 400.

Match the file URL on its path and let the existing local branch resolve it. Fetching that endpoint over the network can never succeed for a local or a remote instance, because it requires an authenticated user, so the host comparison only decided which way the request failed. Access control is unchanged: the local branch still goes through get_file_content_by_id, which enforces owner, admin or shared access.

Fixes #29220

### DIFF
--- a/backend/open_webui/routers/images.py
+++ b/backend/open_webui/routers/images.py
@@ -919,11 +919,8 @@ async def image_edits(
 
             if data.startswith('http://') or data.startswith('https://'):
                 parsed = urlparse(data)
-                if (
-                    parsed.netloc == urlparse(str(request.base_url)).netloc
-                    and parsed.path.startswith('/api/v1/files/')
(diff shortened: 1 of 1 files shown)


kbruegge:
Looks good as far as I can tell. Not sure if the tool itself needs to be updated with clearer instructions for the model as well. Haven't checked.

Issue #29716, opened 2026-09-05
issue: OpenAPI tool server sends path and query parameters in the JSON request body


fab4fun:
### Before Submitting

- [x] I searched open and closed issues and discussions for an existing report.
- [x] I checked whether this is already fixed on the `dev` branch or latest source.
- [x] I understand that maintainers want a well-written issue before any code pull request.
- [x] I am using the latest available version of Open WebUI for my install method.
- [x] This is not a security vulnerability.

### Installation Method

Docker

### Open WebUI Version

v0.11.3

### Operating System

Windows 11

### Browser

Chrome

### Summary

When an OpenAPI tool-server endpoint has a **path parameter** *and* a **request body**, Open WebUI sends the path parameter **inside the JSON request body**. Servers that enforce strict schema validation (`additionalProperties: false`) reject the request with a `422` (e.g. `unexpected property`).

**Read operations (GET) work fine. Write operations (POST/PUT/PATCH) fail.**

### Expected Behavior

The path parameter is substituted into the URL only, and the JSON body contains **only** the properties declared by the `requestBody` schema.

### Actual Behavior

The server responds with `422`. For Vikunja:

```json
{"error": "unexpected property", "property": "task", ...}
```

The path param `task` appears in the request body.

### Steps to Reproduce

1. Add an OpenAPI tool server whose spec includes an endpoint with:
   - a path parameter (e.g. `POST /tasks/{task}/relations`), and
   - a request body whose schema has `additionalProperties: false`.
   - (Vikunja is a good repro: `POST /tasks/{task}/relations` with the `TaskRelation` schema, where `task_id` is `readOnly` and documented as *"Set from the URL path; ignored in the request body."*)
2. Enable the tool in a chat and ask the model to call that endpoint (e.g. "make task X a subtask of task Y").
3. The tool call fails.

### Logs, Screenshots, and Config

tool integration settings:
Edit Connection

Type
OpenAPI
Name
Vikunja
ID (optional)
auto
Description
Task Management
URL
http://192.168.2.22:3456/api/v2

Auth

Bearer
API Key
•••••••••••••••••••••••••••••••••••••••••••

Advanced

Access Control
Function Name Filter List
[...]

PR #29717, opened 2026-09-05
fix: stop sending OpenAPI tool server path and query parameters in the request body

Classic298:
Tool calls to an OpenAPI tool server put every argument the model returned into the JSON request body, including the parameters that were already substituted into the URL. Servers that validate their input strictly (additionalProperties: false) answered 422 "unexpected property", so reads worked and every write through an endpoint with a path or query parameter failed.

The body is now built from the model's arguments minus the operation's declared parameters, keeping any name the requestBody schema declares as a property of its own, so an endpoint that wants the resource id in the body as well as in the path still gets it.

The filter only runs when the resolved body schema lists its properties. A free-form, composed or non-JSON body offers nothing to check a name against, so those requests go out exactly as before.

src/lib/apis/index.ts carries the same request builder for direct tool server connections and had the same bug, so it gets the same fix.

Fixes #29716

### DIFF
--- a/backend/open_webui/utils/tools.py
+++ b/backend/open_webui/utils/tools.py
@@ -1668,6 +1668,7 @@ async def execute_tool_server(
         path_params = {}
         query_params = {}
         body_params = {}
+        declared_param_names = set()
 
         # Merge path-level and operation-level parameters for execution.
         path_level_params = methods.get('parameters', [])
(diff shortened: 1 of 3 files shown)


fab4fun:
Hmm, that frontend build error seems completely unrelated to the change

Issue #29721, opened 2026-09-06
feat: drop verified dead weight from the Docker images


Classic298:
The `:main` and `:slim` images carry a few hundred MB of files that nothing in the image uses. I audited both, verified every item against the code and the installed packages, and measured the result with the real CI workflow on my fork (same runners, same day, main and slim, amd64 and arm64).

Baseline is current `dev`. Measurement branches: `docker-size-baseline` and `docker-size-freelunch` on my fork, the size step is in the workflow there.

### Free lunch (nothing uses it, zero behaviour change)

| item | on-disk saving | why it is dead |
|---|---|---|
| `open_webui/static` copied twice | 68.7 MB (measured layer) | The `chgrp`/`chmod -R g=u` RUN after `COPY ./backend .` rewrites every file into a second layer. Doing the permission fix in a prep stage and copying once keeps the OpenShift fix and drops the duplicate. |
| GUI `opencv-python` installed next to `opencv-python-headless` | 110 to 195 MB (wheel contents) | `rapidocr` depends on the GUI `opencv_python`, so uv installs both; both ship `cv2/`, last writer wins. rapidocr only uses core/imgproc/imgcodecs, headless covers it. Excluding the GUI build during install (#29730) was closed as the wrong layer for it. Asked rapidocr to make the OpenCV build selectable instead: RapidAI/RapidOCR#737. Parked until that lands. |
| `google-api-python-client`, `google-auth-httplib2`, `google-auth-oauthlib` | 100 MB (wheel contents) | Added for Google Drive in 2024, never imported by the backend. The Drive picker is frontend-only (gapi from apis.google.com). Nothing installed requires them. |
| `python3-dev` | ~55 MB (Debian) | Installs a second Debian interpreter. The image's Python is `/usr/local`, which already ships its headers. Nothing in the pinned tree compiles anyway: every compiled pin has cp311/abi3 wheels for x86_64 and aarch64. |
| `uv` left in the final image | 50 MB (binary) | Only needed for the install RUN. Mount it from the uv image for that RUN instead of pip-installing it. |
| hashed copies of `ort-wasm-simd-threaded.jsep.wasm` | 43 MB (measured layer) | Vite emits two hashed copies of the onnxruntime wasm next to the `/wasm/` copy the kokoro worker is hardcoded to load. The hashed ones are only referenced from a fallback path that never runs. |
[...]

PR #29727, opened 2026-09-06
chore: drop the duplicate onnxruntime wasm copies from the frontend build

Classic298:
The frontend build gets 43 MB smaller uncompressed (about 10 MB on the compressed pull), in the Docker image and in the PyPI wheel alike. Vite emits two hashed copies of ort-wasm-simd-threaded.jsep.wasm (21.6 MB each) under _app/immutable besides the copy in /wasm/ that viteStaticCopy places.

Both hashed copies are referenced from shipped code, but only from the branch onnxruntime takes when wasmPaths is unset. src/lib/workers/kokoro.worker.ts sets wasmPaths to /wasm/, and the main-thread import in the audio settings relies on the transformers.js default, which points at its CDN. With wasmPaths set, onnxruntime resolves the runtime through that prefix and never through the hashed files, so nothing fetches them.

The delete runs as a Node step at the end of npm run build, after the SvelteKit adapter has written build/, which is the command the Dockerfile, the hatch wheel build and the frontend CI all call. A Rollup hook could drop the asset during the build instead, but the two copies come out of the main bundle and the worker bundle, which register plugins separately, so one post-build step is the smaller change. build:watch is not covered; it never returns, and its output is not shipped.

Part of #29721.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
 	"scripts": {
 		"dev": "npm run pyodide:fetch && vite dev --host",
 		"dev:5050": "npm run pyodide:fetch && vite dev --port 5050",
-		"build": "npm run pyodide:fetch && vite build",
+		"build": "npm run pyodide:fetch && vite build && node scripts/drop-duplicate-ort-wasm.js",
 		"build:watch": "npm run pyodide:fetch && vite build --watch",
 		"preview": "vite preview",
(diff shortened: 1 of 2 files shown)


tjbck:
f1c803d36bae9d429eb630ad26853c9d939262b5

PR #29730, opened 2026-09-06
chore: stop installing the GUI opencv build alongside headless

Classic298:
rapidocr depends on opencv_python, the GUI build, which the pinned opencv-python-headless does not satisfy, so the image install resolves the newest GUI build (currently opencv-python 5.0.0.93) next to opencv-python-headless 4.13.0.92. Both write the same cv2 package, so today's cv2 is whichever installed last, 5.0 or 4.13, and the GUI build adds about 115 MiB (x86_64) or 86 MiB (aarch64) of files headless does not have, mostly bundled Qt, plus a libGL requirement the image does not install.

rapidocr only uses the core, imgproc and imgcodecs modules, which headless provides, so the GUI build is now excluded from the requirements install through uv's excludes file. cv2 in the image is then deterministically opencv-python-headless 4.13.0.92, which satisfies rapidocr's version floor, a major version below the 5.0 some current builds end up with.

This covers the Docker image only: pip installs of the published package and uv sync from uv.lock still resolve both builds. Inside the image, pip check now reports rapidocr's declared opencv-python as missing.

Part of #29721.


Classic298:
Closing this. Overriding rapidocr's declared dependency from the Dockerfile is the wrong layer for it, the fix belongs in rapidocr. They declined switching to opencv-python-headless outright in RapidAI/RapidOCR#185, so the ask there is now an extra that lets downstream projects opt out of the GUI build. Until that lands the image keeps installing both opencv builds.

Classic298:
https://github.com/RapidAI/RapidOCR/issues/737

Classic298:
Upstream ask is RapidAI/RapidOCR#737.

Issue #29722, opened 2026-09-06
bug: server killed by SIGSEGV during reranking on Apple Silicon (MPS)


dopeearthalien:
### Bug Report

**Summary**

On macOS with Apple Silicon, the server process is killed by the OS partway through answering any question that retrieves from a knowledge base with hybrid search + reranking enabled. It is not an exception — there is no Python traceback and no graceful error. The process dies, the client sees `RemoteDisconnected` / connection refused, and the answer is lost.

The crash is in PyTorch's MPS backend, and it appears to be caused by Open WebUI calling the cross-encoder reranker from more than one thread at the same time.

### Environment

- Open WebUI **0.11.3** (pip install, native, not Docker)
- macOS 15 (Darwin 25.5.0), Apple Silicon, 48 GB
- Python 3.11.15
- torch **2.14.0**
- Reranker: `BAAI/bge-reranker-v2-m3`
- Embeddings: `bge-m3` via Ollama (so torch is used **only** by the reranker)
- Hybrid search: on. Top-K 8.

### Steps to reproduce

1. Create a knowledge base and add documents (mine: 71 files, ~800k characters).
2. Enable hybrid search and set a reranking model.
3. Attach the knowledge base to a model and ask any question that triggers retrieval.

### What happens

The last thing logged is the start of the hybrid search, then the first reranker batch, then the process is gone:

```
INFO | open_webui.retrieval.utils:query_collection_with_hybrid_search:836 - Starting hybrid search for 2 queries in 1 collections...
Batches:   0%|          | 0/1 [00:00<?, ?it/s]
Batches:   0%|          | 0/1 [00:00<?, ?it/s][A
<process dies — no traceback>
```

### Crash reports

Two crashes, different signals, identical faulting frame. Both are a device/dtype cast (`_to_copy`) reaching the MPS unary kernel.

**1. `Python-2026-09-06-125214.ips`**

```
exception   : EXC_BREAKPOINT  (SIGTRAP)
termination : Trace/BPT trap: 5

libsystem_malloc.dylib  _xzm_xzone_malloc_freelist_outlined
libc++abi.dylib         operator new(unsigned long)
libtorch_cpu.dylib      std::__1::__hash_table<...>::__emplace_unique_key_args<...>
libtorch_cpu.dylib      at::native::mps::MetalShaderLibrary::exec_unary_kernel(at::TensorIteratorBase&, ...)
[...]

PR #29733, opened 2026-09-06
fix(retrieval): serialize local embedding and reranking on MPS

Classic298:
On Apple Silicon the server process is killed outright (SIGSEGV or SIGTRAP, no traceback) partway through answering any question that retrieves from a knowledge base with hybrid search and a local reranking model. The client sees a dropped connection and the answer is lost.

Hybrid search fans its queries out concurrently and every task calls the same shared local model on a worker thread. Torch's Metal shader cache is a process-wide singleton whose lookup tables have no lock, so two of those threads racing inside it corrupt the cache and take the process down with it.

Guard the local SentenceTransformer and CrossEncoder calls with a shared lock that is only a real lock when the selected device is MPS. CPU and CUDA installs keep the concurrency they have today, and external reranking endpoints are untouched. Reranking several queries on a Mac now runs one at a time, which is the cost of the process staying alive.

Verified by driving the real hybrid-search fan-out with 16 concurrent queries: peak simultaneous entries into the local model drops from 16 to 1 on MPS, stays at 16 on CPU, and the returned documents, scores and ordering are byte-identical in every case.

Fixes #29722

### DIFF
--- a/backend/open_webui/retrieval/utils.py
+++ b/backend/open_webui/retrieval/utils.py
@@ -5,7 +5,9 @@ import hashlib
 import logging
 import os
 import re
+import threading
 import time
+from contextlib import nullcontext
 from typing import Awaitable, Optional, Union
 from urllib.parse import quote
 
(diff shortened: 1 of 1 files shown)


Classic298:
Superseded by #29735

Issue #29761, opened 2026-09-07
bug: context usage estimator counts base64 image data URIs in message files as text, causing spurious auto-compaction


tastyrice90:
### Before Submitting

- [x] I searched open and closed issues and discussions for an existing report.
- [x] I checked whether this is already fixed on the `dev` branch or latest source.
- [x] I understand that maintainers want a well-written issue before any code pull request.
- [x] I am using the latest available version of Open WebUI for my install method.
- [x] This is not a security vulnerability.

### Installation Method

Docker

### Open WebUI Version

0.11.3 — ghcr.io/open-webui/open-webui@sha256:3e52522170dea0b587a976a7429f8fa6edf7be7d32cc3a1404b9087abb131f69

### Operating System

Linux — Fedora (kernel 7.1.4-204.fc44.x86_64)

### Browser

Brave

### Summary

When a multimodal agent views an image inside Open Terminal via read_file, the engine processes it normally (bounded vision tokens, <8k per image), but OWUI's context meter estimates the persisted base64 string at len ÷ 4 "tokens" (~1M for a 3MB image). With context compaction enabled, the next request triggers a spurious auto-compaction. The raw base64 is also persisted inside the chat JSON — twice — which balloons browser tab RAM (released only on tab close) and database/session payloads. Related to #29208; PR #29665 fixes the model-facing leak but not this meter/persistence behavior.

### Expected Behavior

Viewing an image via read_file should add its real, bounded token cost (the engine's vision encoding, <8k) to the context meter — not chars ÷ 4 of the stored base64.
Auto-compaction should not fire when the engine's actual context is far below the configured threshold (every real request in this backend's history is ≤600k tokens; typical agentic requests ≤40k, threshold 350k).
Browser tab RAM should not grow by roughly the image size for every image the agent views.

### Actual Behavior

Agent viewed a 3,092,378-byte JPG via read_file in Open Terminal. The tool-result message persisted as 4,197,680 bytes (~4.12M base64 chars). The context meter's estimate for that single message = 4,197,680 ÷ 4 ≈ 1,049,420 "tokens" — larger than the engine's entire 450,560-token KV pool, i.e. an estimate that is physically impossible on this backend.
Auto-compaction fired on the next request (log line below), even though engine-side prompt tokens stayed normal.
chat_message.usage was empty for every message in the repro chat — no real usage data is stored — so get_chat_context_usage falls back to the len ÷ 4 estimate path, which is what counts the persisted base64.
[...]

PR #29765, opened 2026-09-07
fix: store tool result images as file references

Classic298:
A tool that returned an image wrote the entire base64 data URI into the chat, and stored it twice: once on the message row and again in the chat's own history map. A single 2 MB image cost 5.6 MB of database, and the same payload was sent to the browser with every chat load, so tab memory grew by roughly the image size per image and only came back when the tab was closed. Long image-heavy chats reached tens of megabytes of JSON.

The image is now uploaded once and the chat keeps a short file reference, the same way an uploaded image is already stored. The base64 is rebuilt only while a provider request is being assembled, so what the model receives is unchanged.

Measured on a 2 MB tool image, with an identical request reaching the provider in both cases:

| | stored bytes | base64 copies |
| --- | --- | --- |
| before | 5,594,322 | 2 |
| after | 1,962 | 0 |

Fixes #29761

### DIFF
--- a/backend/open_webui/utils/files.py
+++ b/backend/open_webui/utils/files.py
@@ -32,6 +32,7 @@ from open_webui.storage.provider import Storage
 
 BASE64_IMAGE_URL_PREFIX = re.compile(r'data:image/\w+;base64,', re.IGNORECASE)
 MARKDOWN_IMAGE_URL_PATTERN = re.compile(r'!\[(.*?)\]\((.+?)\)', re.IGNORECASE)
+FILE_CONTENT_URL_PATTERN = re.compile(r'^/api/v1/files/([^/?#]+)/content')
 
 # Extension-based MIME fallback, only used when ENABLE_IMAGE_CONTENT_TYPE_EXTENSION_FALLBACK is True.
 _IMAGE_MIME_FALLBACK = {
@@ -90,7 +91,8 @@ async def get_image_base64_from_url(url: str, user=None) -> Optional[str]:
         else:
             # Non-URL string — treat as file_id. Delegate to the canonical
             # file-ID resolver which enforces ownership/access checks.
-            return await get_image_base64_from_file_id(url, user=user)
+            file_id_match = FILE_CONTENT_URL_PATTERN.match(url)
+            return await get_image_base64_from_file_id(file_id_match.group(1) if file_id_match else url, user=user)
 
     except Exception:
         return None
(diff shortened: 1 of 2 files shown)


tastyrice90:
Test result: fix confirmed working — built from fix/tool-image-file-refs (PR head 7c7a14e, base dev) and tested end-to-end.

Setup: isolated test instance — fresh SQLite volume (no Postgres/Redis), official ghcr.io/open-webui/open-terminal container, port-isolated from my production stack. Vision-capable model via SGLang (450,560-token KV pool), compaction enabled at token_threshold = 350,000 in the Admin panel. Control run: the exact same 0.11.3 image digest as my production instance, on a fresh SQLite volume, sharing nothing with production except the engine — which rules the backend out as a factor.

Procedure: fresh chat → agent views a 3,092,378-byte JPG via Open Terminal read_file → multiple follow-up turns (ping/pong ×5).

Results (same 3MB-image repro in all three environments):

Production incident (0.11.3, Postgres): chat JSON 4,201,936 bytes; auto-compaction fired (dropped=2, kept=3); meter ~1M "tokens" estimate; vision worked.
Control (0.11.3, fresh SQLite, no fix): chat JSON 4,140,969 bytes; auto-compaction fired (dropped=2, kept=3); same meter behavior; vision worked.
With PR #29765: chat JSON 18,016 bytes (~230× smaller); zero compaction events across 5+ turns; meter normal; vision worked — the agent read spine text straight off the photo.
Notes:

The unfixed builds compact identically on Postgres and on SQLite (same compact_messages_for_request log line, same dropped=2 kept=3 signature) — the bug is backend-independent, and the PR eliminates it on both paths.
Storage for the same 3MB-image scenario dropped ~230× at the chat level — consistent with the PR's own 2MB-image measurement (5,594,322 → 1,962 stored bytes, 2 → 0 base64 copies). The entire test-instance DB (647 KB) is now smaller than a single unfixed bloated chat (4.1 MB).
Vision is unaffected: file references are re-expanded to base64 at request assembly, so the model still receives the image.
Note for anyone scripting against tool outputs: with the PR, read_file's tool result is a short confirmation ("Image file read successfully") with the image attached at message level rather than embedded in the tool-result string.
Thanks for the quick turnaround — happy to re-test if anything changes on the branch.

tjbck:
d372bec70427fe2d568e052ce5e1529e2ad41da9

Issue #29842, opened 2026-09-09
fix: inline <thinking> tag shown raw while streaming, and line breaks lost between single tildes in reasoning text


bitsofinfo:
### Before Submitting

- [x] I searched open and closed issues and discussions for an existing report.
- [x] I checked whether this is already fixed on the `dev` branch or latest source.
- [x] I understand that maintainers want a well-written issue before any code pull request.
- [x] I am using the latest available version of Open WebUI for my install method.
- [x] This is not a security vulnerability.

### Installation Method

Docker

### Open WebUI Version

0.11.3

### Operating System

linux

### Browser

chrome latest

### Summary

**Open WebUI version:** 0.11.3 (regressed from 0.10.2)
**Deployment:** Docker, OpenAI-compatible upstream connection
**Related:** #29786

## TL;DR

An OpenAI-compatible backend streams its reasoning/progress as an inline
`<thinking>…</thinking>` block in `content`, and keeps the block **open for several
seconds**, emitting a small update roughly once per second (a progress heartbeat) — the
normal shape for a slow / tool-using model.

On **0.11.3** two things go wrong; both were fine on **0.10.2**:

1. **Raw open-tag exposure** — the literal text `<thinking>` and everything after it is
   shown as **plain text** for the entire time the block is open. It only collapses into
   the reasoning panel once `</thinking>` arrives (stream completes).
2. **Interior newlines dropped** — on completion, every reasoning line that streamed in
   **while the block was still actively updating** is merged onto **one line** (its
   leading `\n` is gone). Only the **last couple of lines** (the ones streamed *after* the
   updates stopped) and the **final answer** (plain content after `</thinking>`) keep
   their newlines.

Both are caused by the same 0.11 change: the streaming inline-tag detector was rewritten
from a **whole-buffer regex scan** (0.10.2) into an **incremental, offset-based scanner**
(`tag_output_handler` in `backend/open_webui/utils/middleware.py`).

https://github.com/user-attachments/assets/25e677d1-17b1-425c-83c2-87fcbe55d1f1

[owui-repro.zip](https://github.com/user-attachments/files/32017452/owui-repro.zip)

## Environment / how the upstream streams

The upstream is a plain OpenAI-compatible `/v1/chat/completions` SSE stream. Each delta is
`choices[0].delta.content`. Reasoning is surfaced **inline** as one `<thinking>` block:

- The block **opens** in the first delta: `"<thinking>\n<emoji> starting: …"`.
[...]

PR #29858, opened 2026-09-09
fix: stop streaming the raw reasoning tag into the thinking panel

Classic298:
When a model streams its reasoning inline as <think> or <thinking> text, the chunk that opens the block is forwarded to the browser as-is, so the literal tag sits at the top of the thinking panel for as long as the block is open, and the chunk that closes it lands in the answer with the closing tag still in it. Both only clean up when the response finishes and the final output replaces what the browser accumulated. Regressed in v0.11.1 with the per-chunk delta events; v0.10.2 resent the already cleaned output instead.

The backend already strips the tags from its own output items in that same chunk. Now, whenever the tag scanner opens or closes an item, the cleaned output is resent whole (the same snapshot event the handler already sends at the end of a response) instead of the raw delta, so the thinking panel shows clean text from the first chunk and the answer never carries the closing tag. Chunks without a tag transition still stream as plain deltas, unchanged. Channel chats adopt that snapshot too, and the browser now applies text deltas to a code interpreter block's code so it keeps streaming after the tag.

Fixes #29842

### DIFF
--- a/backend/open_webui/socket/main.py
+++ b/backend/open_webui/socket/main.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import copy
 import logging
 import random
 import sys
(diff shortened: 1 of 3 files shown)


Classic298:
Already in dev, this rebases to an empty diff. Closing.

Issue #29879, opened 2026-09-10
issue: MCP tools/list pagination is ignored, so only the first page is discovered


killecaptron:
### Before Submitting

- [x] I searched open and closed issues and discussions for an existing report.
- [x] I checked whether this is already fixed on the `dev` branch or latest source.
- [x] I understand that maintainers want a well-written issue before any code pull request.
- [x] I am using the latest available version of Open WebUI for my install method.
- [x] This is not a security vulnerability.

### Installation Method

Docker

### Open WebUI Version

v0.11.3; also present in current `main` at commit `0a7c15832fb30b1903753e83f81dc7d27e5b0944`

### Operating System

Ubuntu, Docker deployment

### Browser

Not browser-specific

### Ollama Version

Not involved

### Summary

Open WebUI only discovers the first page of tools exposed by a paginated MCP Streamable HTTP server. The MCP specification permits `tools/list` responses to contain a `nextCursor`, but `MCPClient.list_tool_specs()` calls `self.session.list_tools()` exactly once and discards the response cursor.

This was reproduced with an MCP server exposing 44 tools in pages of 20. Pagination-aware MCP clients discover all 44 tools, while Open WebUI exposes exactly the first 20. The server connection and tools on the first page otherwise work normally.

### Expected Behavior

Open WebUI should continue calling `tools/list` with the returned `nextCursor` until the server omits the cursor, then expose the combined tool list.

### Actual Behavior

Only the tools in the first `tools/list` response are available in Open WebUI. Tools from every later page are missing from the integration and cannot be selected or called.

The function-name filter does not help because it is applied after discovery; the later pages were never requested.

### Steps to Reproduce

1. Run any MCP Streamable HTTP server that exposes more tools than its configured page size. For example, expose 44 tools and return 20 per `tools/list` response.
2. Confirm that the first `tools/list` response contains 20 tools and a non-empty `nextCursor`.
3. Add the endpoint in Open WebUI under Admin Settings -> Integrations as `MCP Streamable HTTP`.
4. Leave the function-name filter empty and save the connection.
5. Inspect the tools made available by the integration. Only the first 20 are present.
6. Connect with a pagination-aware MCP client and follow `nextCursor`. All 44 tools are returned by the same server.

### Logs, Screenshots, and Config

No server or browser error is emitted because the first response is a valid, successful MCP response.

The cause is visible in `backend/open_webui/utils/mcp/client.py` in both v0.11.3 and current `main`:

```python
[...]

PR #29919, opened 2026-09-11
fix(mcp): follow tools/list pagination so all pages of tools are discovered

SIDDARTHAREDDY8:
Supersedes #29912, which was closed for targeting `main` — retargeted to `dev` per the contribution requirements.

Fixes #29879.

`MCPClient.list_tool_specs()` called `session.list_tools()` exactly once and discarded the response cursor, so paginated MCP servers only exposed their first page of tools (e.g. 20 of 44). It now follows the MCP-spec `nextCursor` across `tools/list` calls until the server omits it, accumulating every page before building tool specs.

Single-page servers are unaffected: with no cursor returned, exactly one `list_tools` call is made, as before.

Testing confirmation (per the PR checklist):
- Drove the patched `list_tool_specs` against a stubbed session returning the issue repro shape (44 tools in pages of 20/20/4) — all 44 discovered, cursors followed in order.
- Edge cases checked: single-page server (exactly one `list_tools` call), empty tool list, and not-connected client.
- Change is confined to the pagination loop (1 file, +11/-2); no other call paths touched.

### DIFF
--- a/backend/open_webui/utils/mcp/client.py
+++ b/backend/open_webui/utils/mcp/client.py
@@ -89,8 +89,17 @@ class MCPClient:
         if not self.session:
             raise RuntimeError('MCP client is not connected.')
 
-        result = await self.session.list_tools()
-        tools = result.tools
+        # The MCP spec allows tools/list responses to be paginated via
+        # nextCursor. Follow the cursor until the server omits it so every
+        # page of tools is discovered, not just the first one.
+        tools = []
+        cursor = None
+        while True:
+            result = await self.session.list_tools(cursor)
+            tools.extend(result.tools)
+            cursor = result.nextCursor
+            if not cursor:
+                break
 
         tool_specs = []
         for tool in tools:


silentoplayz:
This branch was created from `main` rather than `dev`, so the PR carries 46 release merge commits and the regression workflow commit that is on `main` but not yet on `dev`. Only the last commit, a5f0bd8, is the pagination change, and it touches one file.

Rebasing that one commit onto the current `dev` brings the PR down to a single commit and a single file:

```bash
git fetch upstream dev
git checkout oss-bhai-2026-09-11
git rebase --onto upstream/dev a5f0bd8~1 oss-bhai-2026-09-11
git push --force-with-lease origin oss-bhai-2026-09-11
```

`upstream` here is the open-webui/open-webui remote. After the push, the Commits tab should show only your commit and the Files tab only `backend/open_webui/utils/mcp/client.py`.

SIDDARTHAREDDY8:
Good catch — rebased the branch onto `dev`. The PR is now a single commit touching only `backend/open_webui/utils/mcp/client.py`.

Issue #29944, opened 2026-09-12
bug: Spellcheck right-click corrections intermittently fail in Chromium browsers


gmanbaron74:
### Before Submitting

- [x] I searched open and closed issues and discussions for an existing report.
- [x] I checked whether this is already fixed on the `dev` branch or latest source.
- [x] I understand that maintainers want a well-written issue before any code pull request.
- [x] I am using the latest available version of Open WebUI for my install method.
- [x] This is not a security vulnerability.

### Installation Method

Docker

### Open WebUI Version

v0.11.3

### Operating System

Windows 11

### Browser

Brave 1.95.101 (Official Build) (64-bit) Chromium 153.0.8010.37  Microsoft Edge 153.0.4234.32 (Official build) (64-bit)  Firefox 155.0.1

### Ollama Version

N/A — issue occurs in the Open WebUI frontend and is independent of model/provider.

### Summary

Browser spellcheck correctly detects misspelled words in the Open WebUI chat input and displays spelling suggestions. However, when a suggested correction is selected from the browser's right-click context menu, the correction intermittently fails to stick or is reverted.

The issue occurs in both Brave and Microsoft Edge on Windows 11. It also occurs in a Brave Private Window.

The same test works correctly in Firefox. Manually correcting the word using the keyboard also works correctly.

The issue occurs with Rich Text Input both enabled and disabled.

### Expected Behavior

When a misspelled word is right-clicked and a spelling suggestion is selected from the browser's native context menu, the selected correction should replace the misspelled word and remain in the chat input.

### Actual Behavior

The browser detects the misspelled word and offers the correct spelling suggestion, but selecting the suggestion intermittently fails to update the Open WebUI chat input correctly. The original misspelling may remain or the correction may be reverted.

Manual keyboard editing works normally.

This behaviour has been reproduced in Brave and Microsoft Edge, but not Firefox.

### Steps to Reproduce

1. Run Open WebUI v0.11.3.
2. Open Open WebUI on Windows 11 using Brave or Microsoft Edge.
3. Start a new chat.
4. Type a deliberately misspelled word into the chat input, for example "recieve".
5. Wait for the browser spellchecker to underline the word.
6. Right-click the misspelled word.
7. Select the browser's suggested correction ("receive").
8. Repeat this with several misspelled words.
9. Observe that some right-click spelling corrections fail to stick or are reverted.

Additional testing:

- Reproduced in Brave normal browsing mode.
- Reproduced in Brave Private Window.
[...]

PR #29952, opened 2026-09-12
fix: apply the selection highlight only when the editor is unfocused and keep it out of the chat input

silentoplayz:
# Pull Request

## Checklist

- [x] This PR links to a well-described, confirmed Issue or active Discussion: Closes #29944

## Summary

In Chromium browsers, picking a spelling suggestion from the right-click menu in the chat input often does nothing, or the misspelling comes back. Firefox is fine, and typing the correction by hand is fine everywhere.

Chromium keeps its spelling markers on the text node of the word. A probe on the chat input shows what happens on right-click. About twenty milliseconds after the click, the paragraph's single text node is replaced by three nodes. That is the editor rendering an inline decoration over the selected word, and the markers die with the old node. The context menu was built from those markers, so the suggestion it offers has nothing left to replace.

The decoration is the selection highlight added for the notes editor, which should show only while the editor is unfocused. Its check reads `this.editor.focused`, a property the TipTap editor does not have, so the check is always false and the highlight is applied on every non-empty selection, focused or not. Chromium's right-click selects the word, and that is enough to trigger it.

Two lines fix it. The check reads `isFocused`, which the editor does expose, so the highlight appears only when focus has left the editor, as intended for notes. The extension is also left out of the chat input, which has no use for it, so a platform whose context menu takes focus from the page cannot bring the redraw back there.

```diff
-							const { focused } = this.editor;
+							const { isFocused } = this.editor;

-							if (focused || selection.empty) {
+							if (isFocused || selection.empty) {
 ...
-				SelectionDecoration,
+				...(messageInput ? [] : [SelectionDecoration]),
```

## Testing

1. Chrome on Linux, new chat, a sentence with eleven misspelled words, wait for the underlines, right-click each and pick the suggestion. Current `dev` drops the markers on right-click and loses corrections. This branch keeps all eleven. The probe log on this branch shows each right-click followed by no DOM change at all, then one `insertReplacementText` and one text change per correction.
2. Same with Rich Text Input off. Same result on both.
3. Notes editor: select text, open the edit panel so the editor loses focus. The selection highlight appears. Click back into the editor with the selection kept. The highlight goes away.

## Screenshots

| Surface | Before | After |
|---|---|---|
| Chat input after picking a spelling suggestion from the Chrome context menu | https://github.com/user-attachments/assets/c2be33af-7f06-430b-bb22-312f86f9bafe| https://github.com/user-attachments/assets/d018151b-c229-41d4-80b5-44c51d3e8735 |

### Fixed

- Spelling corrections chosen from the browser's right-click menu now stick in the chat input on Chromium browsers.

## Additional Context

The extension was introduced in 3af2938ef for note selection edits, two days after the move to TipTap 3, and has read the non-existent property since then. Probe log from the chat input on `dev`, right-click on a misspelled word with the underline present:

```
mousedown button=2
focus
contextmenu
DOM childList in <p> added=1 removed=0
DOM childList in <p> added=1 removed=0
DOM childList in <p> added=1 removed=0
DOM childList in <p> added=0 removed=1
```

### DIFF
--- a/src/lib/components/common/RichTextInput.svelte
+++ b/src/lib/components/common/RichTextInput.svelte
@@ -717,9 +717,9 @@
 					props: {
 						decorations: (state) => {
 							const { selection } = state;
-							const { focused } = this.editor;
+							const { isFocused } = this.editor;
 
-							if (focused || selection.empty) {
+							if (isFocused || selection.empty) {
 								return null;
 							}
 
@@ -823,7 +823,7 @@
 				...(messageInput ? [PromptItalic] : []),
 				...(dragHandle ? [ListItemDragHandle] : []),
 				Placeholder.configure({ placeholder: () => _placeholder, showOnlyWhenEditable: false }),
-				SelectionDecoration,
+				...(messageInput ? [] : [SelectionDecoration]),
 
 				...(richText
 					? [


tjbck:
Thanks!

Issue #29808, opened 2026-09-09
issue: DoclingLoader ignores conversion status in HTTP 200 responses


vzd3v:
### Before Submitting

- [x] I searched open and closed issues and discussions for an existing report.
- [x] I checked whether this is already fixed on the `dev` branch or latest source.
- [x] I understand that maintainers want a well-written issue before any code pull request.
- [x] I am using the latest available version of Open WebUI for this reproduction.
- [x] This is not a security vulnerability.

### Installation Method

Git Clone

### Open WebUI Version

`v0.11.3`; also reproduced from `main` at `0a7c15832fb30b1903753e83f81dc7d27e5b0944`

### Operating System

Ubuntu 24.04

### Browser

Not applicable

### Ollama Version

Not applicable

### Summary

`DoclingLoader` treats every HTTP 2xx response as a successful conversion and does not inspect Docling's top-level `status` or `errors`. Docling Serve returns conversion outcomes (`success`, `partial_success`, `skipped`, or `failure`) inside an HTTP 200 JSON response.

For `status="failure"` with empty Markdown, Open WebUI returns a document containing the literal placeholder `<No text content found>`, which is non-empty and can be embedded/indexed. A contract-valid `md_content: null` instead raises an unrelated `TypeError`.

### Expected Behavior

Accept a result only when `status` is `success` or `partial_success` and `document.md_content` is a non-empty string. For `failure`, `skipped`, missing/unknown status, or missing/null/blank Markdown, raise a clear extraction error and include safe details from Docling's `errors` array. Do not create or index placeholder content.

### Actual Behavior

The loader checks only `r.ok`, then executes:

```python
md_content = document_data.get("md_content", "")
text = md_content or "<No text content found>"
```

It never reads `result["status"]` or `result["errors"]`. Empty Markdown is accepted as the placeholder. Null Markdown fails later at `page_break_marker in md_content` with `TypeError: argument of type 'NoneType' is not iterable`.

### Steps to Reproduce

Run this from the Open WebUI backend environment on `v0.11.3` or current `main`:

```python
import json
import tempfile
from unittest.mock import Mock, patch

from open_webui.retrieval.loaders.main import DoclingLoader

payloads = [
    {
        "document": {"filename": "sample.pdf", "md_content": ""},
        "status": "failure",
        "errors": [{
            "component_type": "document_backend",
            "module_name": "reproduction",
            "error_message": "synthetic conversion failure",
        }],
        "processing_time": 0.01,
[...]

PR #29990, opened 2026-09-14
fix(mcp): paginate tools/list until nextCursor is exhausted

### LINKED ISSUE #29879: issue: MCP tools/list pagination is ignored, so only the first page is discovered

Yi-111-a:
## Summary
- Follow MCP `tools/list` pagination via `session.list_tools(cursor=...)` until `nextCursor` is exhausted (with a page cap and repeated-cursor guard).
- Wire pagination into `MCPClient.list_tool_specs` so tool discovery is not limited to the first page.
- Add unit tests covering multi-page, single-page, cursor-loop stop, and `list_tool_specs` aggregation.

Fixes #29879

## Test plan
- [ ] Unit tests in `backend/tests/test_mcp_tools_list_pagination.py` pass
- [ ] Against an MCP server that returns multiple `tools/list` pages, all tools appear in Open WebUI tool specs
- [ ] Against a single-page server, behavior is unchanged


Yi-111-a:
Temporarily closing to free the repository 1-open-PR limit for non-writers so a separate Fixes #29808 PR can be opened from the same fork. This MCP pagination work remains on branch `fix/issue-29879` and can be reopened afterward.

Issue #30088, opened 2026-09-17
issue: an arena model fails with "Model not found" when the models it includes are disabled


silentoplayz:
### Before Submitting

- [x] I searched open and closed issues and discussions for an existing report.
- [x] I checked whether this is already fixed on the `dev` branch or latest source.
- [x] I understand that maintainers want a well-written issue before any code pull request.
- [x] I am using the latest available version of Open WebUI for my install method.
- [x] This is not a security vulnerability.

### Installation Method

Git Clone

### Open WebUI Version

dev at `58b36765a7c20f5943a3180bd289de48876d0878`

### Operating System

Ubuntu 26.04.1 LTS

### Browser

Firefox 155.0

### Ollama Version

0.34.0

### Summary

An arena model that includes only specific models still picks from those models after they are disabled. When every model it includes is disabled, each message sent to the arena model fails with the error "Model not found".

### Expected Behavior

An arena model picks only from the models it includes that are still available. When none of them are available, it does not try to use a disabled model.

### Actual Behavior

The reply to every message shows the error "Model not found", and no model answers.

### Steps to Reproduce

1. Go to Settings, then Evaluations under Admin, and turn on Arena Models.
2. Add an arena model, leave its model filter on include, add two models to it, and save.
3. Go to Models under Admin and turn off both of those models.
4. Start a new chat, select the arena model, and send any message.
5. The reply shows the error "Model not found".

### Logs, Screenshots, and Config

The error on the reply:

<img width="891" height="334" alt="Image" src="https://github.com/user-attachments/assets/75979e04-6a49-4f0d-9643-8e8bd8505911" />

I did not capture browser console or server logs.

### Additional Information

Anchored to `58b36765a7c20f5943a3180bd289de48876d0878`.

Turning a model off removes it from the server's list of available models.
https://github.com/open-webui/open-webui/blob/58b36765a7c20f5943a3180bd289de48876d0878/backend/open_webui/utils/models.py#L183-L206

An arena model on include picks at random from the model IDs saved in its settings without checking that list, and the chat completion picks again the same way. The picked model is then looked up and not found.
https://github.com/open-webui/open-webui/blob/58b36765a7c20f5943a3180bd289de48876d0878/backend/open_webui/utils/middleware.py#L2384-L2408
https://github.com/open-webui/open-webui/blob/58b36765a7c20f5943a3180bd289de48876d0878/backend/open_webui/utils/chat.py#L229-L249
[...]

PR #30090, opened 2026-09-17
fix: pick only available models for an arena model that includes specific models

silentoplayz:
# Pull Request

## Maintainer Request

This fixes my own report, #30088.

## Checklist

- [x] I have read and I understand the [contribution policy](https://docs.openwebui.com/contributing/#submit-code).
- [x] This PR links to a well-described, confirmed Issue or active Discussion: Closes #30088
- [x] I have not added or rewritten automated tests, fixtures, snapshots, or testing infrastructure unless a maintainer explicitly requested them.

## Summary

An arena model that includes specific models now picks only from the ones that are still available. Before, it picked from every model saved in its settings, so a disabled model could be picked. When every included model was disabled, each message failed with "Model not found". When only some were disabled, messages failed at random whenever a disabled one was picked.

**Root cause.** Turning a model off removes it from the server's list of available models:
https://github.com/open-webui/open-webui/blob/58b36765a7c20f5943a3180bd289de48876d0878/backend/open_webui/utils/models.py#L183-L206

In exclude mode, an arena model builds its choices from that list. In include mode, it picked straight from the saved IDs without checking them against the list. The same pick runs when the chat payload is processed and again as a fallback during the chat completion:
https://github.com/open-webui/open-webui/blob/58b36765a7c20f5943a3180bd289de48876d0878/backend/open_webui/utils/middleware.py#L2384-L2408
https://github.com/open-webui/open-webui/blob/58b36765a7c20f5943a3180bd289de48876d0878/backend/open_webui/utils/chat.py#L229-L249

**Fix.** In both places, include mode now keeps only the saved IDs found among the available models. It is the counterpart of the exclude filter right above it, with `in` in place of `not in`.

```diff
                 if available_model.get('owned_by') != 'arena' and available_model['id'] not in arena_model_ids
             ]
+        elif arena_model_ids:
+            arena_model_ids = [
+                available_model['id']
+                for available_model in request.app.state.MODELS.values()
+                if available_model.get('owned_by') != 'arena' and available_model['id'] in arena_model_ids
+            ]
```

The same six lines are added to the fallback in `generate_chat_completion`. Two files changed, with 12 lines added.

## Verification

I reproduced the bug on `dev` first. I added an arena model that includes two models, turned both models off in the admin Models settings, and sent a message to the arena model. The reply showed the error "Model not found". The full reproduction is in #30088.

After, on this branch, I tested the same setup locally. With both included models off, messages to the arena model get replies. With one of them turned back on, the replies come from that model.

Existing checks on this branch: both changed files compile, and ruff format reports no changes to the added lines. The formatting ruff suggests elsewhere in `middleware.py` is already on `dev`.

As supporting evidence, the arena selection code was taken from both changed files and run 200 times per case against a stub model list. It covered five cases:

- all included models missing
- one included model missing
- all included models available
- exclude mode
- no model list

On this branch, every pick in every case is an available model. On `dev`, both cases with a missing model pick it. With one model missing, it is picked on some runs and an available model on others.

### Fixed

- Arena models that include specific models no longer pick a disabled model, which made messages fail with "Model not found".

## Screenshots

| Surface | Before | After |
|---|---|---|
| Reply from an arena model whose included models are disabled | image | image |

## Additional Context

When none of the included models are available, the filtered list is empty. The existing fallback then picks from all available models, which is what exclude mode already does when it excludes every model. For regular users, the picked model still goes through the normal access check before the reply is generated:
https://github.com/open-webui/open-webui/blob/58b36765a7c20f5943a3180bd289de48876d0878/backend/open_webui/utils/chat.py#L209-L213


tjbck:
Arena model config should be updated.

PR #30138, opened 2026-09-18
feat: add Staan as a native web search provider

Classic298:
European deployments that need EU data residency have no hosted web search option out of the box: every plug-and-play provider shipped today is US-based, and the only sovereign alternative is self-hosting SearXNG, which means running and maintaining that infrastructure yourself. Staan (staan.ai) is a European search API with EU data residency, so adding it gives those deployments a drop-in choice.

It is configured like any other provider, through the admin UI or STAAN_API_KEY, STAAN_MARKET and STAAN_MAX_SNIPPETS. Market sets the region and language of the results and defaults to en-us. Max snippets asks Staan to fetch each result page and return semantically scored chunks of it, which get merged into that result's snippet so retrieval has more to work with; leaving it at 0 uses the plain search endpoint.

Wired the same way as Tavily and Exa, with domain filtering going through the shared get_filtered_results.

Requested in #26006.


dsisysteme:
Thank you for integrating this feature.

I just tested this commit, and it looks like there is an extra parameter in the request to Staan.

There is no  "count" parameter supported by the Staan API. Staan currently returns 10 results, and this value cannot be configured in the request.

In the code, there is already a filtering step that only passes the requested count results to the model, which limits the context size.
`for result in results[:count]`

So the count parameter should simply be removed from the request to Staan:
`params = {'q': query, 'market': market, 'count': count}`  => `params = {'q': query, 'market': market}`

Currently, this causes 400 errors when performing searches with Staan.

Issue #30133, opened 2026-09-18
issue: a pgvector read that returns nothing never gives its connection back


runyournode:
### Before Submitting

- [x] I searched open and closed issues and discussions for an existing report.
- [x] I checked whether this is already fixed on the `dev` branch or latest source.
- [x] I understand that maintainers want a well-written issue before any code pull request.
- [x] I am using the latest available version of Open WebUI for my install method.
- [x] This is not a security vulnerability.

### Installation Method

Docker

### Open WebUI Version

ghcr.io/open-webui/open-webui@sha256:0bd34b02c04554bc03433e942cee457a8969148ac21f4864082e3e14f40c987a

### Operating System

Ubuntu 24.04 (WSL2)

## What happens

In `backend/open_webui/retrieval/vector/dbs/pgvector.py`, `PgvectorClient.search`, `.query` and `.get` return early when the result set is empty, skipping the `self.session.rollback()` the populated path performs:

```python
results = result_proxy.all()
...
if not results:
    return SearchResult(ids=ids, ...)      # no rollback

for row in results:
    ...
self.session.rollback()  # read-only transaction
return SearchResult(...)
```

The session keeps its transaction open and, with it, the connection it checked out of the pool.

## Why one thread never sees it

A single thread reusing a single session runs the next statement on the same open transaction.
`query_collection` fans out over `asyncio.to_thread`, and `PgvectorClient.session` is a `scoped_session` — thread-local — so every thread gets its own Session.
A thread that finishes while still holding a connection never gives it back.

Twelve threads issuing one empty search each, against the default pool (size 5, overflow 10): **12 connections checked out, 0 free**.
The next burst waits out the 30 s checkout timeout and fails with `QueuePool limit of size 5 overflow 10 reached`.
Vector search is down and nothing in the application log says why.

## Direction of causation

Checked that it is not the other way round (exhausted pool → timeouts → empty results): 3 threads against a pool of 15, where no timeout can occur, the only variable being whether the searched collection has rows.

| | rows returned | checked out | free |
|---|---|---|---|
| 3 threads, non-empty | 10, 10, 10 | 0 | 3 |
| 3 threads, **empty** | 0, 0, 0 | **3** | **0** |
| 3 threads, non-empty again | 10, 10, 10 | 3 | 3 |

Only the empty case retains its connections. The third row — three checked out *and* three free — is a leak, not exhaustion.
[...]

PR #30139, opened 2026-09-18
fix: release the pgvector connection when a read returns no rows

Classic298:
A pgvector search, query or get that finds nothing returns before the rollback that ends its read-only transaction, so the session keeps the connection it checked out of the pool. Because retrieval fans out over worker threads and the session is thread-local, every thread that runs an empty read holds a connection for the lifetime of that thread.

For users this shows up as vector search dying after a while with "QueuePool limit of size 5 overflow 10 reached, connection timed out", with no way back other than a restart. Empty reads are routine: an empty knowledge base, a file whose chunks were deleted, or a metadata filter that matches nothing all produce one.

Rolling back before each of the three early returns puts the connection back. Measured against PostgreSQL 16 with three threads on a pool of 15, so no timeout is possible: an empty collection left 3 connections checked out and 0 free, and with this change 0 checked out and 3 free. Twelve empty reads on a default-shaped pool left all 12 checked out before, 0 after.

Fixes #30133


Classic298:
Was superseeded

Issue #30135, opened 2026-09-18
issue: on the shared pgvector index, a knowledge-base search can lose most of its neighbours (the loss grows with the table)


runyournode:
### Before Submitting

- [x] I searched open and closed issues and discussions for an existing report.
- [x] I checked whether this is already fixed on the `dev` branch or latest source.
- [x] I understand that maintainers want a well-written issue before any code pull request.
- [x] I am using the latest available version of Open WebUI for my install method.
- [x] This is not a security vulnerability.

### Installation Method

Docker

### Open WebUI Version

ghcr.io/open-webui/open-webui@sha256:0bd34b02c04554bc03433e942cee457a8969148ac21f4864082e3e14f40c987a

### Operating System

Ubuntu 24.04 (WSL2)

## What happens

The pgvector backend keeps every collection's chunks in one table, `document_chunk`, under one vector index, and scopes a search with `WHERE collection_name = …`.
pgvector applies that predicate after the ANN walk: the index returns its `ef_search` nearest candidates from the whole table, and only those that belong to the collection are kept.
A knowledge base that holds 1.5 % of the rows owns about 1.5 % of the candidates, so most of the ten neighbours a query asks for are never examined.
Nothing is logged and no error is raised.

The loss is governed by the collection's share of the table.
Every knowledge base gets worse as the install grows, the small ones first.

## Measured

1M rows of 1024 dimensions, 20 knowledge bases of 15k rows each, the remaining 70 % in per-file, per-memory and per-web-search collections[^1].
recall@10 against an exact kNN over the same collection, 200 queries per point, through `PgvectorClient.search()`, HNSW `m=16`:

| `ef_search` | recall@10 | worst decile | SQL ms |
|---|---|---|---|
| 10 | 0.143 | 0.00 | 3.0 |
| **40 (default)** | **0.269** | **0.00** | 4.7 |
| 100 | 0.459 | 0.20 | 7.0 |
| 400 | 0.937 | 0.80 | 9.3 |

At the default setting three of every four correct neighbours are missing, and one query in ten returns nothing relevant.
At 3M rows the default gives 0.284.
IVFFlat, the default index method, gives 0.835 at its default `probes = 1`; at any higher `probes` the planner drops the index for an exact sort — correct, at 46 ms (1M) and 135 ms (3M) per search.
Raising `ef_search` widens the candidate list and helps slowly, at the cost shown; it does not change the mechanism.

## What pgvector offers, and Open WebUI does not use

[...]

PR #30142, opened 2026-09-18
fix: pgvector reads leak their connection and lose most of their neighbours

Classic298:
Two defects on the pgvector read path. A search, query or get that finds nothing returns before the rollback that ends its read-only transaction, so the session keeps the connection it checked out; retrieval fans out over worker threads and the session is thread-local, so every thread that runs an empty read holds a connection for the lifetime of that thread. Users see vector search die after a while with "QueuePool limit of size 5 overflow 10 reached" and no way back other than a restart. Rolling back before the three early returns puts the connection back: measured on PostgreSQL 16, twelve empty reads on a default-shaped pool left all 12 connections checked out before and 0 after.

All collections also share one table under one vector index, and the WHERE collection_name filter is applied after the index walk, so a knowledge base holding a small share of the rows keeps only a small share of its neighbours, silently. pgvector 0.8 added iterative scans for this: the scan keeps going until enough rows pass the filter. This sets it per search so it cannot leak into other sessions, and only when the installed extension supports it, since setting it on pgvector 0.7 would make every search raise. PGVECTOR_ITERATIVE_SCAN turns it off or picks strict_order; it defaults to relaxed_order because the shipped behaviour is silently wrong results, and the cost is about a millisecond per search.

Measured through PgvectorClient.search on PostgreSQL 17 and pgvector 0.8, 101500 rows of 384 dimensions with the knowledge base at 1.5% of the table, hnsw m=16, recall@10 against an exact scan over 40 queries: 0.070 at 4.9 ms before, 0.970 at 7.2 ms after.

Fixes #30133
Fixes #30135


tjbck:
@Classic298 conflicts

Classic298:
@tjbck rebased on dev, conflicts resolved.

Issue #30195, opened 2026-09-19
fix: ask_user badges the first option as Recommended without the model ever recommending it


kirin-xiao:
### Before Submitting

- [x] I searched open and closed issues and discussions for an existing request.
- [x] I checked whether this already exists on the `dev` branch or latest source.
- [x] I understand that maintainers want a well-written issue or discussion before any code pull request.
- [x] This request is not a security vulnerability.

### Problem

The `ask_user` card always badges the first option as "Recommended" (`src/lib/components/chat/AskUserCard.svelte`: `{#if optionIndex === 0}` renders `{$i18n.t('Recommended')}`), but nothing tells the model this. The `ask_user` docstring (`backend/open_webui/tools/builtin.py`) says nothing about option order or recommendations, and `normalize_ask_user_request` (`backend/open_webui/utils/ask_user.py`) preserves the model's order.

So the badge really means "whichever option was listed first". It can be wrong (e.g. a model listing options cheapest-first), and a model that does want to recommend an option has no way to know it should order it first.

OpenCode's equivalent `question` tool handles this explicitly in its tool description:

> - If you recommend a specific option, make that the first option in the list and add "(Recommended)" at the end of the label

### Desired Behavior

Make the badge meaningful, e.g. add an equivalent instruction to the `ask_user` tool description so the model deliberately orders its recommended option first.

Note the card already renders a "Recommended" chip on option 0, so an explicit label suffix would be redundant here. Either drop the automatic chip and rely on the model's label, or keep the chip and ask the model only to order the option first.

### Why This Matters

Agents cannot express a recommendation through `ask_user`, and users may act on a "Recommended" label the model never intended.

### Examples or References

- OpenCode `question` tool description quoted above.
- `backend/open_webui/tools/builtin.py` (`ask_user` docstring).
- `src/lib/components/chat/AskUserCard.svelte` (badge on `optionIndex === 0`).

### Alternatives or Workarounds

None for users or model authors: the built-in tool description is not user-controlled.

PR #30196, opened 2026-09-19
fix: tell ask_user models that the first option is shown as Recommended

Classic298:
The ask_user card badges the first option of every question as "Recommended", but nothing ever told the model that. The model picks whatever order it likes, so the badge really means "listed first" and users act on a recommendation the model never made.

The ask_user tool description now states that the first option is labelled Recommended and that the model should list the option it recommends first, so the badge reflects an actual choice.

Kept the badge and instructed the model instead of adding a per-option "recommended" flag: the flag would need a schema change, validation in the request normalizer and a frontend change, for the same result in the common case. Dropping the badge was the other option, but it removes a useful affordance rather than fixing it.

Verified that the added line reaches the model by running the docstring through the tool-spec builder and checking the generated OpenAI function schema.

Fixes #30195


kirin-xiao:
There might be cases where the model does not want to recommend anything.

Issue #30198, opened 2026-09-19
issue: the temporary chat URL parameter is ignored when the New Chat page is reached without a page load


silentoplayz:
### Before Submitting

- [x] I searched open and closed issues and discussions for an existing report.
- [x] I reproduced this bug on the latest release AND on the current `dev` branch, right before submitting this report. I did not just check an old version or rely on a check from days ago.
- [x] I understand that maintainers want a well-written issue before any code pull request.
- [x] This is not a security vulnerability.

### Installation Method

Git Clone

### Open WebUI Version

dev commit 621ed6de7

### Operating System

Linux

### Browser

Firefox

### Summary

Returning to the New Chat page while its address still carries the temporary chat parameter gives an ordinary chat, not a temporary one. The address and the page disagree, and anything typed there is saved.

### Expected Behavior

The New Chat page is in temporary mode whenever its address says so, however the page was reached.

### Actual Behavior

The address reads the temporary chat parameter, the temporary chat icon is off, and the chat is saved like any other. Reloading the page restores temporary mode, because a full page load is the only path that reads the parameter.

### Steps to Reproduce

1. On the New Chat page, click the temporary chat button. The address gains the temporary chat parameter and the icon turns on.
2. Click a saved chat in the sidebar. Temporary mode switches off, which is correct for that chat.
3. Press the browser Back button.
4. The New Chat page comes back with the temporary chat parameter still in the address, but temporary mode is off.
5. Press F5 on that same address: temporary mode comes back.

### Logs, Screenshots, and Config

Nothing is written to the browser console or the server log. Default configuration, admin account.

### Additional Information

The temporary chat parameter is read in exactly one place, inside the layout's mount handler:

https://github.com/open-webui/open-webui/blob/621ed6de7f85d216be6adde7fcebdbc601aad735/src/routes/%28app%29/+layout.svelte#L359-L362

Every other URL parameter is read by `initNewChat`, which runs each time the New Chat page is opened:

https://github.com/open-webui/open-webui/blob/621ed6de7f85d216be6adde7fcebdbc601aad735/src/lib/components/chat/Chat.svelte#L2158-L2176

That handler already covers `youtube`, `load-url`, `web-search`, `image-generation`, `code-interpreter` and `tools`. The layout mounts once per full page load, so reaching the address by any in-app navigation never re-reads the parameter. Opening a saved chat switches temporary mode off, and nothing switches it back on.

[...]

PR #30200, opened 2026-09-19
fix: honor the temporary chat URL parameter when the New Chat page is reached without a page load

silentoplayz:
# Pull Request

## Maintainer Request

This fixes my own report, #30198.

## Checklist

- [x] I have read and I understand the [contribution policy](https://docs.openwebui.com/contributing/#submit-code).
- [x] This PR links to a well-described, confirmed Issue or active Discussion: Closes #30198
- [x] I have not added or rewritten automated tests, fixtures, snapshots, or testing infrastructure unless a maintainer explicitly requested them.

## Summary

The New Chat page now enters temporary mode whenever its address carries the temporary chat parameter, however the page was reached. Before, only a full page load applied it, so returning to that address in the app gave an ordinary chat whose messages were saved.

**Root cause.** The parameter is read in one place, the layout's mount handler:
https://github.com/open-webui/open-webui/blob/621ed6de7f85d216be6adde7fcebdbc601aad735/src/routes/%28app%29/+layout.svelte#L359-L362

Every other URL parameter is read by `initNewChat`, which runs whenever the New Chat page is opened:
https://github.com/open-webui/open-webui/blob/621ed6de7f85d216be6adde7fcebdbc601aad735/src/lib/components/chat/Chat.svelte#L2158-L2176

The layout mounts once per full page load, so an in-app navigation to that address never re-reads the parameter. Opening a saved chat switches temporary mode off:
https://github.com/open-webui/open-webui/blob/621ed6de7f85d216be6adde7fcebdbc601aad735/src/lib/components/chat/Chat.svelte#L2260-L2264

and nothing switches it back on when the user returns.

**Fix.** Read the parameter where the sibling parameters are read, immediately above the guard that forces temporary mode off for users who may not use it, so that guard still has the last word:

```diff
+		if ($page.url.searchParams.get('temporary-chat') === 'true') {
+			await temporaryChatEnabled.set(true);
+		}
+
 		if ($user?.role !== 'admin' && !$user?.permissions?.chat?.temporary) {
 			await temporaryChatEnabled.set(false);
 		}
```

The layout's existing read is left alone; it still covers the first load.

## Verification

I reproduced the bug on `dev` first: turn on temporary chat from the New Chat page, open a saved chat, press Back. The address still carried the parameter while temporary mode was off. The full reproduction is in #30198.

After, on this branch, I ran these checks locally:

1. The same steps: the New Chat page comes back in temporary mode, matching the address.
2. Reloading that address still gives temporary mode.
3. Turning temporary chat off still removes the parameter from the address and leaves an ordinary new chat.
4. A user in a group that may not use temporary chats still cannot get temporary mode from the address.

Because Back only reaches the New Chat page once the router's history state survives an address rewrite, I tested this branch with #30199 merged in locally as well; the two changes are independent and this one is filed on its own.

Existing checks on this branch: Prettier reports the changed file as formatted. I did not run `npm run build`.

No screenshots: the change has no visual surface of its own; the difference is whether temporary mode matches the address, which the steps above describe.

### Fixed

- The temporary chat URL parameter is applied when the New Chat page is reached without a page load.

## Additional Context

The change is four lines in `initNewChat` and adds no state. It is client-side, so it lands the same on every deployment; the permission guard that follows it is unchanged, so groups that may not use temporary chats are unaffected. Reverting it restores the previous behaviour and nothing else. It coexists with the layout's own read, which still handles the first page load, and with the toggle, which strips the parameter when temporary mode is switched off.


silentoplayz:
Superseded by 67adde319 on dev, which covers the same behaviour. Thanks for carrying it across.

Issue #29989, opened 2026-09-14
issue: Integrations button becomes invisible on mobile when UI Scale is 1.3x or higher


kirin-xiao:
### Before Submitting

- [x] I searched open and closed issues and discussions for an existing report.
- [x] I checked whether this is already fixed on the `dev` branch or latest source.
- [x] I understand that maintainers want a well-written issue before any code pull request.
- [x] I am using the latest available version of Open WebUI for my install method.
- [x] This is not a security vulnerability.

### Installation Method

Docker

### Open WebUI Version

`dev` @ `55b7343be`

### Operating System

Android (Samsung Galaxy S8+)

### Browser

Chrome 152

### Ollama Version

N/A

### Summary

On mobile, at `Settings > Interface > UI Scale` of 1.3x or higher (1.2x on higher DPR screens), the `Integrations` button in the composer's bottom row becomes completely invisible, with the model name appearing to cover it. It is clipped, not just off-screen: it is untappable.

### Expected Behavior

`Integrations` stays visible and tappable at every `UI Scale` value.

### Actual Behavior

The button is gone and untappable. The model selector's cap is rem based, so it grows with `UI Scale` (`10rem` = 160px at 1x, 208px at 1.3x), and the left control group is `flex-1` (basis 0), so it collapses and clips the `Integrations` button inside its `overflow-x-auto scrollbar-none` strip. No scrollbar hints the button still exists.

### Steps to Reproduce

1. On mobile (Galaxy S8+, Chrome 152, 360px viewport), open the chat page.
2. `Settings > Interface > UI Scale` to `1.2x`.
3. Return to the chat: the `Integrations` button between `+` and the model selector is gone, replaced by the model name.
4. Tap its position: nothing opens. At 1x it reappears.

### Logs, Screenshots, and Config

<img width="430" height="167" alt="Image" src="https://github.com/user-attachments/assets/72e88ab7-f0e6-4083-9258-fb503650a010" />

Layout only; no console or server output. Only `UI Scale` changed from defaults.

Relevant code in `src/lib/components/chat/MessageInput.svelte`:

- `:2191` bottom row `flex justify-between`
- `:2192` left group `ml-1 self-end flex items-center flex-1 min-w-0` (basis 0)
- `:2269` chip strip `flex flex-1 items-center min-w-0 overflow-x-auto scrollbar-none`
- `:2304` `Integrations` button inside that strip
[...]

PR #30374, opened 2026-09-22
fix: keep the composer's + and Integrations buttons visible at larger UI Scale on mobile

Classic298:
On a phone with UI Scale at 1.2x or higher, the model selector in the chat composer covers the Integrations button, so it cannot be seen or tapped. From 1.3x it covers the + button too, which leaves no way to attach files or open integrations.

The model selector's width cap is rem based, so it grows with UI Scale, and the left button group could shrink to nothing, so the selector took the space first. The + and Integrations buttons now sit in a group that keeps its width, and only the toggled chips stay in the scrolling strip. When space runs out the model name truncates, and chips scroll as before.

Measured in the real app at a 360px mobile viewport: before, Integrations is covered at 1.2x and both buttons at 1.3x and 1.5x; after, both are tappable at every scale, with the model name truncated (114px at 1.3x). Button positions at 1x and on desktop are unchanged to the pixel, with and without chips.

Fixes #29989


tjbck:
@Classic298  screenshots?

Classic298:
@tjbck

| | Before | After |
|---|---|---|
| 1x |  |  |
| 1.3x |  |  |
| 1.5x |  |  |

